PR #74, opened 2015-01-12
internalScratchpads transient to help serialize

MasseGuillaume:


xeno-by:
We can probably do that, but that won't help with serialization - once the trees are stripped off their internalScratchpads, all semantic APIs will stop working. That won't be the case in the future, but the technology that is going to enable this is not even designed, just theorized about.

MasseGuillaume:
Make sense. The idea behind the serialization was to have an quick way to develop scala meta. I find the scala repl (Scala Kata in my case) really usefull to experiment/prototype. How hard is it to make scalameta a standalone program (eg. not a scala plugin)?

xeno-by:
Got time for a hangout call? Would be faster to explain this face to face (you might have questions that would be hard for me to anticipate in advance).

xeno-by:
https://groups.google.com/forum/#!topic/scalameta/cElRkk0-kU8

PR #100, opened 2015-02-12
introduces meta.Pat.Type

xeno-by:
Now I remember why we didn't introduce Pat.Var a year ago.

Because it would require a symmetric change to express type variable patterns,
and that would make it impossible to use meta.Type types for type patterns.
That was deemed too inconvenient and rejected.

Now I think that we should've followed that through, because pattern types
seem to form a distinct syntactic group, with their own peculiarities.
An argument that types and pattern types look the same syntactically
is compelling and seductive, but Term.Apply and Pat.Extract also look
the same, and we didn't unify them, so I think that there is more to
tree unification that just syntactic likeness.

Even though this change creates an inconvenience (it's no longer possible
to use Type and Pat.Type interchangeably), that inconvenience is arguably minor,
because one could always imagine having a bidirectional conversion, with
the Type => Pat.Type direction being always successful (+ a Liftable)
and Pat.Type => Type being potentially fallible.

### DIFF
--- a/docs/quasiquotes.md
+++ b/docs/quasiquotes.md
@@ -107,6 +107,26 @@
  Sequence Wildcard | `p"_*"`
  Pattern           | `p"$pat"`
 
+## Type Patterns (meta.Pat.Type)
+
+                   | Quasiquote
+-------------------|------------------------------
+ Wildcard          | `pt"_"`
+ Var               | `pt"name"`
+ Name              | `pt"`name`"`
+ Selection         | `pt"$ref.$tname"`
+ Projection        | `pt"$ptpe#$tname"`
+ Singleton         | `pt"$ref.type"`
+ Application       | `pt"$ptpe[..$ptpes]`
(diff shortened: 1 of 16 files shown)


xeno-by:
https://github.com/scalameta/scalameta/pull/101 fixes several blunders that I made in this pull request.

densh:
If patterns types are not regular types then you need yet another interpolator to express them.

docs/quasiquotes.md
```diff
@@ -107,6 +107,26 @@
  Sequence Wildcard | `p"_*"`
  Pattern           | `p"$pat"`
 
+## Type Patterns (meta.Pat.Type)
+
+                   | Quasiquote
+-------------------|------------------------------
+ Wildcard          | `pt"_"`
+ Var               | `pt"name"`
```
densh: `name` is not var, it's a val. Overloading of notation doesn't work.

PR #117, opened 2015-03-04
TQL: bug fix in TraverserBuilder

mdemarne:
1. The TQL TraverserBuilder prevented any traversal of Templates. This was due to the way stats are setup into templates as Option[Seq[Stat]], which was not handled by the previous version of TQL. This is now corrected and the Obey test case passes as usual.

2. Correcting ObeyRuleSuite for TQL: we can't compare source code as strings directly, due to formatting differences between the source code and the prettyprinter output. The test now checks if the raw of the trees are equals. A small case was also added for testing purposes.

This is the continuation of https://github.com/begeric/TQL-scalameta/pull/2.

### DIFF
--- a/foundation/src/main/scala/org/scalameta/tql/TraverserBuilder.scala
+++ b/foundation/src/main/scala/org/scalameta/tql/TraverserBuilder.scala
@@ -9,8 +9,11 @@ class TraverserBuilderMacros(val c: Context) extends AdtReflection {
   import c.universe._
   val XtensionQuasiquoteTerm = "shadow scala.meta quasiquotes"
 
+  /* Get all leaves for the root (i.e. all Tree cases) */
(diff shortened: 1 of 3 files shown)


foundation/src/main/scala/org/scalameta/tql/TraverserBuilder.scala
```diff
@@ -42,12 +49,6 @@ class TraverserBuilderMacros(val c: Context) extends AdtReflection {
       None
   }
 
-  def buildImpl[T : c.WeakTypeTag, A : c.WeakTypeTag](f: c.Tree, objs: c.Tree*): c.Tree = {
```
mdemarne: This function was not called. BuildImplDeletage is used instead it seems.

xeno-by:
Merged as https://github.com/scalameta/scalameta/pull/120

PR #123, opened 2015-03-09
introduces Context.tpe(Term.Param)

xeno-by:
Doesn't look like we can express this kind of reflection using the
other methods of the context API, so we'll have to introduce this method,
at least for the time being.


vjovanov:
Thx, I needed this.

PR #128, opened 2015-04-04
Adding line position to Origin

mdemarne:
- Addition of line positions to origin
- Addition of a benchmark suite to test it using Defn.Def, Defn.Class, Term.Match.

### DIFF
--- a/scalameta/src/main/scala/scala/meta/Origin.scala
+++ b/scalameta/src/main/scala/scala/meta/Origin.scala
@@ -10,6 +10,8 @@ import org.scalameta.invariants._
   def dialect: Dialect
   def start: Int
   def end: Int
+  def startLine: Int
+  def endLine: Int
   def tokens: Seq[Token]
 }
 
@@ -19,6 +21,8 @@ object Origin {
     val dialect = scala.meta.dialects.Scala211
     val start = 0
     val end = -1
+    val startLine = 0
+    val endLine = -1
     def tokens = Nil
   }
 
@@ -35,6 +39,13 @@ object Origin {
     private implicit lazy val thisDialect: Dialect = this.dialect
     val start = input.tokens.apply(startTokenPos).start
     val end = if (endTokenPos != -1) input.tokens.apply(endTokenPos).end else -1
+    def findLine(x: Int, default: Int) = input match {
+      case Input.None => default
+      case els => els.content.take(x + 1).count(_ == '\n')
+    }
+    val startLine = findLine(start, 0)
+    val endLine = findLine(end, -1)
+
     def tokens: Seq[Token] = input.tokens(dialect).slice(startTokenPos, endTokenPos + 1)
   }
 
@@ -46,6 +57,8 @@ object Origin {
     def dialect = tree.origin.dialect
     def start = tree.origin.start
     def end = tree.origin.end
+    def startLine = tree.origin.startLine
+    def endLine = tree.origin.endLine
     def tokens = tree.origin.tokens
   }
 }
--- a/scalameta/src/main/scala/scala/meta/Origin.scala
+++ b/scalameta/src/main/scala/scala/meta/Origin.scala
@@ -39,7 +39,7 @@ object Origin {
     private implicit lazy val thisDialect: Dialect = this.dialect
     val start = input.tokens.apply(startTokenPos).start
     val end = if (endTokenPos != -1) input.tokens.apply(endTokenPos).end else -1
-    def findLine(x: Int, default: Int) = input match {
+    private def findLine(x: Int, default: Int) = input match {
       case Input.None => default
       case els => els.content.take(x + 1).count(_ == '\n')
     }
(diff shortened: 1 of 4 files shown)


tests/src/test/scala/parser/LinePositionSuite.scala
```diff
@@ -0,0 +1,68 @@
...
+  var dir = new File(new File(System.getProperty("sbt.paths.tests.source")).getAbsolutePath)
+  def isProjectRoot(dir: File) = dir != null && new File(dir.getAbsolutePath + File.separatorChar + "project" + File.separatorChar + "build.scala").exists
+  while (dir != null && !isProjectRoot(dir)) dir = dir.getParentFile
+  test("ProjectDir (" + dir.getAbsolutePath + ")")(assert(isProjectRoot(dir)))
+
+  val ignoredFiles = List(
+    "build.scala",
+    "Token.scala",
+    "Api.scala"
+  )
```
xeno-by: What's wrong with these files?

scalameta/src/main/scala/scala/meta/Origin.scala
```diff
@@ -10,6 +10,8 @@ import org.scalameta.invariants._
   def dialect: Dialect
   def start: Int
   def end: Int
+  def startLine: Int
+  def endLine: Int
```
xeno-by: I think it'd be useful to also add these methods to https://github.com/xeno-by/scalameta/blob/master/scalameta/src/main/scala/scala/meta/syntactic/Api.scala#L48.

tests/src/test/scala/parser/LinePositionSuite.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  val findAllDefn = topDown(collect {
+    case clss: impl.Defn.Class =>
```
xeno-by: Usually one just calls those `x`, which obviates the need to come up with weird names :)

tests/src/test/scala/parser/LinePositionSuite.scala
mdemarne: Build.scala issue some parse exception, probably due to its special syntax:

>[info]   scala.meta.ParseException: XML literals are not supported at < (1751..1751)
>[info]   at scala.meta.internal.parsers.Reporter$class.syntaxError(Reporter.scala:13)
>[info]   at scala.meta.internal.parsers.Reporter$$anon$1.syntaxError(Reporter.scala:18)
>[info]   at scala.meta.internal.parsers.Parser.xmlLiteral(Parsers.scala:1230)
>[info]   at scala.meta.internal.parsers.Parser$$anonfun$simpleExpr$1.apply(Parsers.scala:1254)
>[info]   at scala.meta.internal.parsers.Parser$$anonfun$simpleExpr$1.apply(Parsers.scala:1245)
>[info]   at scala.meta.internal.parsers.Parser.atPos(Parsers.scala:277)
>[info]   at scala.meta.internal.parsers.Parser.autoPos(Parsers.scala:294)
>[info]   at scala.meta.internal.parsers.Parser.simpleExpr(Parsers.scala:1245)
>[info]   at scala.meta.internal.parsers.Parser.prefixExpr(Parsers.scala:1221)
>[info]   at scala.meta.internal.parsers.Parser.postfixExpr(Parsers.scala:1207)


Token.scala fails, because apparently the name of

```scala
@token class `macro`(start: Int) extends Keywords
```

is named "macro" without `.

/semantic/Api.scala contains something of the form:

```scala
internalFilter[T](that => {
              def thisDenot = thisName.denot.require[h.Denotation.Precomputed]
              def thatDenot = that.require[impl.Member].name.require[impl.Name].denot.require[h.Denotation.Precomputed]
              scala.util.Try(thisDenot.symbol == thatDenot.symbol).getOrElse(false)
            }) match {
		...
	    }
```

The starting line of the match contains `internalFilter[T](that => {` only, and therefore does not pass the test for `match`. To keep the test simple, I decided to avoid checking the Api.scala files.

xeno-by:
Thanks! It probably makes sense to create a dedicated class called `Position` and give it methods like `offset`, `line`, `column`, but let's think about this later.

scalameta/src/main/scala/scala/meta/Origin.scala
```diff
@@ -35,6 +39,13 @@ object Origin {
     private implicit lazy val thisDialect: Dialect = this.dialect
     val start = input.tokens.apply(startTokenPos).start
     val end = if (endTokenPos != -1) input.tokens.apply(endTokenPos).end else -1
+    private def findLine(x: Int, default: Int) = input match {
+      case Input.None => default
+      case els => els.content.take(x + 1).count(_ == '\n')
+    }
+    val startLine = findLine(start, 0)
+    val endLine = findLine(end, -1)
```
densh: This looks like `O(n)` every time you call it where `n` is number of preceding tokens. And for all trees it's going to be `O(n!)`.

mdemarne: A simply way to avoid that would be to have `lazy vals`, since startLine and endLine might not be used much.

densh: I think making it a `lazy val` or even `def` should be an improvement.

mdemarne: Done in https://github.com/scalameta/scalameta/pull/129.

PR #129, opened 2015-04-07
endLine and startLine as lazy vals for performance improvement

mdemarne:
Changing endLine and startLine to lazy vals for performance improvement.


densh:
LGTM

xeno-by:
Nice catch @densh!

PR #132, opened 2015-05-04
a big refactoring of tokens, origins, inputs and positions

xeno-by:
review @mdemarne


xeno-by:
Summary of the most important changes:
1) `Vector[Token]` => `Tokens` (the underlying collection is still a vector of tokens, but now it's encapsulated under a decent name; the API hasn't changed - the usual stuff like `:+`, `map` or `slice` works)
2) Removed `Token.dialect`, `Token.index`, `Token.prototype`. Tokens are now fat wrappers over the tuple of `(Input, Int, Int)`. In the future, I'm going to make Token a value class over a long, and Tokens will become a bitstream of delta-encoded starts.
3) Refactored origins. There are now `Origin.Parsed` and `Origin.Synthetic`. The former is used as usual, the latter is lazily assigned to every tree that doesn't have an explicitly specified origin. @mdemarne, the logic that you'll need to implement lives in https://github.com/xeno-by/scalameta/blob/master/scalameta/src/main/scala/scala/meta/internal/ui/InferTokens.scala, which is called to fill in tokens of a synthetic origin.
4) Refactored inputs. There are now `Input.Real` (and a bunch of its subclasses with concrete `def content: Array[Char]`) and `Input.Virtual` (that contains an underlying collection of tokens). Since tokens no longer need an index, we can create virtual inputs at will and not worry about their tokens not belonging to the same input. The new distinction between inputs is propagated into some type signatures (e.g. `Token.input` is now typed as `Input.Real`), which has made things more precise.
5) Encapsulated `start` and `end` into `Point`, which has `input`, `offset`, `line` and `column`. A start and an end can be combined into `Position`.
6) Implemented initial support for XML literals. We now have dedicated tokens for them, as well as a tree representation (encoding via string interpolations). Current support is pretty stupid - it detects starts and ends of XML literals and then stuffs everything in-between into a static interpolation part, but that at least allows stuff to parse successfully.

xeno-by:
@mdemarne, the logic that you'll need to implement lives in https://github.com/xeno-by/scalameta/blob/master/scalameta/src/main/scala/scala/meta/internal/ui/InferTokens.scala, which is called to fill in tokens of `Origin.Synthetic`.

As we discussed, the tokens that you'll be creating will combine the tokens from subtrees that come from `Origin.Parsed` as well as synthetic tokens that you'll create similarly to what `ShowCode` does at the moment. I think that you'll be able to use @duhemm's token quasiquotes for this purpose - we'll need to discuss that tomorrow.

Please let me know whether these changes make sense.

xeno-by:
What annoys me somewhat is that even if we have a tree with `Origin.Parsed`, then we still have to change the origin to `Origin.Synthetic` if we insert it somewhere (in order to accommodate indentation). If we come up with a solution that would allow to leave original tokens alone, it would be really cool.

docs/ast.md
```diff
@@ -306,18 +246,24 @@ Here's the same code after typechecking with helper def macros expanded:
       scala.Some.apply[scala.meta.Tree](Impl.this.internalParent)
     else
       scala.None;
-    def origin: scala.meta.Origin = Impl.this.internalOrigin;
+    def origin: scala.meta.Origin = {
```
mdemarne: Makes sense! Off topic: I have checked this with the bug I have with https://github.com/mdemarne/scalameta/blob/f0aeadf75fe49a5f009160b1a06ce9e3620ff60f/tests/src/test/scala/tql/TQLOriginSuite.scala, and it's going through there, so I would think that somehow something is not done properly in copying a tree, i.e. https://github.com/mdemarne/scalameta/blob/master/scalameta/src/main/scala/scala/meta/internal/tql/TraverserHelper.scala#L25 is sometimes true; the tree has changed (even with collect { ... } which copies the tree as is in https://github.com/mdemarne/scalameta/blob/master/scalameta/src/main/scala/scala/meta/tql/Combinators.scala#L174).

scalameta/src/main/scala/scala/meta/Origin.scala
```diff
@@ -8,57 +8,31 @@ import org.scalameta.invariants._
...
+  def tokens: Tokens
 }
 
 object Origin {
-  @leaf object None extends Origin {
-    val input = Input.None
-    val dialect = scala.meta.dialects.Scala211
-    val start = 0
-    val end = -1
-    val startLine = 0
```
mdemarne: XXXXX code is using this for their engine.

scalameta/src/main/scala/scala/meta/syntactic/Position.scala
```diff
@@ -0,0 +1,63 @@
...
+  @leaf class Virtual(input: Input.Virtual) extends Position {
+    lazy val start = Point.Virtual(input)
+    lazy val end = Point.Virtual(input)
+  }
+}
+
+@root trait Point {
+  def input: Input
+  def offset: Int
+  def line: Int
```
mdemarne: Oh ok, great, I could use that for XXXXX! Forget my last comment.

mdemarne:
LGTM!

PR #133, opened 2015-05-07
Add sequence deconstruction to `Tokens`

Duhemm:
I think it makes sense to allow deconstruction of `Tokens` just like other `Seq`s (after all, it is a `Seq`).

```scala
tokens match { case Tokens(first, second, rest @ _*) => ??? }
```

### DIFF
--- a/scalameta/src/main/scala/scala/meta/syntactic/Tokens.scala
+++ b/scalameta/src/main/scala/scala/meta/syntactic/Tokens.scala
@@ -33,6 +33,7 @@ object Tokens {
     if ((tokens: Seq[Token]).isInstanceOf[Tokens]) tokens.asInstanceOf[Tokens]
     else new Tokens(tokens: _*)
   }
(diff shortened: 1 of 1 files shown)


scalameta/src/main/scala/scala/meta/syntactic/Tokens.scala
```diff
@@ -33,6 +33,7 @@ object Tokens {
     if ((tokens: Seq[Token]).isInstanceOf[Tokens]) tokens.asInstanceOf[Tokens]
     else new Tokens(tokens: _*)
   }
+  def unapplySeq(tokens: Tokens): Option[Seq[Token]] = Some(tokens)
```
xeno-by: Change this to `Some[Seq[Token]]`, and I'll be merging. Nice catch!

Duhemm: Updated

xeno-by:
Thanks!

xeno-by:
Rebuilt the world and published the results.

PR #138, opened 2015-05-27
miscellaneous fixed for @mdemarne

xeno-by:


scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -1070,17 +1070,40 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
...
+      case token: Literal.Long =>
+        val value = if (isNegated) -token.value else token.value
+        if (value > Long.MaxValue) syntaxError("integer number too large", at = token)
+        else if (value < Long.MinValue) syntaxError("integer number too small", at = token)
+        else Lit.Long(value.toLong)
+      case token: Literal.Float =>
+        val value = if (isNegated) -token.value else token.value
+        if (value > Float.MaxValue) syntaxError("floating point number too large", at = token)
+        else if (value < Float.MinValue) syntaxError("floating point number too small", at = token)
+        else Lit.Float(value.toLong)
```
mdemarne: Hey, I think there is a small error there:

```
scala> "23.45F".parse[Term].show[Raw]
res5: String = Lit.Float(23.0f)
```
I guess it should be `value.toFloat`, and should be changed  for Double (below) as well.

xeno-by: Fixed in https://github.com/scalameta/scalameta/pull/139. I shouldnt've skipped the tests, regardless of how obvious they seemed.

PR #140, opened 2015-05-30
Token quasiquote

Duhemm:

PR #141, opened 2015-06-01
Inferring Synthetic tokens

mdemarne:
`InferTokens` now:
- Produces the same amount or fewer parentheses as show[Code]
- Replaces show[Code]
- Is able to prettyprint, parse, prettyprint again the entire codebase of `scalameta` without error
- Is able to infer top-level comments coming from Source (see `InferPartialSuite.scala`)
- Is able to keep previous layout informations  (see `InferPartialSuite.scala`)


mdemarne:
A few notes:
- I did not use `Lift`, since the logic of `Tree.tks` is not simply returning tokens, but also inferring parentheses, which is a logic present in this source file as well. I hope this is fine, otherwise I guess it could be extracted.
- I removed the test comparing parentheses from show[Code], since here it is replaced by `inferToken`
- There is one tiny case with `(a + b + c) && ...` which does not put back parentheses. Syntactically speaking it is valid, since the precedence of `+` is stronger than the one of `&&`, but it's just a bit less nice to read.

project/build.scala
```diff
@@ -146,7 +146,8 @@ object build extends Build {
     publishableSettings: _*
   ) settings (
     libraryDependencies <+= (scalaVersion)("org.scala-lang" % "scala-reflect" % _ % "provided"),
-    libraryDependencies <+= (scalaVersion)("org.scala-lang" % "scala-compiler" % _ % "provided")
+    libraryDependencies <+= (scalaVersion)("org.scala-lang" % "scala-compiler" % _ % "provided"),
+    libraryDependencies += "plugin" %% "plugin" % "0.1.0-SNAPSHOT" // TODO: change/remove once Martin's work is integrated to SM.
```
mdemarne: Note: this will have to be remove prior to the pull request to be accepted (Travis will also not be happy). #140 should be accepted first.

PR #142, opened 2015-06-01
resubmission of #140

xeno-by:
/cc @Duhemm @mdemarne

### DIFF
--- a/scalameta/src/main/scala/scala/meta/syntactic/Token.scala
+++ b/scalameta/src/main/scala/scala/meta/syntactic/Token.scala
@@ -157,9 +157,8 @@ trait TokenLiftables extends adt.Liftables {
   val c: scala.reflect.macros.blackbox.Context
   override lazy val u: c.universe.type = c.universe
 
-  private val XtensionQuasiquoteTerm = "shadow scala.meta quasiquotes"
-
   import c.universe._
+  private val XtensionQuasiquoteTerm = "shadow scala.meta quasiquotes"
 
   implicit lazy val liftContent: Liftable[Content] = Liftable[Content] { content =>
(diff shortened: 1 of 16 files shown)


xeno-by:
I'm merging this asap, because we need to get @mdemarne's stuff in, and it depends on this PR.

@Duhemm, could you review my changes and comment on them? Also, it would be great if you could address https://github.com/scalameta/scalameta/pull/142. This is a user-facing API, so unlike in the case of things for internal consumption we probably can't get away w/o hygiene.

@mdemarne, could you rebase on top of this?

scalameta/src/main/scala/scala/meta/syntactic/Token.scala
```diff
@@ -153,3 +152,88 @@ object Token {
...
+//
+// Due to an unfortunate limitation of knownDirectSubclasses,
+// all macro applications that depend on that API
+// must come after the classes that they call the API on.
+//
+// That's why if we put TokenLiftables elsewhere, we might run into troubles
+// depending on how the file and its enclosing directories are called.
+// To combat that, we have TokenLiftables right here, guaranteeing that there won't be problems
+// if someone wants to refactor/rename something later.
+trait TokenLiftables extends tokens.Liftables {
```
Duhemm: If I understand correctly, the redirection mechanism in `customMatcher` will see that we're trying to lift `Unquote` and use directly `liftUnquote` instead of generating a new one which would require a Liftable for `Any`? Nice!

xeno-by: Yes, that's right. Ideally, we shouldn't have this `customMatcher` at all and the materializer should simply account for the implicits in scope. In the future, I'll get to implementing that.

PR #170, opened 2015-06-09
resubmission of https://github.com/scalameta/scalameta/pull/141

xeno-by:


xeno-by:
Supersedes https://github.com/scalameta/scalameta/pull/141, but we still need to review that one. /cc @mdemarne

PR #176, opened 2015-06-20
fixes #175

### LINKED ISSUE #175: Weird warning on `val q"$lit" = q"42"`
```
import scala.meta._

object Test {
    val q"$lit" = q"42"
  }
}
```

```
12:19 ~/Projects/core/sandbox (master)$ s
Test.scala:8: warning: patterns after a variable pattern cannot match (SLS 8.1.1)
    val q"$lit" = q"42"
        ^
Test.scala:8: warning: unreachable code due to variable pattern 'quasiquote$macro$2$hole$0' on line 8
    val q"$lit" = q"42"
        ^
Test.scala:8: warning: unreachable code
    val q"$lit" = q"42"
        ^
three warnings found
```

xeno-by:
Apparently the pattern matcher is unhappy if we generate code that looks like:
`... match { case <freshname> => Some(...); case _ => None }`, therefore I had to tune
the reification macro to avoid such codegen in rare edge cases.


Dveim:
LGTM

PR #202, opened 2015-07-26
Remaining tests from spec#Expressions

Dveim:


scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -623,12 +635,6 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
   @inline final def commaSeparated[T <: Tree : AstMetadata](part: => T): List[T] =
     tokenSeparated[`,`, T](sepFirst = false, part)
```
xeno-by: I wonder why there was a necessity to change `caseSeparated`, but there wasn't a need to change `commaSeparated`. Both seem to need support for ellipses.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -1715,6 +1726,10 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
 
   def enumerator(isFirst: Boolean, allowNestedIf: Boolean = true): List[Enumerator] =
     if (token.is[`if`] && !isFirst) autoPos(Enumerator.Guard(guard().get)) :: Nil
+    else if (token.is[Ellipsis]) {
+      next()
+      constructR1Quasi[Enumerator.Generator](expr())
+    }
```
xeno-by: Does this handle things like `for (..$blah; $blah) { ... }`?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -562,6 +562,18 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
...
+    val tree = {  
+      val q = ctor match {
+        case q: Quasi if q.rank == 0 => q
+        case _ => syntaxError("impossible to construct rank-1 quasiquote", at = token)
+      }
+      atPos(q, q)(implicitly[AstMetadata[T]].quasi(q.rank, q.tree))
+    }
+    implicitly[AstMetadata[T]].quasi(1, tree) :: Nil
+  }
+
```
xeno-by: Why can't `ctor` be typed as `=> T`?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -562,6 +562,18 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
     }
   }
 
+  // corresponding quasiquote (rank 1) for expressions like `{..$smth}`
+  def constructR1Quasi[T <: Tree : AstMetadata](ctor: => Tree): List[T] = {
```
xeno-by: I think I'd prefer this method merged into `ellipsis`, and `ellipsis` gaining an additional flag `skippingEllipsis = true` or something, if you see what I mean.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: When I used `constructR1Quasi`, both methods `caseSeparated` and `readAnnots` got 0 usages (as Intellij says), so I removed them. But `commaSeparated` had at least 10 usages, and was no real need to rewrite them all.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: It does not. However, is there any case where such functionality would be useful? `blah` = `blahs.last`. 
I think it's not so hard to add it, if necessary.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: Consider `cases ++= constructR1Quasi[Case](expr())`. Here `expr` is `Term`, so is `T`. But `Case` is not `Term`, and that need to be explicitly stated.

Here `expr()` is a way to get ellipsis' variable name (e.g. `..$smth` => `smth`), and `Case` is a type for final Quasi.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: That was my first approximation of this method. But I didn't get any fruitful results, because additional boolean parameter brings too much noise, and `ellipsis` becomes almost unreadable.

For example, consider `val q"for (..$enumerators) yield $expr" = q"for (a <- as; b <- bs) yield foo(x, y)"`.
Before `constructR1Quasi`/`ellipsis` current token is `$enumerators`, and that is not applicable to `ellipsis` and would require some step back in iterator. Or issue with `T/Tree` below.

So, `constructR1Quasi` can be viewed as more suitable method for several situations.

xeno-by: Why's the current token positioned at `$enumerators`? From what I can see, you're explicitly rewinding the token stream before calling `constructR1Quasi`, but if you didn't do that, then you wouldn't have to special-case at least this difference with `ellipsis`.

I understand that this is done to skip `case` and `@` tokens in the case of pattern matching cases and annotations, but how about we do this differently? `ellipsis` could gain additional parameter that says which tokens it needs to skip, so that we rewrite:

```
if (token.is[Ellipsis]) {
  nextTwice()
  annots ++= constructR1Quasi[Mod.Annot](expr())
} else {
``` 

As something like:

```
if (token.is[Ellipsis]) {
  annots += ellipsis(1, skip = `@`)(expr())
} else {
```

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: So the output of the `constructR1Quasi...` call here is going to be `Case.Quasi(1, Term.Quasi(0, ...))`? Isn't this misleading? Shouldn't the second argument of an 1+ rank quasi be a tree with the same type as the enclosing quasi?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: Can `tokenSeparated` be changed to accommodate the suggested structure of `constructR1Quasi`? It seems that it could be designed to tell the underlying `ellipsis` to skip the provided tokens if necessary. As a side-effect, we'll also get code reuse as a bonus.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -1661,19 +1667,14 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
...
-        })
-    }
+    Case(pattern().require[Pat], guard(), {
+      accept[`=>`]
+      autoPos(blockStatSeq() match {
+        case List(q: impl.Quasi) => Term.Block(List(q))
+        case List(blk: Term.Block) => blk
+        case stats => Term.Block(stats)
+      })
+    })
```
xeno-by: That's a good change. I was looking forward to it since I put this change in place.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -2912,7 +2943,7 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
    */
   def blockStatSeq(): List[Stat] = {
     val stats = new ListBuffer[Stat]
-    while (!token.is[StatSeqEnd] && !token.is[CaseDefEnd]) {
+    while (!token.is[StatSeqEnd] && !token.is[CaseDefEnd] && !ahead(token.is[`case`])) {
```
xeno-by: Why's this necessary?

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
```diff
@@ -208,10 +209,18 @@ class QuasiquoteSuite extends FunSuite {
     assert(params(1).show[Syntax] === "y: String")
   }
 
-//  test("q\"{ ..$stats }\"") {
-//    val stats = List(q"val x = 1", q"val y = 2")
-//    q"{ ..$stats }"
-//  }
+  test("1 q\"{ ..$stats }\"") {
+    val stats = List(q"val x = 1", q"val y = 2")
+    assert(q"{ ..$stats }".show[Syntax].replace("\n","") === "{  val x = 1  val y = 2}")
```
xeno-by: Why replace?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: To handle situations like `val q"$expr match { case bar => baz; ..case $casez; case q => w}" = q"foo match { case bar => baz; case _ => foo ; case q => w }"`.

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
Dveim: Updated with margins, like in ErrorSuite.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: I don't know how to pass `@` as a parameter, getting compile errors. So, with
`def ellipsis[T <: Tree : AstMetadata](rank: Int, body: => Tree, skip: Any = 42): T = autoPos { ...`
````annots += ellipsis(1, expr(), skip = `@`)```
,I get weird `not found: value <none>
[error]         annots += ellipsis(1, expr(), skip = 2)
[error]                           ^
`

Issue #203, opened 2015-07-29
Missing value when unquoting enumerators


Dveim:
`val q"for (a <- as; if $cond; ..$enumz) bar" = q"for (a <- as; if foo; b <- bs) bar"`
produces following error: `not found: value quasiquote$macro$20$hole$0`

Here is full output with -Dquasiquote.debug:

```
scala> val q"for (a <- as; if $cond; ..$enumz) bar" = q"for (a <- as; if foo; b <- bs) bar" 

Adhoc(List(BOF (0..0), for (0..3),   (3..4), ( (4..5), a (5..6),   (6..7), <- (7..9),   (9..10), as (10..12), ; (12..13),   (13..14), if (14..16),   (16..17), foo (17..20), ; (20..21),   (21..22), b (22..23),   (23..24), <- (24..26),   (26..27), bs (27..29), ) (29..30),   (30..31), bar (31..34), EOF (34..34)))
Adhoc(List(BOF (0..0), for (0..3),   (3..4), ( (4..5), a (5..6),   (6..7), <- (7..9),   (9..10), as (10..12), ; (12..13),   (13..14), if (14..16),   (16..17), foo (17..20), ; (20..21),   (21..22), b (22..23),   (23..24), <- (24..26),   (26..27), bs (27..29), ) (29..30),   (30..31), bar (31..34), EOF (34..34)))
for (a <- as; if foo; b <- bs) bar
Term.For(List(Enumerator.Generator(Pat.Var.Term(Term.Name("a")), Term.Name("as")), Enumerator.Guard(Term.Name("foo")), Enumerator.Generator(Pat.Var.Term(Term.Name("b")), Term.Name("bs"))), Term.Name("bar"))
[...]

Issue #199, opened 2015-07-23
Correct processing of optional terms


Dveim:
Operations like `val q"try $a finally $b" = q"try x finally y"` should return Option-wrapped types, where spec says. So, in example above, `b` should be `Some[scala.meta.Term]` instead of `scala.meta.Term`.

This issue affects pattern-matching unquoting with non-existent parts in right right. For example, in `val q"try $a finally $b" = q"try x"` `b` should return `None`, whereas at the moment MatchError occurs.

The bug probably is inside reificator.

PR #212, opened 2015-08-03
Support for unquotings in Types and more

Dveim:


scalameta/src/main/scala/scala/meta/Trees.scala
```diff
@@ -170,7 +175,7 @@ package scala.meta.internal.ast {
     @ast class While(expr: Term, body: Term) extends Term
     @ast class Do(body: Term, expr: Term) extends Term
     @ast class For(enums: Seq[Enumerator] @nonEmpty, body: Term) extends Term with Scope {
-      require(enums.head.isInstanceOf[Enumerator.Generator])
+      require(enums.head.isInstanceOf[Enumerator.Generator] || enums.head.isInstanceOf[Enumerator.Generator.Quasi])
```
xeno-by: Do we really need this? Isn't `Enumerator.Generator.Quasi` an `Enumerator.Generator`?

scalameta/src/main/scala/scala/meta/Trees.scala
```diff
@@ -195,7 +200,7 @@ package scala.meta.internal.ast {
       // require(keywords.contains(value) ==> isBackquoted)
     }
     @ast class Select(qual: Term.Ref, name: Type.Name) extends Type.Ref with Pat.Type.Ref {
-      require(qual.isPath || qual.isInstanceOf[Term.Super])
+      require(qual.isPath || qual.isInstanceOf[Term.Super] || qual.isInstanceOf[Term.Quasi])
```
xeno-by: Why do we allow `qual` to be `Term.Quasi`? Why not make it `become` a `Term.Ref.Quasi`?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -562,16 +563,22 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
...
+          implicitly[AstMetadata[T]].quasi(0, unquote.tree)
+        } else ahead {
+          implicitly[AstMetadata[T]].quasi(0, unquote.tree)
+        }
       case _ =>
         unreachable(debug(token))
     }
   }
 
+  def unquote[T <: Tree : AstMetadata]: T = unquote[T](advance = true) // to write `unquote[T]` without round braces
```
xeno-by: Round braces are called parentheses :)

scalameta/src/main/scala/scala/meta/Trees.scala
Dveim: It is, removed.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -660,9 +669,12 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
     }
   }
 
-  def makeTuplePatParens(bodyf: => List[Pat.Arg]): Pat = autoPos {
-    val body = inParens(if (token.is[`)`]) Nil else bodyf.map(_.require[Pat]))
-    makeTuple[Pat](body, () => Lit.Unit(), Pat.Tuple(_))
+  def makeTuplePatParens(): Pat = {
```
xeno-by: Why does this function no longer take an argument?

xeno-by: Also, is it going to work for `p"()"`?

Dveim: I moved it inside `makeTuplePatParens`, since there was only one explicit usage.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -1961,7 +2005,7 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
       case _: Xml.Start =>
         xmlPat()
       case _: `(` =>
-        makeTuplePatParens(noSeq.patterns())
+        makeTuplePatParens()
```
xeno-by: This change is, I guess, the reason for the change in the signature of `makeTuplePatParens`. So I'm just trying to understand the motivation. The thing is that in the current shape `makeTuplePatParens` is different from other make-tuple functions, so it'd be nice to have a good reason for such a discrepancy.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -969,7 +998,9 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
       if (advance) next()
       res
     case token: Unquote =>
-      unquote[T]
+      unquote[T](advance)
```
xeno-by: Isn't this equivalent to `if (advance) unquote[T] else ahead(unquote[T])`? If yes, then why have a special version of `unquote` and `ellipsis`, if we can localize this special case to a single function?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -1682,10 +1708,23 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
...
-    val cases = caseSeparated { caseClause() }
+    val cases = new ListBuffer[Case]
+    while (token.is[CaseIntro] || token.is[Ellipsis]) {
+      if (token.is[Ellipsis]) {
+        cases += ellipsis(1, unquote[Case], accept[`case`])
+        if (token.is[StatSep]) next()
+      } else if (token.is[`case`] && ahead(token.is[Unquote])) {
+        next()
+        cases += unquote[Case]
+        if (token.is[StatSep]) next()
```
xeno-by: Why is this `StatSep`-based skipping necessary here?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -1682,10 +1708,23 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
    *  }}}
    */
   def caseClauses(): List[Case] = {
-    val cases = caseSeparated { caseClause() }
+    val cases = new ListBuffer[Case]
+    while (token.is[CaseIntro] || token.is[Ellipsis]) {
```
xeno-by: Where did this `CaseIntro` come from?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: About method `name`: I thought that would be strange, if `advance` didn't have influence on both cases, `Unquote` and `Ellipsis`.
And about method `unquote`: since I have had to add that `advance: Boolean = true`, compiler did not allow simple `unquote[T]` elsewhere, only with parentheses. So I added special version of `unquote`, to preserve old syntax.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -1661,19 +1692,14 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
 
   def caseClause(): Case = atPos(in.prevTokenPos, auto) {
     require(token.isNot[`case`] && debug(token))
-    token match {
-      case token: Unquote if token.prev.is[Ellipsis] =>
-        unquote[Case]
```
xeno-by: Would it be possible to retain this pattern match clause in `caseClause` (more precisely, changing it to `case token: Unquote if token.prev.is[Case]`) and then migrating `caseClauses` back to `caseSeparated`?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: To allow `;` after cases. 
`..case $as; ..case $bs`

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: It is used to handle situations like
`val q"$expr match { case bar => baz; ..case $casez; case q => w}" = q"foo match { case bar => baz; case _ => foo ; case q => w }"`

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -2911,7 +2966,7 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
    */
   def blockStatSeq(): List[Stat] = {
     val stats = new ListBuffer[Stat]
-    while (!token.is[StatSeqEnd] && !token.is[CaseDefEnd]) {
+    while (!token.is[StatSeqEnd] && !token.is[CaseDefEnd] && !ahead(token.is[`case`])) {
```
xeno-by: So `{ case class ... }` isn't going to work?

tests/src/test/scala/quasiquotes/ErrorSuite.scala
```diff
@@ -297,9 +297,11 @@ class ErrorSuite extends FunSuite {
       val tpes = List(q"T")
       q"expr: ..$tpes"
     """) === """
-      |<macro>:5: identifier expected but ellipsis found
+      |<macro>:5: type mismatch when unquoting;
+      | found   : List[scala.meta.Term.Name]
+      | required: scala.collection.immutable.Seq[scala.meta.Type]
```
xeno-by: Hmm. Why did this error message change?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: Sorry, cannot understand this example. Why it should?

tests/src/test/scala/quasiquotes/ErrorSuite.scala
xeno-by: Also, what's it mean really? What if the user passed a sequence of types there? Is it going to compiler?

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
```diff
@@ -192,9 +183,19 @@ class QuasiquoteSuite extends FunSuite {
     assert(tpe.show[Syntax] === "Double")
   }
 
-//  test("q\"$expr: ..@$expr\"") {
-//    val q"$a: ..@$b" = q"foo: @bar @baz"
-//  }
+  test("1 q\"$expr: ..@annots\"") {
```
xeno-by: Shouldn't there be a dollar in the name of the test?

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
```diff
@@ -192,9 +183,19 @@ class QuasiquoteSuite extends FunSuite {
...
+    val q"$exprr: @q ..@$annotz @$ar" = q"foo: @q @w @e @r"
+    assert(exprr.show[Syntax] === "foo")
+    assert(annotz.toString === "List(@w, @e)")
+    assert(annotz(0).show[Syntax] === "@w")
+    assert(annotz(1).show[Syntax] === "@e")
+    assert(ar.show[Syntax] === "@r")
+  }
+
+  test("2 q\"$expr: ..@annots\"") {
```
xeno-by: And here as well.

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
```diff
@@ -220,20 +235,46 @@ class QuasiquoteSuite extends FunSuite {
...
-//  }
+  test("3 q\"$expr match { ..case $cases }\"") {
+    val q"$expr match { ..case $casez }" = q"foo match { case bar => baz; case _ => foo }"
+    assert(expr.show[Syntax] === "foo")
+    assert(casez.toString === "List(case bar => baz, case _ => foo)")
+    assert(casez(0).show[Syntax] === "case bar => baz")
+    assert(casez(1).show[Syntax] === "case _ => foo")
+  }
+
+  // todo change to expropt (and test it) after issue #199 resolved
```
xeno-by: Could you change the comment to say `TODO: ...` to be in line with other todos in the repo?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: But `case x => y; case q => w` is not case-separated. There is one extra `case`.

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
```diff
@@ -276,16 +313,65 @@ class QuasiquoteSuite extends FunSuite {
...
+    val b = enumerator"b <- bs"
+    val ab = List(a,b)
+    assert(q"for (..$ab) foo".show[Syntax] === "for (a <- as; b <- bs) foo")
+  }
 
-//  test("q\"for (..$enumerators) yield $expr\"") {
-//    val q"for (..$enumerators) yield $expr" =  q"for (x <- xs; y <- ys) yield foo(x, y)" // fixme test is broken, so even does not compile
+//  test("3 q\"for (..$enumerators) $expr\"") {
+//    val q"for (a <- as; if $cond; ..$enums) bar" = q"for (a <- as; if foo; b <- bs) bar" // fixme does not compile
 //  }
```
xeno-by: Could you link https://github.com/scalameta/scalameta/issues/203 in a comment to this test?

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
```diff
@@ -276,16 +313,65 @@ class QuasiquoteSuite extends FunSuite {
...
+  }
+
+  test("3 q\"new { ..$stat } with ..$exprs { $param => ..$stats }\"") {
+    val q"new X with T { $self => def m = 42}" = q"new X with T { def m = 42 }"
+    assert(self.show[Structure] === "Term.Param(Nil, Name.Anonymous(), None, None)")
+  }
+
+//  test("4 q\"new { ..$stat } with ..$exprs { $param => ..$stats }\"") {
+//    val q"new {..$stats; val b = 4} with $a {$selff => ..$statz}" = q"new {val a = 2; val b = 4}" // todo fails to compile, uncomment after issue #199 resolved
 //  }
```
xeno-by: Why's this test disabled?

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
```diff
@@ -306,4 +392,286 @@ class QuasiquoteSuite extends FunSuite {
...
+
+  test("arg\"$expr: _*\"") {
+    val expr = q"foo"
+    assert(arg"$expr: _*".show[Syntax] === "foo: _*")
+  }
+
+  test("arg\"$expr\"") {
+    val expr = q"foo"
+    assert(q"$expr".show[Syntax] === "foo")
+  }
```
xeno-by: Why's it say `q"$expr"`?

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
```diff
@@ -306,4 +392,286 @@ class QuasiquoteSuite extends FunSuite {
...
+    val tname = t"Y"
+    val tpe2 = t"Z"
+    assert(t"$tpe1 $tname $tpe2".show[Syntax] === "X Y Z")
+  }
+
+  test("t\"(..$atpes) => $tpe\"") {
+    val atpes = List(t"X", t"Y")
+    val tpe = t"Z"
+    assert(t"(..$atpes) => $tpe".show[Syntax] === "(X, Y) => Z")
+  }
```
xeno-by: Could we have a test when `..$atpes` are actually `Type.Arg`s here?

tests/src/test/scala/quasiquotes/ErrorSuite.scala
Dveim: I don't know exactly, why this changed, but new error message seems to be informative.
For list of types user'd get `found   : scala.collection.immutable.Seq[scala.meta.internal.ast.Type]
 required: scala.meta.internal.ast.Type
       q"expr: ..$tpes"
                 ^
`

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
```diff
@@ -306,4 +392,286 @@ class QuasiquoteSuite extends FunSuite {
...
+    assert(t"$tpe forSome { ..$stats }".show[Syntax] === "X forSome { val a: A; val b: B }")
+  }
+
+  test("t\"$tpe ..@$annots\"") {
+    val tpe = t"X"
+    val annots = List(mod"@a", mod"@b")
+    assert(t"$tpe ..@$annots".show[Syntax] === "X @a @b")
+  }
+
+  test("t\"_ >: $tpeopt <: $tpeopt\"") {
```
xeno-by: Why does it say `opt` if you're splicing regular types?

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
```diff
@@ -306,4 +392,286 @@ class QuasiquoteSuite extends FunSuite {
...
+    assert(tparams(0).show[Syntax] === "X")
+    assert(tparams(1).show[Syntax] === "Y")
+    assert(tpe.show[Syntax] === "(X, Int) => Y")
+  }
+
+  test("2 t\"[..$tparams] => $tpe\"") {
+    val tparams = List(tparam"X", tparam"Y")
+    val tpe = t"Z"
+    assert(t"[..$tparams] => $tpe".show[Syntax] === "[X, Y] => Z")
+  }
```
xeno-by: What is the principle according to which some cases get construction/deconstruction tests, and some just get a construction test? :)

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
Dveim: Changed. And same error in one of the last tests in the file.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: But isn't `!ahead(token.is[`case`])` going to be false, which will prompt an early exit from the loop?

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
Dveim: Changed.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: There's the `sepFirst` flag that allows to accommodate this.

scalameta/src/main/scala/scala/meta/Trees.scala
Dveim: Changed.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: I'm trying to push unification, because then we'll have less places to change when time will come to make adjustments to our quasiquotes (e.g. when complex ellipses become a thing).

tests/src/test/scala/quasiquotes/ErrorSuite.scala
xeno-by: Okay, what if `tpes` was not `List(q"T")`, but `List(t"T")`. Would it compile then?

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
Dveim: Done.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: Yeah, sure the suggested change would have to be made to both branches. Then the `unquote` method would go back to its old form - without the parameter and without multiple overloads.

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
Dveim: Because it does not compile, and waits for issue to be resolved.

xeno-by: I see. Would be cool if you could mention that.

xeno-by: Ah I see. Could you move the explaining comment to be above the test?

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
Dveim: Typo, fixed.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: What if there are multiple statseps here? Also, why statsep and not `;` directly?

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
Dveim: Here they are `Type.Name`, isn't it `Type.Arg`? They both have `t` StringContext for creating.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: Why `CaseIntro` and not `case` like it was before?

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
xeno-by: Yeah, sure, I mean could we have `Type.Arg`s that aren't `Type`, so that we really test that function types can accept type args as their arguments?

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
Dveim: Probably I should add one more comment
`// TODO change to expropt (and test it) after issue #199 resolved`

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: If unification doesn't work here, then sure no problem - I'm already almost okay with `caseSeparated` going away.

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
Dveim: I check all cases for both type of tests in repl. For those who have (or had before) some issues, I explicitly add test.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: Changed.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: In old form, `unquote` always drives iterator forward. However, sometimes that is not desired behaviour, so I added extra parameter with default value, for the future cases.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: 1) I thought about that, but it would require some extra changes in code (for cases when scala supports multiple `;`, but this parser doesn't) just for nothing. `;;;;;` is useless action.
2) Because new line might happen inside cases.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: Same reason as with `StatSep` vs `;`. `CaseIntro` is most general form of `case` that can be applied here.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: I'll try investigate in moving `tokenSeparated` usage back.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: I still don't understand. `q{case class A}` works.

tests/src/test/scala/quasiquotes/ErrorSuite.scala
Dveim: For list of types it says that wants a single type. Agree, so previous message is lying, and need to be changed.

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
Dveim: Added, typo.

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
Dveim: Done.

tests/src/test/scala/quasiquotes/QuasiquoteSuite.scala
Dveim: Changed test.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: 1) I agree that this solution looks good enough, but what about consistency with the grammar spec? Scala supports stuff like `foo match { case x => y;;;;;; case z => t;;;; }`, so why not make it support the same in quasiquotes?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: Oh it does. Then it's fine.

tests/src/test/scala/quasiquotes/ErrorSuite.scala
xeno-by: I think the new message comes from the fact that you changed `name` to accept ellipses in addition to accepting unquotes. Why was that necessary?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: I suggest we discuss this in voice. Ping me when you're up for that.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: One of tests didn't pass, and analysis showed that issue can be resolved by adding that checking.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: Changed `if` to `while`.

xeno-by:
Allright, even though I'm not completely satisfied with `caseClauses`, let's leave that one for future work. LGTM.

Issue #216, opened 2015-08-11
inconsistent construction of Pat.Var.XXX nodes


Dveim:
We need easy ways to create `Pat.Var.Term` and `Pat.Var.Type`, and the spec promises it to us in the form of `p"name"` and `pt"name"`. 

However, if we dig deeper, we'll find that even though `p"x"` is a `Pat.Var.Term`, `p"X"` is a `Term.Name` and both `pt"x"` and `pt"X"` are `Type.Name`. This is a direct consequence of how Scala's parser works: `Pat.Var.Term` with capitalized name is only allowed in the lhs of vals and vars, whereas `Pat.Var.Type` is only allowed in type arguments.

xeno-by:
Also need to support `pt"_"` that currently gives an error.

PR #228, opened 2015-08-20
Remaining cases from specification

Dveim:


scalameta/src/main/scala/scala/meta/Trees.scala
```diff
@@ -71,6 +71,7 @@ package scala.meta {
   }
 
   object Ctor {
+    @branch trait Call extends Term
     @branch trait Ref extends Term.Ref
```
xeno-by: `extends Term.Ref with Ctor.Call`

scalameta/src/main/scala/scala/meta/Trees.scala
```diff
@@ -460,9 +483,9 @@ package scala.meta.internal.ast {
                          body: Term) extends Ctor with Stat {
       require(body.isCtorBody)
     }
-    @branch trait Ref extends api.Ctor.Ref with impl.Term.Ref
+    @branch trait Ref extends api.Ctor.Ref with impl.Term.Ref with impl.Ctor.Call
     val Name = Ref.Name
-    type Name = Ref.Name
+    type Name = Ref.Name with Ctor.Call
```
xeno-by: This is unnecessary

scalameta/src/main/scala/scala/meta/internal/parsers/Helpers.scala
```diff
@@ -75,11 +76,13 @@ private[meta] object Helpers {
     }
     def isCtorBody: Boolean = {
       def isSuperCall(tree: Tree): Boolean = tree match {
+        case _: Term.This => true
```
xeno-by: Why's this necessary?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -50,7 +50,7 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
   def parseStat(): Stat = parseRule(parser => parser.statSeq[Stat] {
     case token if token.is[`import`] => importStmt()
     case token if token.is[`package `] => packageOrPackageObjectDef()
-    case token if token.is[TokenClass.DefIntro] => nonLocalDefOrDcl()
+    case token if token.is[TokenClass.DefIntro] || token.is[Ellipsis] => nonLocalDefOrDcl()
```
xeno-by: Why can't this check be merged into `is[DefIntro]`?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -634,8 +644,8 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
     ts.toList
   }
 
-  @inline final def commaSeparated[T <: Tree : AstMetadata](part: => T): List[T] =
-    tokenSeparated[`,`, T](sepFirst = false, part)
+  @inline final def commaSeparated[T <: Tree : AstMetadata](part: => T, useUnquoteForEllipsis: Boolean = false): List[T] =
+    tokenSeparated[`,`, T](sepFirst = false, part, useUnquoteForEllipsis)
```
xeno-by: Why's this parameter necessary here?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -976,15 +988,28 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
...
-        case Type.Function(params, res) => atPos(tpe, tpe)(Pat.Type.Function(params.map(p => loop(p.require[Type])), loop(res)))
+        case Type.Function(params, res) => atPos(tpe, tpe)(Pat.Type.Function(params.map {
+          case q @ Type.Arg.Quasi(1, Type.Arg.Quasi(0, _)) => q.become[Pat.Type.Quasi]
+          case p => loop(p.require[Type])
+        }, loop(res)))
         case Type.Tuple(elements) => atPos(tpe, tpe)(Pat.Type.Tuple(elements.map(loop)))
         case Type.Compound(tpes, refinement) => atPos(tpe, tpe)(Pat.Type.Compound(tpes.map(loop), refinement))
         case Type.Existential(underlying, quants) => atPos(tpe, tpe)(Pat.Type.Existential(loop(underlying), quants))
         case Type.Annotate(underlying, annots) => atPos(tpe, tpe)(Pat.Type.Annotate(loop(underlying), annots))
+        case Type.Placeholder(Type.Bounds(None, None)) => convert(tpe)
```
xeno-by: Why's this clause necessary?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -976,15 +988,28 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
...
+      val t: Type = {
+        val t = if (token.is[`(`]) tupleInfixType() else compoundType()
+        def mkOp(t1: Type) = atPos(t, t1)(Type.ApplyInfix(t, typeName(), t1))
+        token match {
+          case _: `forSome` => next(); atPos(t, t)(Type.Existential(t, existentialStats()))
+          case _: Unquote | _: Ident if !isRawBar => infixTypeRest(mkOp(compoundType()), InfixMode.LeftOp)
+          case _ => t
+        }
+      }
+      loop(t)
```
xeno-by: Could you recall the motivation for this change?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -1966,11 +1999,12 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
       // simple diagnostics for this entry point
       simplePattern(token => syntaxError("illegal start of simple pattern", at = token))
     def simplePattern(onError: Token => Nothing): Pat = autoPos(token match {
-      case _: Ident | _: `this` =>
-        val isBackquoted = token.code.endsWith("`")
+      case _: Ident | _: `this` | _: Unquote =>
+        val isBackquoted = parser.isBackquoted
```
xeno-by: Why's there a need to precache this?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -1966,11 +1999,12 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
...
         val sid = stableId()
         val isVarPattern = sid match {
-          case Term.Name(value) => !isBackquoted && value.head.isLower && value.head.isLetter
+          case _: Term.Name.Quasi => false
+          case Term.Name(value) => !isBackquoted && value.head.isLetter && value.head.isLower
```
xeno-by: What was the point in swapping the conditions around? :D

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -2007,8 +2042,6 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
           case p => p.require[Pat]
         }
         makeTuple[Pat](patterns, () => Lit.Unit(), Pat.Tuple(_))
-      case _: Unquote =>
-        unquote[Pat]
```
xeno-by: Okay, so what did this change buy us?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -2111,14 +2149,16 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
...
         if (mod.hasAccessBoundary) mods.filter(_.hasAccessBoundary).foreach(mod => syntaxError("duplicate access qualifier", at = mod))
       }
       validate()
       mods :+ mod
     }
+    def continueLoop = ahead(token.is[`:`] || token.is[`=`] || token.is[EOF])
     def loop(mods: List[Mod]): List[Mod] = token match {
-      case _: Modifier         => loop(appendMod(mods, modifier()))
+      case _: Unquote => if (continueLoop) mods else loop(appendMod(mods, modifier()))
+      case _: Modifier | _: Ellipsis => loop(appendMod(mods, modifier()))
```
xeno-by: Can we build turn this into a tokenclass instead?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -2111,14 +2149,16 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
...
       def validate() = {
         if (isLocal && !mod.tokens.head.is[LocalModifier]) syntaxError("illegal modifier for a local definition", at = mod)
-        mods.foreach(m => if (m.productPrefix == mod.productPrefix) syntaxError("repeated modifier", at = mod))
+        if (!mod.isInstanceOf[Mod.Quasi]) mods.foreach(m => if (m.productPrefix == mod.productPrefix) syntaxError("repeated modifier", at = mod))
         if (mod.hasAccessBoundary) mods.filter(_.hasAccessBoundary).foreach(mod => syntaxError("duplicate access qualifier", at = mod))
       }
       validate()
       mods :+ mod
     }
+    def continueLoop = ahead(token.is[`:`] || token.is[`=`] || token.is[EOF])
```
xeno-by: Where does this heuristic come from?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -2219,36 +2259,44 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
...
-          val isExplicitlyLocal = mods.accessBoundary.map(_.isInstanceOf[Term.This]).getOrElse(false)
-          if (ownerIsCase) isExplicitlyLocal
-          else isExplicitlyLocal || (!isValParam && !isVarParam)
+    val name = termName() match {
+      case q: Term.Name.Quasi => q.become[Term.Param.Name.Quasi]
+      case x => x
+    }
+    val tpt =
+      if (token.isNot[`:`])
+        None
```
xeno-by: How do you mean, `None`? Wasn't this method always returning `Some(...)`?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -2268,7 +2316,7 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
 
   def typeParam(ownerIsType: Boolean, ctxBoundsAllowed: Boolean): Type.Param = autoPos {
     if (token.is[Unquote]) return unquote[Type.Param]
-    var mods: List[Mod] = annots(skipNewLines = true)
+    var mods: List[Mod] = if (token.is[Ellipsis]) List(ellipsis(1, unquote[Mod])) else annots(skipNewLines = true)
```
xeno-by: Is `tparam"@foo ${Mod.Covariant()} T" gonna work?

scalameta/src/main/scala/scala/meta/Trees.scala
Dveim: fixed

scalameta/src/main/scala/scala/meta/Trees.scala
Dveim: fixed

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -2420,7 +2472,7 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
     val isMutable = token.is[`var`]
     next()
     val lhs: List[Pat] = commaSeparated(noSeq.pattern2().require[Pat]).map {
-      // TODO: val `x`: Int should be parsed as Decl.Val(..., Pat.Var.Term(Term.Name("x")), ...)
```
xeno-by: Why remove this comment?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -2658,25 +2710,31 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
...
     }
-    result
-  }
 
-  def constructorCall(tpe: Type, allowArgss: Boolean = true): Term = {
+  def constructorCall(tpe: Type, allowArgss: Boolean = true): Ctor.Call = {
     object Types {
       def unapply(tpes: Seq[Type.Arg]): Option[Seq[Type]] = {
-        if (tpes.forall(_.isInstanceOf[Type])) Some(tpes.map(_.require[Type]))
+        if (tpes.forall(t => t.isInstanceOf[Type] || t.isInstanceOf[Type.Arg.Quasi])) Some(tpes.map {
```
xeno-by: How can `Type.Arg.Quasi` appear here?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -2724,7 +2784,9 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
    */
   def templateParents(): List[Term] = {
     val parents = new ListBuffer[Term]
-    def readAppliedParent() = parents += constructorCall(startModType())
+    def readAppliedParent() =
+      if (token.is[Ellipsis]) parents += ellipsis(1, unquote[Term])
```
xeno-by: `unquote[Term]` or `unquote[Ctor.Call]`?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -2827,9 +2891,10 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
     val stats = new ListBuffer[T]
     while (!token.is[StatSeqEnd]) {
       def isDefinedInEllipsis = { if (token.is[`(`] || token.is[`{`]) next(); statpf.isDefinedAt(token) }
-      if (token.is[Ellipsis] && ahead(isDefinedInEllipsis)) stats += ellipsis(1, statpf(token))
-      else if (statpf.isDefinedAt(token)) stats += statpf(token)
-      else if (!token.is[StatSep]) syntaxError(errorMsg, at = token)
+      if (statpf.isDefinedAt(token) || (token.is[Ellipsis] && ahead(isDefinedInEllipsis)))
+        stats += statpf(token)
+      else if (!token.is[StatSep])
+        syntaxError(errorMsg, at = token)
```
xeno-by: Why this change?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
```diff
@@ -2903,10 +2972,12 @@ private[meta] class Parser(val input: Input)(implicit val dialect: Dialect) { pa
    */
   def templateStats(): List[Stat] = statSeq(templateStat)
   def templateStat: PartialFunction[Token, Stat] = {
+    case token if token.is[Ellipsis] =>
+      ellipsis(1, unquote[Stat])
```
xeno-by: What about handing `_: Unquote` here in the same way like you handled it in `topStat`?

xeno-by: Also, why `token if token.is[Ellipsis]` instead of `_: Ellipsis`?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: Usually ellipsis is parsed with `part`. But sometimes (for example, when parsing paramClauses) it should simple be `ellipsis(1, unquote[...])`, because `part` works ambiguously.

scalameta/src/main/scala/scala/meta/internal/parsers/Helpers.scala
Dveim:
```
val mods = List(mod"private", mod"final")
val paramss = List(param"x: X", param"x: Y")
val expr = q"this(foo, bar)"
q"..$mods def this(..$paramss) = $expr"
....
[info]   org.scalameta.invariants.InvariantFailedException: invariant failed:
[info] when verifying scala.meta.internal.parsers.Helpers.XtensionTermOps(body).isCtorBody
[info] found that scala.meta.internal.parsers.Helpers.XtensionTermOps(body).isCtorBody is false
[info] where body = this(foo, bar)
```

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: Could you give an example?

tests/src/test/scala/quasiquotes/ErrorSuite.scala
```diff
@@ -345,4 +365,102 @@ class ErrorSuite extends FunSuite {
...
+    assert(typecheckError("""
+      import scala.meta._
+      import scala.meta.dialects.Scala211
+      q"private final def this(x: X, y: Y) = foo"
+    """) === """
+      |<macro>:4: this expected but identifier found
+      |      q"private final def this(x: X, y: Y) = foo"
+      |                                             ^
+    """.trim.stripMargin)
+  }
```
xeno-by: Asking here, since github doesn't show the diff for `QuasiquoteSuite.scala`.

xeno-by: Is there any particular order for tests in QuasiquoteSuite.scala? Would it be possible to order everything according to its order of appearance in quasiquote.md?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: In this method https://github.com/Dveim/scalameta/blob/master/scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala#L2220-2220.

Here (https://github.com/Dveim/scalameta/blob/master/scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala#L2251) it would parse ellipsis as mods, what is wrong in this case.

scalameta/src/main/scala/scala/meta/internal/parsers/Helpers.scala
xeno-by: This shouldn't work. The correct quasiquote for expr is `ctor"SomeClass(foo, bar)`.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: Why not have `ellipsis(1, unquote[T])` called every time?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: This is the only case when `Ellipsis` could (possibly) be merged in `TokenClass`. But that would lead to unnecessary checks inside `DefIntro`, because other parsing situations would become broken with naively added `Ellipsis`. So, I decided to put simplest check here, instead of several complicated in `DefIntro`.

scalameta/src/main/scala/scala/meta/internal/parsers/Helpers.scala
Dveim: Removed that `case`.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: This method didn't work correctly with `(` as token, `forSome`, and probably something else.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: `Term.Name.Quasi` is `Term.Name`, but trying to extract `value` from first one leads to error.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: It moved to `case _: Ident | _: `this` | _: Unquote =>`.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: Indeed, seems that method could be simplified.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: I think it is possible, but uniform parsing of Unquotes and Ellipses itself is another refactoring issue. And it requires careful usage analysis (there are much more such cases).

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: Removed.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: To handle cases like `..@$mods $mod1 x: X ...`.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: But no, it handles `pt"_"`.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: Yes, also it didn't handles unquotes. Now it does.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim:
```
scala> import scala.meta._, scala.meta.dialects.Scala211; implicit class tretgfdsvd(a:Any) { def ===(b:Any) = a == b }
import scala.meta._
defined class tretgfdsvd

scala> tparam"@foo ${Mod.Covariant()} T"
<console>:16: error: value Covariant is not a member of object scala.meta.Mod
       tparam"@foo ${Mod.Covariant()} T"
                         ^
```

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: Because that todo is done.
```
scala> q"val `x`: Int".show[Raw]
res2: String = Decl.Val(Nil, List(Pat.Var.Term(Term.Name("x"))), Type.Name("Int"))
```

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: It is necessary to parse `val ctor"(x, ..$tpes) => $tpe" = ctor"(x, y) => z"`.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: `Ctor.Call`, of course. Did that change right before implementing `Ctor.Call` and forgot to review.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: After some previous changes, first and second cases did same job, so I merged them.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: 1) `Unquote` is handled within `DefIntro`.
2) To preserve code style in this method.

tests/src/test/scala/quasiquotes/ErrorSuite.scala
Dveim: I used that order, more or less. First tests are "legacy" ones and other are (almost) in order with specification. Should I reorder QuasiquoteSuite?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: patternTyp() called on `_` should give an error, because stuff like `parse 'x match { case y: _ => }'` produces an error. It's impossible to have a Pat.Var.Type at the top level of patternTyp (only as a type argument), and in the same vein it's impossible to have Pat.Type.Wildcard at the top level.

xeno-by: Also see https://github.com/scalameta/scalameta/issues/216

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: You mean, it was impossible to parse `x match { case _: (x, y) => ??? }`?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: I mean, previously we had `value.head.isLower && value.head.isLetter`. Now we have `value.head.isLetter && value.head.isLower`. What's the reason for this change?

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: You mean, support for stuff like `case Foo[Bar]() => `?

xeno-by: Haha!

```
21:33 ~$ parse 'x match { case foo[bar]() => }'
[[syntax trees at end of parser]]// Scala source: tmpR7L1zs
package <empty> {
  class wrapper extends scala.AnyRef {
    def <init>() = {
      super.<init>();
      ()
    };
    x match {
      case foo[(bar @ <empty>)]() => ()
    }
  }
}
```

This means that Pat.Extract.targs should be `Seq[Pat.Type]`, not `Seq[Type]`.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: This should be handled elsewhere, because putting this logic in `param` means that previously invalid code such as `def foo(x) = ???` will now be parsed successfully.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: I meant `scala.meta.internal.ast.Mod.Covariant()`

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: It would seem so... I wonder what I had in mind back then.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: Hmm I'm having hard time seeing how `ellipsis(1, statpf(token))` and `statpf(token)` could do the same thing :puzzled:

tests/src/test/scala/quasiquotes/ErrorSuite.scala
xeno-by: Yes, that would be great. Otherwise, it'd be hard to maintain the file in the future. You did an amazing job writing all these tests btw.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: This is probably the only thing in this pull request that looked seriously spooky to me. I'd prefer you document it by saying something like "the only things that can come after $mod or $mods are either keywords or $pname/$tpname; the former is easy, but in the case of the latter, we need to take care to not hastily parse those names as modifiers".

xeno-by: Also, speaking of those tokens, what happens in the case of `tparam"..$mods $foo[..."` or `tparam"..$mods $foo <: ..."`? Shouldn't `[`, `<:`, `>:` and `<%` be also included into this list?

scalameta/src/main/scala/scala/meta/Trees.scala
xeno-by: So now, since we have `Ctor.Call`, we can do three things:
1) Update the spec for Template to use `..$ctorcalls` instead of `..$exprs`.
2) Update Template.parents to say `Seq[Ctor.Call]`
3) Update Template to require isCtorCall for all Template.parents (oops, this is already done)

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: Ok, removed that case branch.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: It was impossible to parse `pt"(X, Y) => Z"`.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: Probably copy-pasted from somewhere. Changed.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: But in Trees we have `class Extract(ref: Term.Ref, targs: Seq[impl.Type], args: Seq[Pat.Arg])`.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: Added comment. Also added these tokens in check expression.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim: Checking whether `name` is Unquote should be in this method, I think. Added extra check to avoid situations like in your example.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
Dveim:
```
scala> tparam"@foo ${scala.meta.internal.ast.Mod.Covariant()} T"
<console>:16: error: end of file expected but identifier found
       tparam"@foo ${scala.meta.internal.ast.Mod.Covariant()} T"
                                                              ^

scala> tparam"@foo +T"
res2: scala.meta.Type.Param = @foo +T
```

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: Yeah, that's an oversight.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: Okay, I don't mind, as long as things aren't contradicting the SLS.

scalameta/src/main/scala/scala/meta/internal/parsers/Parsers.scala
xeno-by: I think that's a problem, because the first line clearly fits the `tparam"..$mods $tpname[..$tparams] >: $tpeopt <: $tpeopt <% ..$tpes : ..$tpes"` specification.

scalameta/src/main/scala/scala/meta/Trees.scala
Dveim: Done.

Dveim:
Moved tests.

tests/src/test/scala/quasiquotes/ErrorSuite.scala
Dveim: Moved tests.

PR #232, opened 2015-08-21
Remaining cases from specification (resubmit)

xeno-by:
Supersedes https://github.com/scalameta/scalameta/pull/228


xeno-by:
LGTM

PR #240, opened 2015-09-14
link to scaladoc at sonatype

MasseGuillaume:


xeno-by:
Given that: 1) after modularization (https://github.com/scalameta/scalameta/pull/242) it's currently impossible to produce a unified scaladoc for all the projects, 2) scaladocs need some love (e.g. in grouping related APIs together, so that people don't get overwhelmed by the sheer volume of functionality), I suggest we don't publicly advertise them yet. Wdyt?

MasseGuillaume:
Make sense ;-). Metadoc to the rescue!

Issue #312, opened 2015-11-19
Parser inconsistent with scalac


olafurpg:
I ran the following test on a few open source Scala projects and collected statistics on examples when the scala.meta parser throws a ParseException while scalac is happy.

https://github.com/olafurpg/scalafmt/blob/master/src/test/scala/scala/fmt/ProjectTest.scala

Here is the result:

## Summary:

Successes: 27722, avg time to parse a single file 5.26 ms
Skipped: 1334 (scalac not happy)
Unknown failures: 1
Parse exceptions: 197

## right arrow expected but ( found

* https://github.com/PredictionIO/PredictionIO/blob/0e143c77e9ff0cc7c47a5a1b4ddb85369cd224cc/examples/experimental/scala-recommendations/src/main/scala/Run.scala#L156 cols:34-35
* https://github.com/scala/scala/blob/fe336167e7a9f68c7ea6beb80e2cbb69cea95ff8/src/compiler/scala/tools/nsc/transform/patmat/PatternMatching.scala#L68 cols:28-29
* https://github.com/scala/scala/blob/fe336167e7a9f68c7ea6beb80e2cbb69cea95ff8/test/files/pos/t6022b.scala#L11 cols:36-37

## is not a valid block statement

* https://github.com/scala/scala/blob/fe336167e7a9f68c7ea6beb80e2cbb69cea95ff8/test/pending/shootout/prodcons.scala#L22 cols:11-25

## ) expected but identifier found

* https://github.com/JetBrains/intellij-scala/blob/431b15e6c0e6a3ca1f3ed079813ad4c44c6c79e8/testdata/changeSignature/fromJava/ArrayToVarargs_after.scala#L8 cols:31-32
* https://github.com/akka/akka/blob/a7021827c3974762143920139a72f33064dd78aa/akka-actor/src/main/scala/akka/event/Logging.scala#L1072 cols:69-70
* https://github.com/akka/akka/blob/a7021827c3974762143920139a72f33064dd78aa/akka-actor/src/main/scala/akka/event/Logging.scala#L1072 cols:69-70
* https://github.com/apache/kafka/blob/c7c7f4cfa7e1c385d5f5706161572d657b495b7a/core/src/main/scala/kafka/message/ByteBufferMessageSet.scala#L267 cols:126-127
[...]

PR #339, opened 2016-02-22
some fixes for @olafurpg

xeno-by:


xeno-by:
See https://github.com/scalameta/scalameta/issues/312

PR #395, opened 2016-05-12
removed Default.defaultSettings so the project builds

akhileshs:
It seems that using `Defaults.defaultSettings` is deprecated and causes problems with `ensime-sbt` causing the build to fail.

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -20,7 +20,8 @@ object build extends Build {
     packagedArtifacts := Map.empty,
     unidocProjectFilter in (ScalaUnidoc, unidoc) := inAnyProject,
     aggregate in test := false,
-    test := (test in scalameta in Test).value
+    test := (test in scalameta in Test).value,
+    console := (console in scalameta in Compile).value
   ) aggregate (
     common,
     dialects,
@@ -136,7 +137,7 @@ object build extends Build {
     exposePaths("scalameta", Test): _*
   ) dependsOn (common, dialects, parsers, quasiquotes, tokenizers, transversers, trees)
 
-  lazy val sharedSettings = Defaults.defaultSettings ++ crossVersionSharedSources ++ Seq(
+  lazy val sharedSettings = crossVersionSharedSources ++ Seq(
     scalaVersion := ScalaVersions.max,
     crossScalaVersions := ScalaVersions,
     crossVersion := CrossVersion.binary,


fommil:
can you please type `show scalaVersion` and print the results here? I am not convinced that `root` is the `rootProject`. You might need an extra `override lazy val rootProject = Some(root)`

akhileshs:
Here's the results for `show scalaVersion`: 
`
[info] dialects/*:scalaVersion
[info] 	2.11.8
[info] trees/*:scalaVersion
[info] 	2.11.8
[info] tokens/*:scalaVersion
[info] 	2.11.8
[info] transversers/*:scalaVersion
[info] 	2.11.8
[info] parsers/*:scalaVersion
[info] 	2.11.8
[info] common/*:scalaVersion
[info] 	2.11.8
[info] inputs/*:scalaVersion
[info] 	2.11.8
[info] scalameta/*:scalaVersion
[info] 	2.11.8
[info] quasiquotes/*:scalaVersion
[info] 	2.11.8
[info] tokenizers/*:scalaVersion
[info] 	2.11.8
[info] root/*:scalaVersion
[info] 	2.11.8
`

I tried adding the extra `override lazy val rootProject = Some(root)`. Doesn't seem to work if I use `Defaults.defaultSettings` though.

fommil:
A lot of these settings should probably be in ThisBuild. Nobody ever does it right :-)

project/build.scala
```diff
@@ -7,7 +7,7 @@ import UnidocKeys._
 
 object build extends Build {
   lazy val ScalaVersions = Seq("2.11.8")
-  lazy val LibraryVersion = "0.1.0-SNAPSHOT"
+  lazy val LibraryVersion = "0.1.0-RC5"
```
xeno-by: This shouldn't be in here.

xeno-by:
@fommil So how should we change the build? I'm a bit confused.

fommil:
This change is good enough :-) except the RC bit.

project/build.scala
```diff
@@ -32,7 +32,7 @@ object build extends Build {
     tokens,
     transversers,
     trees
-  )
+  ) dependsOn (scalameta)
```
xeno-by: Why do we need this?

akhileshs: Because, when I do `sbt console`, I can't directly do things like `import scala.meta._`. I'd have to do `scalameta/console` without this.

xeno-by: @fommil What would you advise? I'm kinda scared of dependsOn. Maybe we should do `console := (console in scalameta in Compile).value` instead?

fommil: yeah that might be better, but I'm not sure you can do `in scalameta` can you?

You could just set up an alias command.

xeno-by: Yeah, it does work. @akhileshs could you change this pull request to do that?

xeno-by:
@akhileshs I'll also need you to sign the Scala CLA before merging your contributions: https://www.lightbend.com/contribute/cla/scala.

xeno-by:
Thanks for addressing the feedback. Now the only thing that I'll need is a good commit message.

I suggest:
  * Present tense imperative, i.e. "fix", not "fixed" or "fixes". I've been inconsistent with this myself lately, and that'll be a good motivation for me to follow a consistent guideline, too.
  * Commit description that comes with an explanation for what you did. Why did you add `console...`? Why did you remove `defaultSettings`? These changes are non-obvious enough to warrant an explanation.

xeno-by:
Thank you for your first contribution!

PR #402, opened 2016-05-23
implement drone build

xeno-by:
review @fommil


fommil:
looks good but it says there is a problem with the secrets file and fails the build real quick

PR #408, opened 2016-05-29
implement support for CLA checks in our drone build

xeno-by:


.drone.yml
```diff
@@ -7,6 +7,22 @@ build:
...
+        if [ $? -ne 0 ] ; then exit 1; fi;
+        echo "Pull request submitted by $author";
+        signed=$(curl -s http://www.lightbend.com/contribute/cla/scala/check/$author | jq -r ".signed");
+        if [ $? -ne 0 ] ; then exit 1; fi;
+        if [ "$signed" = "true" ] ; then
+          echo "CLA check for $author successful";
+        else
+          echo "CLA check for $author failed";
+          echo "Please sign the Scala CLA to contribute to scala.meta";
+          echo "Go to https://www.lightbend.com/contribute/cla/scala and then resubmit this pull request";
```
xeno-by: @fommil Is there a different way to restart the build for a regular user? Even closing and reopening the same pull request doesn't make drone restart.

fommil:
You should have admin so you should be able to click rebuild. Make sure you're logged in. Pushing to the branch should work and you can always go into github settings and resend a webhook payload if all else fails.

xeno-by:
@fommil Thanks! But, if I understand correctly, the only way for a new contributor to restart the build after they've signed the CLA is to create a new PR?

Other than that, is everything LGTY?

fommil:
Oh i see, hmm, yeah. I'd advise splitting the CLA checker from the CI build. Otherwise you'll have to restart their builds.

fommil:
In a way that might not be a bad thing as it could avoid DOS attacks!

PR #411, opened 2016-06-01
moves to a different drone image

xeno-by:


.drone.yml
```diff
@@ -1,5 +1,5 @@
 build:
-  image: ensime/ensime:v1.x-cache
+  image: ensime/ensime:v2.x-cache
```
fommil: :smile: I just made them up this weekend to use Azul. `jenv` is available so all you need to specify the java version is a `.java-version` file containing `1.6` or `1.7` etc.

Note that there is currently a regression in OpenJDK which blows up when trying to use tiered compilation --- a default in sbt *sigh* so make sure you have a `.jvmopts` file or you'll get core dumps.

xeno-by:
Looks like this is harder than I thought.

PR #412, opened 2016-06-01
moves to a different drone image that supports Java 1.7

xeno-by:
JDK 1.7 seems to be necessary for Scalatex because of its
transitive dependency on Ammonite which uses java.nio.files.

Note that this doesn't mean that we're going to emit jar files that
can only be run on JVM 1.7+. Scala 2.11.x produces classfiles that
have v50.0 format, which means JVM 1.6+.


.jvmopts
```diff
@@ -0,0 +1 @@
+-XX:-TieredCompilation
```
fommil: I was thinking more along the lines of

```
-XX:MaxPermSize=256m
-Xss2m
-Xms725m
-Xmx725m
```

i.e. specifying your exact requirements.

PR #413, opened 2016-06-01
moves to a different drone image that supports Java 1.7

xeno-by:
JDK 1.7 seems to be necessary for Scalatex because of its
transitive dependency on Ammonite which uses java.nio.files.

Note that this doesn't mean that we're going to emit jar files that
can only be run on JVM 1.7+. Scala 2.11.x produces classfiles that
have v50.0 format, which means JVM 1.6+.

-XX:-TieredCompilation in .jvmopts is necessary to work around
an OpenJDK crash: https://github.com/scalameta/scalameta/pull/411#discussion_r65365666.


.jvmopts
```diff
@@ -1 +1,5 @@
--XX:-TieredCompilation
\ No newline at end of file
+-XX:-TieredCompilation
```
fommil: no need to disable tiered compilation, sbt will use only the config in this file. I guess it doesn't hurt.

PR #414, opened 2016-06-01
force drone to fetch the tag we're looking for

xeno-by:


readme/Readme.scala
```diff
@@ -65,6 +65,7 @@ object Readme {
 
   def versionBadge = {
     def timestampOfTag(tag: String): String = {
+      shell.call(s"git fetch origin refs/tags/$tag:refs/tags/$tag")
```
fommil: or you could set up drone to do this from the beginning http://readme.drone.io/usage/cloning/

xeno-by: I think it's fine. This way we fetch less data.

PR #418, opened 2016-06-03
clarifies our goals towards support for other environments

xeno-by:


xeno-by:
I'm confident that the commit isn't gonna break anything, so I'll merge it without drone's approval. There are some technical problems on the server.

PR #421, opened 2016-06-05
assorted fixes for bugs uncovered during sbt publish-signed

xeno-by:


project/build.scala
```diff
@@ -206,7 +207,10 @@ object build extends Build {
         case ex: Exception if ex.getMessage.contains(nothingToCommit) => println(nothingToCommit)
       }
     },
+    // TODO: doesn't work at the moment, see https://github.com/sbt/sbt-pgp/issues/42
+    publishSigned := publish.value,
```
xeno-by: @fommil maybe you could provide advice here?

xeno-by: More precisely, I can't disable `publish-signed` for the `readme` project. I do not want to publish it, just to use it to generate the website via Scalatex. However, `sbt-pgp` insists that I configure a publishing repository for `readme`, otherwise it errors out when I do `sbt publish-signed`.

xeno-by: Also /cc @lihaoyi

lihaoyi: I usually have a separate project that aggregates all my publishable stuff. Often I have integration tests or other random things that I don't want to publish either

fommil: we aggregate just the bits we want to publish https://github.com/ensime/ensime-server/blob/2.0/project/EnsimeBuild.scala#L243

note also that if you have a custom test configuration, with its own dependencies, you must add them to `test` and your config or it triggers a bug in ivy that means your `main` config will depend on those jars. Probably not relevant for you but a major source of pain for us.

fommil: oh joy, I just realised we are publishing `testutil`... that's probably not right.

PR #423, opened 2016-06-05
sbt publish-signed no longer exits with an error

xeno-by:
By the advice of @fommil and @lihaoyi, I've stopped aggregating readme in root.

Now I explicitly aggregate readme to include it in the most important tasks.
Naturally, I don't feel like copy/pasting code, so I'm only doing this
for `test` and `publish`, but I've already wasted too much time on SBT.

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -24,7 +24,16 @@ object build extends Build {
     packagedArtifacts := Map.empty,
     unidocProjectFilter in (ScalaUnidoc, unidoc) := inAnyProject,
     aggregate in test := false,
-    test := (test in scalameta in Test).value,
+    test := {
+      val runTests = (test in scalameta in Test).value
+      val runDocs = (run in readme in Compile).toTask(" --validate").value
+    },
+    publish := {
+      // Others projects are published automatically because we aggregate.
+      val publishDocs = (publish in readme).value
+    },
+    // TODO: The same thing for publishSigned doesn't work.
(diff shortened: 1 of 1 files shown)


project/build.scala
```diff
@@ -24,7 +24,16 @@ object build extends Build {
...
+    test := {
+      val runTests = (test in scalameta in Test).value
+      val runDocs = (run in readme in Compile).toTask(" --validate").value
+    },
+    publish := {
+      // Others projects are published automatically because we aggregate.
+      val publishDocs = (publish in readme).value
+    },
+    // TODO: The same thing for publishSigned doesn't work.
+    // SBT calls publishSigned on aggregated projects, but ignores everything else.
```
xeno-by: /cc @eed3si9n. Maybe you'll be able to come up with something here. Here's the background that introduces the need for this: https://github.com/scalameta/scalameta/pull/421/files/5be17c403529e85d39cbac3c9f73723d1acac6b2#r65814536.

lihaoyi:
IIRC if you stop aggregating readme in root, `sbt ~readme/run` won't work anymore, which takes away some of the nice edit-refreshing that I have when editing my Scalatex docs.

xeno-by:
I've just checked and, unfortunately, you're correct.

xeno-by:
Hahaha, found a workaround: `sbt 'project readme' ~run` does work. Omg sbt, you're the best.

PR #426, opened 2016-06-09
fix a Lit(()).structure prettyprinting issue

xeno-by:


xeno-by:
Thanks to @akhileshs who discovered this problem.

PR #447, opened 2016-06-30
Remove sbt-dependency-graph-sugar, clashes with newer, global sbt-dependency-graph

dwijnand:


xeno-by:
What do we do instead to achieve https://github.com/scalameta/scalameta/blob/master/notes/modules.md?

dwijnand:
:( I had a look and it seems there's no good answer I can give you..:

* Newer versions of sbt-dependency-graph (SDG) are incompatible with un-updated sbt-dependency-graph-sugar (SDGS)
* Only SDGS can do SVG
* The other outputs of SDG are DOT and GraphML, which GitHub renders as source rather than the graph image

The only reasonable solution I see would be for you to install sbt-dependency-graph-sugar globally and remove it from the build here..

PR #449, opened 2016-07-01
restore modules.svg generation

xeno-by:


dwijnand:
Ah nice one man, and thanks.

Issue #223, opened 2015-08-17
problems with unquoting subtrees into trees


xeno-by:
We have several places in the current implementation of quasiquotes, where unquoting a subtree into a tree is not as easy as writing `... $subtree ...`. We need to scan through the spec, detect all those cases and think what to do with them.

Examples:
- `new $template` doesn't work
- `class C $ctor` or something doesn't work

xeno-by:
Also see #405.

dwijnand:
Ah, good at least to get confirmation that it's truly impossible to define a Term.New. In my code it's one of the few parts that doesn't use quasiquotes but:

```scala
    val ctorref = Ctor.Ref.Name(tname.value)
    val ctorNew = Term.New(Template(Nil, sciSeq(q"$ctorref(..$aexprs)"), anonParam, None))
``

**Comment by xeno-by** [comment]
@dwijnand What exactly would you like to achieve with your code snippet? I have a strong suspicion that your use case will be supported in 2.0.

**Comment by dwijnand** [comment]
Defining a variant of `case class` via macro annotation: https://github.com/dwijnand/dataclasses/blob/e244272b58a5d5abd90a1b6fcd3ca561d17d9766/src/main/scala/dataclasses/internal/DataMacros.scala#L40-L44

**Comment by olafurpg** [comment]
The next release introduces Term.Init, which avoids the need for templates in most basic cases.

```scala
scala> init"Foo"
res3: scala.meta.Init = Foo

scala> q"new $res3"
res4: meta.Term.New = new Foo

// templates are necessary in Term.NewAnonymous
scala> q"new $res5 { val x = 2 }"
res9: meta.Term.NewAnonymous = new Foo(1, 2) { val x = 2 }
```

xeno-by:
@olafurpg and I had a discussion, and we decided to pronounce quasiquotes feature-complete.

There are still some open issues (#46, #223, #251, #280 and others), but they all have two things in common: 1) they require non-trivial design work to be fixed, 2) they aren't popular at all. Given our recent decision to start avoiding unwarranted breaking changes in 2.0, we'll consider quasiquotes good enough at this point and will close all outstanding feature requests.

We will still be fixing bugs (in fact, I've fixed about a dozen of such bugs for 2.0), but without strong evidence our design efforts will be directed elsewhere.

dwijnand:
I never got round to implementing Olaf's suggested Term.Init, but what I was doing here is probably vanilla enough to have been fixed with that feature.

PR #451, opened 2016-07-01
work around #448

### LINKED ISSUE #448: Can't splice in AnyVal type as a parent type
Here's an example:

```scala
package p

import scala.{ collection => sc }, sc.{ immutable => sci }, sci.{ Seq => sciSeq }

import scala.meta._

class bob extends scala.annotation.StaticAnnotation {
  inline def apply(defn: Any) = meta {
    val q"""
      ..$mods0 class $tname[..$tparams] ..$ctorMods (...$paramss0) extends { ..$earlyStats } with ..$ctorcalls {
        $selfParam =>
        ..$stats
      }
    """ = defn

    val name = Term.Name(tname.value)

    val AnyVal = t"_root_.scala.AnyVal"

    val objectDefn = q"""
      object $name {
        final class Builder(val x: Int) extends $AnyVal
      }
    """

    Term.Block(sciSeq(defn.asInstanceOf[Stat], objectDefn))
  }
}
```

You can simulate the bug by running this:

```scala
package p

@bob class Foo(x: Int)

object Test {
    println(new Foo.Builder(1))
  }
}
```

xeno-by:
We can't fully fix #448 at the moment because of the syntactic ambiguity
described in #223.

When the user writes q"class C extends $foo", we can't really know whether
they meant to insert/extract a constructor or a template.

Therefore, what we expect people to do now is to write q"class C extends $foo {}",
in which case they unambiguously point out that they want to work with a parent.
It turns out that this didn't work before, and now I fixed that.
I also improved the error message to clearly explain what needs to be done.

I hope that in the future we'll be able to do better here.

### DIFF
--- a/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -112,6 +112,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
   def parseCase(): Case = parseRule{parser => parser.accept[KwCase]; parser.caseClause()}
   def parseCtorCall(): Ctor.Call = parseRule(_.constructorCall(typ(), allowArgss = true))
   def parseTemplate(): Template = parseRule(_.template())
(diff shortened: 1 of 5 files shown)


dwijnand:
Excellent. Thanks.

Issue #399, opened 2016-05-19
Retain formatting upon performing unrelated source transformations


akhileshs:
This issue forms the the first goal of #379.

As discussed over there, if we had, 

``` scala
scala> "{ /* hello */ def foo(bar: Int) = ??? }".parse[Term].get
res0: scala.meta.Term = { /* hello */ def foo(bar: Int) = ??? }

scala> res0.transform { case q"bar" => q"baz" } 
res1: scala.meta.Tree =
{
  def foo(baz: Int) = ???
}
```

The result of the transformation should retain the original source formatting and other relevant details (comments in this case).

xeno-by:
I'm closing this as a duplicate of #379.

PR #457, opened 2016-07-28
source code aware transformations

akhileshs:
This is a proof of concept solution to #399.


xeno-by:
It looks like pushing this PR further will require a significant design effort that I don't have time for right now. At the moment, the main and only priority is shipping the semantic API asap, so I'll close this PR for the time being.

@akhileshs Thank you very much for your attempt at solving the issue! If you'd like to revisit your prototype in the future, I'll do my best to support you as much as the roadmap permits.

akhileshs:
@xeno-by No worries! Yeah, this definitely requires a conceptual redesign.

PR #464, opened 2016-08-14
fix #462

### LINKED ISSUE #462: Compile error  unquoting Term.Select in package statement
Compiling
```scala
val ref = q"p.a" 
val stats = List(q"class A", q"object B") 
val packageTree = q"package $ref { ..$stats }" 
```

Results in the error message
`
Error:(161, 31) type mismatch when unquoting; found : scala.meta.Term.Select required: scala.meta.Term.Name val packageTree = q"package $ref { ..$stats }"
`

However this compiles
```scala
val ref = q"p" 
val stats = List(q"class A", q"object B") 
val packageTree = q"package $ref { ..$stats }"
```

solomono:
In Issue #462  unquoting a Term.Select results in a compiler error. 
The ScalametaParser creates a Quasi of the wrong type  when unquoting the ref attribute of the Pkg tree. This occurs when the variable that is unquoted is a Term.Select.

The following call sequence results in the creation of the Pkg.ref Quasi tree
  ```scala
    def packageDef(): Pkg
       def qualId(): Term.Ref
          def termName(advance: Boolean = true) Term.Name
              def name[T <: Tree : AllowedName : AstInfo](ctor: String => T, advance: Boolean): T 
                 def unquote[T](advance)
```
This previously always returned a  Term.Name.Quasi.
This value is appropriate if the variable unquoted  is a Term.Name,  However if it is a Term.Select,
the error in #462 results.
We fix this in  qualId( which is presently only called in regard to creating a Pkg tree) 
by converting the Term.Name.Quasi to a Term.Ref.Quasi.
Note that this issue may potentially arise in other cases where termName or name is being called in which the result could potentially be a Quasi, and where Term.Ref is allowable. 
I have added a test to confirm the fix

### DIFF
--- a/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -970,7 +970,12 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
           else {
             next()
             accept[KwType]
-            Type.Singleton(ref)
+            val refOrQuasi = ref match {
+              case quasi: Term.Name.Quasi =>
+                quasi.become[Term.Ref.Quasi]
+              case ref => ref
+            }
+            Type.Singleton(refOrQuasi)
           }
       }))
     }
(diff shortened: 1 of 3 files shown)


solomono:
I signed the  CLA

xeno-by:
This is a very thorough fix! Thanks a lot for your contribution.

I will publish 1.1.0-SNAPSHOT with the fix momentarily.

PR #469, opened 2016-09-08
Properly tokenize xml literals, fixes #356.

### LINKED ISSUE #356: Support for xml literals
Currently, neither our tokenizer, nor our parser correctly handle xml literals. 

Whenever our code encounters an xml literal, we just skip over it, treating the literal as an opaque blob. This solution enough for most use cases, since it is very rarely necessary to manipulate contents of xml literals, but of course this is insufficient in general case.

Your task here would be to:
1. Change the tokenizer to produce [xml interpolation tokens](https://github.com/xeno-by/scalameta/blob/65ff98f81ca9112289b74b7832acb39a2c36222e/scalameta/tokens/src/main/scala/scala/meta/tokens/Token.scala#L62) upon [encountering xml literals](https://github.com/xeno-by/scalameta/blob/65ff98f81ca9112289b74b7832acb39a2c36222e/scalameta/tokenizers/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala#L138). 
2. Make sure that the parser correctly parses and then prettyprints xml literals.

olafurpg:
This commit makes xml literals tokenize like string interpolators.
The solution works roughly like this:

In LegacyScanner:
1) when encountering an xml literal, run a fastparse xml parser to get
   the positions of scala expressions inside the literals.  Populate a
   `xmlLits: Map[Int, Int]` with start positions of the scala
   expressions as keys and their respective end positions as values.
2) in nextToken(), if the current offset has a key in `xmlLits`.
   return an XMLLIT LegacyTokenData and scan until the end of the
   xml literal part. Otherwise, proceed as normal.

In ScalametaTokenizer: we use a similar approach as with interpolated
strings.  When encountering an XMLLIT LegacyTokenData, we emit a
Token.Xml.Part and then run ScalametaTokenizer.loop on the embedded
scala expression up until until the end curly brace.  Then we repeat.

PS. This commit includes some debugging tools that were indispensable while editing LegacyScanner. They might come in handy later.


scalameta/scalameta/src/test/scala/scala/meta/tests/DiffAssertions.scala
```diff
@@ -0,0 +1,124 @@
...
+  * To use, create a class like this:
+  *
+  * {{{
+  *   class MySuite extends FunSuite with DiffAssertions
+  * }}}
+  *
+  * Before:
+  * {{{
+  *   [info]   "...ar</foo> [0..14)
+  *   [info]    [1[5..15])
```
xeno-by: `[1[5`?

scalameta/scalameta/src/test/scala/scala/meta/tests/DiffAssertions.scala
```diff
@@ -0,0 +1,124 @@
...
+  *   [info]     [14..15)
+  *   [info]   EOF [15..." did not equal "...ar</foo> [0..14)
+  *   [info]    [1[4..14])
```
xeno-by: `[1[4`?

scalameta/scalameta/src/test/scala/scala/meta/tests/parsers/XmlSuite.scala
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  test("simple xml literal - 1") {
+    assert(
+      tokensStructure(
+        tokenize("<foo>bar</foo>")) ===
+          """|
+             |          BOF [0..0) ----> class scala.meta.tokens.Token$BOF
+             |              [0..0) ----> class scala.meta.tokens.Token$Xml$Start
+             |<foo>bar</foo> [0..14) ----> class scala.meta.tokens.Token$Xml$Part
```
xeno-by: Is vertical alignment correct here?

scalameta/scalameta/src/test/scala/scala/meta/tests/parsers/XmlSuite.scala
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  test("simple xml literal - 2") {
+    assert(
+      tokensStructure(
+        tokenize("<foo>bar</foo> ")) ===
+          """|
+             |          BOF [0..0) ----> class scala.meta.tokens.Token$BOF
+             |              [0..0) ----> class scala.meta.tokens.Token$Xml$Start
+             |<foo>bar</foo> [0..14) ----> class scala.meta.tokens.Token$Xml$Part
```
xeno-by: Same question as above.

scalameta/tokenizers/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala
```diff
@@ -228,7 +240,9 @@ class LegacyScanner(input: Input, dialect: Dialect) {
       if (lastOffset > 0 && buf(lastOffset) == '\n' && buf(lastOffset - 1) == '\r') {
         lastOffset -= 1
       }
-      if (inStringInterpolation) fetchStringPart() else fetchToken()
+      if (inStringInterpolation) fetchStringPart()
+      else if (inXmlLiteralPart) fetchXmlPart()
```
xeno-by: Maybe `inXmlLiteral`?

scalameta/tokenizers/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
```diff
@@ -236,8 +237,35 @@ class ScalametaTokenizer(input: Input, dialect: Dialect) {
...
+            nextToken()
+            emitContents() // Continue until we reach the final xml part.
+          } else {
+            // We have reached the final xml part.
+            nextToken()
+          }
+        }
+        emitContents()
+        tokens += Token.Xml.End(input, dialect, lastXmlPart, lastXmlPart)
+        legacyIndex -= 1 // I don't know why, but this seems to do the trick.
```
xeno-by: Lol

scalameta/tokenizers/src/main/scala/scala/meta/internal/tokenizers/XmlParser.scala
```diff
@@ -0,0 +1,161 @@
+package scala.meta.internal.tokenizers
+
+/**
+  * Copy-pasta from this lihaoyi comment:
+  * https://github.com/scalameta/fastparse/pull/1#issuecomment-244940542*/
```
xeno-by: Thank you @lihaoyi!

xeno-by:
I suggest we keep the sourcecode-based part in a branch. In the future, we can consider making it part of scalameta contrib. The rest looks good! Thanks a lot for your trouble of interfacing with the legacy scanner.

scalameta/tokenizers/src/main/scala/scala/meta/internal/tokenizers/XmlParser.scala
```diff
@@ -0,0 +1,161 @@
...
+import scala.annotation.switch
+
+import fastparse.all._
+
+class XmlParser(Block: P0,
+                Patterns: P0 = Fail,
+                WL: P0 = CharIn("\n ").rep(),
+                WS: P0 = " ".rep(),
+                HexNum: P0 = Fail) {
```
lihaoyi: Not sure if you want to solve this now, but these default values should really be something else.

Otherwise, XML literals including `CharRef`s will fail, XML literals with `\t`s inside will fail, and XML literal-pattern-matching will fail.

I'm sure you'll bump into this once you start trying to write unit tests and parsing legacy code, but if you can fix it up front maybe you should just do it. If not, kicking the can down the road also works.

scalameta/tokenizers/src/main/scala/scala/meta/internal/tokenizers/XmlParser.scala
```diff
@@ -0,0 +1,161 @@
...
+    var current = 0
+    var curlyBraceCount = 1
+    while (curlyBraceCount > 0) {
+      current += 1
+      (cfg.input(index + current): @switch) match {
+        case '}' => curlyBraceCount -= 1
+        case '{' => curlyBraceCount += 1
+        case _ =>
+      }
+    }
```
lihaoyi: This will totally break once someone puts a curly brace in a string `"}"` or `"{"` inside your XML literal.

Why is it necessary though? Can't the scala-meta parser parse the expression starting from the specified index? (This is not a rhetorical question, I actually don't know what the scala-meta parser can do)

olafurpg: Good catch! I've become sloppy after using the mega tests in scalafmt to catch bugs like this for me : )

The correct solution here is probably to use a lookahead char reader in the (terrible) LegacyScanner. That should allow us to read one token at a time from a given offset respecting backticks/quotes/... . To my best knowledge, the scala.meta tokenizer doesn't expose a cleaner streaming lexer api, @xeno-by ?

lihaoyi: Surely scala-meta has some sort of 

```scala
(code: String, startIndex: Int) => Option[Int /*successIndex*/]
```

API you can call? That's all you need here really. I mean I know it has a `parse` method that's `String => Option[Success]`, and all we need here is to be able to specify a start index for the parse. We shouldn't need to deal with lexing at all here if we have a working parser

xeno-by:
@olafurpg What's our status with this PR?

scalameta/scalameta/src/test/scala/scala/meta/tests/parsers/XmlSuite.scala
```diff
@@ -0,0 +1,148 @@
...
+           |                [0..0) ----> class scala.meta.tokens.Token$Xml$Start
+           |<foo>bar</foo> [0..14) ----> class scala.meta.tokens.Token$Xml$Part
+           |              [14..14) ----> class scala.meta.tokens.Token$Xml$End
+           |              [14..15) ----> class scala.meta.tokens.Token$Space
+           |          EOF [15..15) ----> class scala.meta.tokens.Token$EOF
+           |""".stripMargin.trim.replace("QQQ", "\"\"\""))
+  }
+
+  // TODO(olafur) currently crashes, see https://github.com/scalameta/scalameta/pull/469/files/e2324548b3896d5e7b0993f4063316602d9583ff#r78293500
+  ignore("curlies inside scala expr") {
```
olafurpg: @xeno-by I need to fix this failing test first. I don't know yet when I'll get to it.

scalameta/tokenizers/src/main/scala/scala/meta/internal/tokenizers/XmlParser.scala
olafurpg: We can't use the parser because the tokenizer doesn't depend on the parser. In cases when the input doesn't parse it can be handy to at least have a functioning tokenizer.

I've fixed the implementation to use the LegacyScanner so that expressions like these

```
<foo>{ "{" + `{`}</foo>
```

are accounted for.

xeno-by:
@olafurpg Ping

xeno-by:
@olafurpg It looks like pushing this PR further requires time investments that neither of us can afford right now. Let's close it for the time being.

olafurpg:
Sure thing. I still hope to get back to this since I want to reformat xml literals in scalafmt : ) cc/ https://github.com/olafurpg/scalafmt/issues/662

PR #471, opened 2016-09-11
Allow unquoting options

### LINKED ISSUE #455: Consider adding support for unquoting Option's of trees
Given a situation where I have a `defopt: Option[Defn.Def]` to splice in, it would be nice if this could be done with either `$defopt` or `..$defopt`.

dwijnand:
Fixes #455


xeno-by:
LGTM. I'll pull this in manually.

PR #472, opened 2016-09-11
Fix #458: Recognize quasiquotes-generated lambda functions

### LINKED ISSUE #458: q"($name: $tpe) => ..." doesn't work
```
scala> val name = q"x"
name: meta.Term.Name = x

scala> val tpe = t"T"
tpe: meta.Type.Name = T

scala> q"($name: $tpe) => ???"
<console>:32: error: ; expected but => found
       q"($name: $tpe) => ???"
                       ^
```

jvican:
When parsing quasiquotes, the internal structure of the tree changes. The
previous ascribe case defs were not accounting for the fact that the name and
the type could be quasi.

### DIFF
--- a/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -726,6 +726,9 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
       Some(atPos(tree, tree)(Term.Param(Nil, name, None, None)))
     case name: Term.Placeholder =>
       Some(atPos(tree, tree)(Term.Param(Nil, atPos(name, name)(Name.Anonymous()), None, None)))
+    case Term.Ascribe(quasiName: Term.Quasi, tpt) =>
+      val name = quasiName.become[Term.Name.Quasi]
+      Some(atPos(tree, tree)(Term.Param(Nil, name, Some(tpt), None)))
     case Term.Ascribe(name: Term.Name, tpt) =>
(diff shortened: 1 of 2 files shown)


scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -726,6 +726,9 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
       Some(atPos(tree, tree)(Term.Param(Nil, name, None, None)))
     case name: Term.Placeholder =>
       Some(atPos(tree, tree)(Term.Param(Nil, atPos(name, name)(Name.Anonymous()), None, None)))
+    case Term.Ascribe(quasiName: Term.Quasi, tpe: Type.Quasi) =>
```
xeno-by: This should be just `tpt`, not `tpe: Type.Quasi`.

scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1660,6 +1663,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
               case Term.Quasi(0, _) => true
               case Term.Quasi(1, ParamLike()) => true
               case NameLike() => true
+              case Term.Ascribe(ParamLike(), _) => true
```
xeno-by: This should be as restrictive as possible, i.e. `_: Term.Quasi` instead of `ParamLike()`

xeno-by:
Thank you for your great work!

Issue #470, opened 2016-09-11
scala.meta parser doesn't validate allowed modifiers


manub:
```
scala> q"lazy class X{}"
res6: meta.Defn.Class = lazy class X {}

scala> lazy class X{}
<console>:1: error: lazy not allowed here. Only vals can be lazy
lazy class X{}
```

This also happens with:

```
scala> override class X {}
<console>:14: error: `override' modifier not allowed for classes
       override class X {}
                      ^

scala> q"override class X{}"
res5: meta.Defn.Class = override class X {}
```

xeno-by:
It's not just `lazy class` and `override class`. It's a lot of other stuff. We need to comprehensively validate these cases in the scala.meta parser.

jvican:
I'll pick this up.

jvican:
I have a fix for this that I'll push as soon as I finish the modifiers test suite, which I'm doing to prevent these errors from happening again.

PR #473, opened 2016-09-11
Allows for inline modifier. Fixes #466.

### LINKED ISSUE #466: Support Inline as a modifier
For scalameta paradise, the ability to generate ```inline def apply(defn: Defn) = meta {}``` blocks is ideal, or even just to use inline by itself. Is it possible to have a trait the adds this modifier to a dialect.

I presume the inline modifier would need to be part of scalameta and exposed as a dialect.

My gut feeling is that I want to write ```@leaf implicit object DottyParadise extends Dotty with Paradise```

This would syntax would be beneficial for future support of other compiler plugins too, such as KindProjector etc.

I'm unsure whether the actual definition of the trait would come from the scalameta/scalameta or scalameta/paradise, but support for extensions to dialects would need to come from scalameta

manub:
- A new Dialect for Paradise, which allows inline, has been created.
 - Defined inline modifier only, valid only for Dotty and Paradise. Other dialects work as before.

As part of this, #470 was discovered. `inline` is affected by the same bug, but with @xeno-by it was decided to leave this as part of a separate PR as the work to fix that issue is significant.

### DIFF
--- a/notes/quasiquotes.md
+++ b/notes/quasiquotes.md
@@ -206,6 +206,7 @@ Literal  | `q"$lit"` (construction only), `q"${lit: Lit}"` (also deconstruction)
  Lazy             | `mod"lazy"`
  Val Param        | `mod"valparam"`
  Var Param        | `mod"varparam"`
+ Inline           | `mod"inline"`
 
 ## Enumerators (meta.Enum)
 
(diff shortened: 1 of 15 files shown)


manub:
I signed the Scala CLA.

PR #475, opened 2016-09-11
Allows for inline modifier. Fixes #466.

### LINKED ISSUE #466: Support Inline as a modifier

manub:
- A new Dialect for Paradise, which allows inline, has been created.
 - Defined inline modifier only, valid only for Dotty and Paradise. Other dialects work as before.

### DIFF
--- a/scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
@@ -29,6 +29,9 @@ import scala.compat.Platform.EOL
   // might go ahead and drop support completely.
   def allowXmlLiterals: Boolean
 
+  // Are inline vals and defs supported by this dialect?
+  def allowInline: Boolean
+
   // Permission to parse unquotes.
   // Necessary to support quasiquotes, e.g. `q"$x + $y"`.
   def allowUnquotes: Boolean = metalevel.isQuoted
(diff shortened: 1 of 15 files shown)


xeno-by:
Brilliant work, @manub! It was a pleasure to hack with you today.

scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -88,17 +94,31 @@ package object dialects {
...
     def allowXmlLiterals = Scala210.allowXmlLiterals
+    def allowInline = false
     def allowSpliceUnderscore = true // SI-7715, only fixed in 2.11.0-M5
     def allowToplevelTerms = Scala210.allowToplevelTerms
     def toplevelSeparator = Scala210.toplevelSeparator
     def metalevel = Metalevel.Zero
     private def writeReplace(): AnyRef = new Dialect.SerializationProxy(this)
   }
 
+  @leaf implicit object Paradise extends Dialect {
```
DavidDudson: Just a really quick comment.

This seems to depend heavily on Scala 2.11. With Scala 2.12 in RC now, 2.12 is just around the corner. Can we change the name of this to Paradise_211? Also can we write the same code to work for both Paradise_211 and Paradise_212? @xeno-by correct me if I'm confused. I may have missed something in the Pre-SIP

xeno-by: I didn't want to overengineer this, since for the time being 2.11 is going to be the only supported Scala version for Paradise 3.x. But migration is actually a valid concern. I'll rename this to `Paradise211` now.

DavidDudson: Fair enough. Thanks

PR #480, opened 2016-09-13
Fix #476: Add support for arg quasiquote

### LINKED ISSUE #476: Case class are not parsed on LHS quasiquotes
The following example doesn't work:
```scala
val q"case class $className(arg: Int)" = q"case class A(arg: Int)"
```

jvican:


jvican:
Bad reference to issue. Sorry. Closing.

PR #481, opened 2016-09-13
Fix #468: Add support for arg quasiquote

### LINKED ISSUE #468: quasiquote are unable to get parameter list
It seems hard to get parameter list using quasiquote. See the code below:
```scala
@ import scala.meta._
@ val q"case class $name($list)" = q"case class A(a:Int)"
cmd21.sc:1: identifier expected but ) found
val q"case class $name($list)" = q"case class A(a:Int)"
                            ^
Compilation Failed
```

This code works well with the quasiquote in current scala
```scala
@ import scala.reflect.runtime.universe._
@ val q"case class $name($list)" = q"case class A(a:Int)" 
name: TypeName = A
list: ValDef = <caseaccessor> <paramaccessor> val a: Int = _
```

But if there are multiple parameters, neither works:
```scala
@ import scala.meta._
@ val q"case class $name($list)" = q"case class A(a:Int,b:String)" 
cmd21.sc:1: identifier expected but ) found
val q"case class $name($list)" = q"case class A(a:Int,b:String)"
                            ^
Compilation Failed
```
```scala
@ import scala.reflect.runtime.universe._
@ val q"case class $name($list)" = q"case class A(a:Int,b:String)" 
scala.MatchError: case class A extends scala.Product with scala.Serializable {
  <caseaccessor> <paramaccessor> val a: Int = _;
  <caseaccessor> <paramaccessor> val b: String = _;
  def <init>(a: Int, b: String) = {
    super.<init>();
    ()
  }
} (of class scala.reflect.internal.Trees$ClassDef)
  $sess.cmd2$.<init>(cmd2.sc:1)
  $sess.cmd2$.<clinit>(cmd2.sc:-1)
```
I see a similar issue https://github.com/scalameta/scalameta/issues/458, but I'm not sure whether it is the same problem as described above.

jvican:


jvican:
I believe that the latest refactoring in the sbt build have broken the CI build @xeno-by.

jvican:
Closing this and reopening in another PR to see if the spurious Git error in CI goes away.

Issue #269, opened 2015-10-17
support for parsing dotty


xeno-by:
Shouldn't be very hard given that there are not that many changes from scalac. @darkdimius is there a list of changes somewhere?

smarter:
Here are the ones I can think of:
- union types
- trait parameters: http://docs.scala-lang.org/sips/pending/trait-parameters.html
- repeated by-name parameters: http://docs.scala-lang.org/sips/pending/repeated-byname.html
- literal singleton types: http://docs.scala-lang.org/sips/pending/42.type.html#comment-2211350713

And here's the stuff that is not valid anymore:
- procedure syntax
- view bounds

DarkDimius:
>And here's the stuff that is not valid anymore:

self types on objects.

smarter:
Also, implicit val and def must have an explicit type annotation now.

xeno-by:
This should probably be the moment when we split Type.Compound.

xeno-by:
@odersky suggested that we should take a look at https://github.com/lampepfl/dotty/blob/master/docs/SyntaxSummary.txt.

jvican:
The following features are all available for the hackathon. Just pick one and hack on it!
  
To add:
- [ ] Trait parameters
SIP (for docs): http://docs.scala-lang.org/sips/pending/trait-parameters.html
A good place for start is  https://github.com/scalameta/scalameta/blob/master/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L3061.
Final result should look like `classDef` here: https://github.com/scalameta/scalameta/blob/master/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L3074.

- [ ] Repeated by-name parameters
SIP (for docs): http://docs.scala-lang.org/sips/pending/repeated-byname.html
A good place for start is https://github.com/scalameta/scalameta/blob/master/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L2680.

- [ ] Literal singleton types (aka SIP-23 or 42.type).
Description of what singleton types are: https://github.com/scala/scala/pull/5310
A good way to start is by modifying the expected identifier to accept integers. E.g. `"val a: 42 = 42".parse[Stat].get` fails.
[...]

PR #482, opened 2016-09-14
Support & Types in Dotty

howyp:
Adds support for intersection types from Dotty, contributing to #269.


howyp:
Scala CLA signed now

smarter:
Note that in Dotty, the type `A with B` is a valid but deprecated way to write `A & B`, except when in a extends clause like `class X extends A with B` where the linearization order matters (and writing `class X extends A & B` would be invalid).

xeno-by:
Thanks a lot for your contribution!

I tried to restart the CI job, but without particular success. I'm travelling now, so I don't have much time, so how about you resubmit the PR please?

howyp:
yeah, I couldn't get the job to restart either! re-creating the PR now.

PR #498, opened 2016-09-30
Allow trait parameters in Dotty dialect

### LINKED ISSUE #493: Parse trait parameters in Dotty dialect
Dotty allows trait parameters:
```scala
trait Foo(bar: Int)
```
SIP (for docs): http://docs.scala-lang.org/sips/pending/trait-parameters.html
  
Scala meta has the concept of dialect for different Scala syntax (think sbt, scala or dotty). The dotty dialect does not yet support trait parameters.
  
A good place for its implementation is  https://github.com/scalameta/scalameta/blob/master/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L3061.
Use `classDef` as an inspiration: https://github.com/scalameta/scalameta/blob/master/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L3074.

cb372:
Fixes #493

* Add a flag to `Dialect`
* Update logic in `ScalametaParser` to look at the flag
* Comment out an invariant check in AST class because it's no longer an invariant
* Add a test

Commenting out the invariant check is a shame, but it was unavoidable because it's not currently possible to access the parser's dialect in order to check the appropriate flag.

### DIFF
--- a/scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
@@ -51,6 +51,10 @@ import scala.compat.Platform.EOL
   // What level of quoting are we at?
   // The underlying data structure captures additional information necessary for parsing.
   def metalevel: Metalevel
+
+  // Are trait allowed to have parameters?
+  // They are in Dotty, but not in Scala 2.12 or older.
+  def allowTraitParameters: Boolean
 }
 
 package object dialects {
@@ -63,6 +67,7 @@ package object dialects {
     def allowToplevelTerms = false
     def toplevelSeparator = ""
     def metalevel = Metalevel.Zero
+    def allowTraitParameters = false
     private def writeReplace(): AnyRef = new Dialect.SerializationProxy(this)
   }
 
@@ -75,6 +80,7 @@ package object dialects {
     def allowToplevelTerms = true
     def toplevelSeparator = EOL
     def metalevel = Metalevel.Zero
+    def allowTraitParameters = false
     private def writeReplace(): AnyRef = new Dialect.SerializationProxy(this)
   }
 
@@ -87,6 +93,7 @@ package object dialects {
     def allowToplevelTerms = true
     def toplevelSeparator = ""
     def metalevel = Metalevel.Zero
+    def allowTraitParameters = false
     private def writeReplace(): AnyRef = new Dialect.SerializationProxy(this)
   }
 
@@ -99,6 +106,7 @@ package object dialects {
     def allowToplevelTerms = Scala210.allowToplevelTerms
     def toplevelSeparator = Scala210.toplevelSeparator
     def metalevel = Metalevel.Zero
+    def allowTraitParameters = false
     private def writeReplace(): AnyRef = new Dialect.SerializationProxy(this)
   }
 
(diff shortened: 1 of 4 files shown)


scalameta/trees/src/main/scala/scala/meta/Trees.scala
```diff
@@ -373,7 +373,14 @@ object Defn {
                    templ: Template) extends Defn with Member.Type {
     // TODO: hardcoded in the @ast macro, find out a better way
     // require(templ.stats.getOrElse(Nil).forall(!_.is[Ctor]))
-    require(ctor.mods.isEmpty && ctor.paramss.isEmpty)
+
+    // TODO this doesn't work because the Dialect in implicit
```
olafurpg: We don't have a dialect at this point.

According to the spec it seems mods are allowed http://docs.scala-lang.org/sips/pending/trait-parameters.html

scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3108,7 +3108,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
   // therefore, I'm going to use `Term.Name("this")` here for the time being
 
   def primaryCtor(owner: TemplateOwner): Ctor.Primary = autoPos {
-    if (owner.isClass) {
+    if (owner.isClass || (owner.isTrait && dialect.allowTraitParameters)) {
```
olafurpg: "trait or object don't take params" message

scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -75,6 +80,7 @@ package object dialects {
     def allowToplevelTerms = true
     def toplevelSeparator = EOL
     def metalevel = Metalevel.Zero
+    def allowTraitParameters = false
```
olafurpg: Scala210.allowTrait...

jvican:
Just for the record, some other runtime checks should be created to check that `traits` cannot have params **if** the Dotty dialect is not used. I don't consider this to be in the scope of this PR because it requires a more fundamental fix in meta.

olafurpg:
LGTM 👍

scalameta/trees/src/main/scala/scala/meta/Trees.scala
xeno-by: I don't think that we can have dialects in trees, unfortunately.

What if, for example, the tree is created manually, without any kind of parsing or quasiquoting? What dialect do we choose then?

However, what I think should be done is prohibiting trees like and types, or types, non-trivial trait constructors in contexts, where Scala211 is expected. For example, if you try to prettyprint such a tree in a Scala211 context, an exception must be thrown.

Wdyt?

olafurpg: > For example, if you try to prettyprint such a tree in a Scala211 context, an exception must be thrown.

I agree, it seems we need to do the validation in some other place. However, I think scala.meta should **never** throw exceptions. I would prefer to have a `.validate(dialect)` method that I can call before `.syntax`. We can discuss different strategies later.

PR #500, opened 2016-10-03
Fix #496: Remove view bounds support in Dotty dialect

### LINKED ISSUE #496: Remove view bounds support in Dotty dialect
View bounds have been completely removed in Dotty. The Dotty parser still supports them, so a good idea would be to remove such support. The Dotty dialect should fail for something like:
```scala
def foo[T <% Whatever](t: T) = ???
```
  
Current implementation: https://github.com/scalameta/scalameta/blob/master/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L2782

jonas:
References #269

### DIFF
--- a/scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
@@ -44,6 +44,10 @@ import scala.compat.Platform.EOL
   // Necessary to support popular script-like DSLs.
   def allowToplevelTerms: Boolean
 
+  // Are view bounds supported by this dialect?
+  // Removed in Dotty.
+  def allowViewBounds: Boolean
+
   // What kind of separator is necessary to split top-level statements?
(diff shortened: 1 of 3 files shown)


scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -78,6 +83,7 @@ package object dialects {
     def allowInline = false
     def allowSpliceUnderscore = Scala210.allowSpliceUnderscore
     def allowToplevelTerms = true
+    def allowViewBounds = true
```
olafurpg: Please reference `Scala210.allowViewBounds` in all dialects where the setting is true.

jonas: OK, missed that pattern. Made `Paradise211` reference `Scala211`.

xeno-by:
LGTM

PR #485, opened 2016-09-18
Fix #470 and add test suite for modifiers

jvican:
Ready for review.

PR #507, opened 2016-10-10
Fix #470 and add test suite for modifiers

### LINKED ISSUE #470: scala.meta parser doesn't validate allowed modifiers

xeno-by:
Supersedes https://github.com/scalameta/scalameta/pull/485


jvican:
:+1:

PR #510, opened 2016-10-12
Fix #46: Enable comments inside quasiquotes

### LINKED ISSUE #46: Comments in quasiquotes
``` scala
q"""
object $name {
$unquotee // single line
/* multi
   lines */
"""
```

@densh pretty please :)

jvican:
Quasiquotes preserve now their parsed origin and `tokens` and `syntax` allow you
to access any comments from the original source.

The fix consists in tokenizing, getting the token positions, creating
the origin `Origin.Parsed` and setting it to term quasiquotes. Pattern
quasiquotes cannot benefit from this, because comments are only
preserved in the token level, not the tree level. Therefore, one cannot
pattern match on a quasiquote with a comment (because of `unapply`).

There are some tests failing because `show[Structure]`, implemented
in `TreeStructure.scala`, does not handle `Unquote`. Therefore, it
outputs an empty string when the spliced quasiquote variable should be.
Fixing this requires changing the accessibility of Unquote and
"magically" detecting the value of that spliced variable. The actual
meta tree has all this information, but I find it difficult to map it to
the tokens. Not a trivial task. Maybe we should rethink what `show[Structure]`
means and what's its actual relation with tokens (not very clear to me, atm).

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -78,7 +78,8 @@ object build extends Build {
     base = file("scalameta/inputs")
   ) settings (
     publishableSettings,
-    description := "Scala.meta's APIs for source code in textual format"
+    description := "Scala.meta's APIs for source code in textual format",
+    enableMacros
   ) dependsOn (common)
 
   lazy val parsers = Project(
--- a/scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
@@ -140,14 +140,14 @@ package object dialects {
     def toplevelSeparator = underlying.toplevelSeparator
   }
 
-  @leaf private[meta] class QuasiquoteTerm(underlying: Dialect, multiline: Boolean) extends Quasiquote {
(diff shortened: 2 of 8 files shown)


jvican:
Maybe the good solution is to decouple `Structure` from `tokens`... But I'm unsure until which points this is correct, I understand that scala meta is trying to make quasiquotes a real compile-time construct.
  
My change in `syntax` respects how `"val a = 1 // comment".parse[Source].get.syntax` behaves, instead of outputting the result of the spliced variables. This increases the number of test failures from 9 to 16, but I think it's the right thing to do if we really want "quasiquotes comments" in the syntax.

jvican:
Is there any update on this?

xeno-by:
Not yet, unfortunately. Recently, I've been focusing on the scalac -> meta converter. I'll do my best to get back to this soon.

xeno-by:
It looks like pushing this PR further will require a significant design effort that I don't have time for right now. At the moment, the main and only priority is shipping semantic API asap, so I'll close this PR for the time being.

PR #521, opened 2016-10-15
split Type.Compound into Type.With and Type.Refine

xeno-by:
/cc @olafurpg @liufengyun


xeno-by:
Fast tracking to merge this PR, because people need to base their work off of it. Please let me know if you don't like something here, though.

PR #522, opened 2016-10-18
converter-driven changes

xeno-by:


olafurpg:
Is it worth renaming the import fields now? The roles I won't miss 😺

olafurpg:
The CI has now been fixed so the build should pass.

xeno-by:
@olafurpg Since field names have become a public API, I think it is important to polish them.

changelog/1.3.0.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Trees
+
+  * Type.Compound has been split into `Type.With` and `Type.Refine` in order to provide common ground for compatibility with Dotty. Dotty gets rid of `with` in favor of intersection types, but it still has type refinements, so we needed to extract `Type.Refine`, so that it can be used in both dialects.
+
+    ```
+    scala> t"A with B".structure
+    res0: String = Type.With(Type.Name("A"), Type.Name("B"))
+
+    scala> t"{ def x: Int }".structure
```
olafurpg: Why is it `Type.Refine` and not `Type.Structural`?

jvican: I think it's bencause the spec considers them *Refinements*: http://scala-lang.org/files/archive/spec/2.11/03-types.html#compound-types

xeno-by: The thing in curly braces is called type refinement in the Scala Language Specification.

PR #528, opened 2016-10-31
make it obvious what a pattern is

MasseGuillaume:
I was doing some like:

```scala
private def instrument(source: Source, output: Term.Name): Source = {
    val main = 
      q"""
      object Main {
        val $output = scala.collection.mutable.Map.empty[(Int, Int), (String, String)]
      }
      """
}
```

and I ran into the following error:

```
[error] /home/gui/center/scastie/instumentation/src/main/scala/instumentation/Instument.scala:49: can't unquote a name here, use a pattern instead
[error]       val $output = scala.collection.mutable.Map.empty[(Int, Int), (String, String)]
[error]           ^
```

I want to make obvious that a pattern is a `scala.meta.Pat` in my case `scala.meta.Pat.Var(output)`

solved like this:

```scala
private def instrument(source: Source, output: Term.Name): Source = {
   val pat = Pat.Val.Term(output)
    val main = 
      q"""
      object Main {
        val $pat = scala.collection.mutable.Map.empty[(Int, Int), (String, String)]
      }
      """
}
```

### DIFF
--- a/scalameta/common/src/main/scala/scala/meta/internal/ast/Liftables.scala
+++ b/scalameta/common/src/main/scala/scala/meta/internal/ast/Liftables.scala
@@ -82,7 +82,7 @@ class LiftableMacros(override val c: Context) extends AdtLiftableMacros(c) with 
           pat match {
             case q: _root_.scala.meta.internal.ast.Quasi if unquotesName(q) =>
               val action = if (q.rank == 0) "unquote" else "splice"
-              c.abort(q.pos, "can't " + action + " a name here, use a pattern instead")
+              c.abort(q.pos, "can't " + action + " a name here, use a scala.meta.Pat instead, for example Pat.Var.Term(name: Term.Name)")
             case _ =>
           }
         }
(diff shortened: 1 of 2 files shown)


scalameta/common/src/main/scala/scala/meta/internal/ast/Liftables.scala
```diff
@@ -82,7 +82,7 @@ class LiftableMacros(override val c: Context) extends AdtLiftableMacros(c) with
           pat match {
             case q: _root_.scala.meta.internal.ast.Quasi if unquotesName(q) =>
               val action = if (q.rank == 0) "unquote" else "splice"
-              c.abort(q.pos, "can't " + action + " a name here, use a pattern instead")
+              c.abort(q.pos, "can't " + action + " a name here, use a pattern (scala.meta.Pat) instead")
```
olafurpg: How about giving even more information? Like

```
use a scala.meta.Pat instead, for example Pat.Var.Term(name: Term.Name)
```

xeno-by: Sounds good! @MasseGuillaume, could you update the PR please?

MasseGuillaume: oh it look like I have to update the test also

olafurpg:
LGTM 👍 Thanks @MasseGuillaume!

PR #532, opened 2016-11-09
Add collection API implementations that preserve Tokens type

ShaneDelmore:


ShaneDelmore:
As this is my first PR to scala meta feedback would be greatly appreciated.  I tried to fit in with the existing style as much as I could but let me know if I need to change my coding or testing style.

I would like to add a few methods still to deal with the fact that we will do things that are not standard with collections when manipulating the sequences from both ends, wrapping or unwrapping them but I thought I had done enough that it was time to get some feedback to make sure I was not going in the wrong direction.

xeno-by:
LGTM.

olafurpg:
Thanks @ShaneDelmore !

xeno-by:
Looks like while LGTMing I forgot that now we have a changelog. Fixed my mistake in https://github.com/scalameta/scalameta/pull/538.

PR #533, opened 2016-11-09
Fix 404 link to old workshop

olafurpg:


olafurpg:
Confirmed that the new link is OK.

PR #541, opened 2016-11-12
Reject invalid var/val definitions (fixes #529)

### LINKED ISSUE #529: Parser should not accept: val x: Int = _
```scala
> val x: Int = _
cmd6.sc:264: unbound placeholder parameter
val x: Int = _
             ^
Compilation Failed
> q"val x: Int = _"
res6: Defn.Val = val x: Int = _ // <-- Expected parse error: unbound placeholder parameter
```

cb372:
Parser should not accept the following:

```
@ var x = _
cmd4.sc:160: unbound placeholder parameter
var x = _
        ^
Compilation Failed
```

```
@ val x: Int = _
cmd4.sc:160: unbound placeholder parameter
val x: Int = _
             ^
Compilation Failed
```


olafurpg:
This looks great! Thank you @cb372 for this contribution.

PR #542, opened 2016-11-12
Add a dialect-aware `validate` method to Tree

cb372:
Following on from the discussion on #498 about how to validate trees for a given dialect, I propose adding a `validate` method to `Tree`. This would replace all the `require` statements, so we no longer throw exceptions, and it also allows validation to take account of the dialect, which is currently not possible.

The downside is that it becomes possible to build invalid trees. The user becomes responsible for calling `validate` on any trees they build.

Things to discuss:

* What should the return type of `validate` be? For now I just made it a `Seq` of error messages, with an empty list representing validity.
* Where should `validate` be called? I guess the pretty-printers will want to validate trees before printing them, but I don't have any experience with that area of the codebase.

If this looks reasonable, I can rewrite all the other `require`s similarly.

### DIFF
--- a/scalameta/scalameta/src/test/scala/scala/meta/tests/parsers/DottySuite.scala
+++ b/scalameta/scalameta/src/test/scala/scala/meta/tests/parsers/DottySuite.scala
@@ -40,6 +40,8 @@ class DottySuite extends ParseSuite {
   test("trait parameters are allowed") {
     val tree = dialects.Dotty("trait Foo(bar: Int)").parse[Stat].get
     assert(tree.syntax === "trait Foo(bar: Int)")
+    assert(tree.validate(dialects.Dotty).isEmpty)
+    assert(tree.validate(dialects.Scala211).nonEmpty)
   }
 
   test("view bounds not allowed") {
--- a/scalameta/trees/src/main/scala/scala/meta/Trees.scala
+++ b/scalameta/trees/src/main/scala/scala/meta/Trees.scala
@@ -19,6 +19,9 @@ import scala.meta.internal.ast.Helpers._
   def pos: Position
   def tokens(implicit dialect: Dialect): Tokens
 
+  // TODO have a proper Success/Failure result type?
+  def validate(implicit dialect: Dialect): Seq[String] = Nil
+
   final override def canEqual(that: Any): Boolean = this eq that.asInstanceOf[AnyRef]
   final override def equals(that: Any): Boolean = this eq that.asInstanceOf[AnyRef]
   final override def hashCode: Int = System.identityHashCode(this)
@@ -375,13 +378,11 @@ object Defn {
     // TODO: hardcoded in the @ast macro, find out a better way
     // require(templ.stats.getOrElse(Nil).forall(!_.is[Ctor]))
 
-    // TODO this doesn't work because the Dialect in implicit
-    // scope is the dialect of the host Scala environment
-    // (i.e. Scala211), not the parser's dialect.
-    //require (
-    //  implicitly[Dialect].allowTraitParameters ||
-    //    (ctor.mods.isEmpty && ctor.paramss.isEmpty)
-    //)
+    override def validate(implicit dialect: Dialect): Seq[String] = {
+      if (!dialect.allowTraitParameters && (ctor.mods.nonEmpty || ctor.paramss.nonEmpty))
+        Seq("Trait parameters are not allowed in this dialect")
+      else Nil
+    }
   }
   @ast class Object(mods: Seq[Mod],
                     name: Term.Name,


olafurpg:
This is an interesting proposition. I like the idea of replacing exceptions with regular values. However, some thoughts

* it's now under the responsibility of users to validate Trees, which means they could forget to validate their trees.
* invalid trees won't fail-fast anymore, which might make it harder to track down a bug.
* tree validation is already done in parsing/quasiquotes, so this would only affect manual tree construction
* we might want to introduce a another method `validateAll(implicit dialect: Dialect): Either[Seq[(Tree, Seq[String])], this.type]` that runs validate on all child nodes and includes the culprit tree with it's validation error messages. The error type is quite unwieldy so we might want to introduce some `Error` type with good pretty printing.

I am curious to know what @xeno-by thinks of this change.

cb372:
>tree validation is already done in parsing/quasiquotes, so this would only affect manual tree construction

Interesting. Doing validation in 2 places seems sub-optimal, so maybe the existing validation logic could be moved into here? Could you point me to an example of it?

olafurpg:
Okay, my bad, the same validation is not exactly run during parsing but rather we guard against it. For example,
```
> q"trait T(a: Int)"(dialects.Scala211)
cmd1.sc:1: traits or objects may not have parameters
val res1 = q"trait T(a: Int)"
                    ^
Compilation Failed
> q"trait T(a: Int)"(dialects.Dotty)
res1: Defn.Trait = trait T
```
We can do the validation during parsing because there we require a dialect. Quasiquotes use the parser to my best understanding.

jvican:
Some validation checks need to be done at runtime, because at the time of parsing they cannot be checked. Although the validation in two different places may be suboptimal, it's key that invalid trees are catched as soon as possible, and as far as I know this is a feature of Scala Meta, not an implementation detail.
  
That being said, I would welcome the validate tree function. There could be an explicit option for those users that don't want Scala Meta to check for correct trees and that prefer to manually handle errors. But this option would coexist with the runtime checks that we perform right now, which I would suggest are enabled by default.

xeno-by:
The inability to construct invalid abstract syntax trees (also called "safety by construction" by @densh) is a core feature of Scala Meta. 

In the early days of the project, we tried to enforce this with types, but that was too much notational overhead, so we replaced some compile-time checks with runtime checks during tree creation. There are additional checks [in quasiquotes](https://github.com/xeno-by/scalameta/blob/f9afef2bcdce93cc344d240b7bc0ed96c06a5306/scalameta/common/src/main/scala/scala/meta/internal/ast/Liftables.scala#L75) and [in parser](https://github.com/xeno-by/scalameta/blob/f9afef2bcdce93cc344d240b7bc0ed96c06a5306/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L3162), but they are complementary to the main checks.

@cb372 Could you discuss the benefits of allowing invalid trees a bit more?

olafurpg:
The benefits from my perspective is that users are not surprised by runtime exceptions. It's quite un-idiomatic to use exceptions for data validation in Scala. It might be preferable to allow invalid trees and provide the validation as an optional safety step.

jvican:
From the perspective of a macro writer, I do want meta to throw exceptions at me. Not because I'm a masochist, but because this is what Dotty and Scalac do and stack traces help me **a lot**. I can see the neatness of having a second validation step, though. It would be useful, for instance, if Scala Meta is used as a framework and the input depends on the end user. The developerof the framework may want to control the exceptions and pattern match on concrete error adt nodes.
  
To unify these two approaches, I would propose to introduce two modes. The default mode, which is our current one, will check at runtime the created trees. The second mode will allow to disable runtime checks and let users validate the trees themselves. Ideally, this could be done with macros so that the validation code at the node creation site disappears. Implementation details would need to work out, these are just rough suggestions.

cb372:
I agree that using the type system to make it impossible to create invalid trees in the first place would be ideal, but if that has already been tried and is infeasible then fair enough.

I think I was trying to address two separate issues with this PR:

1. I found it vaguely unsettling that scala.meta might throw exceptions, but I didn't realise that it was by design. I quite like @jvican's suggestion of allowing users to control how validation happens. Having the option to explicitly perform validation might be useful for e.g. IDEs and other tooling.

2. There are some cases in which it's impossible to do validation at tree construction time, because the tree is valid in some dialects and invalid in others. e.g. `trait Foo(a: Int)` is valid in Dotty but invalid in Scala 2.11, and vice versa for `object A { self: A => }`. I can think of a couple of ways to handle this:

    * Add a validate method like I did in this PR, which validates the tree in the context of a given dialect
    * Give the tree a dialect at construction time. In other words, instead of creating a generic tree and then trying to use it in a given dialect, specify the dialect up front. (I imagine this would be quite a big design change for scala.meta.)

densh:
> The benefits from my perspective is that users are not surprised by runtime exceptions.

Before, in Scala Reflect days, one would often get an obscure error somewhere in backend that are caused by bugs in your macro. There are very limited checks on AST well-formedness for synthetic (i.e. compiler or macro generated) code making failures extremely late and *truly surprising*. 

Failing early on nonsensical trees is a killer feature as it tells you precisely where you screwed up at the time you did it. Delaying the checks would allow you to create complicated blob of code without clear understanding which part exactly is at fault to fail validation. That would bring us back to the place where we started.

jvican:
I think that the implicit approach, while may give some suprises, it's the adequate design for Meta. I like the fact that some of the semantics are separated from the tree structure, because validation/transformation requires the *interpretation* of the tree nodes. You can only interpret them in the presence of `dialect`, which give you a context. IMO, this makes sense to me.
  
I would like to know first what @olafurpg and @xeno-by think about the introduction of the two validations modes, but I would lean towards adding validation methods that depend on a `dialect` and that return concrete errors represented in an ADT. I would put all these in a package called `validation, instead of defining them right next to the tree nodes definition. Also, I think it would make sense to explore ways to achieve this with macros, so that the runtime logic can be removed when the user opts in for manual validation.

jvican:
> Failing early on nonsensical trees is a killer feature as it tells you precisely where you screwed up at the time you did it. Delaying the checks would allow you to create complicated blob of code without clear understanding which part exactly is at fault to fail validation. That would bring us back to the place where we started.
  
Yeah, I agree. Making this the default validation mode makes sense. But there are some compelling use cases for developers that may want to use Meta in their framework/IDE. In the light of these benefits, do you think adding an opt-in, exception-free validation step would make sense? @densh

olafurpg:
@densh I agree that failing early on invalid trees is a killer feature of scala.meta and less surprising than getting cryptic error messages in the scalac backend. My comment above was only to say that runtime errors may nevertheless still surprise users.

Some thoughts:

- I'm in favor of @cb372's approach in this PR to provide an opt-in validation step with the presence of a dialect.
- I suggest to keep the non-dialect related validations as runtime checks in the constructor. Why?
  - Adding a dialect in the constructor would be a huge change in the scala.meta api and I'm not convinced (yet) that the benefits outweighs the introduced complexity.
  - Moving the non-dialect-aware validations to the `validate` method would delay critical validation to catch bugs that, for example, manifest themselves later as cryptic error messages in the scalac backend.
- I would like the dialect-aware validation to run recursively on all children nodes and the error message should attach the offending tree nodes. Question: should the validation accumulate all errors or fail on first error?
- Here is a commented out check that requires a dialect: https://github.com/cb372/scalameta/blob/cf026df6b4676ba80abb8d4f661b58c789459fc6/scalameta/trees/src/main/scala/scala/meta/Trees.scala#L507

xeno-by:
Hi everyone! Since we haven't arrived at a concrete conclusion in quite a while, I'll be closing this PR and moving the discussion to a dedicated issue: https://github.com/scalameta/scalameta/issues/599.

I'd love to revisit this matter personally in the future, but at the moment my main and only priority is shipping the semantic API asap. Let's stay in touch.

PR #546, opened 2016-11-15
Fix definition for class

mpollmeier:
I'm new to scalameta, but this looks like a typo. The proposed version matches fine on my class, the previous results in a MatchError

### DIFF
--- a/notes/quasiquotes.md
+++ b/notes/quasiquotes.md
@@ -153,7 +153,7 @@ Literal  | `q"$lit"` (construction only), `q"${lit: Lit}"` (also deconstruction)
  Def            | `q"..$mods def $name[..$tparams](...$paramss): $tpeopt = $expr"`
  Macro          | `q"..$mods def $name[..$tparams](...$paramss): $tpeopt = macro $expr"`
  Type           | `q"..$mods type $tname[..$tparams] = $tpe"`
- Class          | `q"..$mods class $tname[..$tparams] ..$mods (...$paramss) extends $template"`
+ Class          | `q"..$mods class $tname[..$tparams] (...$paramss) extends $template"`
  Trait          | `q"..$mods trait $tname[..$tparams] extends $template"`
  Object         | `q"..$mods object $name extends $template"`
  Package Object | `q"package object $name extends $template"`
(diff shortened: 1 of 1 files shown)


olafurpg:
Thanks for this PR! The primary constructor for a class can have modifiers

```scala
> val q"..$classMods class $tname[..$tparams] ..$constructorMods (...$paramss) extends $template" = q"class A private (a: Int)"
classMods: collection.immutable.Seq[Mod] = List()
tname: Type.Name = A
tparams: collection.immutable.Seq[Type.Param] = List()
constructorMods: collection.immutable.Seq[Mod] = List(private) // <-- here
paramss: collection.immutable.Seq[collection.immutable.Seq[Term.Param]] = List(List(a: Int))
template: Template =
```

I agree the docs should not define `mods` twice, however. Any idea what would be better name for the second `mods`?

mpollmeier:
That makes sense, thanks!

So we should change the naming in the docs since it can't ever work right, since it's using the same name for two variables. I would not change the first `mods` since it's nicely aligned with the other examples in this section :)
How about we just rename the second one to `constructorMods` or `ctorMods`? I can update the PR - just signed the CLA as well.

olafurpg:
I think `ctorMods` is more consistent with the scala.meta naming convention. LGTM 👍 thanks!

notes/quasiquotes.md
```diff
@@ -153,7 +153,7 @@ Literal  | `q"$lit"` (construction only), `q"${lit: Lit}"` (also deconstruction)
  Def            | `q"..$mods def $name[..$tparams](...$paramss): $tpeopt = $expr"`
  Macro          | `q"..$mods def $name[..$tparams](...$paramss): $tpeopt = macro $expr"`
  Type           | `q"..$mods type $tname[..$tparams] = $tpe"`
- Class          | `q"..$mods class $tname[..$tparams] ..$mods (...$paramss) extends $template"`
+ Class          | `q"..$mods class $tname[..$tparams] $ctorMods (...$paramss) extends $template"`
```
xeno-by: Is the change from `..$` to `$` accidental or intentional?

xeno-by: /cc @olafurpg

PR #553, opened 2016-11-23
support for Scala 2.12

xeno-by:
Enables cross-compilation against 2.11.8 and 2.12.0 with the same codebase.

At the moment, `crossScalaVersions` is still set to `Seq(2.11.8)`, because we can't compile the `readme` Scalatex project under 2.12 (https://github.com/lihaoyi/Scalatex/issues/38), so our CI doesn't yet publish 2.12 builds automatically.


xeno-by:
@dwijnand How do I conditionally disable lines 30 and 34 in https://github.com/xeno-by/scalameta/blob/ca99041624600c57136ec5caeb4ab64a96491fa4/project/build.scala when running under Scala 2.12? I tried to wrap these lines in `if (scalaVersion.value.startsWith("2.11"))`, but that didn't help.

dwijnand:
You need to use Def.taskDyn to avoid the side-effect:

http://www.scala-sbt.org/0.13/docs/Howto-Dynamic-Task.html

dwijnand:
Here's a better link http://www.scala-sbt.org/0.13/docs/Tasks.html#Dynamic+Computations+with

xeno-by:
@dwijnand It works, thank you! https://github.com/scalameta/scalameta/pull/554

PR #554, opened 2016-11-23
fix the build for 2.12

xeno-by:


xeno-by:
@fommil Is there way to get java 1.8 out of `ensime/ensime:v2.x-cache`?

fommil:
Use jenv. Create a .java-version file with the version of Java you want.

fommil:
I don't use jenv global, just create the file. It should be the default, do you maybe have one already? Have a look in pcplod's build do so how i do it.

xeno-by:
@fommil Thank you! It worked.

Issue #561, opened 2016-12-13
Regression in parsing wildcard functions


liufengyun:
I just upgraded to 1.4.0 and ScalaMeta is no longer able to parse the following valid code:

```Scala
val f: Int => String = _.toString
```

```
error: unbound placeholder parameter
```

xeno-by:
@cb372 Looks like this regression is caused by https://github.com/scalameta/scalameta/commit/070b9eb84507d82f39f364399b8fd3b549a93ce3. Could you take a look?

olafurpg:
Whoops, I need to be more careful reviewing. In fact, I think a test suite parsing millions of lines of Scala code would have caught this regression. I opened #562.

I can also take a look at this @liufengyun if you like?

liufengyun:
Thanks @olafurpg , that's very nice of you to take a look at this, I'm now focusing on another Dotty issue.

xeno-by:
Fixed in 1.5.0.563.

PR #563, opened 2016-12-13
Fix parsing wildcard functions

cb372:
This is a fix for #561, the regression in parsing `val f: Int => String = _.toString` which was introduced in #541.

It also fixes parsing of `var f: Int => String = _.toString`, which was broken even before #541.

The underlying problem was that the decision on whether a statement had an RHS was based on whether the first token after the `=` was an underscore. But this is not sufficient, because an RHS expression can start with an underscore.

My solution is to first parse the RHS as an expression, then check if the resulting tree is `Placeholder`. Happy to take expert advice on whether this is a reasonable strategy.

@xeno-by @olafurpg review please
@liufengyun Sorry for the regression


xeno-by:
LGTM

liufengyun:
Thanks for the fix @cb372 👍

olafurpg:
Thanks @cb372 ! :)

PR #566, opened 2016-12-14
2.x rebase

xeno-by:


xeno-by:
Improves upon https://github.com/scalameta/scalameta/pull/564 by adding `semantic` to the builds aggregated by `root`. Not doing that was an oversight.

scalameta/semantic/src/main/scala/scala/meta/semantic/Errors.scala
```diff
@@ -0,0 +1,37 @@
+package scala.meta
+package semantic
+
+import org.scalameta.adt._
+import org.scalameta.data._
+import scala.meta.inputs._
+import scala.meta.internal.inputs._
+
+@root trait Completed[+T] {
```
xeno-by: Internally, scalac calls the process of obtaining semantic information `completion`. Therefore, I just took this name and rolled with it. 

Judging by your reaction, this name is obscure, which is not good. Any ideas about replacing it?

PR #568, opened 2016-12-15
Fix infinite recursion.

olafurpg:
While working on building a scala.meta testkit, I noticed that the
pretty-printer would get stuck in an infnite loop. I can't find an
example that triggered the infinite loop, but this commit fixes it.

All tests still pass.


xeno-by:
LGTM

PR #570, opened 2016-12-19
support Unlift.unapply

liufengyun:
Unlift.unapply is required in order to implement Quasiquotes in Dotty. Review @xeno-by .


xeno-by:
LGTM

xeno-by:
This will become available after http://drone.geirsson.com:8001/scalameta/scalameta/80 finishes. Scroll down the build log to find out the build number.

PR #573, opened 2016-12-20
support lift/unlift String <-> Name

liufengyun:
Lift/unlift String <-> Name. Review @xeno-by .

### DIFF
--- a/scalameta/quasiquotes/src/main/scala/scala/meta/quasiquotes/Lift.scala
+++ b/scalameta/quasiquotes/src/main/scala/scala/meta/quasiquotes/Lift.scala
@@ -24,6 +24,8 @@ object Lift {
   implicit def liftNull[I >: Lit]: Lift[Null, I]             = Lift{ x => Lit(x) }
   implicit def liftUnit[I >: Lit]: Lift[Unit, I]             = Lift{ x => Lit(x) }
 
+  implicit def liftString2Name[O <: String, I >: Name]: Lift[O, I] = Lift{ x => Term.Name(x) }
+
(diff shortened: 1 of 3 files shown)


xeno-by:
Hi @liufengyun! Could you elaborate on the motivation behind this change?

liufengyun:
@xeno-by The motivation is the following code:

```Scala
  def generate(tps: Seq[Type]): Seq[Term.Param] =
     for {
       (tp, index) <- tps.zipWithIndex
       pname = Term.Name(name(index))
    } yield param"$pname : $tp"
```

https://github.com/liufengyun/eden/commit/fd8757e23e013f483e84e5fea7dc6e2727089216#diff-21fcd9fd298722101162657b26c281b6R87

liufengyun:
It seems impossible to harmonize `liftString2Name` and `liftString`, I'm not sure if it is a good idea to introduce the implicits.

xeno-by:
Honestly, I'm not a fan of the idea to lift strings to names. First, because of the conflicts between `liftString2Name` and `liftString`.

Secondly, because of the previous negative experience with a similar lifting in scalac codebase. During the 2.10 lifecycle, we used to have similar implicit conversions, and they sometimes created a lot of confusion. Gradually, the conversions were phased out, until they were evicted from (almost?) everywhere in the codebase.

I agree that the current syntax is a bit boilerplatey, but at least it's unambiguous. I think, in this particular case, given the reasons above, it makes sense to choose explicitness.

/cc @olafurpg

liufengyun:
@xeno-by Thanks for the explanation, it makes perfect sense to me. Close this for now.

Issue #572, opened 2016-12-20
toString generates invalid code


iTakeshi:
Hi, 

```scala
val tree=
  Term.Apply(
    Term.Name("foo"),
    Seq(
      Term.Function(
        Seq(Term.Param(Seq(Mod.Implicit()), Term.Name("i"), Some(Type.Name("Int")), None)),
        Lit(())
      )
    )
  )
println(tree.toString)
```
will be `foo(implicit i: Int => ())`. This code results in a compile error.
```
<console>:1: '=>' expected but ')' found.
foo(implicit i: Int => ())
                         ^
```

This should be printed as `foo { implicit i: Int => () }`.

olafurpg:
The pretty printer contains a ton of bugs, see https://github.com/scalameta/scalameta/issues/567#issuecomment-267061270. I suspect it's not too hard to fix most of these, it boils down to finding the right branch in TreeSyntax (e.g., [here](https://github.com/scalameta/scalameta/blob/ff054d55f28c63c860f56fea74ae67adbc7fc791/scalameta/trees/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala#L234)) and then add a test in syntax suite like [here](https://github.com/scalameta/scalameta/blob/ff054d55f28c63c860f56fea74ae67adbc7fc791/scalameta/scalameta/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala#L470-L473).

iTakeshi:
I'm just inspecting around there and found the cause, preparing PR. Thanks @olafurpg

iTakeshi:
solved by #574

PR #574, opened 2016-12-20
fix "foo { i: Int => () }".toString

iTakeshi:
Term.Apply with function as its arg is printed using braces instead of parenthesis.
solves #572

### DIFF
--- a/scalameta/scalameta/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
+++ b/scalameta/scalameta/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
@@ -472,11 +472,13 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
     assert(t"((a, b), c) => c".syntax == "((a, b), c) => c")
   }
 
-  test("Term.Apply(_, Seq(Term.Function(...))) #572") {
+  test("Term.Apply(_, Seq(Term.Function(...))) #572, #574") {
     import scala.collection.immutable.Seq
-    val tree1= Term.Apply(Term.Name("foo"), Seq(Term.Function(Seq(Term.Param(Seq(), Term.Name("i"), Some(Type.Name("Int")), None)), Lit(()))))
-    val tree2= Term.Apply(Term.Name("foo"), Seq(Term.Function(Seq(Term.Param(Seq(Mod.Implicit()), Term.Name("i"), Some(Type.Name("Int")), None)), Lit(()))))
-    assert(tree1.toString == "foo { (i: Int) => () }")
-    assert(tree2.toString == "foo { implicit i: Int => () }")
+    val tree1 = Term.Apply(Term.Name("foo"), Seq(Term.Function(Seq(Term.Param(Seq(), Term.Name("i"), Some(Type.Name("Int")), None)), Lit(()))))
+    val tree2 = Term.Apply(Term.Name("foo"), Seq(Term.Function(Seq(Term.Param(Seq(Mod.Implicit()), Term.Name("i"), Some(Type.Name("Int")), None)), Lit(()))))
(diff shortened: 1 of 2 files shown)


scalameta/scalameta/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
```diff
@@ -471,4 +471,12 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
     assert(t"((a, b)) => c".syntax    == "((a, b)) => c")
     assert(t"((a, b), c) => c".syntax == "((a, b), c) => c")
   }
+
+  test("Term.Apply(_, Seq(Term.Function(...))) #572") {
+    import scala.collection.immutable.Seq
+    val tree1= Term.Apply(Term.Name("foo"), Seq(Term.Function(Seq(Term.Param(Seq(), Term.Name("i"), Some(Type.Name("Int")), None)), Lit(()))))
+    val tree2= Term.Apply(Term.Name("foo"), Seq(Term.Function(Seq(Term.Param(Seq(Mod.Implicit()), Term.Name("i"), Some(Type.Name("Int")), None)), Lit(()))))
+    assert(tree1.toString == "foo { (i: Int) => () }")
```
olafurpg: - Any reason to use `Term.Apply(...)` instead of quasiquotes? It's the same, but I think quasiquotes are more readable. 
- also, `.syntax` instead of `.toString`

scalameta/trees/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -518,8 +518,9 @@ object TreeSyntax {
 
       // Multiples and optionals
       implicit def syntaxArgs: Syntax[Seq[Term.Arg]] = Syntax {
-        case (b: Term.Block) :: Nil => s(" ", b)
-        case args                   => s("(", r(args, ", "), ")")
+        case (b: Term.Block) :: Nil    => s(" ", b)
+        case (f: Term.Function) :: Nil => s(" { ", f, " }")
```
olafurpg: Should we maybe restrict this to Term.Function where all params have a defined `decltpe`?

```scala
// OK
@ List(1).map(implicit a => a + 2)
res24: List[Int] = List(3)
// Not OK
@ List(1).map(implicit a: Int => a + 2)
<console>:1: identifier expected but integer literal found.
List(1).map(implicit a: Int => a + 2)                                   ^
```

scalameta/scalameta/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
iTakeshi: 1. using quasiquote, it is printed with `{}` even if this PR isn't merged. using `Term.Apply`, on the other hand, it is printed with `()` when this patch isn't applied.
2. got it

scalameta/trees/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
iTakeshi: I think you're right. updating.

olafurpg:
1. Makes sense. 

LGTM, thanks!

iTakeshi:
Thanks!
Is it on the bintray? (https://bintray.com/scalameta/maven/scalameta/1.5.0.571)

olafurpg:
Sry just saw now. It'll be published to bintray after ci completes on merge
into master with version number 3.0.0.574 (this pr number).
On Tue, 20 Dec 2016 at 17:01, iTakeshi <notifications@github.com> wrote:

Thanks!


Is it on the bintray? (
https://bintray.com/scalameta/maven/scalameta/1.5.0.571)




—
You are receiving this because you modified the open/close state.


Reply to this email directly, view it on GitHub
<https://github.com/scalameta/scalameta/pull/574#issuecomment-268297574>,
or mute the thread

PR #575, opened 2016-12-23
change signature of meta

liufengyun:
I think it makes more sense to have `meta: Any => Nothing`, so that user can any meaningful type annotations of def macros.


xeno-by:
LGTM

PR #579, opened 2016-12-28
make withOrigin available

liufengyun:
In order to carry positions in the conversion, we need to make `withOrigin` available for usage. Review @xeno-by .


xeno-by:
Would it be possible to move eden into a scala.meta-based package instead? E.g. `scala.meta.eden`?

I'm a bit cautious about exposing this method as part of the public API, because it's quite low-level.

liufengyun:
@xeno-by That makes sense, I'll rename the namespace. Eventually, I'd like eden to be part of the `scalameta` ecosystem, when it reaches the quality requirement.

PR #581, opened 2016-12-31
Fix TreeSyntax for macro definitions

eatkins:
I wasn't able to add def macros in an annotation macro because the generated tree for Defn.Macro contained a syntax error.

scala> import scala.meta._
import scala.meta._

scala> q"def foo = macro fooImpl"
res0: meta.Defn.Macro = def foo: = macro fooImpl

scala> q"def foo: Int = macro fooImpl"
res1: meta.Defn.Macro = def foo: : Int = macro fooImpl

After making this change, I was able to add def macros in my annotation
macros and the syntax looks correct in the console.

scala> import scala.meta._
import scala.meta._

scala> q"def foo = macro fooImpl"
res0: meta.Defn.Macro = def foo = macro fooImpl

scala> q"def foo: Int = macro fooImpl"
res1: meta.Defn.Macro = def foo: Int = macro fooImpl

I have no idea if there are some tests that should be updated
or if a new test should be added.


olafurpg:
Thank you for this contribution! A good place to test this would be here: https://github.com/scalameta/scalameta/blob/master/scalameta/scalameta/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala#L470-L473

I think it would be good to include a test for both with and without a `decltpe`, I couldn't see any tests for either. 

Related #567

eatkins:
Updated with a test.

Issue #583, opened 2017-01-05
17 Quasiquotes' constructed and deconstructed types don't match


tabdulradi:
I noticed couple of quasiquotes have incompatible constructed/deconstructed. 
```
[error] DeConstructionTypesSuite.scala:22: type mismatch when unquoting;
[error]  found   : scala.meta.Term.Arg
[error]  required: scala.meta.Term
[error]     case q"$expr1 $name[..$tpes] $expr2" => q"$expr1 $name[..$tpes] $expr2"
[error]                                                                     ^
[error] DeConstructionTypesSuite.scala:44: type mismatch when unquoting;
[error]  found   : scala.collection.immutable.Seq[scala.meta.Enumerator]
[error]  required: scala.collection.immutable.Seq[scala.meta.Enumerator.Generator]
[error]     case q"for (..$enumeratorsnel) $expr" => q"for (..$enumeratorsnel) $expr"
[error]                                                       ^
[error] DeConstructionTypesSuite.scala:45: type mismatch when unquoting;
[error]  found   : scala.collection.immutable.Seq[scala.meta.Enumerator]
[error]  required: scala.collection.immutable.Seq[scala.meta.Enumerator.Generator]
[error]     case q"for (..$enumeratorsnel) yield $expr" => q"for (..$enumeratorsnel) yield $expr"
[error]                                                             ^
[error] DeConstructionTypesSuite.scala:70: type mismatch when unquoting;
[error]  found   : scala.meta.Tree
[error]  required: scala.meta.Type
[error]     case targ"$tpe" => targ"$tpe"
[error]                             ^
[error] DeConstructionTypesSuite.scala:72: type mismatch when unquoting;
[error]  found   : scala.meta.Tree
[error]  required: scala.meta.Pat
[error]     case p"$pname" => p"$pname"
[error]                         ^
[error] DeConstructionTypesSuite.scala:73: type mismatch when unquoting;
[error]  found   : scala.meta.Pat.Var.Term
[error]  required: scala.meta.Stat
[error]     case q"${name: Pat.Var.Term}" => q"${name: Pat.Var.Term}"
[error]                                        ^
[error] DeConstructionTypesSuite.scala:78: type mismatch when unquoting;
[error]  found   : scala.collection.immutable.Seq[scala.meta.Pat.Arg]
[error]  required: scala.collection.immutable.Seq[scala.meta.Pat]
[...]

PR #584, opened 2017-01-05
Tests quasiquotes can be constructed and deconstucted using same pattern

tabdulradi:
Commented lines are not compiling. Issue reported here https://github.com/scalameta/scalameta/issues/583


tabdulradi:
I used some multi-cursor magic to extract the quasiquotes and write the test. A script with a simple regex should be able to do the trick. However, the generated test will need manual editing to remove a couple of patterns who work only for construction.

olafurpg:
I'll just go ahead and merge this then : ) Thank you for your contribution!

PR #588, opened 2017-01-21
Extract strings from lits in Pat.Interpolate

### LINKED ISSUE #587: Pat.Interpolate Syntax does not extract Strings from Lits
Term.Interpolate does.

This caused https://github.com/scalameta/paradise/issues/155

PR Inbound

DavidDudson:
Fixes #587

### DIFF
--- a/scalameta/trees/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
+++ b/scalameta/trees/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
@@ -354,11 +354,13 @@ object TreeSyntax {
             case pats       => s(pats)
           }))
         case t: Pat.Interpolate      =>
-          val zipped = t.parts.zip(t.args).map {
+          val parts = t.parts.map{ case Lit(part: String) => part }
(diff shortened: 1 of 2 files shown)


scalameta/trees/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -354,11 +354,13 @@ object TreeSyntax {
             case pats       => s(pats)
           }))
         case t: Pat.Interpolate      =>
-          val zipped = t.parts.zip(t.args).map {
+          val parts = t.parts.map{ case Lit(part: String) => part }
```
olafurpg: Just in case, can we add a fallback case when `part` is not a string?

olafurpg: to avoid matcherror

olafurpg: Nevermind, there's no fallback in term.interpolate. LGTM

PR #589, opened 2017-01-22
2.x

xeno-by:
Sets the baseline for my today's nirvana porting.


xeno-by:
The failure is spurious. I'm merging anyway, since we'll be republishing a more recent version soon.

PR #590, opened 2017-01-22
catching up with infrastructural debt

xeno-by:
review @olafurpg


xeno-by:
@olafurpg It's green, so I'm merging to unblock further work down the line. Please let me know if you have comments about this PR, and we'll fix them in subsequent submissions.

olafurpg:
LGTM, code health PRs like this are unrewarding but pay off big time. Thanks for adding me to developers :) 

At some point, I would like to reformat everything with scalafmt. But it's not urgent at all.

build.sbt
```diff
@@ -0,0 +1,436 @@
...
+  bintrayOrganization := Some("scalameta"),
+  publishArtifact in Compile := true,
+  publishArtifact in Test := false,
+  {
+    val publishingStatus = {
+      if (shouldPublishToBintray) "publishing to Bintray"
+      else if (shouldPublishToSonatype) "publishing to Sonatype"
+      else "publishing disabled"
+    }
+    println(s"[info] Welcome to scala.meta $LibraryVersion ($publishingStatus)")
```
xeno-by: @dwijnand What's the right way of doing this?

dwijnand: Try using [sLog](http://www.scala-sbt.org/release/sxr/sbt/Keys.scala.html#sbt.Keys.sLog), but you'll have to be in the RHS of `publish in Compile`. Not sure if outside or inside Def.taskDyn though..

PR #593, opened 2017-01-23
Update testkit.

olafurpg:
- Explain in changelog
- Run property tests to avoid regression


xeno-by:
```
[error] (run-main-0) java.io.IOException: Cannot run program "wget": error=2, No such file or directory
java.io.IOException: Cannot run program "wget": error=2, No such file or directory
```

olafurpg:
Hurray! Finally got this past CI : ) 

Seems the property test takes 80s to run, which I think it totally OK given it might catch tricky corner case bugs in pretty-printer and parser.

.drone.yml
```diff
@@ -10,7 +10,7 @@ build:
       - COURSIER_CACHE=/drone/cache/coursier
     commands:
       - ./bin/checkCLA.sh
-      - sbt -Dsbt.ivy.home=/drone/cache/ivy2 ++$SCALA_VERSION test doc
+      - sbt -Dsbt.ivy.home=/drone/cache/ivy2 ++$SCALA_VERSION test doc "testkit/test:runMain  scala.meta.testkit.ScalametaParserPropertyTest"
```
xeno-by: Do we need the double whitespace between `runMain` and `scala.meta...`?

.jvmopts
```diff
@@ -1,4 +1,8 @@
--XX:-TieredCompilation
--Xss2m
--Xms725m
--Xmx725m
+-Xms4000m
+-Xmx55g
+-Xss4m
+-XX:MaxPermSize=2000m
```
xeno-by: Why do we need the MaxPermSize option if .jvmopts already specify JVM 1.8?

changelog/1.5.0.md
```diff
@@ -2,8 +2,10 @@
 
 ## `testkit`
 
-  * We have introduced `testkit`, an infrastructure to test metaprograms like a pro.
-    See [https://github.com/scalameta/scalameta/commit/506c284fd389fcbd12c47550a8527a9d051aa3b4](https://github.com/scalameta/scalameta/commit/506c284fd389fcbd12c47550a8527a9d051aa3b4) for more information.
+  * `testkit` is a new module that can be used to run property-based tests
```
xeno-by: What's a property-based test?

xeno-by:
Agreed, the slowdown is totally justified. Apart from the minor comments above, LGTM.

.drone.yml
olafurpg: Done

.jvmopts
olafurpg: Good catch, I switched to a jdk 7 docker image in a temporary commit

olafurpg: removed.

changelog/1.5.0.md
olafurpg: [See here](http://letmegooglethat.com/?q=property+based+testing)

olafurpg: hehe :p sry, do you think it needs more explanation?

xeno-by: Just a URL should be enough, so that we're sure that people reading the notes understand what's going on :)

xeno-by:
I'll merge the PR, because I'll be overhauling the build today. If you'd like to follow up on changing the changelog, please submit it as a separate PR. Thanks a lot for your work!

olafurpg:
I'll update the changelog tomorrow!

PR #595, opened 2017-01-25
revive scalahost

xeno-by:
This commit moves the converter part of scalameta/paradise to our old friend
scalameta/scalahost.

Several projects, including scalafix, are in need of the converter without
the additional burden of the entire paradise, so I’m creating a dedicated
library for that.

Moreover, this will be the place to implement our upcoming semantic API
courtesy of early experiments in scalafix and nirvana.


olafurpg:
I am a big fan monorepos, so I welcome this change. However, this change introduces a few tooling problems

1. intellij doesn't understand the tree macro annotations so all code that uses Tree and quasiquotes (e.g., ToMTree) is now all red squiggly marks.
  - Workaround: depend on published artifact (1.5.0) of scalameta in scalahost.
2. we used scalafmt in paradise but it's not enabled here. I personally feel more productive not having to manipulate syntactic trivia
  - Workaround: enable scalafmt for scalahost only with `project.includeFilters = ["scalahost.*"] see https://olafurpg.github.io/scalafmt/#project (I can do this in a separate PR)
3. CI will run everything on every PR, including updates to docs/markdown files. This OK for now but may slow down PRs. I wish sbt offered better support to detect which targets to test given a diff, similar to what's possible with hermetic build tools such as pants/bazel.
  - Workaround: nothing for now. Might be worth experimenting with some bash script to detect what sbt test commands to run given a diff.

olafurpg:
Here's a patch to fix the build failure https://github.com/olafurpg/scalameta/commit/2f1780e0eb1a722595e757e0d611a01690216550

olafurpg:
I added a commit to my scalahost branch to enable scalafmt for scalahost only https://github.com/scalameta/scalameta/compare/master...olafurpg:scalahost?expand=1

xeno-by:
Thanks for the fixes! Cherrypicked them to my branch.

As for IntelliJ, let's leave it as it is for the time being until it becomes a blocker. As you mentioned during our private discussion, we can also try to reimplement the `@ast` annotation with scala.meta, in which case IntelliJ shouldn't have any problems with it.

olafurpg:
Alternatively, as this is only a problem during development, we could make it easy to update the scalahost build to depend on a scalameta published artifact. It's not a critical issue atm, so we can leave it for now.

xeno-by:
Omg, finally green!

olafurpg:
Haha, yes, getting a green CI can take quite some time. It's a shame how everything needs to be re-run on every PR even if you're just editing a comment.

Issue #537, opened 2016-11-09
scalameta-contrib


olafurpg:
The scala.meta trees and tokens APIs can be a bit low-level sometimes. It would be nice to have a library with utility methods for common operations, for example

- `trailingComments`/`leadingComments` on tree nodes, see https://gist.github.com/olafurpg/7be8f1d5aa9d01853e9ce514838710cc#gistcomment-1803194
- `Whitespace`/`Keyword`/... token extractors, see https://github.com/olafurpg/scalafmt/blob/master/core/src/main/scala/org/scalafmt/util/TokenClasses.scala
- tree extractors classes: https://github.com/olafurpg/scalafmt/blob/master/core/src/main/scala/org/scalafmt/util/TreeClasses.scala
- `toType/toPat/toTerm` from `Term.Name/Pat.Var.Term/Type.Name`
- `def parents(tree: Tree): List[Tree]` for all parents of a tree
- `Patch` to do high-performance token level transformations, something like an optimized and safer variant of https://github.com/scalacenter/scalafix/blob/master/core/src/main/scala/scalafix/util/Patch.scala
- pretty print hack: https://gist.github.com/olafurpg/00ce1601a7e213141f25
- maybe something from scalafmt TreeOps/TokenOps 

I'm sure there's a ton of other useful ideas that could be bundled into this library.
Would anyone be interested in using/contributing to such a module?

cc/ @ShaneDelmore

ShaneDelmore:
I haven't worked with it enough to have a ton of good ideas yet but brainstorming for a few minutes I come up with this:

On high performance patches, Is this backed by a mutable or immutable array?  Would we save allocations by having multiple types of patches:  Say Add | Remove | Replace.  Replace could be only same sized replacement so no need to shrink or grow the array.  Add/Remove could possibly be stacked in such a way that multiples could be stacked say an Add and Remove both of same length squashed into a Replace?

Due to the nature of syntax wrapping and unwrapping is more common than the usual collections API, more Ops working on both ends of the stream at once could be really useful.  I posed a couple of small examples to https://gist.github.com/ShaneDelmore/50bee845755d8cf81946ee2fff3efe9e

Examples;

def takeOuter(n: Int) = (take(n), takeRight(n))

def takeOuterWhile(forwardPredicate, reversePredicate): (Tokens, Tokens)
[...]

PR #624, opened 2017-01-31
Create contrib module

olafurpg:
cc/ #537 @DavidDudson 

Contains:

- Equal typeclass for type-safe comparison
- Collection like operations on trees
- Utility to find leading/trailing comments


scalameta/contrib/src/main/scala/scala/meta/contrib/ContribApi.scala
```diff
@@ -0,0 +1,50 @@
...
+    def contains[F[_ <: Tree]](toFind: Tree)(implicit conv: Tree => F[Tree],
+                                             eqEv: Equal[F[Tree]]): Boolean =
+      find(_.equal[F](toFind)).nonEmpty
+  }
+
+}
+
+object TreeOps {
+  @tailrec
+  final def parents(tree: Tree, accum: Seq[Tree] = Seq.empty[Tree]): Seq[Tree] = {
```
DavidDudson: Can we name this ancestors instead. And the have the opposite too.

- parent == this tree's parent
- children == this tree's children
- descendants = this tree's children, and grandchildren etc.
- ancestors = this trees parents, and grandparents etc.

DavidDudson: Also it seems weird that this is in a treeops object, whereas `collectFirst` is not, where `collectFirst` is longer

olafurpg: Good points.

1. Sure.
2. I was thinking of putting all the methods in an object and make the pimps optional. The pimps could for example be added with a `syntax` import, as is convention in other libs.

DavidDudson: Actually, can you hold up on modifying these. I've been extending this PR and it makes alot of my comments redundant. I'm about to push

olafurpg: Sorryy, just pushed a small update. I look forward to see your changes tomorrow, I'm hitting bed now :)

DavidDudson: https://github.com/DavidDudson/scalameta/commit/80aa21b6fb5d725a5107481666616b1a396953b6

Depending on your opinions on this, I may create a PR for your PR. Or else i can keep it seperate

xeno-by:
I agree that there's a big need in a place where we could put a grab bag of utilities without polluting the main API. Once certain utilities get popular, we can consider graduating them into scalameta/scalameta. If they don't get popular, they can stay in scalameta/contrib without confusing the newcomers.

I'm totally up for merging this, apart from one comment. I suggest that we follow the convention that's already in place in the main library and don't require extra imports apart from `import scala.meta.contrib._`.

scalameta/contrib/src/main/scala/scala/meta/contrib/ContribApi.scala
```diff
@@ -0,0 +1,50 @@
+package scala.meta.contrib
+
+import scala.language.higherKinds
+import scala.language.implicitConversions
+
+import scala.annotation.tailrec
+import scala.collection.immutable.Seq
+import scala.meta._
+
+trait ContribApi {
```
xeno-by: How about merging this trait into `package.scala`?

olafurpg:
I updated the PR to include more collections operations from @DavidDudson's commit (took the liberty to include you as an author to the commit). I'm happy with the PR as it is now, feel free to merge @xeno-by.

DavidDudson:
@olafurpg LGTM

xeno-by:
LGTM

PR #626, opened 2017-01-31
Add logger utility

olafurpg:
When hacking on scala.meta I've desperately missed being able to println a variable during development and see the file name + line numer + variable name + variable value

```scala
val x = 1
logger.elem(x)
// Foo.scala:10 [x]: 1
```

The macros are copied from lihaoyi/sourcecode to avoid adding a new dependency. The library is really small however so I personally don't really mind adding it as a dependency.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -51,6 +51,7 @@ lazy val common = Project(
   base = file("scalameta/common")
 ) settings (
   publishableSettings,
+  libraryDependencies += "com.lihaoyi" %% "sourcecode" % "0.1.3",
   description := "Bag of private and public helpers used in scala.meta's APIs and implementations",
   enableMacros
 )
--- a/scalameta/common/src/main/scala/org/scalameta/logger.scala
+++ b/scalameta/common/src/main/scala/org/scalameta/logger.scala
@@ -1,0 +1,35 @@
+package org.scalameta
+
+class FileLine(val file: sourcecode.File, val line: sourcecode.Line) {
+  override def toString: String = {
+    val shortFilename = file.value.replaceAll("(.*/|\\.scala)", "")
+    Console.GREEN + s"$shortFilename:${line.value}" + Console.RESET
+  }
+}
+
+object FileLine {
+  implicit def generate(implicit file: sourcecode.File, line: sourcecode.Line): FileLine =
+    new FileLine(file, line)
+}
+
+object logger {
+  /** Same as println except includes the file+line number of call-site. */
+  def debug(x: Any)(implicit fileLine: FileLine): Unit = {
+    println(s"$fileLine $x")
+  }
+
+  /** Prints out the value with and it's source code representation
+    *
+    * Example: logger.elem(x) // prints "MyFile:24 [x]: 42"
+    **/
+  def elem(values: sourcecode.Text[Any]*)(implicit fileLine: FileLine): Unit = {
+    values.foreach { t =>
+      val value = {
+        val str = s"${t.value}"
+        if (str.contains("\n")) s"\n$str"
+        else str
+      }
+      println(s"$fileLine [${t.source}]: $value")
+    }
+  }
+}


scalameta/common/src/main/scala/org/scalameta/internal/logger.scala
```diff
@@ -0,0 +1,95 @@
...
+}
+
+case class FileLine(file: File, line: Line)
+object FileLine {
+  implicit def generate(implicit file: File, line: Line): FileLine = FileLine(file, line)
+}
+
+case class Text[T](value: T, source: String)
+object Text {
+  implicit def generate[T](v: T): Text[T] = macro Impls.text[T]
```
olafurpg: Any idea @xeno-by why this line fails doc generation? 

http://drone.geirsson.com:8001/scalameta/scalameta/170

scalameta/common/src/main/scala/org/scalameta/internal/logger.scala
```diff
@@ -0,0 +1,95 @@
...
+    else {
+      val file = fileLine.file.value.replaceAll("(.*/|\\.scala)", "")
+      val line = fileLine.line.value
+      green(s"$file:$line")
+    }
+  }
+  def info(x: Any)(implicit fileLine: FileLine): Unit = {
+    println(s"$fileAndLine $x")
+  }
+  def elem(texts: Text[Any]*)(implicit fileLine: FileLine): Unit = {
```
xeno-by: What's the story behind this method name?

scalameta/common/src/main/scala/org/scalameta/internal/logger.scala
```diff
@@ -0,0 +1,95 @@
...
+    val parser = g.newUnitParser(fileContent.drop(start))
+    parser.expr()
+    val end  = parser.in.lastOffset
+    val txt  = fileContent.slice(start, start + end)
+    val tree = q"""${c.prefix}(${v.tree}, $txt)"""
+    c.Expr[Text[T]](tree)
+  }
+}
+
+object logger {
```
xeno-by: How about putting this object in here? https://github.com/scalameta/scalameta/blob/e5136bed4289e98c2e852162d73f63a7b6988479/scalameta/common/src/main/scala/org/scalameta/package.scala#L29

scalameta/common/src/main/scala/org/scalameta/internal/logger.scala
```diff
@@ -0,0 +1,95 @@
+// The macro implementations are copied from https://github.com/lihaoyi/sourcecode
```
xeno-by: Let's directly depend on sourcecode. That'll avoid both the scaladoc issue and the potential licensing trouble (i'm not a lawyer).

scalameta/scalameta/src/test/scala/scala/meta/tests/internal/LoggerTest.scala
```diff
@@ -0,0 +1,14 @@
...
+import org.scalameta.internal.FileLine
+
+class LoggerTest extends org.scalatest.FunSuite {
+  test("basic") {
+    val fileLine = FileLine.generate
+    assert(fileLine.file.value.endsWith("LoggerTest.scala"))
+    assert(fileLine.line.value == 7)
+    import org.scalameta.internal.logger
+    logger.elem(fileLine)
+    logger.info(fileLine)
```
xeno-by: How do we test the behavior of these `logger` calls? Just eyeballing, or I'm missing something?

scalameta/common/src/main/scala/org/scalameta/internal/logger.scala
xeno-by: Do we still need this comment?

olafurpg:
I updated this PR to add a dependency on sourcecode and removed the copied macros. I also moved the logger to package `org.scalameta`.

scalameta/common/src/main/scala/org/scalameta/internal/logger.scala
olafurpg: I added a docstring explaining, this method is AWESOME. I'm sure you'll love it!

scalameta/scalameta/src/test/scala/scala/meta/tests/internal/LoggerTest.scala
olafurpg: Removed the entire test, since the impl was trivial once we added a dependency on sourcecode.

xeno-by:
LGTM

Issue #612, opened 2017-01-30
Functionality to work with types


xeno-by:
Prerequisites:
  - [x] Online mirror with support for Ref.symbol (#606)
  - [ ] Term.tpe (#611)

Following #611, we'd like to implement support for the following operations on types, inspired by our experience with scala.reflect:
  - [ ] `Type.=:=`
  - [ ] `Type.<:<`
  - [ ] Other operations, as requested, given compelling use cases

Much like in the case of #609, we may need to create a separate ticket to handle types that refer to separately compiled local symbols.

xeno-by:
@olafurpg @DavidDudson @ShaneDelmore Please share your wishlist of semantic APIs that you'd like to see supported for types.

xeno-by:
@liufengyun Could you provide some thoughts on the semantic APIs that you'd like to see here?

ShaneDelmore:
@xeno-by I believe =:= and <:< cover all of my needs as long as they work properly with type aliases.

liufengyun:
I think (1)`<:<`, (2)`=:=`, (3) create types from string or quasiquote (4) convert type to string or tree  would suffice in the beginning.

For writing macros, I'm unaware of scenarios where people need `Type.members`.  `Type.widen/dealias` are too much knowledge to the programmer, and we can wait for use cases of them to emerge. I think it's good to be conservative and restrictive about features.

xeno-by:
@ShaneDelmore @liufengyun I've updated the implementation plan according to your feedback. Thank you for the push for minimalism!

xeno-by:
Recently, we decided to limit the surface of our work to actionable items only - either bugs in existing code or feature requests that are warranted by the roadmap. 

While semantic API is definitely on our roadmap, we're going to figure out desired APIs from use cases, not from reflecting on scala.reflect and pondering what functionality from there we should add to Scalameta. That's why I'll be closing this issue and letting feature requests be submitted naturally by our users.

Issue #611, opened 2017-01-30
Term.tpe


xeno-by:
Prerequisites:
  - [x] Online mirror with support for Ref.symbol (#606)

Following `Ref.symbol` (#606, #607, #608) and other symbol APIs (#609), we'd also like to support `Term.tpe` that returns a type of a given term in Scala source code.

Usage example:
```scala
implicit val m = Mirror(...)
q"List(1, 2, 3)".tpe // t"Int"
```

Implementation plan:
  - [ ] Upgrade the specification of semantic database (#605) to include support for types
  - [ ] Add support for types to the `CompilationUnit => Database` functionality. I suspect that we may have to be very mindful of performance here, since the number of terms greatly exceeds the number of names.
  - [ ] Add support for types in the semantic db persistence functionality
  - [ ] Figure out which scala.reflect types we want to support. See #273 for relevant discussion.
  - [ ] Add a converter that goes from scala.reflect types to scala.meta types. Unlike with symbols, here we won't be able to get away with just strings.

xeno-by:
Recently, we decided to limit the surface of our work to actionable items only - either bugs in existing code or feature requests that are warranted by the roadmap. 

While semantic API is definitely on our roadmap, we're going to figure out desired APIs from use cases, not from reflecting on scala.reflect and pondering what functionality from there we should add to Scalameta. That's why I'll be closing this issue and letting feature requests be submitted naturally by our users.

xeno-by:
Superseded by https://github.com/scalameta/scalameta/issues/1212.

Issue #609, opened 2017-01-30
Functionality to work with symbols


xeno-by:
Prerequisites:
  - [x] Online mirror with support for Ref.symbol (#606)

We start implementing the semantic API with a very simple `Ref.symbol` functionality (#606, #607, #608). While this API suffers from the same issues that `Tree.symbol` has in scala.reflect (dedicated API required to inspect signatures, forgetfulness with respect to prefixes, etc), it's easy to implement and can provide immediate wins, powering important use cases.

Another easy win is porting some of the scala.reflect `Symbol` APIs, providing scala.meta users with basic means to perform many semantic operations. Some inspiration for potential APIs can be drawn from [a previous prototype of semantic APIs](https://github.com/scalameta/scalameta/blob/7d6f917de39f8bfdbeb1f87d145f9797803d9ebb/scalameta/semantic/src/main/scala/scala/meta/semantic/Api.scala#L179-L308).

Usage example:
```scala
implicit val m = Mirror(...)
val sym = ... // scala.meta.semantic.Symbol
println(sym.name) // Int
println(sym.fullName) // scala.Int
println(sym.members) // List(getClass, ##, asInstanceOf, ...)
...
```

Implementation plan:
  - [ ] Figure out the list of APIs that we'd like to support. So far, I've got requests for:
    * `Symbol.toString` (we probably don't want to reveal internal ids in a public API)
    * `Symbol.syntax/structure` (these APIs are too commonly used to be missing)
    * `Symbol.name: Ref` (unwraps the symbol id to reveal a name, probably wrapped in an appropriate `XXX.Name` tree)
    * `Symbol.fullName: Ref` (a fully-qualified path for global symbols, the same as `name` for local symbols)
    * `Symbol.members: Seq[Symbol]` (I'd also add some helpers that filter this list by name and/or kind)
    * Other operations, as requested, given compelling use cases
  - [ ] As far as I can see, the implementation of symbol APIs will be different depending on whether the symbols are local or global:
    * Global symbols should be present in the underlying `Global` (in case of #606) or in a dynamically created `Global` (in case of #607). It should be easy to go from a UID to a corresponding `g.Symbol` and then perform a semantic operation on that symbol.
    * Local symbols that aren't present in `g`'s symbol table (because they were compiled separately and then loaded from a semantic database by #607) may pose a problem, but for them we can create a separate ticket (also see #612 for a related situation).
[...]

Issue #607, opened 2017-01-30
Offline mirror with support for Ref.symbol


xeno-by:
**DISCLAIMER:** `Ref.symbol` is only going to be available in `scala.meta.semantic.v1`. The concept of symbols has a benefit of being easy to implement, but has a bunch of known issues discovered in the scala.reflect experiment. Therefore, in `scala.meta.semantic.v2`, we're planning to remove symbols and `Ref.symbol` and replace them with something else (#610).

------------

Prerequisites:
  - [x] Online mirror with support for Ref.symbol (#606)

In this ticket, we'll be implementing persistence for our semantic db (#605), and a mirror that can load previously persisted databases. Much like the mirror from #606, this mirror will only support one semantic API, `Ref.symbol`, but in a different way, by:
  * Taking a classpath that contains one or more persisted databases and a sourcepath that contains sources referred to in those databases
  * Loading persisted databases and merging them into one semantic database 
  * Using the semantic db and just the semantic db to reply to semantic queries.

Usage example:
```scala
implicit val m = Mirror("<classpath>", "<sourcepath>")
// the same as in #606
```

Implementation plan:
  - [ ] Specify persistence format for semantic databases in #605 
    * @densh suggested that we persist semantic databases on per-classfile basis, similarly to sjsir and nir files from Scala.js and Scala Native. Otherwise, incremental compilation may be problematic.
    * @stuhood suggested that we'll need to integrate with sbt to have incremental compilation working correctly in all cases: https://github.com/sbt/zinc/issues/125.
  - [ ] Update the `Location.uri` specification in #605 to make sure we can correlate locations used to create semantic dbs and locations coming from sourcepaths when loading them
  - [x] Implement semantic persistence by hijacking the scalac backend similarly to [our previous approaches at TASTY support](https://github.com/scalameta/scalameta/blob/7d6f917de39f8bfdbeb1f87d145f9797803d9ebb/scalahost/src/main/scala/scala/meta/internal/hosts/scalac/HijackBackend.scala).
  - [x] Create a classpath/sourcepath-based implementation of `Mirror`
  - [x] Add testing infrastructure

xeno-by:
Implemented an initial version of semantic DB persistence in 7c44e77. The format is an extremely inefficient text-based representation, locations are still specified via absolute paths. Nonetheless, the functionality is already useful. I'm closing this issue, moving further discussion to #605.

jvican:
[...]

PR #636, opened 2017-02-01
initial support for semantic API

xeno-by:


xeno-by:
Yesterday, I hacked up a prototype of an online mirror (#606). Today I implemented the first prototype of an offline mirror, i.e. a mirror that can load previously persisted semantic databases (#607).

Here's a quick demo. Let's start with a standalone library of Scala code.

```
$ cat Library.scala
object Library {
  def doSomething = {
    val list = List(1, 2, 3)
    println(list)
  }
}
```

Now, we compile this library with the `scalahost` plugin enabled.

```
$ scalac -Yrangepos -Xplugin:...scalahost.jar Library.scala
```

After compilation finishes, along with the usual classfiles, we can see a `.semanticdb` file that contains semantic information about the program that's been compiled (#605). For the purposes of this prototype, the file is placed simply at the root of the output directory and is given a pseudo-unique name.

```
$ ls
1485999295626.semanticdb	Library.class			Library$.class
Library.scala
```

The `.semanticdb` file contains a serialized representation of the semantic database corresponding to the compiled program. To save time, I didn't do anything fancy for the serialization format just yet. In fact, for the purposes of this prototype, I just dumped already implemented debug output to the output file.

```
$ cat 1485999295626.semanticdb
/Users/xeno_by/Projects/Meta1x/sandbox/Library.scala
[7..14): Library => _empty_.Library.
[23..34): doSomething => _empty_.Library.doSomething()V.
[47..51): list => file:/Users/xeno_by/Projects/Meta1x/sandbox/Library.scala@47
[54..58): List => _root_.scala.collection.immutable.List.apply(Lscala/collection/Seq;)Lscala/collection/immutable/List;.
[72..79): println => _root_.scala.Predef.println(Ljava/lang/Object;)V.
[80..84): list => file:/Users/xeno_by/Projects/Meta1x/sandbox/Library.scala@47
```

Now, interesting things begin. Below we create a mirror, using a classpath (that contains the `.semanticdb` file) and a sourcepath (that contains the sources of the originally compiled program). Having this mirror in hand, we traverse the original file and then resolve all `println` names inside it.

```
$ cat Test.scala
import scala.meta._

object Test extends App {
  implicit val mirror = Mirror(classpath = ".", sourcepath = ".")

  val library = new java.io.File("Library.scala").parse[Source].get
  library collect {
    case name @ Term.Name("println") =>
      println(name.symbol)
  }
}

$ scalac Test.scala && scala Test
Symbol(_root_.scala.Predef.println(Ljava/lang/Object;)V.)
```

As you can see, the mirror is capable of telling us that `println` actually refers to `Predef.println`. More precisely, it's the `(Any)Unit` overload. Even though it looks simple, this is the functionality that hasn't been previously available in scala.meta, so all in all it's pretty huge :)

My next step is building a user-friendly semantic API on top of the low-level semantic database. Check out #609, #611 and #612 to see what's coming up next.

/cc @olafurpg @ShaneDelmore @liufengyun @DavidDudson

ShaneDelmore:
I won't have a chance to try this out until this weekend but it is looking great so far.

xeno-by:
@olafurpg Finally, I'm happy enough about this PR to just merge it. I haven't yet started working on #609, but the accumulated progress is significant enough anyway.

See https://github.com/xeno-by/scalameta/blob/9379f98f6d67deeb29088f40e37e8ad82d6b73b5/scalahost/src/test/scala/scala/meta/tests/scalahost/v1/DatabaseSuite.scala for examples of things that already work.

PR #637, opened 2017-02-02
WIP Prototype implementation of Statements typeclass

DavidDudson:
Basically this pimps all the trees that have a stat block with methods such as

- appendStats
- getStats
- replaceStats

and extra methods such as:

- defDefs (returning all defs contained in that thing)
- valDefs
- typeDefs

and many more.

If this approach is liked by the maintainers I will add other typeclasses for similar things, such as modifers and parameters.

This will make it possible to do the following

Initial:
```
class Foo {
  val baz: Int = 2
  def bar: Int = {
    println("foo")
    2
  }
}
```

```clazz.defDefs.map(_.replaceStat[Structurally](List(2), List(3)))```

Result:
```
class Foo {
  val baz: Int = 2
  def bar: Int = {
    println("foo")
    3
  }
}
```

### DIFF
--- a/scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala
+++ b/scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala
@@ -1,0 +1,105 @@
+package scala.meta.contrib.implicits
+
+import scala.collection.immutable.Seq
+import scala.language.higherKinds
+import scala.language.implicitConversions
+import scala.meta._
+import scala.meta.contrib.Equal
+
+/**
+  * Typeclass to deal with scalameta trees which have an expected set of statements.
+  *
+  * For trees that have multiple stat lists (such as Defn.Class) the main stats are used.
+  *
+  * This allows you to do things like add a statement to the
+  * body of a class without digging through the low level meta api
+  *
+  * TODO: Consider using a meta variant of simulacrum to
+  * generate the boilerplate once one is avaliable
+  */
+trait Statements[A] {
+  def getStats(a: A): Seq[Stat]
+  def replaceStats(a: A, newStats: Seq[Stat]): A
+}
+
+object Statements {
+
+  trait Ops[A] {
+    val self: A
+    val typeClassInstance: Statements[A]
+
+    def getStats: Seq[Stat] =
+      typeClassInstance.getStats(self)
+
+    def replaceStats(newStats: Seq[Stat]): A =
+      typeClassInstance.replaceStats(self, newStats)
+
+// DAVETODO: Get F param working (Aka Structurally/Syntactically)
+//    Is this going to be useful?
+//    def replaceStatWith[F[_ <: Tree]](stat: Stat, replacement: Stat)(
+//        implicit conv: Tree => F[Tree], eq: Equal[F[Tree]]): A = {
+//
+//      val newStats: Seq[Stat] = getStats.collect {
+//        case candidate if eq.equal(candidate, stat) => replacement
+//        case other => other
+//      }
+//
+//      replaceStats(newStats)
+//    }
+//    def removeStat[F[_]](stat: Stat)(implicit eq: Equal[F[A]]): A =
+//      removeStats[F](stat :: Nil)
+//
+//    def removeStats[F[_]](stats: Seq[Stat])(
+//      implicit conv: Tree => F[Tree], eq: Equal[F[Tree]]): A = {
+//
+//      val filtered: Seq[Stat] = getStats.filter(stat => stats.exists(eq.equals(_, stat)))
+//      replaceStats(filtered)
+//    }
+
+
+    def appendStat(stat: Stat): A =
+      replaceStats(getStats :+ stat)
+
+    def appendStats(stats: Seq[Stat]): A =
+      replaceStats(getStats ++ stats)
+
+    def prependStat(stat: Stat): A =
+      replaceStats(stat +: getStats)
+
+    def prependStats(stats: Seq[Stat]): A =
+      replaceStats(stats ++ getStats)
+
+    def deleteAllStats(): A =
+      replaceStats(Nil)
+
+    def typeDefs(): Seq[Defn.Type] =
+      getStats.collect { case t: Defn.Type => t}
+
+    def defDefs(): Seq[Defn.Def] =
+      getStats.collect { case d: Defn.Def => d}
+
+    def valDefs(): Seq[Defn.Val] =
+      getStats.collect { case v: Defn.Val => v}
+
+    def varDefs(): Seq[Defn.Var] =
+      getStats.collect { case v: Defn.Var => v}
+
+    def traitDefs(): Seq[Defn.Trait] =
+      getStats.collect { case t: Defn.Trait => t}
+
+    def objectDefs(): Seq[Defn.Object] =
+      getStats.collect { case o: Defn.Object => o}
+
+    def classDefs(): Seq[Defn.Class] =
+      getStats.collect { case c: Defn.Class => c}
+  }
+
+  trait ToStatementOps {
+    implicit def toStatementsOps[A](target: A)(implicit tc: Statements[A]): Ops[A] = new Ops[A] {
+      override val self: A = target
+      override val typeClassInstance: Statements[A] = tc
+    }
+  }
+
+  def apply[A](implicit instance: Statements[A]): Statements[A] = instance
+}
--- a/scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala
+++ b/scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala
@@ -19,87 +19,42 @@ import scala.meta.contrib.Equal
   */
 trait Statements[A] {
   def getStats(a: A): Seq[Stat]
-  def replaceStats(a: A, newStats: Seq[Stat]): A
+  def withStats(a: A, newStats: Seq[Stat]): A
 }
 
-object Statements {
+trait StatementsTypeClass {
 
-  trait Ops[A] {
-    val self: A
-    val typeClassInstance: Statements[A]
+  implicit def statementsTypeClass[A](implicit ev: Statements[A]): Statements[A] = ev
 
+  implicit class XtensionStatements[A](a: A)(implicit ev: Statements[A]) {
     def getStats: Seq[Stat] =
-      typeClassInstance.getStats(self)
+      ev.getStats(a)
 
-    def replaceStats(newStats: Seq[Stat]): A =
-      typeClassInstance.replaceStats(self, newStats)
+    def withStats(newStats: Seq[Stat]): A =
+      ev.withStats(a, newStats)
 
-// DAVETODO: Get F param working (Aka Structurally/Syntactically)
-//    Is this going to be useful?
-//    def replaceStatWith[F[_ <: Tree]](stat: Stat, replacement: Stat)(
-//        implicit conv: Tree => F[Tree], eq: Equal[F[Tree]]): A = {
-//
-//      val newStats: Seq[Stat] = getStats.collect {
-//        case candidate if eq.equal(candidate, stat) => replacement
-//        case other => other
-//      }
-//
-//      replaceStats(newStats)
-//    }
-//    def removeStat[F[_]](stat: Stat)(implicit eq: Equal[F[A]]): A =
-//      removeStats[F](stat :: Nil)
-//
-//    def removeStats[F[_]](stats: Seq[Stat])(
-//      implicit conv: Tree => F[Tree], eq: Equal[F[Tree]]): A = {
-//
-//      val filtered: Seq[Stat] = getStats.filter(stat => stats.exists(eq.equals(_, stat)))
-//      replaceStats(filtered)
-//    }
+    def updateStats(f: Seq[Stat] => Seq[Stat]): A =
+      withStats(f(getStats))
 
-
-    def appendStat(stat: Stat): A =
-      replaceStats(getStats :+ stat)
-
-    def appendStats(stats: Seq[Stat]): A =
-      replaceStats(getStats ++ stats)
-
-    def prependStat(stat: Stat): A =
-      replaceStats(stat +: getStats)
-
-    def prependStats(stats: Seq[Stat]): A =
-      replaceStats(stats ++ getStats)
-
-    def deleteAllStats(): A =
-      replaceStats(Nil)
-
-    def typeDefs(): Seq[Defn.Type] =
+    def types(): Seq[Defn.Type] =
       getStats.collect { case t: Defn.Type => t}
 
-    def defDefs(): Seq[Defn.Def] =
+    def defs(): Seq[Defn.Def] =
       getStats.collect { case d: Defn.Def => d}
 
-    def valDefs(): Seq[Defn.Val] =
+    def vals(): Seq[Defn.Val] =
       getStats.collect { case v: Defn.Val => v}
 
-    def varDefs(): Seq[Defn.Var] =
+    def vars(): Seq[Defn.Var] =
       getStats.collect { case v: Defn.Var => v}
 
-    def traitDefs(): Seq[Defn.Trait] =
+    def traits(): Seq[Defn.Trait] =
       getStats.collect { case t: Defn.Trait => t}
 
-    def objectDefs(): Seq[Defn.Object] =
+    def objects(): Seq[Defn.Object] =
       getStats.collect { case o: Defn.Object => o}
 
-    def classDefs(): Seq[Defn.Class] =
+    def classes(): Seq[Defn.Class] =
       getStats.collect { case c: Defn.Class => c}
   }
-
-  trait ToStatementOps {
-    implicit def toStatementsOps[A](target: A)(implicit tc: Statements[A]): Ops[A] = new Ops[A] {
-      override val self: A = target
-      override val typeClassInstance: Statements[A] = tc
-    }
-  }
-
-  def apply[A](implicit instance: Statements[A]): Statements[A] = instance
 }
--- a/scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/implicits.scala
+++ b/scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/implicits.scala
@@ -1,0 +1,3 @@
+package scala.meta.contrib.implicits
+
+trait implicits extends Statements.ToStatementOps
--- a/scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/implicits.scala
+++ b/scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/implicits.scala
@@ -1,3 +1,3 @@
 package scala.meta.contrib.implicits
 
-trait implicits extends Statements.ToStatementOps
+trait implicits extends StatementsTypeClass
--- a/scalameta/contrib/src/main/scala/scala/meta/contrib/instances/StatementsInstances.scala
+++ b/scalameta/contrib/src/main/scala/scala/meta/contrib/instances/StatementsInstances.scala
@@ -1,0 +1,117 @@
+package scala.meta.contrib.instances
+
+import scala.collection.immutable.Seq
+import scala.meta._
+import scala.meta.contrib._
+import scala.meta.contrib.implicits.Statements
+
+trait StatementsInstances {
+  implicit val termBlockStats: Statements[Term.Block] = new Statements[Term.Block] {
+    override def getStats(block: Term.Block): Seq[Stat] = block.stats
+
+    override def replaceStats(block: Term.Block, newStats: Seq[Stat]): Term.Block =
+      block.copy(stats = newStats)
+  }
+
+  implicit val packageStats: Statements[Pkg] = new Statements[Pkg] {
+    override def getStats(pkg: Pkg): Seq[Stat] = pkg.stats
+
+    // TODO: Packages have restrictions that all others dont.
+    // aka top level stats only
+    override def replaceStats(pkg: Pkg, newStats: Seq[Stat]): Pkg =
+      pkg.copy(stats = newStats)
+  }
+
+  implicit val templStats: Statements[Template] = new Statements[Template] {
+    override def getStats(templ: Template): Seq[Stat] =
+      templ.stats match {
+        case Some(stats) => stats
+        case None => Nil
+      }
+
+    override def replaceStats(templ: Template, newStats: Seq[Stat]): Template = {
+      newStats match {
+        case Nil => templ.copy(stats = None)
+        case s => templ.copy(stats = Some(s))
+      }
+    }
+  }
+
+  implicit val classStats: Statements[Defn.Class] = new Statements[Defn.Class] {
+    override def getStats(clazz: Defn.Class): Seq[Stat] =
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

**Comment by DavidDudson** [comment]
@xeno-by @ShaneDelmore @olafurpg Thoughts welcome.

**Comment by DavidDudson** [comment]
Not i just realised classes are broken in this patch, since `valDefs` and `varDefs` wont inlcude those defined in ctor parameters

File: scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala [hunk 1]
```diff
@@ -0,0 +1,105 @@
...
+  *
+  * This allows you to do things like add a statement to the
+  * body of a class without digging through the low level meta api
+  *
+  * TODO: Consider using a meta variant of simulacrum to
+  * generate the boilerplate once one is avaliable
+  */
+trait Statements[A] {
+  def getStats(a: A): Seq[Stat]
+  def replaceStats(a: A, newStats: Seq[Stat]): A
```
olafurpg: How about `withStats`? I feel like "replace" sounds like a mutative operatation

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+  * Typeclass to deal with scalameta trees which have an expected set of statements.
+  *
+  * For trees that have multiple stat lists (such as Defn.Class) the main stats are used.
+  *
+  * This allows you to do things like add a statement to the
+  * body of a class without digging through the low level meta api
+  *
+  * TODO: Consider using a meta variant of simulacrum to
```
olafurpg: Could you explain the value in organizing type-classes the simulacrum way? Compared to 

```scala
trait TypeClass[T] {
  def op(e: T): Foo
}
object TypeClass {
  implicit val XTypeClass: TypeClass[X]
  ...
}

package object contrib {
  implicit class XtensionTypeClass[A](e: A)(implicit ev: TypeClass[A]) {
    def op: Foo = ???
    ...
  }
}}}
}
```

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala
```diff
@@ -0,0 +1,105 @@
...
+//
+//    def removeStats[F[_]](stats: Seq[Stat])(
+//      implicit conv: Tree => F[Tree], eq: Equal[F[Tree]]): A = {
+//
+//      val filtered: Seq[Stat] = getStats.filter(stat => stats.exists(eq.equals(_, stat)))
+//      replaceStats(filtered)
+//    }
+
+
+    def appendStat(stat: Stat): A =
```
olafurpg: I'm kinda skeptical about adding so many utility methods. Why?

- it mixes the idea of "extract Seq[Stat] from tree" with useful Seq operations. These methods could in theory be type parameterized and work with any `Seq[T]`.
- Where do you draw the line which method is useful or not?
- it clogs up autocompletion lists

I think one of the key design goals of the collections api was to
- support the widest number of operations
- with the smallest possible vocabulary.
- (personal interpretation) without too much ceremony. For example, contains/exists/forall are in theory unnecessary but really useful in practice).

I think we should try and strive for the same goal in contrib.

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/implicits.scala
```diff
@@ -0,0 +1,3 @@
+package scala.meta.contrib.implicits
+
+trait implicits extends Statements.ToStatementOps
```
olafurpg: I like the idea of hiding implicits behind an optional `implicit` import. It seems to in convention with other libs such as cats/scalaz. If we're gonna go this route, we should move the Xtension methods in scala.meta.contrib package to the `implicits` too (I can do that in a separate PR)

scalameta/contrib/src/main/scala/scala/meta/contrib/instances/StatementsInstances.scala
```diff
@@ -0,0 +1,117 @@
...
+
+trait StatementsInstances {
+  implicit val termBlockStats: Statements[Term.Block] = new Statements[Term.Block] {
+    override def getStats(block: Term.Block): Seq[Stat] = block.stats
+
+    override def replaceStats(block: Term.Block, newStats: Seq[Stat]): Term.Block =
+      block.copy(stats = newStats)
+  }
+
+  implicit val packageStats: Statements[Pkg] = new Statements[Pkg] {
```
olafurpg: Might be good to also include `Source`

scalameta/contrib/src/main/scala/scala/meta/contrib/instances/StatementsInstances.scala
```diff
@@ -0,0 +1,117 @@
...
+    override def getStats(pkg: Pkg): Seq[Stat] = pkg.stats
```
olafurpg: Crazy idea, handle nested Pkg?

```scala
@ clangPrint(source"""package a
           package b
           class C""".structure)
res2: String = """
Source(Seq(Pkg(Term.Name("a"),
               Seq(Pkg(Term.Name("b"),
                       Seq(Defn.Class(Nil, Type.Name("C"), Nil, Ctor.Primary(Nil, Ctor.Ref.Name("this"), Nil),
                                      Template(Nil, Nil, Term.Param(Nil, Name.Anonymous(), None, None), None))))))))
"""
```

That is, if `stats` has only a single `Pkg`, then concat with the stats of that Pkg. If so, `replaceStats` should handle that case too, which to wrap the tail into the `head: Pkg` element.

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala
ShaneDelmore: I agree on separating out Seq operations from getting a new tree with different stats.  Having multiple operations of type Seq[Stat] => A makes me want to separate out Seq[Stat] => Seq[Stat] and A.copy(stats = newStats)

I think many of these could be self.copy(stats = operationsFromSeqStatToSeqStat(originalStats)

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala
```diff
@@ -0,0 +1,105 @@
...
+  def replaceStats(a: A, newStats: Seq[Stat]): A
+}
+
+object Statements {
+
+  trait Ops[A] {
+    val self: A
+    val typeClassInstance: Statements[A]
+
+    def getStats: Seq[Stat] =
```
ShaneDelmore: def stats?

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala
```diff
@@ -0,0 +1,105 @@
...
+      typeClassInstance.getStats(self)
+
+    def replaceStats(newStats: Seq[Stat]): A =
```
ShaneDelmore: def withStats

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala
```diff
@@ -0,0 +1,105 @@
...
+
+    def traitDefs(): Seq[Defn.Trait] =
+      getStats.collect { case t: Defn.Trait => t}
+
+    def objectDefs(): Seq[Defn.Object] =
+      getStats.collect { case o: Defn.Object => o}
+
+    def classDefs(): Seq[Defn.Class] =
+      getStats.collect { case c: Defn.Class => c}
+  }
```
ShaneDelmore: The collect variants here do feel useful, I find myself writing these repeatedly.

scalameta/contrib/src/main/scala/scala/meta/contrib/instances/StatementsInstances.scala
```diff
@@ -0,0 +1,117 @@
...
+    override def replaceStats(pkg: Pkg, newStats: Seq[Stat]): Pkg =
+      pkg.copy(stats = newStats)
+  }
+
+  implicit val templStats: Statements[Template] = new Statements[Template] {
+    override def getStats(templ: Template): Seq[Stat] =
+      templ.stats match {
+        case Some(stats) => stats
+        case None => Nil
+      }
```
ShaneDelmore: I might skip the pattern matching here and use
```temp.stats.getOrElse(Nil)```

File: scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala (hunk 9 above)
Comment (olafurpg) [inline scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala]: I agree. 

Just an idea, though, this could be implemented as a single method if we can find a good method name

```scala
// psuedo code
template.stats.extract[Defn.Val] // or something other than "extract"
implicit class XtensionStats(stats: Seq[Stat]) {
  def extract[T: ClassTag] = stats.collect {
    case t if implicitly[ClassTag[T]].runtimeClass.isInstance(t) => t.asInstanceOf[T]
  }
}
```

No tree types have type params, so the ClassTag will always work.

ShaneDelmore: A variant of collect is probably the least surprising way to go.
collectTypes, collectInstances?

scalameta/contrib/src/main/scala/scala/meta/contrib/instances/StatementsInstances.scala
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+
+  implicit val objectStats: Statements[Defn.Object] = new Statements[Defn.Object] {
+    override def getStats(obj: Defn.Object): Seq[Stat] =
+      obj.templ.getStats
+
+    override def replaceStats(obj: Defn.Object, newStats: Seq[Stat]): Defn.Object =
+      obj.copy(templ = obj.templ.replaceStats(newStats))
+  }
```
ShaneDelmore: withStats?  
This looks like sugar to prevent needing to do the copy repeatedly.  What about preventing the need for getStats and copy by creating withMappedStats that takes a function from Seq[Stat] => Seq[Stat]?

olafurpg: I think that's a great idea, how about `updateStats(f: Seq[Stat] => Seq[Stat])`? In [monadic-html](https://github.com/OlivierBlanvillain/monadic-html) we have an `update` method with a similar signature on `Rx[T]`, all immutable.

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/implicits.scala
xeno-by: Let's discuss the two approaches. I understand the theoretical appeal of modularizing things, but in my opinion the overhead of requiring two (or more, if we have multiple modules) imports is significant. Could you guys elaborate on what the modularization buys us?

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala
xeno-by: I second this suggestion.

xeno-by:
How about renaming `fooDefs()` to just `foos`? I think this will make things more lightweight without the loss in clarity.

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala
olafurpg: After more thinking, I think it's OK to have custom methods `vals/var/defs/classes/traits/objects/defns` to pull out members of those types. I agree with @xeno-by about calling the methods `foos` instead of `fooDefs`

scalameta/contrib/src/main/scala/scala/meta/contrib/instances/StatementsInstances.scala
xeno-by: I think `withStats` sounds great.

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala
DavidDudson: We can do it that way. there's really no specific reason for me doing it that way

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala
DavidDudson: That's fair. I'll create a small subset

scalameta/contrib/src/main/scala/scala/meta/contrib/instances/StatementsInstances.scala
DavidDudson: Ah yes, I knew i would forget something!

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala
DavidDudson: cant, it breaks when the tree itself implements a field stats. Thats the only reason i switched it

xeno-by: What do you mean by breaks?

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/implicits.scala
DavidDudson: The idea i general is that the use can import instances, without the implicit parts.

eg. the user can import only instances then write:

```Statements[Defn.Val].getStats(someVal)```

or create pimps dependant on `Statements` without importing all the imports.
note that in this case I have still injected both instances and implicits into the package object. its purely there for an a la carte importing similar to how Cats does it.

I personally will always use only `import scala.meta.contrib._`

I'm quite happy with whatever is decided in terms of this pattern

File: scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala (hunk 9 above)
Comment (DavidDudson) [inline scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala]: I was 50/50 about foos vs fooDefs, for example how do you differentiate between Defn.Foo's and Decl.Foo's? or do we just not support both

**Comment by DavidDudson** [comment]
I think i have taken most of the feedback into account.

I will be away for ~5 days on holiday and will have limited internet access so anyone is welcome to pick up this PR and continue it if they cannot wait.

**Comment by xeno-by** [comment]
@DavidDudson Have fun! Looking forward to your return.

File: scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala (hunk 7 above)
Comment (DavidDudson) [inline scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/Statements.scala]: Breaks is probably the wrong word.

Template.stats returns `Option[Seq[Stats]]` so we cant shadow it with a method returning `Seq[Stat]`
Given the custom behaviour @olafurpg suggested for `Pkg` means that `Pkg.stats` and `Pkg.getStats` are semantically different.

File: scalameta/contrib/src/main/scala/scala/meta/contrib/instances/StatementsInstances.scala (hunk 6 above)
Comment (olafurpg) [inline scalameta/contrib/src/main/scala/scala/meta/contrib/instances/StatementsInstances.scala]: I've changed my mind on this. I think we shouldn't hide any extra logic in this Statements type class. It should be focused on  extracting stats.

File: scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/implicits.scala (hunk 4 above)
Comment (olafurpg) [inline scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/implicits.scala]: I think we should opt for a similar strategy as scalameta and aim for a single a la carte import scala.meta.contrib._

**Comment by olafurpg** [comment]
Enjoy your vacation @DavidDudson! I've lost track of the hidden comments, GH review ui is terrible for large PRs.

I was thinking, should there be `Statements` instances for types that already have `stats` fields? If we drop Pkg, Source and Template, and Term.Block, then the method could be named `.stats` without shadowing problems.

**Comment by DavidDudson** [comment]
If you remove the instance for those you remove the ability to use all of the implicit s that require that instance. I have some ideas of how to get around this restriction but we shall see how it turns out

**Comment by olafurpg** [comment]
We could separate stats and withStats into two type classes. Pkg, Source,... could have instances for withStats. That would make it possible to provide a unified interface for all these types. We could then pimp Seq[t] for useful utilities, if we want.

**Comment by xeno-by** [comment]
Am I right in thinking that only Template is a problem here? What if we change its stats to be Seq[Stat], just like everywhere else?

**Comment by olafurpg** [comment]
I don't think Template is necessarily the problem here. If template.stats returned `Seq[Stat]`, how would you distinguish?

```scala
@ q"class a {}".templ.stats
res3: Option[collection.immutable.Seq[Stat]] = Some(List())
@ q"class a".templ.stats
res4: Option[collection.immutable.Seq[Stat]] = None
```

Although, if `stats` was renamed to something like `optStats` (or similar), then we could extend Template with a `stats: Seq[Stat]` method.

DavidDudson:
@olafurpg I don't think meta needs to distinguish between syntax details. I think syntax should provide the canonical formatting. If the user has parsed the stats, they will get the appropriate info at the token level. If they have have not and wish their default to be with brackets they should use scalafmt to show the syntax how they desire.

Note: What @xeno-by suggested and what I agree with, is having the same logic as `Pkg`

The below are identical:

```
"package foo".parseSource.get.structure
"package foo {}".parseSource.get.structure
```

I don't mind what we do as long as its consistent, but I much prefer making `templ.stats` return `Seq[Stat]`

I cant see any reason why it would matter if syntax emits `trait Foo {}` or `trait Foo`

DavidDudson:
@olafurpg I do really like the idea of splitting the copying from the extraction though. I had also thought of that.

olafurpg:
@DavidDudson There is semantic difference between the two

```scala
@ q"new X {}".structure
res11: String = """
Term.New(Template(Nil, Seq(Ctor.Ref.Name("X")), Term.Param(Nil, Name.Anonymous(), None, None), Some(Nil)))
"""
@ q"new X".structure
res12: String = """
Term.New(Template(Nil, Seq(Ctor.Ref.Name("X")), Term.Param(Nil, Name.Anonymous(), None, None), None))
"""
```

Example

```scala
@ {
    sealed abstract class A
    new A
  }
cmd14.sc:2: class A is abstract; cannot be instantiated
  val res14_1 = new A
                ^
Compilation Failed
@ {
    sealed abstract class A
    new A {}
  }
defined class A
res14_1: A = $sess.cmd14$$anon$1@4fe84a0b
```

For this reason, I think it's best to keep `templ.stats` as `Option[Seq[Stat]]`, unless we can model it somehow differently.

> I do really like the idea of splitting the copying from the extraction though. 

Sounds good. With that change, I suggest we

- rename `getStats`  to `stats`, remove Pkg/Source/Template instances for `Statements`
- rename `Statements` into `HasStatements`(?)
- create `CanCopyStatements`(?) Naming is hard, with this convention the `Equal` typeclass should really be `HasEquality`, so I don't know what's best here, suggestions welcome.
- pimp Seq[T <: Tree] with prepend/append/... (?)

DavidDudson:
@olafurpg Ah! Of course :man_facepalming: I completely forgot anon classes. So template optStats only matters when inside `New`, but not inside `Class/Trait/Object`?

olafurpg:
> So template optStats only matters when inside New, but not inside Class/Trait/Object?

Yes, I think so. Maybe we can change the ast for Templates inside `New` then. The migration cost should be taken into account if we change the AST. I don't personally mind template.stats returning an option.

DavidDudson:
@olafurpg Yeah I don't think we should change the ast's without apprehension as they break user code. I'll see if we can easily get around the restriction first

olafurpg:
Any chance to get this in for the upcoming 1.6?

DavidDudson:
Yeap. I have this afternoon off so I sure can! I completely forgot that this wasn't merged with the other stuff

DavidDudson:
Sorry all, I had troubles building meta on my laptop. Finally got round to setting up my desktop for dev. 

Heres the issues.

I think we may need 3 typeclasses.

1. `Update[A, B]`

Depends on an instance of `Extract[A, B]`

`A` = Tree in question
`B` = Thing we can replace

eg. `Update[Class, Seq[Stat]]`, `Replace[Class, Seq[Mod]]`

Provides the various methods form replacing/updating implicitly.

Use:

```
// Type parameter would probably be optional as it can be inferred. I need to test this
clazz.replace[Seq[Stat]](newStats)
clazz.replace(newStats)
clazz.update(old => old :+ q"println(\"Hello world\")")
clazz.update[Mod](old => old :+ Mod.Inline())
```

2. `Provider[A]`

This allows us to add pimps to a subset of `Extract[?, B]`

Would be specifically used for adding pimps for `stats` to any `Tree` which don't directly have a `stat: Seq[Stat]` property.

```
clazz.stats
```

2. `Extract[A, B]` // Is this just `Read[A, B]`???
Is used purely as a typeclass, to be used by `Update`

No pimps would be used directly by this. As this would include classes that already have stats, even with a different signature.

`Extract[Template, Seq[Stat]].extract(template) == Seq[Stat] // not Optiion[Seq[Stat]]`

We could possibly combine `Extract` and `Update` 

Overall this would allow us to support Mods/Stats/Annotations all from the same set of typeclasses, for all possible meta Tree's

DavidDudson:
Some extra thoughts

* We could also use `Extract/Provide` for implementing `defs/vals` etc.
* We could expose `extract/read` as a pimp with only 1 type parameter `clazz.extract[Seq[Stat]]`
* I wonder if even having `stats/defs/vars` is actually worth it.
  * We will end up having ~5-10 extra methods
  * most of the time `clazz.update` will be used
  * The other times `clazz.extract` might be good enough
  * `val stats: Seq[Stat] = clazz.extract` would actually infer correctly.
* AFAIK we will always want to return `Seq[A]`. So we might want to make the extract method return `Seq[A]` rather then A == Seq[A]

Overall i think there's 3 purposes to this changeset

- Provide helper methods for retrieving data from Tree's that conceptually should be there (eg. stats on a class)
- Provide helper methods for retrieving specific data from otherwise generic ones (eg. extracting all defs from a class via its stats)
- Provide helper methods for updating/replacing the above 2 at this higher level api

DavidDudson:
@xeno-by @olafurpg @ShaneDelmore Any opinions on this before I create the 2nd prototype?

olafurpg:
@DavidDudson I like the idea of abstracting over things with `Extract` and `Replace` (is `Update` a better name?). My concern is whether adding another type parameter

1. messes up with implicit inference
2. has worse auto-complete support in intellij

Do you maybe have any insight on that? I'm a total noob when it comes to designing such type-heavy interfaces.

DavidDudson:
@olafurpg I'm fairly noobish with type heavy stuff too. If necessary I will ask for help from the cats gitter channel.

The typeclasses are sort of there as internal constructs. The user only cares about the pimps. So I doubt we will affect users, unlesss they interact directly with the typeclasses.

olafurpg:
I say go ahead with that design and we will learn :) if it works then
perfect, as it not only solves the statement issue but also modifiers.

DavidDudson:
Move any discussion to #741

PR #639, opened 2017-02-05
Updated commons.io dependency

ShaneDelmore:
I was unable to build due to 
`Corpus.scala:7: object io is not a member of package org.apache.commons
[error] import org.apache.commons.io.FileUtils`

When troubleshooting resolution I noticed scalameta depended on the old location of commons-io was using an old version.  This change is just an update to the latest version of commons-io.  All tests pass locally for me now.


olafurpg:
LGTM, thank you for fixing this! The old dependency was copied from the scalafmt build.

PR #649, opened 2017-02-07
change scalafmt config

xeno-by:
Vertical alignment is now disabled. From personal experience, it was too much
of a disruption with little benefit.

stripMargin is now assumed to be String.stripMargin from scala-library.jar.
Allegedly, this should lead to better-looking multiline strings with margins.

### DIFF
--- a/scalahost/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
+++ b/scalahost/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
@@ -13,11 +13,11 @@ trait ScalahostPipeline { self: ScalahostPlugin =>
 
   object ScalahostComponent extends PluginComponent {
     val global: ScalahostPipeline.this.global.type = ScalahostPipeline.this.global
-    val runsAfter                                  = List("typer")
-    override val runsRightAfter                    = Some("typer")
-    val phaseName                                  = "scalameta"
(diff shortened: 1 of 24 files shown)


scalahost/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
```diff
@@ -13,11 +13,11 @@ trait ScalahostPipeline { self: ScalahostPlugin =>
...
-    val runsAfter                                  = List("typer")
-    override val runsRightAfter                    = Some("typer")
-    val phaseName                                  = "scalameta"
-    override val description                       = "compute the scala.meta semantic database"
-    def newPhase(_prev: Phase)                     = new ScalahostPhase(_prev)
+    val runsAfter = List("typer")
+    override val runsRightAfter = Some("typer")
+    val phaseName = "scalameta"
+    override val description = "compute the scala.meta semantic database"
+    def newPhase(_prev: Phase) = new ScalahostPhase(_prev)
```
xeno-by: This is so much better now /cc @olafurpg

olafurpg: I'm glad you like it. Please continue reporting what you dislike with scalafmt. 

It might be useful to configure a git hook to run scalafmt before committing.

Issue #645, opened 2017-02-06
Enable Literal Types syntax for Scala 212 when version 2.12.3 is released


xavier-fernandez:
The PR https://github.com/scalameta/scalameta/pull/640 is due to be merged, enabling Literal Types syntax on Dotty. 

In spite type literals are not yet available on Scala 2.12, it is scheduled [to have them supported by Scala 2.12.2](https://github.com/scala/scala/pull/5310). When Scala 2.12.2 is released, the Scala 2.12 dialect should enable the flag for supporting literal types: https://github.com/scalameta/scalameta/pull/640/files#L175

xeno-by:
Also, we shouldn't forget to enable trailing commas as well.

xeno-by:
Trailing commas are already enabled for `Scala212`. Only literal types are pending.

xeno-by:
Literal types got delayed again: https://github.com/scala/scala/pull/5310#issuecomment-311830956. 

Given the evident uncertainty that surrounds planning of some of the advanced language features, I think it is best to close this issue (and issues like this one) and be reactive rather than proactive in accommodating language changes.

PR #660, opened 2017-02-09
Add support for trailing commas, fixes #657

### LINKED ISSUE #657: Trailing commas in 2.12.2
PR was just merged https://github.com/scala/scala/pull/5245.

We need to allow trailing commas in 2.12.2 but not previous version. much like literal type syntax

olafurpg:
cc/ https://github.com/scala/scala-dev/issues/299

### DIFF
--- a/scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
@@ -71,6 +71,9 @@ import scala.compat.Platform.EOL
 
   // Are literal types allowed, i.e. is `val a : 42 = 42` legal or not?
   def allowLiteralTypes: Boolean
+
+  // Are trailing commas allowed? SIP-27.
+  def allowTrailingComma: Boolean
 }
 
 package object dialects {
@@ -88,6 +91,7 @@ package object dialects {
     def allowAndTypes = false
     def allowTraitParameters = false
     def allowLiteralTypes: Boolean = false
+    def allowTrailingComma: Boolean = false
     private def writeReplace(): AnyRef = new Dialect.SerializationProxy(this)
   }
 
@@ -105,6 +109,7 @@ package object dialects {
     def allowOrTypes = Scala210.allowOrTypes
     def allowTraitParameters = Scala210.allowTraitParameters
     def allowLiteralTypes: Boolean = Scala210.allowLiteralTypes
+    def allowTrailingComma: Boolean = Scala210.allowLiteralTypes
     private def writeReplace(): AnyRef = new Dialect.SerializationProxy(this)
   }
 
@@ -122,6 +127,7 @@ package object dialects {
     def allowOrTypes = Scala210.allowOrTypes
     def allowTraitParameters = Scala210.allowTraitParameters
     def allowLiteralTypes: Boolean = Scala210.allowLiteralTypes
+    def allowTrailingComma: Boolean = Scala210.allowLiteralTypes
     private def writeReplace(): AnyRef = new Dialect.SerializationProxy(this)
   }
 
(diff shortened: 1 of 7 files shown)


scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -105,6 +109,7 @@ package object dialects {
     def allowOrTypes = Scala210.allowOrTypes
     def allowTraitParameters = Scala210.allowTraitParameters
     def allowLiteralTypes: Boolean = Scala210.allowLiteralTypes
+    def allowTrailingComma: Boolean = Scala210.allowLiteralTypes
```
dwijnand: is the RHS here intentional?

olafurpg: Yes. This setup is quite verbose but it reads easily.

scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -173,6 +181,7 @@ package object dialects {
     def allowOrTypes = Scala211.allowOrTypes
     def allowTraitParameters = Scala211.allowTraitParameters
     def allowLiteralTypes: Boolean = false // Scheduled to be included in 2.12.2
+    def allowTrailingComma: Boolean = true // Schedule to be included in 2.12.2
```
dwijnand: both allowTrailingComma and allowLiteralTypes and scheduled for 2.12.2, but one is false and the other is true?

scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -122,6 +127,7 @@ package object dialects {
     def allowOrTypes = Scala210.allowOrTypes
     def allowTraitParameters = Scala210.allowTraitParameters
     def allowLiteralTypes: Boolean = Scala210.allowLiteralTypes
+    def allowTrailingComma: Boolean = Scala210.allowLiteralTypes
```
dwijnand: So this dialect is what we'll use in sbt 1, which will use Scala 2.12.2+ and therefore _will_ allow trailing commas. But in sbt between 0.13.7 and 1.0.0 (`[0.13.7,1.0.0)`) it's not. Can we model _that_ here?

scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
dwijnand: You sure you don't mean `Scala210. allowTrailingComma`?

scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
olafurpg: Good question. I think both should be true, but don't want to change allowLiteralTypes in this PR.

scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
olafurpg: There will be a custom Sbt100 (or something like that, maybe `Sbt1`) dialect where we can set all the fields as desired.

dwijnand: ok cool

scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
olafurpg: Oh boy -_- good catch, thanks! Fixed now

dwijnand: Copy-pasta, the arch-enemy of "quite verbose but it reads easily" :-P

olafurpg: Haha you're totally right. I actually hate this, was trying to find something nice to say about it. Opened https://github.com/scalameta/scalameta/issues/661

scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
xeno-by: Let's make both false and then update both to true when 2.12 is released (see https://github.com/scalameta/scalameta/issues/645).

olafurpg:
How am I supposed to test the functionality if no dialect supports trailling commas? By having an "AllowEverything" dialect ;) (I've wanted this forever!) I took the liberty to add such a dialect.

xeno-by:
@olafurpg Lol, I see :)

PR #659, opened 2017-02-09
Add 'Typelevel Scala' dialects

xavier-fernandez:
- The dialects are equal to their corresponding mainstream Scala version
  with the literal types syntax enabled.

PR #662, opened 2017-02-09
DRY up the dialects. Fixes #661

### LINKED ISSUE #661: Reorganize dialects
I propose to change `Dialect` from an `@root` ADT into a case class (or `@data`). Benefits

- Possible to create custom dialect with `.copy`, eliminates the need for https://github.com/scalameta/scalameta/pull/653
- Less boilerplate in adding new dialects, preventing copy-pasta errors https://github.com/scalameta/scalameta/pull/660/commits/cf8618b30cb9a26f69eb53547b82bd8d66957cd9
- More readable code, for example, `val Scala212 = Scala211.copy(allowTrailingComma = true, allowLiteralTypes = true)`

I suspect dialects will only continue to become more complicated with Typelevel/Lightbend scala + sbt + paradise + "next big thing"

dwijnand:
An initial solution, before Olaf's concrete suggestions in #661.

### DIFF
--- a/scalameta/common/src/main/scala/scala/meta/internal/dialects/Macros.scala
+++ b/scalameta/common/src/main/scala/scala/meta/internal/dialects/Macros.scala
@@ -18,11 +18,10 @@ class Macros(val c: Context) extends MacroHelpers with AdtReflection {
   lazy val Scala212 = mirror.staticModule("scala.meta.dialects.Scala212")
   lazy val Dotty = mirror.staticModule("scala.meta.dialects.Dotty")
 
-  def all: c.Tree = {
-    val dialects = Dialect.asRoot.allLeafs
-    val relevantDialects = dialects.filter(!_.prefix.startsWith("Quasiquote"))
-    q"${relevantDialects.map(leaf => hygienicRef(leaf.sym)).toList}"
-  }
+//  def all: c.Tree = {
+//    val dialects = Dialect.asAdt.allVals
+//    q"${dialects.map(dialect => hygienicRef(dialect.sym)).toList}"
+//  }
 
   def current: c.Tree = {
     val version = scala.util.Properties.versionNumberString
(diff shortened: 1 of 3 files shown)


xavier-fernandez:
Will rebase the PR https://github.com/scalameta/scalameta/pull/659 onto this once it is merged :)

dwijnand:
"But, but, sbt is, like, _not_ Scala"

Yes and no, it's `Scala210` but with `allowToplevelTerms = true`:

```scala
@leaf implicit object Sbt0137 extends DelegatingDialect {
  def name = "Sbt0137"
  def delegate = Scala210
  override def allowToplevelTerms = true
}
```

olafurpg:
I personally would love to have `.copy`, since for example in scalafmt I want to use a dialect that allows pretty much everything. It could be as simple as

```scala
@ case class Dialect2(allowTopLevelTerms: Boolean)(implicit private val sourceName: sourcecode.Name) {
    def name: String = sourceName.value
  }
defined class Dialect2
@ object Dialect2 {
    implicit val Scala210 = new Dialect2(
      allowTopLevelTerms = false
    )
    implicit val Sbt0137 = Scala210.copy(allowTopLevelTerms = true)
  }
defined object Dialect2
@ Dialect2.Sb
Sbt0137
@ Dialect2.Sbt0137.name
res7: String = "Sbt0137"
```

PS. sourcecode is already a dependency

xeno-by:
I second @olafurpg's suggestion.

olafurpg:
After trying to get the case class approach to work I've changed my mind and now I'm leaning more towards the approach taken by @dwijnand in this PR. See https://github.com/scalameta/scalameta/commit/1a451a339e9305a613e751de296f47915a5fc923#commitcomment-20830673 for where I got stuck with the other approach.

dwijnand:
Started working on this. But changing the dialects to be values instead of types requires a few things to change. Particularly I've got to find a way to rework `Dialect.all`.

dwijnand:
Right, it's nothing to do with using `@data` or with making sure `Dialect.all` returns the standard dialects: I'm hitting the compile errors you were hitting now.

dwijnand:
@xeno-by Have you taken this task or would you like for me to keep looking at it?

xeno-by:
@dwijnand I will fix the related macro errors and follow this through.

xeno-by:
Superseded by #678.

PR #678, opened 2017-02-12
refactor dialects

xeno-by:
Supersedes #662. /cc @dwijnand


dwijnand:
Thanks :)

dwijnand:
Technically I think you should define Dotty in terms of Scala211 instead fo 212. At the moment, as you say yourself, they're equal. But when 2.12.2 comes out with features, those aren't (necessarily) going to be available for Dotty.

scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -12,327 +13,194 @@ import scala.compat.Platform.EOL
...
+  // Normally none is required, but scripts may have their own rules.
+  toplevelSeparator: String
+) {
+  // Are unquotes ($x) and splices (..$xs, ...$xss) allowed?
+  def allowUnquotes: Boolean = allowTermUnquotes || allowPatUnquotes
+
+  // Dialects have reference equality semantics,
+  // because sometimes dialects representing distinct Scala versions
+  // can be structurally equal to each other.
+  override def canEqual(that: Any): Boolean = this eq that.asInstanceOf[AnyRef]
```
olafurpg: What if we add a field `name: String` to Dialect and manually maintain `Dialect.all: Seq[Dialect]`? Then we could get away with custom equality, custom toString and jvm reflection (which won't work on Scala.js). From my POV, Dialect should be a stupid configuration object.

scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -12,327 +13,194 @@ import scala.compat.Platform.EOL
...
+    allowTrailingCommas = false,
+    allowTraitParameters = false,
+    allowViewBounds = true,
+    allowWithTypes = true,
+    allowXmlLiterals = true, // Not even deprecated yet, so we need to support xml literals
+    toplevelSeparator = ""
+  )
+
+  implicit val Sbt0136 = Scala210.copy(
+    allowToplevelTerms = true,
```
olafurpg: YES 🙏  This reads much better 😸

scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -12,327 +13,194 @@ import scala.compat.Platform.EOL
...
   }
+}
 
-  @leaf private[meta] class QuasiquotePat(underlying: Dialect, multiline: Boolean) extends Quasiquote {
-    require(!underlying.isInstanceOf[Quasiquote])
-    def qualifier = s"QuasiquotePat"
-    private def writeReplace(): AnyRef = new Dialect.SerializationProxy(this)
-    override def toString = name
+object Dialect extends InternalDialect {
+  private[meta] lazy val standardByName: Map[String, Dialect] = {
```
olafurpg: Is this necessary? See comment above on `name: String`.

scalameta/scalameta/src/test/scala/scala/meta/tests/parsers/TrailingCommaSuite.scala
```diff
@@ -2,13 +2,10 @@ package scala.meta
...
 package parsers
 
-import scala.collection.immutable.Seq
-import scala.meta.dialects.AllowEverything
-
-import scala.meta.tokenizers.TokenizeException
-
 // Tests from https://github.com/scala/scala/pull/5245/files
 class TrailingCommaSuite extends ParseSuite {
+  implicit val Scala2122 = scala.meta.dialects.Scala212.copy(allowTrailingCommas = true)
```
olafurpg: Hurray, copy is awesome.

scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
xeno-by: If we add `Dialect.name` then people may accidentally break dialects with `Dialect.copy`, changing attributes and forgetting to change a name.

olafurpg: The sourcecode.Name macro can take care of that. Otherwise I don't see why this is a big problem.

olafurpg: See https://github.com/scalameta/scalameta/pull/662#issuecomment-278653974 for what I mean

scalameta/common/src/main/scala/scala/meta/internal/dialects/Macros.scala
```diff
@@ -13,18 +13,13 @@ class Macros(val c: Context) extends MacroHelpers with AdtReflection {
...
-  lazy val Scala212 = mirror.staticModule("scala.meta.dialects.Scala212")
-  lazy val Dotty = mirror.staticModule("scala.meta.dialects.Dotty")
-
-  def all: c.Tree = {
-    val dialects = Dialect.asRoot.allLeafs
-    val relevantDialects = dialects.filter(!_.prefix.startsWith("Quasiquote"))
-    q"${relevantDialects.map(leaf => hygienicRef(leaf.sym)).toList}"
-  }
+  lazy val dialects = mirror.staticModule("scala.meta.dialects.package")
+  lazy val Scala210 = dialects.info.member(TermName("Scala210"))
```
xeno-by: This is now dead code. Removed.

xeno-by:
@dwijnand Good point! Done.

scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala
xeno-by: Wouldn't custom equality and custom toString still be a thing? If name is automatically populate by a macro, then it won't be part of the case class signature, so the autogenerated implementations won't do.

xeno-by: As for using JVM reflection, I don't mind changing `Dialect.standard` to be populated manually.

olafurpg: > If name is automatically populated by a macro, then it won't be part of the case class signature,

Not sure I follow.

> Wouldn't custom equality and custom toString still be a thing?

Sure, custom equality and toString are fine by me. My suggestion was only to make Dialect as easy as possible to understand by making it a plain `@data` class, but I don't really mind the customization.

> As for using JVM reflection, I don't mind changing Dialect.standard to be populated manually.

We can keep reflection for now, but add a comment for our future future self that this can easily be replaced with a manual list to support Scala.js.

xeno-by: Added a commit that removes reflection, but uses it in tests to make sure that the map is populated correctly. I think that the code actually looks better now.

xeno-by: As for autopopulated names, I was referring to this code: https://github.com/scalameta/scalameta/blob/1a451a339e9305a613e751de296f47915a5fc923/scalameta/dialects/src/main/scala/scala/meta/dialects/package.scala#L132. In that case, `name` is in the second parameter list of the case class, so it won't participate in equals/hashCode, etc.

Issue #681, opened 2017-02-14
Set scalaVersion to latest in scalahost-sbt


olafurpg:
Currently, only 2.11.8 and 2.12.1 are published. This means that projects building against for example 2.12.0 get the error

```
org.scalameta#scalahost_2.12.0;1.6.0-680: not found
```

xeno-by:
Tbh I'd like to limit ourselves to just the latest minor versions of 2.11/2.12. As paradise shows, supporting everything gets old very quickly. Wdyt?

olafurpg:
I agree, would be nice if scalahost-sbt could automatically set the scala version to latest then. I currently get resolve errors running on community build.

Issue #666, opened 2017-02-10
consider generating one semanticdb file per compilation unit


xeno-by:
To quote @olafurpg:

```
- Easy to track progress, just check out many .semanticdb files there are
- Easy to zip everything into a bundle in bash. The zip can be distributed for further analysis
- Easy to open semantic mirror for single source file
```

olafurpg:
Here is a prototype implementation for this

https://github.com/scalameta/scalameta/compare/master...olafurpg:settings.d?expand=1

xeno-by:
Fixed via https://github.com/scalameta/scalameta/pull/808

Issue #665, opened 2017-02-09
unsupported file <no file>


olafurpg:
While compiling scalactic with scalahost 

```scala
java.lang.RuntimeException: unsupported file <no file>
        at scala.sys.package$.error(package.scala:27)
        at scala.meta.internal.scalahost.v1.LocationOps$XtensionGFileAddr.toAddr(LocationOps.scala:22)
        at scala.meta.internal.scalahost.v1.LocationOps$XtensionGSourceFileAddr.toAddr(LocationOps.scala:16)
        at scala.meta.internal.scalahost.v1.online.SymbolOps$XtensionGSymbolMSymbol.toSemantic(SymbolOps.scala:28)
        at scala.meta.internal.scalahost.v1.online.DatabaseOps$XtensionCompilationUnitDatabase$$anonfun$toDatabase$1$traverser$4$.scala$meta$internal$scalahost$v1$online$DatabaseOps$XtensionCompilationUnitDatabase$$anonfun$traverser$$success$1(DatabaseOps.scala:139)
        at scala.meta.internal.scalahost.v1.online.DatabaseOps$XtensionCompilationUnitDatabase$$anonfun$toDatabase$1$traverser$4$.tryMend$1(DatabaseOps.scala:167)
        at scala.meta.internal.scalahost.v1.online.DatabaseOps$XtensionCompilationUnitDatabase$$anonfun$toDatabase$1$traverser$4$.tryFindMtree(DatabaseOps.scala:219)
        at scala.meta.internal.scalahost.v1.online.DatabaseOps$XtensionCompilationUnitDatabase$$anonfun$toDatabase$1$traverser$4$.traverse(DatabaseOps.scala:271)
        at scala.meta.internal.scalahost.v1.online.DatabaseOps$XtensionCompilationUnitDatabase$$anonfun$toDatabase$1$traverser$4$.traverse(DatabaseOps.scala:133)
        at scala.reflect.internal.Trees$class.traverseComponents$1(Trees.scala:1294)
        at scala.reflect.internal.Trees$class.itraverse(Trees.scala:1330)
        at scala.reflect.internal.SymbolTable.itraverse(SymbolTable.scala:16)
        at scala.reflect.internal.SymbolTable.itraverse(SymbolTable.scala:16)
        at scala.reflect.api.Trees$Traverser.traverse(Trees.scala:2475)
[...]

Issue #650, opened 2017-02-07
semanticdb is dumped at the directory where sbt is run


xeno-by:
It looks like scalac instantiated by sbt has an incorrect path in `global.settings.d.value`, and that confuses scalahost (https://github.com/scalameta/scalameta/blob/f26801820116511945d867d83f2a3f3312d8bf44/scalahost/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala#L29).

@dwijnand I'd appreciate your advice.

dwijnand:
I don't know global.settings.d.value, but I can spend some time looking into it. I dont know scalahost, so could you tell me what the steps are to reproduce the error? sbt run at the root of the repo?

xeno-by:
Thanks for your help!

The reproduction may be a little tricky, so I'd just like to ask for your knowledge in this area. How does sbt communicate the target directory to the hosted scalac instance?

dwijnand:
It's defined in term of the project's `compile:classDirectory`, which by default is defined as:

```scala
classDirectory := crossTarget.value / (prefix(configuration.value.name) + "classes")
```

So typically `target/scala-2.12/classes`.

On the other end of the spectrum global.settings.d is set (via settings.outputDirs) like so: https://github.com/sbt/sbt/blob/v0.13.13/compile/interface/src/main/scala/xsbt/CompilerInterface.scala#L74

Ping if you need any more, happy to help.

PR #683, opened 2017-02-14
Create scalahost sbt plugin, fixes #673.

### LINKED ISSUE #673: Create scalahost-sbt plugin
Prototype here: https://gist.github.com/olafurpg/9c7cf9ddfb299d5e44d511cf99772d66

Features:
- setup the compiler plugin with correct scalacOption for >2.10 projects
- make it possible to pass in path to local jar via system properties for quick testing.

olafurpg:
- Injects scalahost compiler plugin for 2.11/2.12 projects
- Ignores 2.10 projects
- Defaults to build version.value, but possible to override with -Dscalahost.version
- Possible to use local assembly jar for fast iteration speeds during development.


build.sbt
```diff
@@ -282,8 +321,6 @@ lazy val readme = scalatex.ScalatexReadme(
 
 lazy val sharedSettings = Def.settings(
   scalaVersion := ScalaVersion,
-  crossScalaVersions := ScalaVersions,
-  crossVersion := CrossVersion.binary,
```
xeno-by: How about we keep these lines in place and just override these settings in scalahostSbt? It looks like this will significantly reduce the amount of required changes.

build.sbt
```diff
@@ -21,6 +22,19 @@ lazy val scalametaRoot = Project(
 ) settings (
   sharedSettings,
   unidocSettings,
+  // needs to be a command since scripted requires publishing across scala-versions
+  commands += Command.command("scripted") { state =>
```
xeno-by: What does "scripted" mean?

build.sbt
```diff
@@ -21,6 +22,19 @@ lazy val scalametaRoot = Project(
...
+    "such publishLocal" ::
+    "very scalahostSbt/scripted" ::
+    state
+  },
+  commands += Command.command("testAll") { state =>
```
xeno-by: How about we rename `testAll` to something different, e.g. `ci`, so it doesn't make an impression of something that just runs all tests aggregated across projects?

build.sbt
```diff
@@ -201,11 +216,35 @@ lazy val scalahost = Project(
   }
 ) dependsOn (scalameta, testkit % Test)
 
+lazy val scalahostSbt = Project(
+  id   = "scalahostSbt",
+  base = file("sbt-plugins/scalahost")
```
xeno-by: Why not have it under `scalahostSbt` like you do in scalafmt and scalafix?

sbt-plugins/scalahost/src/sbt-test/sbt-scalahost/helloworld/test
```diff
@@ -0,0 +1,5 @@
+# compile should not affect scalafix
+> dumpsemanticdb
+# TODO(olafur) verify that 2.11 and 2.12 got dumped
```
xeno-by: Do you want to verify this in this PR, or plan to submit a follow-up?

sbt-plugins/scalahost/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
```diff
@@ -0,0 +1,41 @@
...
+  override def trigger: PluginTrigger = AllRequirements
+  // Defaults to builds version.value, but with possibility to override.
+  val scalahostVersion: String = sys.props.getOrElse("scalahost.version", BuildInfo.version)
+  // If set to output of scalahost/package, uses -Xplugin:/path/to/scalahost.jar
+  // instead of addCompilerPlugin. When testing scalahost locally, this provides a
+  // much faster iteration speed compared to publishLocal.
+  // NOTE. Only possible with one scalaVersion, not cross-build projects.
+  val scalahostPackageJar: Option[String] = sys.props.get("scalahost.packagejar")
+  override def projectSettings = Seq(
+    commands += Command.command("dumpsemanticdb")(s => "test:compile" :: s),
```
xeno-by: Why do we need a separate command for this if simply running `compile` should be enough?

build.sbt
olafurpg: Good idea, done.

build.sbt
olafurpg: Good idea, I started this change before I realised it was only two setting keys.

sbt-plugins/scalahost/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
olafurpg: Before this PR, I only ran test:compile for >2.11 scala version projects and printed out useful info. Since I don't do that anymore, I've remove this command.

build.sbt
olafurpg: You mean scalahost-sbt? That's fine too, I was unsure where to put it.

sbt-plugins/scalahost/src/sbt-test/sbt-scalahost/helloworld/test
olafurpg: This is a bug, see #650. I added a link to the issue in the test.

olafurpg: I prefer not to block this PR by this bug. My personal use-case for this sbt plugin is to use #666, which I hope to bundle into a PR tomorrow.

build.sbt
```diff
@@ -325,6 +365,8 @@ lazy val mergeSettings = Def.settings(
 )
 
 def computePreReleaseVersion(LibrarySeries: String): String = {
+  // useful to provide custom version number when running sbt scripted tests, to avoid republishing 2.11/2.12
+  if (sys.props.contains("scalameta.version")) return sys.props("scalameta.version")
```
xeno-by: Shouldn't this then go in `LibraryVersion` at the top of the file?

project/project/plugins.sbt
```diff
@@ -0,0 +1 @@
+addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.0.0-M15")
```
xeno-by: Nice!

scalahost-sbt/src/sbt-test/sbt-scalahost/helloworld/build.sbt
```diff
@@ -0,0 +1,19 @@
...
+  )
+
+lazy val p1 = project.settings(scalaVersion := "2.10.5")
+lazy val p2 = project.settings(scalaVersion := "2.11.8")
+lazy val p3 = project.settings(scalaVersion := "2.12.1")
+
+TaskKey[Unit]("check") := {
+  val contents = IO.read(file("semanticdb"))
+  assert(!contents.contains("p1"))
+  // TODO(olafur) enable once https://github.com/scalameta/scalameta/pull/686 is merged.
```
xeno-by: Done

olafurpg:
I updated the sbt plugin to automatically upgrade old patch scalaVersion, fixing #681. This means that for example 2.12.0 projects will use scalahost:2.12.1. 

I also replaced the `compilerPlugin` "helper" in sbt (it seems broken for 2.12, picks the javadoc jar) with a custom `Xplugin:/scalahost.jar`.

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
```diff
@@ -0,0 +1,66 @@
...
+            .collectFirst {
+              case file if file.getAbsolutePath.matches(jarRegex) => file.getAbsolutePath
+            }
+            .getOrElse {
+              throw new IllegalStateException(
+                s"""Unable to find scalahost compiler plugin jar.
+                   |Please report the output below at https://github.com/scalameta/scalameta/issues
+                   |Scala version: ${scalaVersion.value}
+                   |Cross version: ${crossScalaVersions.value}
+                   |Jar regex: $jarRegex
```
xeno-by: When can this happen?

xeno-by:
LGTM

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
olafurpg: When the update report does not contain scalahost.jar in an expected format.

olafurpg:
BTW. The CI is super slow because it's spending like 10 minutes resolving dependencies. It seems tha the drone cache plugin is disabled when .drone.yml changes (wtf? 😭 )

olafurpg:
I ran this sbt plugin through the community build and most critical errors seems to have been fixed now. https://github.com/scalameta/scalameta/issues/665 appears in a few places. I was able to collect ~2.2 million symbols with the latest scalahost version (v1.6.0-687) compared to ~210k symbols last week.

Issue #685, opened 2017-02-14
Ensure that semanticdb is preserved for all projects in sbt-scalahost


olafurpg:
Currently, compiling two projects in parallel can cause one project to override the semanticdb of another project.

xeno-by:
Duplicate of https://github.com/scalameta/scalameta/issues/650

olafurpg:
Aha, I didn't make the connection that these were related, thanks!

PR #686, opened 2017-02-14
fix #650 #685

### LINKED ISSUE #650: semanticdb is dumped at the directory where sbt is run

xeno-by:


scalahost/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
```diff
@@ -26,7 +26,9 @@ trait ScalahostPipeline { self: ScalahostPlugin =>
 
       override def run(): Unit = {
         super.run()
-        val databaseFile = new File(global.settings.d.value + "/semanticdb")
+        val outputDir =
+          global.settings.outputDirs.getSingleOutput.getOrElse(global.settings.d.value)
+        val databaseFile = new File(outputDir + "/semanticdb")
```
olafurpg: `new File(outputDir + File.separator + "semanticdb")`?

xeno-by: Good point.

Issue #556, opened 2016-11-28
Parser for Scaladoc comments


jonas:
One approach being explored for akka/akka-http#11 is to use scala.meta to extract signatures and documentation for the various Akka HTTP directives. Having support for parsing scaladoc in scala.meta itself would allow tools like scalafmt and scalafix to keep comments formatted properly and updated. In the hope that the parser code developed for Akka HTTP can eventually make it into scala.meta the goal of this ticket is to get pointers on how to go about this.

Current thoughts:

 - Add tokens to model the different Scaladoc tags and elements: `DocToken.Param`, `DocToken.Paragraph` etc.
 - Implement a the internal `Parse` interface taking either `Token.Comment` or `Input`.
 - ...
 - Profit! :)

olafurpg:
Yes 1000x times! Thanks for starting this discussion. I see great potential in using scala.meta to generate docs.

My thoughts

- you can go with something super simple like skipping tokens and implementing a parser with fastparse to build a small DocTree structure. Scala.meta already has a dependency on fastparse for xml literals and I don't expect that to change in the near time.
- you can go with something more fancy that mimics `scala.meta.{Tree,Token}`  

The first option might be easier. If you go for the second option, some thoughts

- I would go with `DocToken`,`DocTree`,`DocParse` as the model for docstrings. This matches `scala.meta.{Token,Tree,Parse}`.
- Decide how granular DocToken should be. For example, you might allow a `DocToken.Word` to be any space separated token like `banana,` (including the comma) and then have a `DocToken.NewParagraph` (or whatever name you prefer) to be double newlines `\n\n`. Such a model would at least make it simple for scalafmt to wrap long docstrings by `maxColumn` without the need for any fancy NLP tokenizer.
- `DocTree` could have nodes like `Param(ident: DocTree.Ident, tree: DocTree)` where the right-hand side tree could be `DocTree.Text` (only words, no links) or `DocTree.Block(stats: Seq[DocTree])` where a stat could be `DocTree.Link(ident: DocTree.Ident)`.

The second approach gets tricky quite quickly, so I think the first approach by creating something simple with fastparse might be worth a shot first.

jonas:
Thank you for the pointers @olafurpg!

Do you have an idea how hard it would be to support comments in quasiquotes? Right now I am matching comments using the token start/end positions in a separate pass. It would be great to eventually have something like this working:

```
q"/** $comment */ def $name[..$tparam](..$args): $restype = $body"
```

[...]

PR #692, opened 2017-02-16
Parser for Scaladoc comments (#556)

xavier-fernandez:
TODO:

- [x] Unit test remaining scaladoc labels
- [x] Unit test multiline doc merging
- [x] Code blocks
- [x] Unit Test code blocks
- [x] Headers
- [x] Unit test Headers
- [x] Paragraph
- [x] Unit test Paragraphs

### DIFF
--- a/scalameta/contrib/src/main/scala/scala/meta/contrib/DocToken.scala
+++ b/scalameta/contrib/src/main/scala/scala/meta/contrib/DocToken.scala
@@ -1,11 +1,12 @@
 package scala.meta.contrib
 
-import DocToken.Kind
+import fastparse.all._
+import fastparse.core.Parsed
 
 /**
   * Represents a scaladoc line.
   */
-case class DocToken(kind: Kind, name: Option[String], body: Option[String]) {
+case class DocToken(kind: DocToken.Kind, name: Option[String], body: Option[String]) {
 
   override def toString: String = {
     ((name, body) match {
@@ -14,6 +15,27 @@ case class DocToken(kind: Kind, name: Option[String], body: Option[String]) {
       case _ => kind.toString
     }).replaceAll("\n", " ")
   }
+
+  /**
+    * Obtains all the references from a [[DocToken]].
+    */
+  def references: Seq[DocToken.Reference] =
+    body
+      .map { b =>
+        {
+          def parseBodyFrom(idx: Int): Seq[DocToken.Reference] = {
+            DocToken.referenceParser
+              .parse(b, idx) match {
+              case p: Parsed.Success[String, _, _] =>
+                Seq(DocToken.Reference(p.value)) ++ parseBodyFrom(p.index)
+              case _ => Seq[DocToken.Reference]()
+            }
+          }
+
+          parseBodyFrom(0)
+        }
+      }
+      .getOrElse(Nil)
 }
 
 /**
@@ -23,6 +45,26 @@ case class DocToken(kind: Kind, name: Option[String], body: Option[String]) {
   * @see http://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html
   */
 object DocToken {
+
+  /**
+    * Parser that for obtaining a class reference like '[[scala.Some]]' from an scaladoc body.
+    */
+  private val referenceParser = P(
+    // Removes the elements previous to the references.
+    ((AnyChar ~ !"[[").rep ~ AnyChar).?
+    // Retrieves the element within a
+      ~ "[[" ~
+      ((AnyChar ~ !"]]").rep ~ AnyChar).!
+      ~ "]]"
+  )
+
+  /**
+    * Represents an documentation code reference, like '[[scala.Some]]'.
+    */
+  case class Reference(to: String) extends AnyVal {
+
+    override def toString: String = s"[[$to]]"
+  }
 
   /**
     * Returns all the labelled token kinds.
--- a/scalameta/contrib/src/main/scala/scala/meta/contrib/DocToken.scala
+++ b/scalameta/contrib/src/main/scala/scala/meta/contrib/DocToken.scala
@@ -27,7 +27,7 @@ object DocToken {
   /**
     * Returns all the labelled token kinds.
     */
-  val labelledTokenKinds: Seq[LabelledKind] =
+  val tagTokenKinds: Seq[TagKind] =
     Seq(
       Constructor,
       Param,
@@ -52,6 +52,16 @@ object DocToken {
     )
 
   /**
+    * Represents a documentation remark.
+    */
+  sealed abstract class Kind
+
+  /**
+    * Represents a labeled documentation remark.
+    */
+  sealed abstract class TagKind(val label: String, val numberParameters: Int) extends Kind
+
+  /**
     * Helper [[DocToken]] apply method.
     */
   def apply(kind: Kind): DocToken = new DocToken(kind, None, None)
@@ -64,122 +74,112 @@ object DocToken {
   /**
     * Helper apply method for named [[DocToken]].
     */
-  def apply(kind: LabelledKind, name: String, body: String): DocToken =
+  def apply(kind: TagKind, name: String, body: String): DocToken =
     new DocToken(kind, Option(name), Option(body))
 
   /**
-    * Represents a documentation remark.
-    */
-  sealed abstract class Kind
-
-  /**
-    * Represents a labeled documentation remark.
-    */
-  sealed abstract class LabelledKind(val label: String, val numberParameters: Int) extends Kind
-
-  /**
-    * Companion object for [[LabelledKind]] containing
+    * Companion object for [[TagKind]] containing
     * its pattern match extractor.
     */
-  object LabelledKind {
-    def unapply(kind: LabelledKind): Option[(String, Int)] =
+  object TagKind {
+    def unapply(kind: TagKind): Option[(String, Int)] =
       Option(kind.label, kind.numberParameters)
   }
 
   /**
     * Placed in the class comment will describe the primary constructor.
     */
-  case object Constructor extends LabelledKind("@constructor", numberParameters = 1)
+  case object Constructor extends TagKind("@constructor", numberParameters = 1)
 
   /**
     * Documents a specific value parameter of a method or class constructor.
     */
-  case object Param extends LabelledKind("@param", numberParameters = 2)
+  case object Param extends TagKind("@param", numberParameters = 2)
 
   /**
     * Documents a specific type parameter of a method, class, trait or abstract type.
     */
-  case object TypeParam extends LabelledKind("@tparam", numberParameters = 2)
+  case object TypeParam extends TagKind("@tparam", numberParameters = 2)
 
   /**
     * Documents the return value of a method.
     */
-  case object Return extends LabelledKind("@returns", numberParameters = 1)
+  case object Return extends TagKind("@returns", numberParameters = 1)
 
   /**
     * Documents an exception type that may be thrown by a method or class constructor.
     */
-  case object Throws extends LabelledKind("@throws", numberParameters = 2)
+  case object Throws extends TagKind("@throws", numberParameters = 2)
 
   /**
     * Points to other sources of information such as external documentation
     * or related entities in the documentation.
     */
-  case object See extends LabelledKind("@see", numberParameters = 1)
+  case object See extends TagKind("@see", numberParameters = 1)
 
   /**
     * Documents pre- and post-conditions as well as other notable requirements
     * or restrictions.
     */
-  case object Note extends LabelledKind("@note", numberParameters = 1)
+  case object Note extends TagKind("@note", numberParameters = 1)
 
   /**
     * Provides example code and related descriptions.
     */
-  case object Example extends LabelledKind("@example", numberParameters = 1)
+  case object Example extends TagKind("@example", numberParameters = 1)
 
   /**
     * Documents a use case of a method, class, trait or abstract type.
     */
-  case object UseCase extends LabelledKind("@usecase", numberParameters = 1)
+  case object UseCase extends TagKind("@usecase", numberParameters = 1)
 
   /**
     * Attributes an entity to one author.
     */
-  case object Author extends LabelledKind("@author", numberParameters = 1)
+  case object Author extends TagKind("@author", numberParameters = 1)
 
   /**
     * The version of the system or API that a class, trait, object or
     * package is part of.
     */
-  case object Version extends LabelledKind("@version", numberParameters = 1)
+  case object Version extends TagKind("@version", numberParameters = 1)
 
   /**
     * The version of the system or API that an entity was first defined in.
     */
-  case object Since extends LabelledKind("@since", numberParameters = 1)
+  case object Since extends TagKind("@since", numberParameters = 1)
 
   /**
     * Documents unimplemented features in an entity.
     */
-  case object Todo extends LabelledKind("@todo", numberParameters = 1)
+  case object Todo extends TagKind("@todo", numberParameters = 1)
 
   /**
     * Marks an entity as deprecated. The message should
     * describe replacement implementation.
     */
-  case object Deprecated extends LabelledKind("@deprecated", numberParameters = 1)
+  case object Deprecated extends TagKind("@deprecated", numberParameters = 1)
 
   /**
     * Like [[Deprecated]] but provides advanced warning of
     * planned changes ahead of deprecation.
     */
-  case object Migration extends LabelledKind("@migration", numberParameters = 1)
+  case object Migration extends TagKind("@migration", numberParameters = 1)
 
   /**
     * Marks the entity as member of the body group.
     */
-  case object Group extends LabelledKind("@group", numberParameters = 1)
+  case object Group extends TagKind("@group", numberParameters = 1)
 
   /**
     * Provide an optional name for the group.
     */
-  case object GroupName extends LabelledKind("@groupname", numberParameters = 2)
+  case object GroupName extends TagKind("@groupname", numberParameters = 2)
 
   /**
     * Adds an optional descriptive text to display under the group name.
     */
-  case object GroupDescription extends LabelledKind("@groupdesc", numberParameters = 2)
+  case object GroupDescription extends TagKind("@groupdesc", numberParameters = 2)
 
   /**
     * Control the order of the group on the page. Defaults to 0. Ungrouped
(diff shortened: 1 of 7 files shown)


xavier-fernandez:
I expect to have this ready for review tomorrow :)

scalameta/contrib/src/main/scala/scala/meta/contrib/DocToken.scala
```diff
@@ -0,0 +1,246 @@
+package scala.meta.contrib
+
+import scala.meta.contrib.DocToken.DocKind
+
+/**
+  * Represents a scaladoc line.
+  */
+case class DocToken(kind: DocKind, name: Option[String], body: String) {
```
olafurpg: I think this ast design could improve. For example, what if there comes a new annotation with 3 arguments? Then this top-definition would need to change.

Here's an alternative
```scala
sealed abstract class DocToken
object DocToken {
  case class Param(name: String, description: String) extends DocToken
  ...
}
```

Fastparse should make it easier to handle multi-line descriptions/bodies, avoiding the need for merging lines and the `copy` macro.

scalameta/contrib/src/main/scala/scala/meta/contrib/DocToken.scala
```diff
@@ -0,0 +1,246 @@
...
+  /**
+    * Helper apply method for named [[DocToken]].
+    */
+  def apply(kind: DocKind, name: String, body: String): DocToken =
+    new DocToken(kind, Option(name), body)
+
+  /**
+    * Trait used for representing each type of documentation label.
+    */
+  sealed trait DocKind {
```
olafurpg: How about `sealed abstract class Kind(val label: String)`? That'd make it possible to do `Param extends Kind("@param")`

scalameta/contrib/src/main/scala/scala/meta/contrib/ScaladocParser.scala
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  /**
+    * Parses a Scaladoc line [[String]] into a [[DocToken]]
+    * with the line [[DocKind]], if it is known.
+    */
+  private[this] def parseScaladocLine(docLine: String): DocToken = {
+    docLine match {
+      // DocConstructor
+      case _ if docLine.startsWith(DocConstructor.label) => prepareSingleParameterToken(DocConstructor, docLine)
```
olafurpg: I'd love to find a way to simplify or abstract over this pattern match.

scalameta/contrib/src/main/scala/scala/meta/contrib/ScaladocParser.scala
```diff
@@ -0,0 +1,120 @@
...
+    require(
+      requirement = trimmedComment.startsWith("/*") && trimmedComment.endsWith("*/"),
+      message = "Input comment is not a Scaladoc"
+    )
+
+    val scaladocSymbols = Seq('/', '*', ' ')
+
+    val lineSeparedDocString: Seq[String] =
+      trimmedComment
+        .split(Array('\n', '\r')) // Splits the Scaladoc in lines
```
olafurpg: `.lines` does the same to get an `Iterator[String]`.

I think we should try to avoid parsing line-by-line however, since bodies and descriptions can be multiline. Fastparse is designed for a use-case like this.

scalameta/contrib/src/main/scala/scala/meta/contrib/DocToken.scala
```diff
@@ -0,0 +1,246 @@
...
+  def append(docToken: DocToken) : DocToken = copy(body = s"$body\n${docToken.body}")
+}
+
+/**
+  * Companion object containing the classes required for describing an ScalaDoc token.
+  *
+  * The available tokens and their documentation are obtained from:
+  * @see http://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html
+  */
+object DocToken {
```
xavier-fernandez: Done

scalameta/contrib/src/main/scala/scala/meta/contrib/ScaladocParser.scala
xavier-fernandez: On the following commit I will use '.lines'.

You are proposing using fastparse for merging the lines before doing the parsing?

xavier-fernandez: @olafurpg I have serious with my fastparser knowledge for this use case, it is the first time I use it and I am not expert enought for this type of parsing.

I have trying to find a partern for CodeBlocks, but I am unable to parse:

```
"""
bambaigimgqg
{{{
mgoqmgiqwmgqw
ignwrmiqmg
mgowqmgow
}}}
gmqwmgqwo
```

so I can obtain the code-block body. Any help for this?

Thank you :)

olafurpg: @xavier-fernandez 

Here's one example way to approach it

```
@ val codeblock = P("{{{" ~ ((AnyChar~ !"}}}").rep ~ AnyChar).! ~ "}}}")
codeblock: Parser[String] = codeblock
@ val doc = """{{{
      def codeBlockBody(foo: Bar)
  }}}
  Bla bla bla
  """"
doc: String = """
{{{
    def codeBlockBody(foo: Bar)
}}}
Bla bla bla
"
"""
@ codeblock.parse(doc)
res42: Parsed[String, Char, String] = Success(
  """

    def codeBlockBody(foo: Bar)

  """,
  39
)
```

The `codeblock` parser starts at the `{{{` and ends at the `}}}` and consumes only the input inside the blocks.

I takes a while get the hang of fastparse but once it clicks it feels very natural. Fastparse is awesome, I highly recommend the excellent docs http://www.lihaoyi.com/fastparse/

xavier-fernandez: It works :)

The only problem is that I still not understand what this is exactly doing: 
`((AnyChar~ !"}}}").rep ~ AnyChar)`

Thank you a lot, with this I will be able to continue working with the PR :)

olafurpg: That part took a while for me to come up with :) there may be a better/more-efficient way to implement it, but it does the trick and performance is not a big issue.

We can break it down into

```scala
AnyChar // any character
val x = AnyChar ~ !"}}}" // any character that is not followed by }}}, see http://www.lihaoyi.com/fastparse/#NegativeLookahead
x.rep // repeat this parser indefinitely
val y = (x.rep ~ AnyChar) // take one extra character because the character followed by }}} is not included in `x`

y.! // "cut" this parser, which means keep the value of this. an uncut parser discards it's output, like the "}}}" parser
```

xavier-fernandez: @olafurpg thank you a lot for your help, now I am getting more familiar with fastparser, I should have known this tool before, it is amazing :)

This is the final parser, it ignores also the code from the beginning, previous to the code block insertion, and it captures the start/end of the code block (For removing the codeblock from the code to parse):

```scala
case class PositiveCodeBlockParserOutput(beginIndex: Int,
                                         codeBlock: DocToken,
                                         endIndex: Int)

val codeBlockParser = {
   P(
      // Code before the code block (ignored)
      ((AnyChar ~ !"{{{").rep ~ AnyChar).?
      // Captures the index where the code block begins.
      ~ Index
      // Code block start
      ~ "{{{"
      // Code within the code block.
      ~ ((AnyChar ~ !"}}}").rep ~ AnyChar).!.map(c => DocToken(CodeBlock, c.trim))
      // Captures the index where the code block ends.
      ~ Index
      // Code block end
      ~ "}}}"
  ).map(PositiveCodeBlockParserOutput.tupled)
}
```

xavier-fernandez:
I think I am pretty satisfied with the parser, as it is right know. Anything else I should improve?

Any (or all) of the following items should be added to this commit scope?

- [ ]  `show[Syntax]` for `Seq[DocToken]`
- [ ]  `show[Structure]` for `Seq[DocToken]`
- [ ] Integrating the parser to the `Contrib` package object
- [ ]  Consider adding ScaladocComment token type

olafurpg:
Awesome work @xavier-fernandez.

- one more feature we might want to support is extracting references, for example ` [[scala.Option]]` (see [here](http://docs.scala-lang.org/style/scaladoc.html)). Any suggestion on how to add that to `DocToken`?
- maybe add a property test just checking that running the parsed on docstrings doesn't crash, something like the test here: https://github.com/scalameta/scalameta/blob/bedd3c004c37366fd72c83c7c6a96da76608eadd/scalameta/contrib/src/test/scala/scala/meta/contrib/EqualSuite.scala#L32-L52
- Structure/Syntax can be left for another PR, if you prefer
- Maybe create an implicit pimp on `Comment` calling https://github.com/scalameta/scalameta/pull/692/files#diff-d14c0bc749e7c915c8926539c90bc121R15


> Consider adding ScaladocComment token type

How would that be different from DocToken?

olafurpg:
The CI checks that the scaladocs can be generated, I think the `[[Comment]]` needs to be fully qualified as in `[[scala.meta.Comment]]`

xavier-fernandez:
Rebased into master.

@olafurpg 

```
One more feature we might want to support is extracting references, for example [[scala.Option]] (see here). Any suggestion on how to add that to DocToken?
```

I would add a method to `DocToken` for obtaining a `Seq` of all that token references. (A description may have multiple references, splitting them in a separated token kind would make working with the structure a pain, IMO).

```
maybe add a property test just checking that running the parsed on docstrings doesn't crash, something like the test here
```

It would never crash, in case it does not know how to parse it fallbacks to the `DocKind` description.

If there is nothing to parse, it returns an empty sequence directly.

```
Structure/Syntax can be left for another PR, if you prefer.
```
It would be the same, but maybe I should wait till this PR is merged: https://github.com/scalameta/scalameta/pull/702

```
Maybe create an implicit pimp on Comment calling https://github.com/scalameta/scalameta/pull/692/files#diff-d14c0bc749e7c915c8926539c90bc121R15
```

I want to do this, but I would rather make it typesafe, separating scaladoc comments and current comments in two different classes.   (Something to do with the ScaladocComment token type)

Makes sense?

olafurpg:
> working with the structure a pain

I agree, it would complicate the api.

> I would add a method to DocToken for obtaining a Seq of all that token references.

I'm not sure how useful that would necessarily be. The use-case I had in mind for extracting `[[X]]` was to generate HTML docs, and then you need to replace those references with html links. Doing that isn't so hard with regexes, however. Maybe we can just skip handling `[[X]]` until it's needed. What do you think?

> It would never crash, in case it does not know how to parse it fallbacks to the DocKind description.

That's exactly the point of the property test 😛 to verify that it does not crash on random input. There can always be some unexpected cases!

> It would be the same, but maybe I should wait till this PR is merged: #702


Yes, let's wait for that to get merged.

> I want to do this, but I would rather make it typesafe, separating scaladoc comments and current comments in two different classes. (Something to do with the ScaladocComment token type)
 
The issue I have with separating `Comment` is that we'll need a classifier to match against scaladoc comments and regular comments. I'd prefer to keep the current `Comment` but maybe we can add some methods on it like `isScaladoc: Boolean` and `contents: String` that remove the leading whitespace+asterisks.

xavier-fernandez:
> I'm not sure how useful that would necessarily be. The use-case I had in mind for extracting [[X]] was to generate HTML docs, and then you need to replace those references with html links. Doing  that isn't so hard with regexes, however. Maybe we can just skip handling [[X]] until it's needed.  What do you think?

Just did a commit implementing it, I did not saw this comment :| This method can be used for performing this replace in the future. I should revert it?

>  That's exactly the point of the property test 😛 to verify that it does not crash on random input. There can always be some unexpected cases!

Yep, I will implement it in the next commit.

> Yes, let's wait for that to get merged.

Perfect :)

> The issue I have with separating Comment is that we'll need a classifier to match against scaladoc comments and regular comments. I'd prefer to keep the current Comment but maybe we can add some methods on it like isScaladoc: Boolean and contents: String that remove the leading whitespace+asterisks.

Ok, makes sense. Making this will allow us to treat a normal comment as a Scaladoc. 

I suppose this `Comment` pimps should be implicit and within a class named `CommentOps`?

xavier-fernandez:
Rebased the code into master so I can use: https://github.com/scalameta/scalameta/pull/702

scalameta/contrib/src/main/scala/scala/meta/contrib/DocToken.scala
```diff
@@ -0,0 +1,275 @@
...
+    }).replaceAll("\n", " ")
+  }
+
+  /**
+    * Obtains all the references from a [[DocToken]].
+    */
+  def references: Seq[DocToken.Reference] =
+    body
+      .map { b =>
+        {
```
olafurpg: Redundant block?

scalameta/contrib/src/main/scala/scala/meta/contrib/DocToken.scala
```diff
@@ -0,0 +1,275 @@
...
+  * @see http://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html
+  */
+object DocToken {
+
+  /**
+    * Parser that for obtaining a class reference like '[[scala.Some]]' from an scaladoc body.
+    */
+  private val referenceParser = P(
+    // Removes the elements previous to the references.
+    ((AnyChar ~ !"[[").rep ~ AnyChar).?
```
olafurpg: I know I was the one to recommend fastparse, but i'd like to mention that regexes are actually perfect for simple stuff like this.

```scala
@ val Reference = "\\[\\[([^]]+)\\]\\]".r
Reference: Regex = \[\[([^]]+)\]\]
@ Reference.findAllMatchIn("This is [[Option]], [banana] is [[cool]].").map(_.group(1)).toList
res36: List[String] = List("Option", "cool")

@ val Codeblock = "(?s)\\{\\{\\{(.*?)\\}\\}\\}".r
Codeblock: Regex = (?s)\{\{\{(.*?)\}\}\}
@ Codeblock.findAllMatchIn(""" Doc
  {{{

    val x = 2
  }}}

  foo

  {{{ val y = 2 }}}

  banana""").map(_.group(1)).toList
res41: List[String] = List(
  """


  val x = 2

  """,
  " val y = 2 "
)
```

olafurpg: Fastparse shines when you have a recursive grammar.

xavier-fernandez: Ok, will change the code for it :)

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/CommentExtensions.scala
```diff
@@ -0,0 +1,43 @@
...
+  implicit class XtensionCommentOps(c: Comment) {
+
+    private[this] val rawSyntax: String = c.show[Syntax].trim
+
+    @inline
+    def isScaladoc: Boolean = {
+      rawSyntax.startsWith("/**") && rawSyntax.endsWith("*/")
+    }
+
+    @inline
```
olafurpg: There's no need to inline such big methods.

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/CommentExtensions.scala
```diff
@@ -0,0 +1,43 @@
...
+trait CommentExtensions {
+
+  private[this] val scaladocSymbols = Seq('/', '*', ' ')
+
+  implicit class XtensionCommentOps(c: Comment) {
+
+    private[this] val rawSyntax: String = c.show[Syntax].trim
+
+    @inline
+    def isScaladoc: Boolean = {
```
olafurpg: I recommend we keep the implementations outside the implicit class as static methods on a `CommentOps` object, just like the TreeExtensions.

scalameta/contrib/src/test/scala/scala/meta/contrib/ScaladocParserSuite.scala
```diff
@@ -218,4 +219,18 @@ class ScaladocParserSuite extends FunSuite {
         Seq(DocToken.Reference(reference1), DocToken.Reference(reference2))
     )
   }
+
+  test("property tests") {
+    // Checks that the parser does not crash with any input
+    val errors = SyntaxAnalysis.onParsed[Tree](ContribSuite.corpus) { ast =>
+      val tokens: Seq[Comment] = ast.tokens.collect { case c: Token.Comment => c }
+
+      if (tokens.map(comment => Try(comment.structure)).exists(_.isFailure)) {
```
olafurpg: Are you calling the scaladoc parser here?

scalameta/contrib/src/test/scala/scala/meta/contrib/ScaladocParserSuite.scala
```diff
@@ -218,4 +219,18 @@ class ScaladocParserSuite extends FunSuite {
...
+        Seq(ast)
```
olafurpg: You can return the culprit comment.

scalameta/contrib/src/test/scala/scala/meta/contrib/ScaladocParserSuite.scala
xavier-fernandez: yes

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/CommentExtensions.scala
```diff
@@ -0,0 +1,43 @@
...
+          .trim
+      } else {
+        rawSyntax
+          .take(rawSyntax.lastIndexWhere(!scaladocSymbols.contains(_)) + 1) // Remove trailing comments symbols
+          .dropWhile(scaladocSymbols.contains) // Removes leading comments symbols
+      }
+    }
+
+    @inline
+    def structure: Seq[DocToken] = ScaladocParser.parseScaladoc(c)
```
olafurpg: This shadows the `structure` method to show the `Structure[Comment]` (requires `import scala.meta._`). I would go with something like `docTokens`, `parseScaladoc`

scalameta/contrib/src/test/scala/scala/meta/contrib/ScaladocParserSuite.scala
xavier-fernandez: It ask me to return a Tree...

How I convert the comment into a tree?

olafurpg:
Scaladoc generation is still broken in CI. Feel free to remove the links if you like.

olafurpg:
@xeno-by @DavidDudson I'm done with my review, see comment above on why I think we're good to merge this PR.

xavier-fernandez:
For me its ok, but I would like to take a look to the `@root/@leaf` approach before the version 1.7.0 :)

DavidDudson:
LGTM for a first impl. we can always improve it later

olafurpg:
Agreed, let's go ahead and merge this then 👏 Congrats @xavier-fernandez on this large contribution

DavidDudson:
@xavier-fernandez Could you please create a ticket with the outstanding work. Just so we can keep track of it all.

xavier-fernandez:
@DavidDudson For the `@root/@leaf` support?

scalameta/contrib/src/main/scala/scala/meta/contrib/DocToken.scala
olafurpg: It seems the regex approach is not cross-platform, it got a failing test while porting contrib to Scala.js. Good to know I can replace it with fastparse :)

stefanobaghino-da:
Is this still in? I can't find this in the docs.

PR #694, opened 2017-02-16
sbt-scalahost no longer adds a library dependency on scalahost

xeno-by:


xeno-by:
Nothing to see here. Move along.

PR #696, opened 2017-02-20
Simplify sbt-semantic-example.

olafurpg:
- Automatic handling of javaOptions/fork/libraryDependencies
- Simpler Mirror construction using -Dscalameta.mirror.{sourcepath,classpath} by convention
- Sourcepath can now be a comma-separated list


scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
```diff
@@ -3,33 +3,46 @@ package scala.meta.scalahost.sbt
...
 import Keys.{version => _, _}
 import org.scalameta.BuildInfo
+import sbt.ScopeFilter.ScopeFilter
 import sbt.plugins.JvmPlugin
 
 object ScalahostSbtPlugin extends AutoPlugin {
+  object autoImport {
+    // use dependsOn(foo % Scalahost) to automatically include the semantic database
+    // of foo in the scala.meta.Mirror() constructor.
+    val Scalahost: Configuration = config("scalahost")
```
xeno-by: How about we call this `Scalameta`?

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
```diff
@@ -46,21 +59,75 @@ object ScalahostSbtPlugin extends AutoPlugin {
...
+          scalahostSourcepath.value
+            .map(_.getAbsolutePath)
+            .mkString(",")
+        val classpath =
+          scalahostClasspath.value
+            .flatMap(_.files.map(_.getAbsolutePath))
+            .mkString(java.io.File.pathSeparator)
+        List(
+          s"-Dscalameta.mirror.sourcepath=$sourcepath",
+          s"-Dscalameta.mirror.classpath=$classpath"
```
xeno-by: What if multiple projects do `scalahostAggregate`?

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
```diff
@@ -46,21 +59,75 @@ object ScalahostSbtPlugin extends AutoPlugin {
...
             "-Xplugin-require:scalahost"
           )
         case _ => Nil
       }
+    }
+  )
+
+  lazy val scalahostHasMirrorSettings: Seq[Def.Setting[_]] = Def.settings(
+    // Automatically set scalahostAggregate to projects listed in dependsOn(x % Scalahost)
+    scalahostAggregate := thisProject.value.dependencies.collect {
```
xeno-by: How about calling this `scalametaDependencies`?

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
```diff
@@ -46,21 +59,75 @@ object ScalahostSbtPlugin extends AutoPlugin {
...
+          s"-Dscalameta.mirror.classpath=$classpath"
+        )
+      }
+    },
+    // fork := true is required to pass along -Dscalameta.mirror.{classpath,sourcepath}
+    fork in run := {
+      if (scalahostAggregate.value.isEmpty) fork.in(run).value
+      else true
     },
+    // automatically depend on scalahost if this project dependsOn(otherProject % Scalahost)
```
xeno-by: Can we do the same for `libraryDependency`?

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
```diff
@@ -46,21 +59,75 @@ object ScalahostSbtPlugin extends AutoPlugin {
...
+    scalahostAggregate := thisProject.value.dependencies.collect {
+      case x if x.configuration.exists(_ == Scalahost.name) => x.project
+    },
+    javaOptions ++= {
+      if (scalahostAggregate.value.isEmpty) Nil
+      else {
+        val sourcepath =
+          scalahostSourcepath.value
+            .map(_.getAbsolutePath)
+            .mkString(",")
```
xeno-by: I suggest we use the same separator as for classpaths.

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
```diff
@@ -46,21 +59,75 @@ object ScalahostSbtPlugin extends AutoPlugin {
...
+    private val MajorMinor = "(\\d+\\.\\d+)..*".r
+    def unapply(arg: String): Option[String] = arg match {
+      case MajorMinor(version) =>
+        BuildInfo.supportedScalaVersions.find(_.startsWith(version))
+      case _ => None
+    }
+  }
+  // Defaults to version.value of in scala.meta's build.sbt
+  private val scalahostVersion: String =
+    sys.props.getOrElse("scalahost.version", BuildInfo.version)
```
xeno-by: When would one use `-Dscalahost.version`?

scalahost/src/main/scala/meta/Mirror.scala
```diff
@@ -5,9 +5,13 @@ import scala.collection.immutable.Seq
...
 import scala.meta.internal.scalahost.v1.offline.{Mirror => OfflineMirror}
+import scala.util.Try
 
 object Mirror {
   def apply(global: Global): Mirror = new OnlineMirror(global)
   def apply(classpath: String, sourcepath: String): Mirror =
     new OfflineMirror(classpath, sourcepath)
+  def loadFromSysProps(): Try[Mirror] = Try {
+    apply(sys.props("scalameta.mirror.classpath"), sys.props("scalameta.mirror.sourcepath"))
+  }
```
xeno-by: I suggest we change the signature to `def apply(): Mirror`. Here's the reasoning:

1) I'd say that `Try` (or `Option`, or `Either`) is useful when the error condition is non-fatal. Can the user do anything about this error? Can they do something alternative to achieve the same result? In our case, both answers are "no", so I'd advocate for using exceptions here.

2) I think that it is very important to have a hassle-free way to instantiate a mirror. How we implement it - via sbt properties, via analyzing `getClass.getClassLoader` - it doesn't matter that much imo. I suggest that we try to make this work via `Mirror()`, and if magic is confusing to the users then we drop it later. I guess we also have to provide a better exception message, e.g. `auto mirror is not configured. use sbt's ``% Scalameta`` to blah`.

/cc @densh

scalahost-sbt/src/sbt-test/sbt-scalahost/semantic-example/project/plugins.sbt
```diff
@@ -0,0 +1 @@
+addSbtPlugin("org.scalameta" % "sbt-scalahost" % sys.props("plugin.version"))
```
xeno-by: Where does `plugin.version` come from?

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
xeno-by: How about we have multiple `-D` properties, as in `-D<projectname>.scalameta.sourcepath`, etc?

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
olafurpg: Done.

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
olafurpg: Done.

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
olafurpg: Done.

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
olafurpg: The sys.props are unique per jvm process, so two independent projects won't clash. The only situation where this scheme fails if project A and project B both have `scalametaDependencies` and A depends on B (or vice-versa). I'm willing to bet this situation is very rare.

Using `-D<projectname>.scalameta.sourcepath` complicates the `Mirror()` constructor, since we don't know the project name.

How about doing both as a compromise? That is, `-D<projectname>.scalameta.sourcepath` and `-Dscalameta.sourcepath`

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
olafurpg: I'm sure we can, but I struggle to see the use case at this moment. Can we leave it to a future issue? This PR is already very useful.

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
olafurpg: I use this all the time for local development on sbt-scalahost. Without it, I would need to republish all packages on every build reload.

scalahost-sbt/src/sbt-test/sbt-scalahost/semantic-example/project/plugins.sbt
olafurpg: See https://github.com/scalameta/scalameta/blob/6ce2620fb37c0ce1b1cae81d547c922be569f9ba/build.sbt#L245

Scripted starts `sbt` in a temporary directory.

scalahost/src/main/scala/meta/Mirror.scala
olafurpg: I reverted back to `apply()` and cleared the config error message.

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
olafurpg: The way I use this is start sbt with `sbt -Dscalahost.version=1.6-SNAPSHOT`.

scalahost-sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
xeno-by: Sure.

xeno-by:
LGTM!

PR #701, opened 2017-02-25
Add name converters

DavidDudson:


olafurpg:
LGTM 👍 

It's a little inconsistent that `Converters` are mixed into  the `contrib` package object in this PR while the collections operations are defined inside `contrib`. Considering that we may end up with a lot of these, I think the best long-term strategy is to use the same pattern as in this PR: define pimps in separate file and mix into `contrib`.

scalameta/contrib/src/main/scala/scala/meta/contrib/Converters.scala
```diff
@@ -0,0 +1,14 @@
+package scala.meta.contrib.conversion
+
+import scala.meta._
+
+trait Converters {
+  implicit class convertName(n: Name) {
+    def asTerm: Term.Name = Term.Name(n.value)
+    def asType: Type.Name = Type.Name(n.value)
+    def asCtorRef: Ctor.Ref.Name = Ctor.Ref.Name(n.value)
+    def asPatVarTerm: Pat.Var.Term = Pat.Var.Term(n.asTerm)
```
xeno-by: I'd suggest renaming this to just `asPat`.

DavidDudson: Sounds good to me

Issue #703, opened 2017-02-26
Referential equality test failure in corpus


DavidDudson:
Change the random number to 650.

Notice that 2 files fail. Both are fully commented out.

Due to transform taking a PartialFunction.

If the input ast does not contain a single Type.Name or Term.Name. The tree will never be transformed. And referential equality will hold.

The below is an example...

```scala
/* Comment */
```

olafurpg:
Good catch! I suspected something like this might happen with the randomized tests, hence I printed out the seed to reproduce.

olafurpg:
Fixed in #704

PR #704, opened 2017-02-26
Fix bug where corpus would break

DavidDudson:


DavidDudson:
#703

olafurpg:
Thanks for taking the time to fix my bug, I appreciate it.

olafurpg:
Need to be more careful when writing property based tests.

PR #711, opened 2017-02-28
Fix #708

### LINKED ISSUE #708: Simpler `Equal` api
Following discussions on gitter here: https://gitter.im/scalameta/scalameta?at=58b4d2391465c46a56c297a8 and here https://gitter.im/scalameta/scalameta?at=58b534eade5049082225abe3

The `Equal` typeclass in contrib is a good start but could improve with

- more discoverability for beginners, for example with `equalStructurally` method. Question, should this method be included in the same `contrib._` import as `equal[T]` 
- more simplicity, e.g., replace `Structurally` with already existing `Structure`

olafurpg:


olafurpg:
One additional question: what do y'all think about forcing `b: A` instead of `b: Tree` in `isEqual`? That means 

```scala
(a: Defn.Val).equal(b: Defn.Def) // compile error, can never be true, awesome to catch this at compile time
(a: Defn.Val).equal(b: Defn) // compile error, might be true, so not so awesome to get compile error
(a: Defn).equal(b: Defn.Val) // OK, as expected
```

olafurpg:
Which approach?

DavidDudson:
Thinking specifically about syntax here... I believe the only time `isEqual[Syntax] == true[Structural] equal == false` are unparsable things. Such as `{}` and `Foo`. I think in both cases typesafe equality would break syntax equality for these kinds of things. 

In saying that I do not care.

In the case of `Name` it would be nice if it was `def isEqual(n: Name)`
in saying that, it's not too hard to write `name.value == otherName.value`
Perhaps we could add a pimp if people desire it.

I say go for it @olafurpg

scalameta/contrib/src/main/scala/scala/meta/contrib/TreeOps.scala
```diff
@@ -4,13 +4,15 @@ package contrib
...
+
 import scala.meta.contrib.equality.Equal
+import scala.meta.contrib.equality.TreeEquality
 import scala.meta.contrib.implicits.Equality._
 
 object TreeOps {
 
-  def contains[F[_ <: Tree]](tree: Tree)(toFind: Tree)(implicit conv: Tree => F[Tree],
-                                                       eqEv: Equal[F[Tree]]): Boolean =
+  def contains[F[x <: Tree] <: TreeEquality[x]](tree: Tree)(
```
DavidDudson: Lowercase X? or is my screen playing tricks on me. Could we have the param as `A` for consistency?

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/TreeExtensions.scala
```diff
@@ -36,8 +38,8 @@ trait TreeExtensions {
       TreeOps.exists(a)(f)
 
     @inline
-    def contains[F[_ <: Tree]](toFind: Tree)(implicit conv: Tree => F[Tree],
-                                             eqEv: Equal[F[Tree]]): Boolean =
+    def contains[F[x <: Tree] <: TreeEquality[x]](toFind: Tree)(implicit conv: Tree => F[Tree],
```
DavidDudson: `F[B <: Tree]`

ShaneDelmore:
@olafurpg Sorry for being unclear, I meant to vote for the typesafe equality (A) approach.

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/TreeExtensions.scala
olafurpg: The reason I make it lowercase is because `x` is not a regular type parameter, it's only scoped inside the type param list. I saw `x` used in the same situation here: https://github.com/OlivierBlanvillain/monadic-html/blob/de78d4b147e39f5fdec300cce4c2b68e3596bf3a/monadic-html/src/main/scala/scala/xml/xml.scala#L180-L182

olafurpg:
BTW. I made an unsuccessful experiment to get rid of the implicit conversions. I think the current equality solution is not fully satisfying since it quite heavy on advanced language features. However, this complexity seems to be mostly hidden to the user now that it defaults to structural equality.

changelog/1.7.0.md
```diff
@@ -0,0 +1,12 @@
...
+
+
+## Upgrade guide
+
+Contrib:
+- `equal` has been renamed to `isEqual` to avoid accidentally using `equals`,
+  which always compiles but does the wrong thing at runtime.
+- `isEqual` defaults to structual equality, so it's not longer necessary to
+  write `[Structurally]` in `a.isEqual[Structurally](b)`.
+- `(a: Subtype).isEqual(b: Supertype)` will no longer compile, either upcast
```
DavidDudson: This is not the only case.

For instance in the case of `Term.Name` and `Type.Name` they are sibling types. Thus switching the order is irrelevant most of the time. This is a typesafe equals. Meaning that the rhs must be the lhs type or a subtype of the lhs type.

olafurpg: Good point, you could argue though that the sibling case is a bug fix rather than a breaking change. Comparing sibling types will always be false in the structural equality case (although not in syntax 🤔 )

olafurpg:
Two LGTMs \o/ I'll go ahead and merge this then :)

PR #713, opened 2017-03-03
Fix for #549 from the Scalasphere Hackaton

mlangc:
Note that this also includes some tweaks to make the project more ScalaIDE friendly.

### DIFF
--- a/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2942,12 +2942,13 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
       val importees = inBraces(commaSeparated(importee()))
 
       if (importees.nonEmpty) {
-        importees.init.foreach { importee =>
-          if (importee.is[Importee.Wildcard]) {
+        importees.init.foreach {
+          case importee: Importee.Wildcard =>
             syntaxError("Wildcard import must be in the last position", importee.pos)
+
+            case _ => ()
           }
         }
-      }
 
       importees
     }
--- a/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2936,8 +2936,21 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
    *  }}}
    */
   def importees(): List[Importee] =
-    if (token.isNot[LeftBrace]) List(importWildcardOrName())
-    else inBraces(commaSeparated(importee()))
+    if (token.isNot[LeftBrace]) {
+      List(importWildcardOrName())
+    } else {
+      val importees = inBraces(commaSeparated(importee()))
+
+      if (importees.nonEmpty) {
+        importees.init.foreach { importee =>
+          if (importee.is[Importee.Wildcard]) {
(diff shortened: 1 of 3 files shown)


mlangc:
I just signed the CLA.

project/plugins.sbt
```diff
@@ -16,4 +16,6 @@ addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.6.1")
 
 addSbtPlugin("com.eed3si9n" % "sbt-doge" % "0.1.5")
 
+addSbtPlugin("com.typesafe.sbteclipse" % "sbteclipse-plugin" % "5.1.0")
```
olafurpg: Can we remove this? You can have it in a scalaide.sbt file that is ignored in you `~/.gitignore`.

scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2936,8 +2936,21 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
...
   def importees(): List[Importee] =
-    if (token.isNot[LeftBrace]) List(importWildcardOrName())
-    else inBraces(commaSeparated(importee()))
+    if (token.isNot[LeftBrace]) {
+      List(importWildcardOrName())
+    } else {
+      val importees = inBraces(commaSeparated(importee()))
+
+      if (importees.nonEmpty) {
+        importees.init.foreach { importee =>
```
olafurpg: Wow, cool, I didn't know about `.init`!

project/plugins.sbt
DavidDudson: I agree with this. Once you remove this I'm happy to merge.

Or have it globally in you .sbt folder

scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2936,8 +2936,21 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
...
+          if (importee.is[Importee.Wildcard]) {
+            syntaxError("Wildcard import must be in the last position", importee.pos)
+          }
+        }
+      }
```
xeno-by: How about `importees.init.collect { case importee: Importee.Wildcard => syntaxError(...) }`?

project/plugins.sbt
xeno-by: I agree, too.

scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
mlangc: This would look a little nicer, but it would also needlessly allocate another list.

mlangc:
As for the additional list created by `init`: I don't see any way of getting rid of this without complicating the code (for example by introducing `var lastImporteeSeenIsWildcard = false` that is then set in `foreach`). Let me know if you prefer something like this. I think that the lists we are dealing here typically have a few elements at best, and that "premature optimisation is the root of all evil".

project/plugins.sbt
mlangc: Removed ;-)

olafurpg:
No worries @mlangc about the extra list, I had just come back from a discussion with Rory about optimizations when reviewing the PR. Most importer lists are tiny (<10), so this is no big concern.

FWIW, I just setup JMH benchmarks in #726 to prepare work on tackling some performance issues in the parser/pretty-printers.

olafurpg:
Congrats @mlangc on your first scala.meta PR! 😄

PR #722, opened 2017-03-04
Update enquote and escape to be more performant

rorygraves:
Logic from escape has been duplicated in enquote for performance.
escape now appears to be unused and may be a candidate for removal.


rorygraves:
This PR is for discussion and will probably require tweaking (e.g. commit comment)

- The only use of `escape` I found was in `enquote` - it might be worth simply removing escape if that is the case.  
- do we need to create explicit benchmark for this?
- I spotted the removal of the TODO comment in the diff - happy to actually fix that as part of this PR (used to say `// TODO: comprehensive handling (e.g. escape triple quotes in triple quotes)` - although I'm not sure what counts as 'comprehensive'

DavidDudson:
I'll request that xeno and olafur review this. Thanks for the contribution!

rorygraves:
@olafurpg I've removed escape and updated the commit message.
A JMH benchmark would be great.

olafurpg:
Congratulations @rorygraves on removing more lines than you added while optimizing the implementation. Doesn't get much better :)

olafurpg:
JMH benchmark setup is ready in #726

rorygraves:
lol, I had not noticed that I had removed more  - even better ;)

PR #723, opened 2017-03-04
Fix Scaladoc parser don't managing non-scaladoc comments as expected

xavier-fernandez:


scalameta/contrib/src/test/scala/scala/meta/contrib/ScaladocParserSuite.scala
```diff
@@ -51,6 +51,31 @@ class ScaladocParserSuite extends FunSuite {
     assert(
       parseString(
         s"""
+         // $expectedBody
```
DavidDudson: Shouldn't all comments that start with `//` or `/*`  be unparsable by the scaladoc parser? That would be my expected behavior. Since they are not scaladocs.

DavidDudson: Also these tests should have their own section. they have nothing to do with indentation

DavidDudson:
I'll merge this and #724 at the same time

scalameta/contrib/src/test/scala/scala/meta/contrib/ScaladocParserSuite.scala
xavier-fernandez: The Scaladoc parser will parse them, no problem with that.

If it's unparseable what I should do? `IllegalArgumentException` with a require, return an empty `Seq`, an` `Option` or an `Either`?

Thank you :)

DavidDudson: Id expect an Option.

`parseString: Option[Seq[DocToken]]`

Once you are using `@root` then it would probably be something like `parseString: Option[DocTree]`

xavier-fernandez: @DavidDudson I did an ammend with this changes :)

xavier-fernandez:
The seams to be running now :)

Issue #716, opened 2017-03-03
Implement Equal[A] for all instances of Show


DavidDudson:
Show tends to be similar to a specific "view" of the language. Comparing these is natural.

It would be as simple as `a.show[X].isEqual(b.show[X])` but could be optimised on an individual basis.

DavidDudson:
Related to #717

olafurpg:
Contrib has been stable for a while now and is working great. Let's defer feature requests like this ticket until we have someone driving the work.

PR #725, opened 2017-03-05
WIP: Contrib equality

### LINKED ISSUE #717: Have types used for Equal and Show the same.
For example...

`Equal[Syntax] and Show[Syntax]`

This removes the amount of knowledge the user requires to use the type classes, thus making the API more user friendly.

DavidDudson:
(Ignore the first commit, that's a separate PR). This is an experimentation of how equality would look if we switched it to use Show.

This would fix #717 and #716 

You can see an example of any `Style` aka subtype of `Show` being comparable by defaulting to `a.show[Style].isEqual(b.show[Style])`

For optimization purposes, there is also an overload specifically for structure (It might be ambiguous, I cannot test it)

Pros:

- `isEqual[Positions]` and `isEqual[Summary]` will work by default

Cons: 

- It a massive change that is not able to be deprecated (Actually we may be able to introduce types for Structurally and Syntactically that just reference Structure and Syntax)

Problems:

- Something in the Higher Kinds is broken (I'm not an expert, basically still a beginner on this stuff)
  - It means you can't actually test any of this...
  - Once that's fixed everything else will be easy
  - This is largely why i stopped and pushed as is.
- There's currently no default, that can be added back later.
- There's currently no way to create sets with this approach
  - You could create a set of `Set[Structure]()` but that would lose the optimizations @olafurpg put in place 
  - It would actually be interesting to see just how much faster Structurally vs Structure is, given that the later creates a string and the former does not. (I guess memory is also increased with Structure)

Anyone is welcome to pick this up and continue working on it. I have other more pressing issues. But this was a good experiment


DavidDudson:
This also could fix #716. Although structural equality must sub type tree

scalameta/contrib/src/main/scala/scala/meta/contrib/instances/EqualityInstances.scala
```diff
@@ -1,26 +1,33 @@
...
+import scala.meta._
+import scala.meta.contrib.Equal
+import scala.meta.prettyprinters.{Show, Structure}
+import scala.language.higherKinds
+
+trait EqualityInstances {
+  implicit def showEquality[F[x] <: Show[x]]: Equal[F] =
+    new Equal[F] {
+      override def isEqual[A](a: A, b: A): Boolean =
+        a.show[F].equals(b.show[F])
```
olafurpg: I suspect the performance penalty of using Show[T] is quite high. For large trees, `.structure` builds MASSIVE strings for the entire tree while Structurally allocates almost nothing and fails fast on the first mismatching tree node.

We'll need benchmarks to confirm how significant the difference is. I was working last night on adding JMH benchmarks to scala.meta, following up from a hacking session with @rorygraves trying to speed up parsing/tree traversals.

DavidDudson: Hmmm... I wonder if we could add a common optimization pattern for show instances. I wonder however if any optimization on string based show instances.

You'll notice Structure is still optimized. This is a just a default implementation as a fallback.

olafurpg:
I am not convinced we should couple equality with `Show[T]`. However, I'd like to simplify the current implementation. I've been thinking about doing something like 

```scala
implicit class XtensionTree[A <: Tree](a: A) {
  def isEqual(b: A)(implicit ev: Equal[A]): Boolean = ev.isEqual(a, b)
}
...
implicit val structuralEquality: Equal[Tree] 
val syntaxEquality: Equal[Tree]
```


In particular, I'd like to get rid of the implicit conversions.

There would be no type parameter on `isEqual`, which is a shame, but I think since this design will instead
1. gain simplicity
2. optimize for the most common use-case (I suspect >98% of tree equality will be structural)

What do you think?

scalameta/contrib/src/main/scala/scala/meta/contrib/Equal.scala
```diff
@@ -0,0 +1,12 @@
+package scala.meta.contrib
+
+import scala.meta.prettyprinters.Show
```
xeno-by: It would seem that `Show` is now used for multiple heterogeneous purposes. How about we move it to a different package?

olafurpg:
I just saw an interesting alternative approach to the `Equal` type-class which be of interest to us: https://github.com/TomasMikula/hasheq/blob/017f289caac398723501b194cd2b36c4584df638/src/main/scala/hasheq/Equiv.scala#L9-L14

Here's an example instance: https://github.com/TomasMikula/hasheq/blob/017f289caac398723501b194cd2b36c4584df638/src/test/scala/hasheq/immutable/HashSetTest.scala#L12-L13

I haven't wrapped my head around how this changes things, but I thought it might be interesting to share.

DavidDudson:
@olafurpg Thanks for sharing, I might investigate a bit. It would be interesting to do a comparison of this vs subtyping.

DavidDudson:
Closing this, but it should still be in the logs for #717 as the comments mentioned above should be kept

scalameta/contrib/src/main/scala/scala/meta/contrib/Equal.scala
DavidDudson: I would say the typeclass `Show` should be in a separate package. The instances can still be defined in prettyprinters.

It might make sense to have:
scala.meta.typeclass
and 
scala.meta.contrib.typeclass

as packages. I'm still undecided though.

PR #726, opened 2017-03-05
Setup JMH benchmarks to test optimizations.

olafurpg:
The benchmarks can be run against 4 files of different sizes (small,
medium, large and extra large).  The files are taken from Scala.js and
Spark, to get a representative sample of "real-world" scala code.

Output from running `./bin/run-benchmarks.sh` is something like this
```
[info] Benchmark                   Mode  Cnt    Score     Error  Units
[info] Micro.ExtraLarge.transform  avgt    3  396.557 ± 596.512  ms/op
[info] Micro.Large.transform       avgt    3   92.412 ± 102.961  ms/op
[info] Micro.Medium.transform      avgt    3   23.696 ±  50.019  ms/op
[info] Micro.Small.transform       avgt    3    2.649 ±   5.275  ms/op
```

cc/ @rorygraves

### DIFF
--- a/bin/run-benchmarks.sh
+++ b/bin/run-benchmarks.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+prefix=${1-Micro}
+iterations=${2-3}
+sbt "benchmarks/jmh:run -rf csv -rff target/jmh-results.csv -i $iterations -wi $iterations -f1 -t1 org.scalameta.benchmarks.$prefix*"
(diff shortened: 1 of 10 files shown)


olafurpg:
Using the new benchmark suite, I was able to get promising performance improvements on `XtensionClassifiable.is`, see here https://github.com/scalameta/scalameta/commit/7a7dde61a8034c61d0b371a974aea0e42c544bb7

To unblock further PRs that use the benchmark suite, I'm gonna go ahead and merge this since I know @xeno-by is busy preparing his public defence.

PR #727, opened 2017-03-05
Fix compilation error in CI.

olafurpg:
```
The outer reference in this type test cannot be checked at run time.
[error]         case c: Token.Comment => c
[error]              ^
```

cc @xavier-fernandez


DavidDudson:
I was just about to do the make a pr doing a similar thing. Can we temporarily ignore the 2 failing tests so I can rebase on top of this?

olafurpg:
I disabled the entire scaladoc suite. @xavier-fernandez, please fix the scaladoc parser asap or I'm afraid we have to remove it from contrib.

xavier-fernandez:
Hi @DavidDudson, can you take a look to this PR?
https://github.com/scalameta/scalameta/pull/728

Thank you :)

PR #729, opened 2017-03-05
Contrib extract

DavidDudson:
This is Phase 1 of 3.

I may want to beef up the test suite a little bit.
The only implicit added is `extract`. More may come at a later point in time.
This can be merged as is, and the others can be pushed separately

### DIFF
--- a/scalameta/contrib/src/main/scala/scala/meta/contrib/Extract.scala
+++ b/scalameta/contrib/src/main/scala/scala/meta/contrib/Extract.scala
@@ -1,0 +1,15 @@
+package scala.meta.contrib
+
+/**
+  * Logical extraction of B from A.
+  *
+  * Meaning that this is supposed to replicate what the use thinks should happen.
+  * Not the actual class representation
+  *
+  * eg. Extract[Defn.Class, Seq[Stat]]
+  *
+  * is actually extracting the stats from the Template, which is a child of Defn.Class.
+  */
+trait Extract[A, B] {
+  def extract(a: A): B
+}
--- a/changelog/1.7.0.md
+++ b/changelog/1.7.0.md
@@ -16,4 +16,4 @@ Contrib:
   - eg. `q"trait Foo[A] { def bar = 2; val baz = 2}".extract[Vals]`
   - In future this will be supplemented by the ability to replace/update based on extract.
   - In future we will also add many more extractors. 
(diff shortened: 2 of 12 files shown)


scalameta/contrib/src/main/scala/scala/meta/contrib/Extract.scala
```diff
@@ -0,0 +1,5 @@
+package scala.meta.contrib
+
+trait Extract[A, B] {
```
olafurpg: Maybe add a docstring?

olafurpg:
BTW. It seems scalafmt is negatively impacting productivity, is there something I can do to improve the workflow/make it go out of our way? Should we disable scalafmt? I'm totally fine with that too. cc/ @xavier-fernandez

xavier-fernandez:
@olafurpg For me Scalafmt is ok, but I would enable it automatically the parsing check when doing `sbt test` on local machines, so it is hard to push unformatted code to the CI.

olafurpg:
@xavier-fernandez I run tests all the time, and I don't want to slow that workflow down by reformatting all the time. I think a less intrusive approach is to put `./bin/scalafmt --diff` in `.git/hooks/pre-commit`

DavidDudson:
Before I push a branch I run the below command

`git rebase --interactive --exec "./bin/scalafmt && git amend"`

It will rebase reformatting at every commit.

So you do not have to do it on commit, only on push. We could add a push hook.

xavier-fernandez:
@DavidDudson I would rather have a push hook for avoiding problems :)

DavidDudson:
@olafurpg The build is broken somehow https://platform-ci.scala-lang.org/scalameta/scalameta/1

olafurpg:
@DavidDudson I am working on fixing the build to work on the new drone instance. Each iteration of setting up the CI is quite long because running sbt on a fresh docker instance is super slow. It would be really nice if sbt used coursier by default

DavidDudson:
@olafurpg this may be of intrest https://github.com/sbt/sbt/issues/2997

olafurpg:
The CI server should be fixed now, you may need to force push or create a new pull request to trigger the new webook.

olafurpg:
It seems you will need to re-create a PR to use the new drone setup.

PR #728, opened 2017-03-05
Fix failing tests from the ScaladocParserSuite

xavier-fernandez:
captura de pantalla 2017-03-05 a las 22 12 27

PR #731, opened 2017-03-06
Fix failing test from the ScaladocParserSuite

xavier-fernandez:
Aparently there is a Drone bug on the https://github.com/scalameta/scalameta/pull/728 build. Following the suggestion from @olafurpg in Gitter I pushed this PR which content fixes are equivalent to the previous PR.


olafurpg:
The drone server seems to be unreliable these days. I had to re-start the build, since it was frozen for some reason.

I opened https://github.com/scalameta/scalameta/issues/732 to upgrade the CI server

olafurpg:
For some reason, the CI is freezing when running the `scalahostSbt/test` command. I don't have time to look at this right now, I will try to get back to it later this week @xavier-fernandez

xavier-fernandez:
@olafurpg, no problem at all :)

olafurpg:
The CI server should be fixed now, you may need to force push or create a new pull request to trigger the new webook.

xavier-fernandez:
@olafurpg the Drone build passed. Thank you a lot :)

olafurpg:
Great! I split the ci into slow and fast, so you should get faster feedback now if the unit tests and doc generation work. The slow ci runs integrations tests like the sbt plugin scripted tests as well as properties of the parser against millions of lines of scala code.

Next step is to intelligently run the affected tests from a diff, but that will have to wait for another day 😄 


It seems the CI broke on master because of some cache plugin issue https://platform-ci.scala-lang.org/scalameta/scalameta/19/1 

I will merge this PR as soon as the CI is green on master :)

xavier-fernandez:
@olafurpg, Perfect, amazing change, this is a huge improvement over the previous CI setting :)

olafurpg:
The cache plugin finally seems to be working, compilation starts after 23 seconds 🎉 see https://platform-ci.scala-lang.org/scalameta/scalameta/23/1

FYI I'll be away skiing for the next couple days so I may be slow to respond to PRs/issues

olafurpg:
Fast tests (2.11 and 2.12) complete in 6.5 minutes! Really nice ^^

xavier-fernandez:
I will be celebrating the Barça 6 - PSG 1 and the fast CI times on Scalameta today :)

@olafurpg Enjoy your trip :)

PR #733, opened 2017-03-07
[DO NOT MERGE] Make XtensionClassifiable a value class.

olafurpg:
(I'm opening this PR just to see if this impacts the large integration test performance on CI)

Profiling common scala.meta programs revealed that the `.is` extension
method was the biggest the source of allocations. A snapshot showed
~1.000.000 allocations from XtensionClassifiable while the second
biggest allocator had 300.000 allocations.

This change moved the implicit class to two package objects (classifiers and meta),
making is possible to make it a value class. Moreover, the Classifiable
evidence was moved from the class constructor to the `.is` method
implicit parameter list.

JMH benchmarks running the tokenizer and several `.is[T]` calls shows a ~30% speedup improvement (of total time, including tokenization)

```
// XtensionClassifiable.is
[info] Micro.ExtraLarge.is  avgt   10  28.101 ± 0.960  ms/op
// XtensionClassifiable value class
[info] Micro.ExtraLarge.is  avgt   10  19.512 ± 0.780  ms/op
```

Before making the XtensionClassifiable a value class, I explored
putting the `is` method on the `Token` and `Tree` traits.
The benchmarks from that change were not as promising:

```
// XtensionClassifiable.is
[info] Micro.ExtraLarge.is  avgt   10  28.101 ± 0.960  ms/op
// Token { def is }
[info] Micro.ExtraLarge.is  avgt   10  49.558 ± 4.823  ms/op
// Token { final def is }
[info] Micro.ExtraLarge.is  avgt   10  52.326 ± 2.968  ms/op
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -288,6 +288,7 @@ lazy val benchmarks = Project(
   )
   .settings(
     sharedSettings,
+    fork in Test := true, // easier to attach to from profilers.
     resourceDirectory in Jmh := (resourceDirectory in Compile).value,
     javaOptions in run ++= Seq(
       "-Djava.net.preferIPv4Stack=true",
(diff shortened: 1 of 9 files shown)


olafurpg:
Closing since I don't like the idea of duplicating the implementation.

PR #736, opened 2017-03-08
Upgrade to drone v0.5

olafurpg:
The template is copy-pasted from:
https://scalacenter.github.io/platform-staging/ci-integration.html#creating-the-drone.yml-file

The CI now runs two parallel tasks: ci-fast and ci-slow. The fast task runs unit tests and doc generation. For most PRs, a green response from this task is enough to merge. The slow tests run large property tests on the parser/tokenizer as well as scripted sbt plugin tests.


olafurpg:
CI passes, ship it! :shipit:

.drone.yml
```diff
@@ -1,20 +1,25 @@
...
+    restore: true
+    mount:
+      - /drone/.ivy2
+      - /drone/.coursier-cache
+      - /drone/.sbt
+      - /drone/.git
+
   tests:
-    image: ensime/ensime:v2.x-cache
+    image: scalacenter/scala-extras:1.0
```
xeno-by: What does `extras` mean?

.drone.yml
```diff
@@ -1,20 +1,25 @@
...
     pull: true
-    environment:
-      - COURSIER_CACHE=/drone/cache/coursier
     commands:
-      - git log | head -n 20
+      - git fetch --tags && git log | head -n 20
```
xeno-by: Why is `fetch --tags` necessary?

.drone.yml
```diff
@@ -28,14 +33,19 @@ build:
...
-  mount:
-    - /drone/cache
-notify:
-  email:
-    from: noreply@drone.geirsson.com
-    host: smtp.mailgun.org
-    username: noreply@drone.geirsson.com
-    password: $MAILGUN_PASSWORD
-    recipients:
-      - olafurpg+scalameta-ci@gmail.com
```
xeno-by: Will this functionality be resurrected?

project/os.scala
```diff
@@ -126,7 +126,7 @@ object git {
   }
 
   def distance(from: String, to: String): Int = {
-    shell.check_output(s"git rev-list $to ^$from --count", cwd = ".").trim.toInt
+    shell.check_output(s"git rev-list $to $from --count", cwd = ".").trim.toInt
```
xeno-by: Why was this necessary?

.drone.yml
olafurpg: Yes, this should be much easier on drone v0.5, I opened #744 to enable email notifications on build failure.

project/os.scala
olafurpg: Because the command kept crashing. I don't know why this behavior changed, it could be caused by the git version installed on the docker image.

.drone.yml
jvican: No, `scala-extras` is not the image meant to be used for the platform, we already have one for that already.
  
This has been named after the fact that it does not only bundle scala and sbt, but contains extra stuff like jekyll, git, and all sort of utils that may be handy to do scripting in sbt.

olafurpg: Maybe `scalacenter/ci` is a better name then? I don't really mind either way

Issue #741, opened 2017-03-10
Typeclasses surrounding Extract/Replace


DavidDudson:
This is a task list ticket of work to be done (Or not, depending on whether it's deemed important).

Extractors:

Allows you extract key information from the structure.

- [ ] Defns (And subclasses)
- [ ] Stats
- [ ] Mods
- [ ] Annotations
- [ ] Names
- [ ] Decls
- [ ] Expressions (Lits/Loops etc.)

Replace:

Gives a very generic functionality to replace anything you can extract.

- [ ] Defns (And subclasses)
- [ ] Stats
- [ ] Mods
- [ ] Annotations
- [ ] Names
- [ ] Decls
- [ ] Expressions (Lits/Loops etc.)

Provide:

Currently only used form Stats, gives Defn's that feel like they should have a stats list, a stats list.

- [ ] Defns

Pimps - Extract:

- `def extract[A]: A` 
- `def has[A]: Boolean` (Contemplating this for Mods. eg. `q"final class Foo".has[Mod.Inline]`
- `def has[A](A => Boolean): Boolean` (Contemplating this for Names. eg. `q"final class Foo".has[Name](_.value.startsWith("Foo"))`

Pimps - Replace:

- `def replace[A](a: A): this.type`
- `def update[A](f: A => A): this.type`

Pimps - Provide:

- `def stats: Seq[Stats]`

xavier-fernandez:
It would be really interesting to be able to extract Comments/Scaladoc as well, so we can use the `Extract` macro for obtaining also this classes, so the user does not need to learn the `AssociatedComments` and `ScaladocParser` API, only its AST.

DavidDudson:
@xavier-fernandez Hmm... I dont know much about the scaladoc tree api or Associated comments, could you give me some examples? Also Is there a spec of what exactly can have scaladoc?

xavier-fernandez:
Hi @DavidDudson, Scaladoc still doesn't have a tree API (Working on it).

But, take a look on: 

https://github.com/scalameta/scalameta/blob/c37cb1cf2166b9696219fc888cafbcb6f18769f9/scalameta/contrib/src/test/scala/scala/meta/contrib/ScaladocParserSuite.scala

And: 

https://github.com/scalameta/scalameta/blob/c37cb1cf2166b9696219fc888cafbcb6f18769f9/scalameta/contrib/src/test/scala/scala/meta/contrib/AssociatedCommentsTest.scala

[...]

PR #742, opened 2017-03-10
Contrib extract 3

DavidDudson:
Superseeds #740


xavier-fernandez:
Really nice job, can't wait to have it on master :)

On the other hand, there is no possibility to extract `Comment`s, right?

DavidDudson:
Hmm, now thats a though. Extracting Comments/Docs could be interesting. Can you add this to https://github.com/scalameta/scalameta/issues/741 and an example use case? It would be nice if we could extend this to scaladoc Tree's so that we can autogenerate scaladoc post macro transformation

changelog/1.7.0.md
```diff
@@ -10,3 +10,10 @@ Contrib:
   write `[Structurally]` in `a.isEqual[Structurally](b)`.
 - `(a: Subtype).isEqual(b: Supertype)` will no longer compile, either upcast
   `a: Supertype` or swap the order to `b.isEqual(a)`
+- The `Extract[A, B]` typeclass has been added.
```
olafurpg: How about? 

> has been added as a helper to extract `Seq[Stat/Mod]` out of `Defn.{Object,Val,...}`. 

Maybe we should move the tutorial page to the `readme` project in this repo, so we can document new features immediately and link to them in the changelog.

changelog/1.7.0.md
```diff
@@ -10,3 +10,10 @@ Contrib:
...
+  - This means that you can extract Stats/Mods/Annotations etc.
+  - eg. `q"@Foo @Bar sealed final class Foo".extract[Annotations]`
```
olafurpg: I'm afraid these examples are not sufficient enough to explain what the method does, I think it would be better to document them fully in the tutorial.

scalameta/contrib/src/main/scala/scala/meta/contrib/Extract.scala
```diff
@@ -0,0 +1,21 @@
+package scala.meta.contrib
+
+/**
+  * Logical extraction of B from A.
+  *
+  * Meaning that this is supposed to replicate what the use thinks should happen.
```
olafurpg: How about:

> The motivation for this typeclass is to extract values out of trees that could have been fields on the classes. However, these values are not included as fields on the tree nodes because of various reasons.

scalameta/contrib/src/main/scala/scala/meta/contrib/Extract.scala
```diff
@@ -0,0 +1,21 @@
...
+  * eg. Extract[Defn.Class, Seq[Stat]]
+  *
+  * is actually extracting the stats from the Template, which is a child of Defn.Class.
+  */
+trait Extract[A, B] {
+  def extract(a: A): B
+}
+
+object Extract {
+  def apply[A, B](f: A => B): Extract[A, B] = new Extract[A, B] {
```
olafurpg: What do you think about making the return type `Seq[B]`? That would avoid the need for all the `type Ts = Seq[T]` aliases. It would make it impossible to extract a single `B`, but I think all of the instances in this PR are for some `Seq[B]`.

scalameta/contrib/src/main/scala/scala/meta/contrib/implicits/ExtractPimps.scala
```diff
@@ -0,0 +1,43 @@
+package scala.meta.contrib.implicits
+
+import scala.meta._
+import scala.meta.contrib.Extract
+import scala.collection.immutable.Seq
+
+trait ExtractPimps {
+
+  // Helper types since Seq has to be immutable
```
olafurpg: See comment above.

changelog/1.7.0.md
```diff
@@ -10,3 +10,10 @@ Contrib:
...
+  - eg. `q"trait Foo[A] { def bar = 2; val baz = 2}".extract[Vals]`
+  - In future this will be supplemented by the ability to replace/update based on extract.
```
DavidDudson: Sounds good

scalameta/contrib/src/main/scala/scala/meta/contrib/Extract.scala
DavidDudson: Sounds good

DavidDudson:
@olafurpg In regards to the `Seq[A]` stuff. We could definitely settle on `Seq`, I'm finding fewer use cases where replacing/extracting/updating a single item is necessary, in theory you could also just pattern match on the `Seq` to ensure that only one item exists.

The cases I could see this being useful is, like `Seq[Mod]` where multiple `Tree` nodes have the same field but do not have the same superclass. named `Tree` nodes have `Member` as a common supertype so they are irrelevant. Can you think of any others currently?

olafurpg:
> Can you think of any others currently?

Can you elaborate?

DavidDudson:
@olafurpg Cases where:

1. We may want to abstract replacement over multiple `Tree` nodes.
2. These `Tree` nodes do not have a common supertype.
3. The field we want to extract/replace is not a `Seq[A]`

`Seq[Mod]` and `Seq[Stat]` are examples of the first 2, but not number 3.
Where we want to abstract over things that have stats, but they don't have a common supertype.
However if we switch the extract impl to return `Seq[A]` it will still work.

`Name` is an example of 1 & 3, but not 2. (Due to being able to abstract over Member).
We want to abstract over things that have names, switching the impl to return Seq would prevent this from being possible. but we can already get the same behavior by just abstracting over `Member` without the need for a typeclass instance.

Does that help explain it?

olafurpg:
I see. Good question.

> Where we want to abstract over things that have stats, but they don't have a common supertype.

What if we handle these on a case-by-case basis with methods like `Member.withName(name)`?

DavidDudson:
@olafurpg Sounds good. We shall just hit them case by case, cover the majority and worry about the minority when someone hits it. I think avoiding the typo issues with `Stat` vs `Stats` is the better path then leaving the implementation too open

olafurpg:
It's a fine art to maximize flexibility while keeping the vocabulary small. I think hardcoding `Seq[A]` for `Extract` is one example where limiting the flexibility is worth the price of avoiding a few dozen type aliases.

changelog/1.7.0.md
```diff
@@ -10,3 +10,10 @@ Contrib:
...
+  - In future we will also add many more extractors.
```
xeno-by: How about `we plan to add...`?

DavidDudson:
It looks like 2.10 scalahost failed? Is this a known failure?

olafurpg:
@DavidDudson It seems the build was running a stale commit. I don't know if that's drone's fault, but I feel like I have to sometimes push twice to make it run on an up-to-date commit, esp. when force pushing :/ (cc/ @jvican)

I fetched your branch, ran `git rebase master` and  force pushed it to your fork just to see if that helped and the build seems to have started now https://platform-ci.scala-lang.org/scalameta/scalameta/96/3 Feel free to do the same and force push over my push, I didn't mean to become an author on all the git commits.

olafurpg:
btw, we're aiming to release 1.7 on April 10th so it would be great to get this in before then ^^

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/implicits/ExtractExtensions.scala
```diff
@@ -0,0 +1,15 @@
+package scala.meta.contrib.implicits
+
+import scala.meta.contrib._
+
+trait ExtractExtensions {
+
+  /**
+    * The motivation for this implicit enrichment is to extract values out of trees
```
olafurpg: Remove? This is covered by the Extract docstring.

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/instances/ExtractAnnotationInstances.scala
```diff
@@ -0,0 +1,11 @@
...
+
+import scala.meta._
+import scala.meta.contrib._
+
+trait ExtractAnnotationInstances {
+  implicit def extractAnnotationsFromMods[A](implicit ev: Extract[A, Mod]): Extract[A, Mod.Annot] =
+    Extract(a => ev.extract(a).collect { case m: Mod.Annot => m })
+}
+
+object ExtractAnnotationInstances
```
olafurpg: `extends ExtractAnno...`

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/instances/ExtractStatSubtypeInstances.scala
```diff
@@ -0,0 +1,53 @@
+package scala.meta.contrib.instances
+
+import scala.meta._
+import scala.meta.contrib._
+
+trait ExtractStatSubtypeInstances {
+  implicit def extractDefsFromStats[A](implicit ev: Extract[A, Stat]): Extract[A, Defn.Def] =
+    Extract(a => ev.extract(a).collect { case d: Defn.Def => d })
```
olafurpg: We can probably abstract over this boilerplate with a ClassTag. However, this is fine by me.

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/instances/ExtractStatSubtypeInstances.scala
```diff
@@ -0,0 +1,53 @@
...
+    Extract(a => ev.extract(a).collect { case v: Decl.Var => v })
+
+  implicit def extractDeclValsFromStats[A](implicit ev: Extract[A, Stat]): Extract[A, Decl.Val] =
+    Extract(a => ev.extract(a).collect { case v: Decl.Val => v })
+
+  implicit def extractDeclTypesFromStats[A](implicit ev: Extract[A, Stat]): Extract[A, Decl.Type] =
+    Extract(a => ev.extract(a).collect { case v: Decl.Type => v })
+}
+
+object ExtractStatSubtypeInstances
```
olafurpg: `extends`

scalameta/contrib/shared/src/test/scala/scala/meta/contrib/StatExtractionTest.scala
```diff
@@ -0,0 +1,45 @@
...
+    assert(stats == Nil)
+  }
+
+  test("Test extract object stats") {
+    val stats = q"object Foo".extract[Stat]
+    assert(stats == Nil)
+  }
+
+  test("Test extract def stats") {
+    val stats = q"def foo = {}".extract[Stat]
```
olafurpg: Add test for `def foo = {2}`?

changelog/1.7.0.md
```diff
@@ -10,3 +10,9 @@ Contrib:
   write `[Structurally]` in `a.isEqual[Structurally](b)`.
 - `(a: Subtype).isEqual(b: Supertype)` will no longer compile, either upcast
   `a: Supertype` or swap the order to `b.isEqual(a)`
+  
+## `contrib`
+- The `Extract[A, B]` typeclass has been added.
+  - This means that you can extract Stats/Mods/Annotations/Defs.
+  - More information about this will be in the tutorial
```
olafurpg: Replace with this line with "See testsuite (link) for more example usages"?

changelog/1.7.0.md
```diff
@@ -10,3 +10,9 @@ Contrib:
   write `[Structurally]` in `a.isEqual[Structurally](b)`.
 - `(a: Subtype).isEqual(b: Supertype)` will no longer compile, either upcast
   `a: Supertype` or swap the order to `b.isEqual(a)`
+  
+## `contrib`
+- The `Extract[A, B]` typeclass has been added.
```
olafurpg: been added for convenient extraction of Stat/Mod/Def/...

changelog/1.7.0.md
```diff
@@ -10,3 +10,9 @@ Contrib:
...
+  - This means that you can extract Stats/Mods/Annotations/Defs.
```
olafurpg: Remove.

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/implicits/ExtractExtensions.scala
DavidDudson: I guess we can. I more thought its a point of contact for people using "Jump to symbol" to try and find out what this method does

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/instances/ExtractStatSubtypeInstances.scala
DavidDudson: Hmmm. we could indeed

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/instances/ExtractStatSubtypeInstances.scala
DavidDudson: Done

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/instances/ExtractAnnotationInstances.scala
DavidDudson: Done

scalameta/contrib/shared/src/test/scala/scala/meta/contrib/StatExtractionTest.scala
DavidDudson: Done

changelog/1.7.0.md
DavidDudson: Done

changelog/1.7.0.md
DavidDudson: Done

changelog/1.7.0.md
DavidDudson: Can we add the link post push? And which test page would the best to show?

DavidDudson:
Okay thats done.

@olafurpg Don't worry about authorship. It's the last thing I'm concerned about.

PR #744, opened 2017-03-11
Enable email plugin for build failures.

olafurpg:
Secrets are available as environment variables in push/tag/deploy
events. With Drone v0.5, it became no longer necessary to do
`environment: YY=$$YY`.


olafurpg:
Wooo hoo! It works 😄 

screen shot 2017-03-13 at 10 37 18

PR #747, opened 2017-03-15
Small terminology change in docs

acjay:
Switching "pimp" terminology for "enrich", per https://groups.google.com/forum/#!topic/scala-internals/kVWg7HKldjs. Many prefer the latter because the former has the potential to offend, and at the very least isn't self-explanatory to people unfamiliar with the decade-old TV show "Pimp My Ride". I hope this is uncontroversial!


DavidDudson:
If it's good enough for Scala internal it's good enough for me. I can merge once the build passes

olafurpg:
LGTM 👍 Thank you @acjay, I agree pimps is not a good name. I personally use "extension methods", but enrichment works too.

PR #748, opened 2017-03-15
Use protobuf to persist semanticdb.

olafurpg:
The current persistence scheme for the semanticdb relies on an ad-hoc
serialization format. This is problematic if we want to add/remove fields
while keeping compatibility with existing semanticdb.

This commit migrates the serialization format to protobuf, which will
make it easier to evolve the schema without breaking compatibility
with existing semanticdb. Moreover, by delegating the serialization task
to an established library, we avoid the implementation cost of
implementing custom serializers and de-serializers.

Why protobuf instead of thrift+scrooge?

1. scalapb supports Scala.js, which may come in handy when
   building web-based tools such as code-search/refactoring UIs.
2. I personally have more experience with protobuf.

Why protobuf instead of scodec?

1. scodec is Scala only, while protobuf has bindings for a lot of
   languages. Someone might want to build tools/applications on top
   of semanticdb using another language than Scala.
2. protobuf has more documentation and is more widely used.
3. I don't have much experience with scodec to judge how hard it would
   be to design a protocoal that makes it easy to add/remove fields from
   the schema without breaking backward compat.

If it turns out that thrift is a better fit for this use-case, I think
it will still be possible to migrate. The semanticdb will only contain
simple field types to start with (mostly int32/string), which are
supported by both protobuf and thrift.


olafurpg:
I kept the proto.Database schema as close as possible to the original Database. However, I propose to make this change

```proto
message Database {
  repeated DatabaseFile files = 1;
}

message DatabaseFile {
  Address address = 1;
  SymbolRef symbols = 2;
}

message Address {
  string path = 1; // Address.file
  string contents = 2; // Address.snippet
}

message SymbolRef {
  Range range = 1;
  string symbol = 2;
}

message Range {
  int32 start = 2;
  int32 end = 3;
}
``` 

Why?

- More compact binary representation (no duplicate file  path for every SymbolRef)
- easier to combine two databases (I'm not 100% sure, but I think we can pretty much concatenate the byte arrays)

scalameta/semantic/src/main/scala/scala/meta/semantic/v1/Database.scala
```diff
@@ -35,77 +31,10 @@ case class Database(
...
-    val fis = new FileInputStream(file)
-    try {
-      val length = fis.available
-      val bytes = new Array[Byte](length)
-      fis.read(bytes)
-      Database(bytes)
-    } finally {
-      fis.close()
-    }
-  }
```
xeno-by: You know, the more I think of it, the more I have a feeling that we can leave these methods here. I suggest we view serialization/deserialization as a core aspect of our data model and that means these facilities should be available in semantic, not scalahost (e.g. to be reused by dottyhost, intellijhost, etc).

scalameta/semantic/src/main/scala/scala/meta/semantic/v1/Symbol.scala
```diff
@@ -225,6 +225,9 @@ object Symbol {
     }
     naiveParser.entryPoint()
   }
+
+  def unapply(arg: String): Option[Symbol] =
+    scala.util.Try(apply(arg)).toOption
 }
```
xeno-by: Neat!

scalahost/src/main/protobuf/semanticdb.proto
```diff
@@ -0,0 +1,34 @@
+syntax = "proto3";
+
+package org.scalameta.semantic.v1.proto;
```
xeno-by: I suggest `scala.meta.internal.semantic.v1.protobuf`. `org.scalameta` is typically used for functionality reusable between projects inside `scalameta/*`.

scalahost/src/main/protobuf/semanticdb.proto
```diff
@@ -0,0 +1,34 @@
...
+  Address address = 1;
+  repeated SymbolRef symbols = 2;
+}
+
+message Address {
+  string path = 1; // Address.file
+  string contents = 2; // Address.snippet
+}
+
+message SymbolRef {
```
xeno-by: `ResolvedName`?

scalahost/src/main/scala/scala/meta/internal/scalahost/v1/offline/Mirror.scala
```diff
@@ -51,7 +51,9 @@ class Mirror(classpath: String, sourcepath: String)
         else Nil
       })
       .sortBy(_.getName)
-    val databases = databaseFiles.map(Database.apply)
+    val databases = databaseFiles.map(SerializationOps.fromFile).collect {
+      case scala.util.Success(db) => db
+    }
```
olafurpg: Done.

scalameta/semantic/src/main/scala/scala/meta/internal/semantic/v1/DatabaseOps.scala
```diff
@@ -7,6 +7,11 @@ import scala.collection.mutable
 import scala.meta.semantic.v1.Database
 import scala.meta.semantic.v1.Location
 import scala.meta.semantic.v1.Symbol
+import scala.util.Try
+
+import java.io.File
+import java.io.FileInputStream
+import java.io.FileOutputStream
```
olafurpg: Nope.

scalameta/semantic/src/main/scala/scala/meta/semantic/v1/Database.scala
olafurpg: That includes supporting new fields such as messages from #749. If the database text-based parser is buggy/not bijective, do you really want to support it? 

I'm sure it's easier to port the protobuf logic to support dottyhost/intellijhost.

scalahost/src/main/protobuf/semanticdb.proto
olafurpg: I opted for `org.scalameta` because scalapb generated the package name 

```
package `scala`.scalameta.semantic.v1...
```

It's a bug in scalapb, I guess I should report it, but I want the lazy route and used org.scalameta as a workaround.

scalameta/semantic/src/main/scala/scala/meta/semantic/v1/Database.scala
xeno-by: Aren't we retiring the text-based parser?

scalahost/src/main/protobuf/semanticdb.proto
xeno-by: Allright I think we can live with the workaround then :)

scalameta/semantic/src/main/scala/scala/meta/semantic/v1/Database.scala
xeno-by: Oh I got it. I mean I suggest to remove the old text-based logic and replace it with the new pb-based logic, so that scalameta/semantic can serialize/deserialize via protobuf.

ShaneDelmore:
Not sure how much value there is in keeping an old semantic db and migrating it when schema changes, but if it is valuable and PB doesn't make it easy it's worth knowing that there are similar formats that have explicit support for schema evolution. Avro is a good example.

xeno-by:
LGTM

olafurpg:
@ShaneDelmore think for example about code search, which uses the semanticdb to support jump to definition. We care greatly about being able to load old semanticdb for legacy code, even if it contains outdated fields.

olafurpg:
I have not used avro, does the Scala codegen bindings support Scala.js?

ShaneDelmore:
Probably no support for Scala.js

Here is an older but useful post on different approaches to schema evolution. In case it helps with schema design regardless of the protocol http://martin.kleppmann.com/2012/12/05/schema-evolution-in-avro-protocol-buffers-thrift.html

olafurpg:
That is an interesting article, thank you for sharing @ShaneDelmore.

PR #752, opened 2017-03-16
Explicitly disable publishing for readme+benchmarks

olafurpg:
While publishing a snapshot to bintray, I noticed that benchmarks was publishing to sonatype, because it uses the default publish settings.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -289,6 +289,7 @@ lazy val benchmarks = Project(
   )
   .settings(
     sharedSettings,
+    noPublish,
     resourceDirectory in Jmh := (resourceDirectory in Compile).value,
     javaOptions in run ++= Seq(
       "-Djava.net.preferIPv4Stack=true",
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -530,6 +532,12 @@ lazy val buildInfoSettings = Def.settings(
   buildInfoObject := "BuildInfo"
 )
 
+lazy val noPublish = Seq(
+  publishArtifact := false,
+  publish := {},
+  publishLocal := {}
+)
```
xeno-by: I suggest we deduplicate this with https://github.com/scalameta/scalameta/blob/d38a0f9cd5e58373967d31e2a9ebe1ee67662690/build.sbt#L50-L51.

olafurpg: I reused noPublish in scalametaRoot: https://github.com/scalameta/scalameta/pull/753/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR24

PR #755, opened 2017-03-16
Make scalametaDependencies taskKey public.

olafurpg:
I want to use this setting in scalafix to rewrite all projects.


xeno-by:
Sure thing, why not.

PR #760, opened 2017-03-22
Shrink scalahost/assembly by 10mb

olafurpg:
Apparently, those two lines added a dependency on the protobuf compiler and a whole boatload of other stuff that we don't need. Now, the scalahost assembly is 14mb instead of 24mb.


olafurpg:
I'm gonna go ahead and merge this to unblock #761

PR #761, opened 2017-03-23
Port to Scala.js

olafurpg:
- `java.io.File` is now `scala.meta.io.AbsolutePath`.
  The io module abstracts over the platform, making it possible
  to hook in node.js modules to write JavaScript command-line apps.
- The `Input.File(file: java.io.File)` constructor is still kept,
  usage of those constructors will lead to linking errors in a JS app.
  The scalametaJS/test will very likely guard us against using those
  constructors in the implementation of the `scalameta` projects.
- The JVM api remains otherwise unchanged :tada:
- `Lit.{Float,Double}.syntax` is different on JVM and JS. Still WIP.
- Binary serialization/deserialization of semanticdb is available on
  both JVM and JS, thanks to Scala.js support in scalapb.
- WeakHashMap is replaced with java.util.Map on JS, we could port the
  implementation to js.WeakMap if memory/GC becomes a problem in
  scala.meta js applications.
- Reflection tests only run on the JVM.
- Scalahost remains JVM only.


scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
```diff
@@ -184,26 +184,26 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
     assert(Dotty(q"trait T(a: Int)").syntax === "trait T(a: Int)")
   }
 
-  test("literalTypes") {
```
olafurpg: Just remembered that I forgot to fix syntax for literals. Will fix asap.

scalameta/io/js/src/main/scala/scala/meta/io/PlatformIO.scala
```diff
@@ -0,0 +1,15 @@
+package scala.meta.io
+
+import java.io.InputStream
+import java.nio.charset.Charset
+
+object PlatformIO {
+  def workingDirectory: AbsolutePath = AbsolutePath(fileSeparator)
+  def read(stream: InputStream, charset: Charset): String =
+    throw new IllegalStateException("Reading from java.io.InputStream is not supported in JavaScript.")
```
olafurpg: These can be replaced sometime with Node.js io libraries.

sjrd: You don't need Node.js libraries for this. You can implement it in portable Scala.js code. We support both `InputStream` and `Charset`, as well as `InputStreamReader`. You simply have to avoid `scala.io`, but everyone should do that anyway.

Wrap the `stream` in a `new InputStreamReader(stream, charset)`, then loop on its `read` method until you exhaust the input, while appending everything in the result string. Something like the following (which is even portable on the JVM too, and faster than `scala.io.Source`):
```scala
def read(stream: InputStream, charset: Charset): String = {
  val reader = new InputStreamReader(stream, charset)
  val buffer = new Array[Char](4096)
  val builder = new java.lang.StringBuilder

  @tailrec def loop(): Unit = {
    val charsRead = reader.read(buffer)
    if (charsRead >= 0) {
      builder.append(buffer, 0, charsRead)
      loop()
    }
  }

  loop()
  reader.close() // check whether you need this
  builder.toString()
}
```

olafurpg: Thanks! `scala.io.Source.fromStream` failed to link so I simply assumed it was not possible. I've moved the method out of the `PlatformIO` object into an `InputStreamIO` object in the `shared/` folder.

olafurpg:
Okay, I think I've managed to clean up everything unexpected that came up in CI :) 

For the record, the CI was green at least once here: https://platform-ci.scala-lang.org/scalameta/scalameta/11 The latest commits have been to clean up some remaining quirks in the port.

It would be great to try and merge this before we merge other large changes in the repo. This PR moved most `.scala` files in the repo so I'd prefer to avoid conflict hell when rebasing.

olafurpg:
The CI is green now. The tests failed last night because of OOM errors.

olafurpg:
The most urgent issue is to fix https://github.com/scalameta/scalameta/issues/762 I gave it a try but hit on cryptic errors in the `@ast` macros.

build.sbt
```diff
@@ -2,379 +2,402 @@ import java.io._
...
-
-lazy val benchmarks = Project(
-    id = "benchmarks",
-    base = file("scalameta/benchmarks")
+lazy val ioJVM = io.jvm
+lazy val ioJS = io.js
+
+lazy val dialects = crossProject
+  .in(file("scalameta/dialects"))
+  .settings(publishableSettings, description := "Scala.meta's dialects", enableMacros)
```
xeno-by: For arguments of `settings`, I'd prefer a uniform style of one-per-line, indented-with-two-spaces.

build.sbt
```diff
@@ -2,379 +2,402 @@ import java.io._
...
+
+lazy val transversers = crossProject
+  .in(file("scalameta/transversers"))
+  .settings(
+    publishableSettings,
+    description := "Scala.meta's traversal and transformation infrastructure for abstract syntax trees",
+    enableMacros)
+  .dependsOn(common, trees)
+lazy val traversersJVM = transversers.jvm
+lazy val traversersJS = transversers.js
```
xeno-by: The `lazy val blahJVM = blah.jvm; lazy val blahJS = blah.js` is getting a little bit old by now :) I wonder whether there is a possibility to abstract this away.

build.sbt
```diff
@@ -383,43 +406,51 @@ lazy val readme = scalatex.ScalatexReadme(
...
+    }
+  },
   version := LibraryVersion,
   organization := "org.scalameta",
   resolvers += Resolver.sonatypeRepo("snapshots"),
   resolvers += Resolver.sonatypeRepo("releases"),
-  addCompilerPlugin("org.scalamacros" % "paradise" % "2.1.0" cross CrossVersion.full),
-  libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.1" % "test",
-  libraryDependencies += "org.scalacheck" %% "scalacheck" % "1.13.4" % "test",
+  addCompilerPlugin(("org.scalamacros" % "paradise" % "2.1.0").cross(CrossVersion.full)),
```
xeno-by: I'm not 100% sure I'm happy about making `cross` invocations non-infix.

build.sbt
```diff
@@ -383,43 +406,51 @@ lazy val readme = scalatex.ScalatexReadme(
...
+  libraryDependencies += "org.scalatest" %%% "scalatest" % "3.0.1" % "test",
+  libraryDependencies += "org.scalacheck" %%% "scalacheck" % "1.13.5" % "test",
   scalacOptions ++= Seq("-deprecation", "-feature", "-unchecked"),
-  scalacOptions in (Compile, doc) ++= Seq("-skip-packages", ""),
-  scalacOptions in (Compile, doc) ++= Seq("-implicits", "-implicits-hide:."),
-  scalacOptions in (Compile, doc) ++= Seq("-groups"),
+  scalacOptions.in(Compile, doc) ++= Seq("-skip-packages", ""),
+  scalacOptions.in(Compile, doc) ++= Seq("-implicits", "-implicits-hide:."),
+  scalacOptions.in(Compile, doc) ++= Seq("-groups"),
```
xeno-by: Same about `in`. These are supposedly idiomatic uses of the SBT DSL copied from the official documentation.

build.sbt
```diff
@@ -457,50 +491,53 @@ def shouldPublishToSonatype: Boolean = {
...
 }
 
+lazy val sonatypeCredentials = os.secret
+  .obtain("sonatype")
+  .map {
+    case (username, password) =>
+      Credentials("Sonatype Nexus Repository Manager", "oss.sonatype.org", username, password)
+  }
+  .toList
+
```
xeno-by: Is this val supposed to be used elsewhere apart from `credentials`?

build.sbt
```diff
@@ -579,14 +621,48 @@ lazy val enableMacros = macroDependencies(hardcore = false)
...
   val backwardCompat210 = {
-    if (scalaVersion.value.startsWith("2.10")) Seq("org.scalamacros" %% "quasiquotes" % "2.1.0")
+    if (scalaVersion.value.startsWith("2.10"))
+      Seq("org.scalamacros" %% "quasiquotes" % "2.1.0")
     else Seq()
   }
   scalaReflect ++ scalaCompiler ++ backwardCompat210
 }
+
+lazy val compilePublish: Def.Initialize[Task[Unit]] = Def.taskDyn {
```
xeno-by: Same question as for `sonatypeCredentials`

project/os.scala
```diff
@@ -5,6 +5,12 @@ import java.io._
 import scala.sys.process._
 import scala.compat.Platform.EOL
 
+object time {
+  // cached here to use consistent time stamp across sbt reloads, which happen
+  // for every ++SCALA_VERSION switch in sbt "very publish"
+  val current = System.currentTimeMillis().toString
```
xeno-by: This isn't quite current. How about renaming to e.g. `time.stamp`?

scalameta/io/js/src/main/scala/scala/meta/io/PlatformIO.scala
```diff
@@ -0,0 +1,12 @@
+package scala.meta.io
+
+import java.nio.charset.Charset
+
+object PlatformIO {
+  def workingDirectory: AbsolutePath = AbsolutePath(fileSeparator)
+  def read(path: AbsolutePath, charset: Charset): String = read(path)
+  def read(path: AbsolutePath): String =
+    throw new IllegalStateException("Reading from an AbsoluteFile is not supported in JavaScript.")
```
xeno-by: What's an `AbsoluteFile`?

scalameta/io/shared/src/main/scala/scala/meta/io/AbsolutePath.scala
```diff
@@ -0,0 +1,30 @@
+package scala.meta.io
+
+import PlatformIO._
+
+// sealed abstract to ensure that Absolute.apply constructor is used.
```
xeno-by: `AbsolutePath`?

scalameta/io/shared/src/main/scala/scala/meta/io/Api.scala
```diff
@@ -0,0 +1,12 @@
+package scala.meta.io
+
+private[meta] trait Api {
+  implicit class XtensionAbsolutePath(path: AbsolutePath) {
+    def read: String = PlatformIO.read(path)
```
xeno-by: `slurp`?

scalameta/quasiquotes/shared/src/main/scala/scala/meta/internal/quasiquotes/ReificationMacros.scala
```diff
@@ -122,10 +126,7 @@ class ReificationMacros(val c: Context) extends AstReflection with AdtLiftables
         }
         curr + 1
       }
-      val result = {
-        if (reflectInput.file.file != null) Input.File(reflectInput.file.file)
-        else Input.String(new String(reflectInput.content)) // NOTE: can happen in REPL or in custom Global
-      }
+      val result = Input.String(new String(reflectInput.content))
```
xeno-by: I'm afraid this may degrade error reporting for invalid quasiquotes. Is it possible to restore the initial version?

scalameta/scalameta/jvm/src/test/scala/scala/meta/tests/prettyprinters/LiteralSyntacticSuite.scala
```diff
@@ -0,0 +1,54 @@
...
+
+import org.scalatest._
+import scala.meta._
+import scala.meta.dialects.Scala211
+import scala.meta.internal.ast._
+
+// TODO: move back into SyntacticSuite. Scala.js does not support runtime checks
+// on primitives like Float/Double/Char/Byte/... See:
+// https://www.scala-js.org/doc/semantics.html#runtime-type-tests-are-based-on-values
+// The TreeSyntax pretty printer relies on runtime type checks.
```
xeno-by: Fixed in https://github.com/scalameta/scalameta/pull/763.

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/PublicSuite.scala
```diff
@@ -118,6 +118,12 @@ class PublicSuite extends FunSuite {
     assert(scala.meta.dialects.ParadiseTypelevel212.toString === "ParadiseTypelevel212")
   }
 
+  test("scala.meta.io.AbsolutePath") {}
+
+  test("scala.meta.io.InputStreamIO") {}
```
xeno-by: Can we move the `IO` thingies to `scala.meta.internal.io`?

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/PublicSuite.scala
```diff
@@ -133,10 +139,10 @@ class PublicSuite extends FunSuite {
   test("scala.meta.inputs.Input.File.toString") {
     import java.io._
     import java.nio.charset.Charset
-    val file = new File("hello.scala")
-    val input1 = Input.File(file, Charset.forName("cp1251"))
-    val input2 = Input.File(file, Charset.forName("UTF-8"))
-    assert(input1.toString == """Input.File(new File("hello.scala"), Charset.forName("windows-1251"))""")
+    val path = AbsolutePath("hello.scala")
```
xeno-by: From our initial discussion, I got the impression that `AbsolutePath` is only going to take absolute paths in its constructor. This line challenges this impression. Could you elaborate?

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/PublicSuite.scala
```diff
@@ -133,10 +139,10 @@ class PublicSuite extends FunSuite {
   test("scala.meta.inputs.Input.File.toString") {
     import java.io._
     import java.nio.charset.Charset
-    val file = new File("hello.scala")
-    val input1 = Input.File(file, Charset.forName("cp1251"))
```
xeno-by: Is there anything wrong with this codepage? :)

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/PublicSuite.scala
```diff
@@ -262,10 +268,12 @@ class PublicSuite extends FunSuite {
...
 
   test("scala.meta.semantic.v1.Address.File.toString") {
-    val syntax = "file:source.scala"
+    val file = "source.scala"
+    val syntax = s"file:$file"
+    val fullPath = AbsolutePath(file).absolute
     val address = scala.meta.semantic.v1.Address(syntax)
-    val scala.meta.semantic.v1.Address.File("source.scala") = address
-    assert(address.toString === syntax)
+    val scala.meta.semantic.v1.Address.File(AbsolutePath(path)) = address
```
xeno-by: Is `path` used anywhere?

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/v1/Location.scala
```diff
@@ -22,23 +26,34 @@ case class Location(addr: Address, start: Int, end: Int) {
 }
 
 @root trait Address {
+  def toProto: proto.Address
```
xeno-by: I wonder whether we want to have serialization methods directly in domain objects.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -401,13 +401,13 @@ object TreeSyntax {
...
-        case Lit(value: Byte)    => m(Literal, s("ByteLiterals.", if (value == 0) "Zero" else if (value > 0) "Plus" + value else "Minus" + value))
-        case Lit(value: Short)   => m(Literal, s("ShortLiterals.", if (value == 0) "Zero" else if (value > 0) "Plus" + value else "Minus" + value))
         case Lit(value: Int)     => m(Literal, s(value.toString))
         case Lit(value: Long)    => m(Literal, s(value.toString + "L"))
-        case Lit(value: Float)   => m(Literal, s(value.toString + "f"))
         case Lit(value: Double)  => m(Literal, s(value.toString + "d"))
+        case Lit(value: Float)   => m(Literal, s(value.toString + "f"))
         case Lit(value: Char)    => m(Literal, s(enquote(value.toString, SingleQuotes)))
+        case Lit(value: Byte)    => m(Literal, s("ByteLiterals.", if (value == 0) "Zero" else if (value > 0) "Plus" + value else "Minus" + value))
+        case Lit(value: Short)   => m(Literal, s("ShortLiterals.", if (value == 0) "Zero" else if (value > 0) "Plus" + value else "Minus" + value))
```
xeno-by: See https://github.com/scalameta/scalameta/pull/763

build.sbt
sjrd: If there were, we would have found it by now.

But unless and until sbt recognizes `CrossProject`s as such (or some abstraction like a `CompositeProject` which could be implemented by `CrossProject`), there is nothing we can do about it.

sjrd: You can upvote: https://github.com/sbt/sbt/issues/3042

build.sbt
```diff
@@ -2,379 +2,402 @@ import java.io._
...
-  },
-  commands += Command.command("ci-slow") { state =>
-    "very scalahost/test:runMain scala.meta.tests.scalahost.converters.LotsOfProjects" ::
-    "such testkit/test:runMain scala.meta.testkit.ScalametaParserPropertyTest" ::
-    "scalahostSbt/test" ::
-    state
-  },
-  commands += Command.command("ci-publish") { state =>
-    "very publish" ::
+name := "scalametaRoot"
```
olafurpg: I removed the `lazy val scalametaRoot` project to get automatic aggregation for all defined projects. If I had kept it, we would need to 
```scala
.aggregate(
ioJVM, ioJS,
inputsJVM, inputsJS,
...
)
```
and it would get worse with native crossbuilding.

build.sbt
```diff
@@ -2,379 +2,402 @@ import java.io._
...
+sharedSettings
+noPublish
+unidocSettings
+commands += Command.command("ci-fast") { state =>
+  if (isScalaJS) "scalametaJS/test" :: state
+  else "test" :: "doc" :: state
+}
+commands += Command.command("ci-slow") { state =>
+  if (isScalaJS) state // do nothing
```
olafurpg: This is actually no longer necessary since I explicitly list all entries in the build matrix. Now it doesn't run `ci-slow` when `SCALA_JS=true`

olafurpg: Removed.

build.sbt
```diff
@@ -2,379 +2,402 @@ import java.io._
...
-  trees
-)
+  }
+}
+commands += Command.command("ci-publish") { state =>
+  for {
+    v <- sys.env.get("SCALA_VERSION")
+    t <- sys.env.get("TEST")
+    sjs <- sys.env.get("SCALA_VERSION")
+    if v == "2.11.8" && t == "ci-fast" && sjs == "false"
```
olafurpg: `very publish` publishes for every project in the full matrix. This line is to guard against publishing multiple times.

I've made this clearer by passing in `PUBLISH=true` which is only true for one entry in the build matrix.

build.sbt
```diff
@@ -2,379 +2,402 @@ import java.io._
...
+  } {
+    RunSbtCommand("very publish")
+  }
+  state
+}
+packagedArtifacts := Map.empty
+unidocProjectFilter.in(ScalaUnidoc, unidoc) := inAnyProject
+aggregate.in(test) := false
+test := {
+  // `test` by default only runs on the JVM
```
olafurpg: Absolutely. For example, if you use `java.io.File`.

However, I left JS out because it's slow, see https://github.com/scala-js/scala-js/issues/2827 Sometimes, incremental can take 3-4 minutes to link before running. The cause seems to be that the Scala.js linker struggles with inlining the thousand `test` method calls to Scalatest. I suggest we leave this for the CI server to catch.

build.sbt
```diff
@@ -2,379 +2,402 @@ import java.io._
...
+  val runScalametaTests = test.in(scalametaJVM, Test).value
+  val runScalahostTests = test.in(scalahost, Test).value
+  val runBenchmarkTests = test.in(benchmarks, Test).value
+  val runContribTests = test.in(contrib, Test).value
+  val runDocs = run.in(readme, Compile).toTask(" --validate").value
+}
+console := console.in(scalametaJVM, Compile).in(Compile).value
```
olafurpg: Good catch.

build.sbt
olafurpg: I thought the saved vertical space would be nice, the build file is already ~600 LOC. But fair enough, done.

build.sbt
olafurpg: Hehe, I ran this
```
scalafmt --config-str "{rewrite.rules=[AvoidInfix],rewrite.neverInfix.excludeFilters=[]}" -i -f build.sbt
```

the default AvoidInfix settings permit `in` to stay infix. I guess `cross` can be excluded too.

build.sbt
olafurpg: The official sbt docs also do
```scala
val foo = (project in file("bar").
  settings(...)
```

I'm strongly against infix application with multiple arguments since they can be hard to parse in my head. Regular Term.Apply is easier to read and it scales more nicely for additional arguments/chained calls. For example, the old `) dependsOn (common)` infix usage made it annoying to append another `.addPlugins (ScalaJSPluginb)` line at the end. For uniformity, I refactored all `in` calls to non-infix. Are you OK with keeping this?

build.sbt
olafurpg: Inlined.

build.sbt
olafurpg: The reason I refactored the build so much is because the crossProject changes made some settings evaluate twice or three times on build reload. While moving things around, I got weird compiler errors when around `Def.taskDyn` which I believe are attributed to being inside the `:=` macro. Even if this is only called once I think it's clearer to move it outside of the long `publishableSettings` sequence. 

I can move it back if you want. I honestly think there's way to much logic in this build.sbt which should be refactored into an sbt plugin, but that's for another PR.

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/PublicSuite.scala
sjrd: Technically, it is *complete* in the sense that it supports all [Standard Charsets](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/Charset.html), as defined by the documentation, which are the only one required to be implemented. It doesn't provide any *other* charset but the standard ones, though, so indeed cp1251 is not supported.

scalameta/quasiquotes/shared/src/main/scala/scala/meta/internal/quasiquotes/ReificationMacros.scala
olafurpg: I added a new `Input.VirtualFile` which is an `AbsolutePath` along with its file contents. That solves the linking problem in Scala.js. I imagine this will be useful other contexts such as for example browser extensions to parse files.

olafurpg:
I added an experimental solution for `Lit.Float/Double`in https://github.com/scalameta/scalameta/pull/761/commits/4649537366ced32647ec79214cfe1d618d6c9db4 the tests are still failing, just wanted to "float" the idea around 😅

olafurpg:
I fixed the `Lit.Float/Double.syntax` problem by making `Lit.Float/Double` accept a string. All unit tests pass now on JS and JVM. This PR is ready for another round of review @xeno-by

olafurpg:
I took the liberty to fix the previously broken pretty printing for doubles and floats,

```scala
// before
Lit.Float(Double.NaN).syntax == "NaNd"
// after
Lit.Float(Double.NaN).syntax == "Double.NaN"
```

build.sbt
xeno-by: I don't have a strong opinion. My only serious concern now is familiarity for newcomers.

.drone.yml
```diff
@@ -57,21 +57,19 @@ pipeline:
 
 matrix:
   include:
-    # 2.11.8
     - TEST: ci-fast
       SCALA_VERSION: 2.11.8
       SCALA_JS: false
+      PUBLISH: true
```
xeno-by: Why is PUBLISH set to true only here?

build.sbt
```diff
@@ -28,24 +28,14 @@ commands += Command.command("ci-fast") { state =>
...
-  for {
-    v <- sys.env.get("SCALA_VERSION")
-    t <- sys.env.get("TEST")
-    sjs <- sys.env.get("SCALA_VERSION")
-    if v == "2.11.8" && t == "ci-fast" && sjs == "false"
-  } {
-    RunSbtCommand("very publish")
-  }
-  state
+  if (isPublish) "very publish" :: state
```
xeno-by: I suggest we rename these environment vars and associated methods to explicitly state that they are CI-related. E.g. `CI_PUBLISH` and `isCiPublish`. Otherwise, it's not quite clear whether one should bother with setting them up when trying to publish from a local machine.

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/PublicSuite.scala
```diff
@@ -194,6 +190,10 @@ class PublicSuite extends FunSuite {
     assert(lhs.pos.toString === """[0..3) in Input.String("foo + bar")""")
   }
 
+  test("scala.meta.io.AbsolutePath.toString") {
+    // n/a
```
xeno-by: Why's that? Maybe we can have a prettyprinter for `AbsolutePath`?

build.sbt
xeno-by: Also, do we need those methods? It looks like they are only used once.

scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
```diff
@@ -48,6 +48,12 @@ object Input {
     }
   }
 
+  @data class VirtualFile(path: AbsolutePath, contents: scala.Predef.String) extends Input {
+    override def location = s"<${path.absolute}>"
+    lazy val chars = contents.toArray
+    override def toString = s"""Input.VirtualFile(AbsolutePath("${path.relative}"), "${contents.take(10)}...")""""
```
xeno-by: The `take(10)` part is a bit inconsistent with prettyprinting of `Input.String`. How about we avoid trimming here?

scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
```diff
@@ -7,8 +7,11 @@ import org.scalameta.data._
 import org.scalameta.invariants._
 import scala.meta.common._
 import scala.meta.internal.inputs._
+import scala.meta.internal.io._
+import scala.meta.io._
 
 trait Input extends Optional with Product with Serializable with InternalInput {
+  def location = "<input>"
```
xeno-by: I must've missed this. Why do we need `Input.location`?

scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
```diff
@@ -48,6 +48,12 @@ object Input {
     }
   }
 
+  @data class VirtualFile(path: AbsolutePath, contents: scala.Predef.String) extends Input {
```
xeno-by: How about just `Input.Virtual`?

scalameta/common/shared/src/main/scala/org/scalameta/internal/MacroHelpers.scala
```diff
@@ -115,6 +115,7 @@ trait MacroHelpers extends DebugFinder
     def unapply(tpe: Type): Option[Type] = {
       if (tpe =:= typeOf[String] ||
           tpe =:= typeOf[scala.Symbol] ||
+          tpe =:= typeOf[scala.BigDecimal] ||
```
xeno-by: Hmm. Why's this necessary?

scalameta/scalameta/jvm/src/test/scala/scala/meta/tests/prettyprinters/PlatformSyntacticSuite.scala
```diff
@@ -0,0 +1,17 @@
...
+package prettyprinters
+
+import org.scalatest._
+import scala.meta._
+import scala.meta.internal.ast._
+
+class PlatformSyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
+  test("JVM has proper floats") {
+    // this gives different results work in a JS runtime
+    assert(Lit.Float(1.40f).show[Syntax] == "1.4f") // trailing comma is lost
```
xeno-by: comma or zero?

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -302,8 +302,10 @@ object Lit {
   def unapply(arg: Lit): Option[scala.Any] = Some(arg.value)
   @ast class Null(value: scala.Any) extends Lit
   @ast class Int(value: scala.Int) extends Lit
-  @ast class Double(format: scala.Predef.String) extends Lit { lazy val value = format.toDouble }
-  @ast class Float(format: scala.Predef.String) extends Lit { lazy val value = format.toFloat }
+  @ast class Double(format: scala.Predef.String) extends Lit { val value = format.toDouble }
+  object Double { def apply(double: scala.Double): Double = Lit.Double(double.toString)  }
+  @ast class Float(format: scala.Predef.String) extends Lit { val value = format.toFloat }
+  object Float { def apply(float: scala.Float): Float = Lit.Float(float.toString)  }
```
xeno-by: Does this mean that `x` in `val Lit.Double(x) = ???` will be typed as `String`?

xeno-by: Also, what about locales?

scalahost/src/test/scala/scala/meta/tests/scalahost/converters/ConverterSuite.scala
```diff
@@ -19,7 +19,7 @@ trait ConverterSuite extends FunSuiteLike {
     def fail(msg: String) = sys.error(s"ReflectToMeta initialization failed: $msg")
     val classpath = System.getProperty("sbt.paths.scalahost.test.classes")
     val pluginpath = System.getProperty("sbt.paths.plugin.jar")
-    val options = "-cp " + classpath + " -Xplugin:" + pluginpath + ":" + classpath + " -Xplugin-require:macroparadise"
+    val options = "-usejavacp -cp " + classpath + " -Xplugin:" + pluginpath + ":" + classpath + " -Xplugin-require:macroparadise"
```
xeno-by: Why's this necessary?

xeno-by:
@olafurpg Some of my comments are marked as outdated. No idea why, because I've just submitted them :)

build.sbt
olafurpg: I've inlined the methods and prepended `CI_` to all ci-related environment variables.

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/PublicSuite.scala
olafurpg: I thought about it but left it out because the test would be equivalent to the implementation.

I've added a prettyprinter test.

olafurpg:
I've updated the PR address all points except the custom `Lit.Float/Double.unapply`. 

Github review UI is not suited for larger PRs like this :'(

olafurpg:
It seems the new `Lit` encoding doesn't work yet with macros, I bumped into this error while porting scalafmt to scala.js

```scala
ConfigReader.scala:50: scala.meta.Lit.type does not take parameters
[error]       val x = q"""val x = "string""""
[error]               ^
```

olafurpg:
False alarm, the quasiquote error came from a half-baked local release. It seems to work fine with the new Lit.X encoding :)

I'm still gonna take the liberty to add `Lit.apply(Int/String/Double/...)` methods for backwards source compability. I believe it will be easy to then provide a scalafix rewrite to migrate existing code to the new encoding.

xeno-by:
Let's merge then :) LGTM! 👍

PR #763, opened 2017-03-24
assorted fixes

xeno-by:


scalameta/trees/shared/src/main/scala/scala/meta/internal/ast/LiteralFormatter.scala
```diff
@@ -0,0 +1,8 @@
+package scala.meta.internal.ast
+
+trait LiteralFormatter[T] {
```
olafurpg: Remove?

xeno-by:
@olafurpg Any ideas about what's wrong with this build?

olafurpg:
The CI server had some memory issues recently causing flaky builds. I hit on the same issue in the Scala.js PR, restarting the build fixed the issue.

The platform-ci is new and we are still learning how to scale it up. We probably need to monitor the server resource usage more closely. I'll discuss with @jvican about making sure we avoid more flaky builds.

olafurpg:
Restarting the build does (fortunately) not help. I'm gonna try and lower the `-Xmx` settings https://github.com/scalameta/scalameta/pull/764

olafurpg:
The flaky build failures are unrelated to this change so I'm gonna go ahead and merge this. Let's hope #764 fixes the flakiness.

jvican:
I've done some changes to the server and now it should perform better. We should not experience the flaky errors. Let me know if they persist.

PR #769, opened 2017-03-28
Fix publishing (hopefully last time).

olafurpg:
The previous docker image was on jdk 7 which caused the publish step to
fail for 2.12.

I also updated the readme/publish step to only run if credentials are
available to push to scalameta.github.io.


build.sbt
```diff
@@ -364,9 +364,12 @@ lazy val readme = scalatex.ScalatexReadme(
...
       os.shell.call(s"git config user.name 'Scalameta Bot'", cwd = repo.getAbsolutePath)
       os.shell.call(s"git commit -m $currentUrl", cwd = repo.getAbsolutePath)
-      val httpAuthentication = os.secret.obtain("github").map{ case (username, password) => s"$username:$password@" }.getOrElse("")
-      val authenticatedUrl = s"https://${httpAuthentication}github.com/scalameta/scalameta.github.com"
-      os.shell.call(s"git push $authenticatedUrl master", cwd = repo.getAbsolutePath)
+      os.secret.obtain("github").foreach {
+        case (username, password) =>
+          val httpAuthentication = s"$username:$password@"
+          val authenticatedUrl = s"https://${httpAuthentication}github.com/scalameta/scalameta.github.com"
+          os.shell.call(s"git push $authenticatedUrl master", cwd = repo.getAbsolutePath)
```
olafurpg: The publish task would hang on this line waiting for a username/password combo to be typed into the repl.

PR #773, opened 2017-03-29
Port contrib to Scala.js

### LINKED ISSUE #770: Crossbuild contrib for scalajs
I don't see any real reason why we shouldn't.

### LINKED ISSUE #771: Contrib tests are not run
I believe this affects any module with tests, that has not been ported to a crossbuilt project.

sbt expects tests to be in src/test, but now they are in shared/src/test.

This means "contrib/test" will run 0 tests.
I presume this means CI is not currently running the contrib tests.

olafurpg:
Fixes #771 and fixes #770


scalameta/contrib/shared/src/main/scala/scala/meta/contrib/DocToken.scala
```diff
@@ -41,7 +48,14 @@ object DocToken {
   /**
     * Parser that for obtaining a class reference from an scaladoc body.
     */
-  private val referenceRegex: Regex = "\\[\\[([^]]+)\\]\\]".r
```
olafurpg: Regexes don't work identically across JS/JVM, see https://www.scala-js.org/doc/semantics.html#regular-expressions

olafurpg:
It seems drone pulls a stale git commit on force push :( my WiFi went down
on top of that so now I'm fixing the router haha!

olafurpg:
Ironically, by running unit tests + doc in all modules, ci-fast has become slower than ci-slow

olafurpg:
FYI @xeno-by see b1997f4941b818a7d025fadd4ab6ebc37af733b9

olafurpg:
Sweet, CI is finally green now. Feel like I've been fighting CI for over a week now 😅 At least I'm quite happy with the setup now, using `tests/test` instead of `plz 2.11.8 test` (which run the same amount of tests) cut CI time from 17 minutes to 9 minutes in ci-fast.

PR #775, opened 2017-03-29
Scalahost split

olafurpg:


build.sbt
```diff
@@ -212,33 +212,36 @@ lazy val scalameta = crossProject
...
+    isFullCrossVersion,
+    libraryDependencies += "org.scala-lang" % "scala-compiler" % scalaVersion.value
+  )
+  .dependsOn(scalametaJVM)
+
+lazy val scalahostNsc = project
+  .in(file("scalahost/nsc"))
+  .settings(
+    moduleName := "scalahost-nsc",
+    description := "Scala 2.x compiler plugin that persists the semantic API on compile.",
```
xeno-by: `the semantic DB`.

build.sbt
```diff
@@ -212,33 +212,36 @@ lazy val scalameta = crossProject
...
+  .in(file("scalahost/core"))
   .settings(
+    moduleName := "scalahost",
+    description := "Scala.meta semantic API integration for Scala 2.x (scalac).",
     publishableSettings,
     hasLargeIntegrationTests,
+    isFullCrossVersion,
+    libraryDependencies += "org.scala-lang" % "scala-compiler" % scalaVersion.value
+  )
+  .dependsOn(scalametaJVM)
```
xeno-by: Why do we need this project at all?

build.sbt
```diff
@@ -629,6 +632,18 @@ lazy val noPublish = Seq(
...
+  unmanagedSourceDirectories.in(Compile) += {
+    // NOTE: sbt 0.13.8 provides cross-version support for Scala sources
+    // (http://www.scala-sbt.org/0.13/docs/sbt-0.13-Tech-Previews.html#Cross-version+support+for+Scala+sources).
+    // Unfortunately, it only includes directories like "scala_2.11" or "scala_2.12",
+    // not "scala_2.11.8" or "scala_2.12.1" that we need.
+    // That's why we have to work around here.
+    val base = sourceDirectory.in(Compile).value
+    base / ("scala-" + scalaVersion.value)
+  }
+)
```
xeno-by: Nice!

xeno-by: Btw how about a newline after this lazy val?

scalahost/core/src/main/scala/scala/meta/internal/scalahost/v1/offline/Mirror.scala
```diff
@@ -87,7 +87,7 @@ class Mirror(classpath: String, sourcepath: String)
         sys.error(
           "failed to locate scalahost.jar, specify the location manually via -Dscalahost.jar")
       }
-      autodetectFromClassloader.orElse(autodetectFromProperties).getOrElse(fail)
+      autodetectFromProperties.orElse(autodetectFromClassloader).getOrElse(fail())
```
xeno-by: Why's that?

scalahost/core/src/main/scala/scala/meta/Mirror.scala
```diff
@@ -1,16 +1,53 @@
...
+    * - a compiler plugin
+    * - the presentation compiler
+    */
+  def apply(global: Global): Mirror = new online.Mirror(global)
+
+  /** Construct offline Mirror from a persisted semantic database.
+    *
+    * @param classpath java.io.File.pathSeparator separated list of jar files
+    *                 or directories containing classfiles and `semanticdb` files.
+    *                 The `semanticdb` files are emitted by the scalahost-nsc
```
xeno-by: Is it called `scalahost-nsc` or `scalahost`?

scalahost/core/src/main/scala/scala/meta/Mirror.scala
```diff
@@ -1,16 +1,53 @@
...
+    * @param sourcepath java.io.File.pathSeparator separated list of
+    *                  Scala source files OR directories containing Scala
+    *                  source files. For example
+    *
+    *                  Example:
+    *                      "src/main/scala:build/Bar.scala".
+    * @param scalahostNscPluginPath File path to the scalahost-nsc compiler plugin fatjar,
+    *                               the same path that is passed in `-Xplugin:/scalahost.jar`.
+    */
+  def apply(classpath: String, sourcepath: String, scalahostNscPluginPath: String): Mirror =
```
xeno-by: Can we do something about the third parameter? This is a pretty serious leak of implementation details.

build.sbt
olafurpg: Scalafix will depend on this project to construct an offline mirror. See https://github.com/scalacenter/scalafix/issues/100 for why scalafix can't depend on scalahost-nsc directly.

build.sbt
olafurpg: Done.

build.sbt
olafurpg: Done.

scalahost/core/src/main/scala/scala/meta/Mirror.scala
olafurpg: I don't see a workaround to get rid of the third parameter. Users that depend on the scalahost-nsc fatjar can use the auto-detection method but in the case of scalafix I will depend on the slim jar.

We could make the auto-detection smarter by finding scalahost.jar in the classpath and look for a sibling scalahost-nsc.jar in the same directory. However, we can never guarantee that will always work so I prefer to leave this choice to the build integration/caller.

build.sbt
olafurpg: Also, paradise will depend on this project. At the moment, `addCompilerPlugin(paradise)` enables scalahost-nsc as well.

scalahost/core/src/main/scala/scala/meta/Mirror.scala
olafurpg: `scalahost-nsc` is the compiler plugin that emits semanticdb. `scalahost` is the core library used by `scalahost-nsc` AND used by libraries like scalafix that construct the offline mirror.

xeno-by:
LGTM

PR #781, opened 2017-03-31
migrate to 2.11.9

xeno-by:


DavidDudson:
The build seems to be broken for another reason?

xeno-by:
Oh right. I haven't updated build secrets. I'll contact @olafurpg tomorrow, as he has the most recent version of our secrets.

build.sbt
```diff
@@ -268,7 +268,7 @@ lazy val scalahostSbt = project
     bintrayRepository := "maven", // sbtPlugin overrides this to sbt-plugins
     testQuick := {
       // runs tests for 2.11 only, avoiding the need to publish for 2.12
-      RunSbtCommand("; plz 2.11.8 publishLocal " +
+      RunSbtCommand(s"; plz $ScalaVersion publishLocal " +
```
olafurpg: This line will fail if the main Scala version becomes 2.12, the scripted test depends on a 2.11 Scala version.

xuwei-k:
https://github.com/scala/scala/pull/5821

xeno-by:
@xuwei-k Thank you for reaching out! We're aware of the issue, and we'll upgrade to 2.11.0 immediately after it's released.

Would you like to have a 2.11.8-based prerelease build in the meanwhile?

PR #783, opened 2017-04-04
publish both stable and pre-release builds to bintray

xeno-by:


build.sbt
```diff
@@ -265,7 +255,7 @@ lazy val scalahostSbt = project
     buildInfoSettings,
     sbt.ScriptedPlugin.scriptedSettings,
     sbtPlugin := true,
-    bintrayRepository := "maven", // sbtPlugin overrides this to sbt-plugins
+    bintrayRepository := "sbt", // sbtPlugin overrides this to sbt-plugins
```
olafurpg: I don't think this is necessary, sbt resolves plugins from maven central. I publish sbt-scalafmt to maven with no problems.

build.sbt
```diff
@@ -499,65 +486,37 @@ lazy val mergeSettings = Def.settings(
...
-def shouldPublishToSonatype: Boolean = {
-  if (!os.secret.obtain("sonatype").isDefined) return false
-  if (sys.props("sbt.prohibit.publish") != null) return false
-  !LibraryVersion.contains("-")
-}
-
 lazy val publishableSettings = Def.settings(
   sharedSettings,
   bintrayOrganization := Some("scalameta"),
+  bintrayRepository := (if (LibraryVersion.contains("-")) "pre-release" else "stable"),
```
olafurpg: Not necessary right now, but I think we should consider removing the custom version scheme in favor to sbt-dynver https://github.com/dwijnand/sbt-dynver

The `+` suffix follows semantic versioning rule 10 about attaching metadata to version numbers http://semver.org/spec/v2.0.0.html

build.sbt
xeno-by: Done

xeno-by:
We plan to merge this soon: https://gitter.im/scalameta/scalameta?at=58e3eba28bb56c2d1175ae0f.

PR #789, opened 2017-04-08
Make sure extract uses immutable.Seq

DavidDudson:
Due to a rebase error, the immutable.Seq imports were dropped.

This reintroduces them


xeno-by:
LGTM

PR #792, opened 2017-04-08
migrate to 2.11.10

xeno-by:


xeno-by:
There's no 2.12.2 build published just yet, so I guess let's deal with it next week.

PR #794, opened 2017-04-11
Fix #793

### LINKED ISSUE #793: ScaladocParser swallows comments symbols within code blocks
While playing around with the upcoming Scaladoc parser I found a behaviour that I would classify as a minor bug.

Given I have Scaladoc where within a code block a comment is used like here:
```
/**
  * {{{
  *   // comment
  * }}}
  */
```
Then the ScaladocParser.parseScaladoc should return a Codeblock that includes the slashes (//).

I'm pretty sure the Bug is within the ScaladocParser and its interaction with [CommentOps](https://github.com/scalameta/scalameta/blob/master/scalameta/contrib/shared/src/main/scala/scala/meta/contrib/CommentOps.scala#L21) where the unconditional removal of all leading comment symbols seems to be problematic. But I'm not sure how to fix this.

Affected version: [current master](https://github.com/scalameta/scalameta/tree/93c9871363177d8a4ed5486d9b701d56adde0e57)

Jentsch:
This should fix #893

### DIFF
--- a/scalameta/contrib/shared/src/test/scala/scala/meta/contrib/ScaladocParserSuite.scala
+++ b/scalameta/contrib/shared/src/test/scala/scala/meta/contrib/ScaladocParserSuite.scala
@@ -110,6 +110,8 @@ class ScaladocParserSuite extends FunSuite {
     val complexCodeBlock =
       """
         |ggmqwogmwogmqwomgq
+        |val x = 1 // sdfdfh
+        |// zzz
         |gmqwgoiqmgoqmwomw
       """.stripMargin.trim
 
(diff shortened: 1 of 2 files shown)


scalameta/contrib/shared/src/test/scala/scala/meta/contrib/ScaladocParserSuite.scala
```diff
@@ -110,6 +110,8 @@ class ScaladocParserSuite extends FunSuite {
     val complexCodeBlock =
       """
         |ggmqwogmwogmqwomgq
+        |val x = 1 // sdfdfh
```
xavier-fernandez: This test is for codeblocks, and will not test the issue...

xavier-fernandez: I was wrong, it test it indeed...

xavier-fernandez:
LGTM

xavier-fernandez:
LGTM

PR #797, opened 2017-04-12
s/Lit.Any/Lit.Null/.

olafurpg:
`Lit(null)` previously caused ambiguous overload to String+Symbol,
this new signature makes `Lit(null)` trigger the correct deprecation
warning message.


xeno-by:
LGTM

PR #800, opened 2017-04-14
Add Select.unapply and Name.unapply.

olafurpg:
Motivation, make this answer simpler: http://stackoverflow.com/a/43415867/1469245

- Name.unapply is added for consistency with other tree unapply extractors.
- Select.unapply is put in contrib because otherwise it would conflict
  with existing code that does `import scala.meta._, Type._/Type._`.


xeno-by:
LGTM

PR #803, opened 2017-04-17
Add support for loading the semanticdb from jars.

### LINKED ISSUE #784: Semantic DB is not loaded from inside a jar on the classpath
A `semanticdb` file located inside a jar on the classpath passed to an offline Mirror is not loaded.

stuhood:
Adds support for loading the semanticdb from jars on the classpath.

I haven't added a test for loading them from a jar, and that should probably happen before merging.

Fixes #784.

### DIFF
--- a/scalahost/core/src/main/scala/scala/meta/internal/scalahost/v1/offline/Mirror.scala
+++ b/scalahost/core/src/main/scala/scala/meta/internal/scalahost/v1/offline/Mirror.scala
@@ -4,6 +4,7 @@ package v1
 package offline
 
 import java.io._
+import java.net.URI
 import scala.{Seq => _}
 import scala.collection.immutable.Seq
 import scala.collection.mutable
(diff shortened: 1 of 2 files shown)


xeno-by:
LGTM, though it would indeed be nice to have a test for this.

stuhood:
@xeno-by : Can you recommend the best test suite to put it in? I didn't see any tests loading the db from classpaths, but perhaps I just didn't look hard enough.

xeno-by:
Yeah, this is an undertested area. I'd create a new suite in https://github.com/scalameta/scalameta/tree/v1.7.0/scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/v1.

stuhood:
This one probably doesn't block the talk, so I'm going to let it idle until next week.

stuhood:
I guess this has completely bitrotted at this point! Closing.

xeno-by:
I've already ported this to the code in my megabranch :)

Issue #804, opened 2017-04-18
Defn.hasMod(mod): Boolean


olafurpg:
Currently, the most concise way to test if a defn has a modifier is

```scala
q"lazy val x = 2".mods.exists(_.isEqual(mod"lazy"))
```

it would be nice if it was possible to write that as 

```scala
q"lazy val x = 2".hasMod(mod"lazy")
```
instead.

xavier-fernandez:
Will take a look to this during this week :)

DavidDudson:
it should just be an implicit on https://github.com/scalameta/scalameta/blob/master/scalameta/contrib/shared/src/main/scala/scala/meta/contrib/implicits/ExtractExtensions.scala

This is also an item on https://github.com/scalameta/scalameta/issues/741

```scala
 def hasMod(mod: Mod)(implicit ev: Extract[A, Mod]): Boolean = ev.extract(a).exists(_.isEqual(mod))
```

Should be all that's needed i believe

olafurpg:
@DavidDudson Good catch! I didn't notice the `.has` helper, I like that syntax. Closing as dupe of #741

DavidDudson:
@olafurpg I like the name hasMod though, for clarity

olafurpg:
which do you prefer between `hasMod[Mod.Inline]` and `has[Mod.Inline]`? The duplicate `Mod` strikes out IMO.

DavidDudson:
@olafurpg how about.... `has[Mod.Inline]` and `hasAnnot("Foo")`

I think we may be able to make `has` even more generic (But I am not sold on the idea just yet)....

`has[Defn.Def]` vs `extract[Defn.Def].nonEmpty` might make sense too.

olafurpg:
I'm not sure about `hasAnnot(String)`, since annotations can also have type + term arguments.

> I think we may be able to make has even more generic (But I am not sold on the idea just yet)....

We are hitting on grey territory with `extract.exists`, IMO. I would like to limit vocabulary as possible. My motivation for `hasMod` was to expose a parallel to scalac's internal `.hasFlag`, which is often handy.

I'm leaning again towards this signature,

```scala
 def hasMod(mod: Mod)(implicit ev: Extract[A, Mod]): Boolean = ev.extract(a).exists(_.isEqual(mod))
```
[...]

PR #807, opened 2017-04-20
Implement 'Defn.hasMod(mod): Boolean' (#804)

xavier-fernandez:
Fix: #804


scalameta/contrib/shared/src/main/scala/scala/meta/contrib/implicits/DefnExtensions.scala
```diff
@@ -0,0 +1,18 @@
+package scala.meta.contrib.instances
+
+import scala.meta.Defn
+import scala.meta._
+import scala.meta.contrib.Extract
+import scala.meta.contrib.implicits.Equality._
+
+trait DefnExtensions {
```
DavidDudson: Why have this in a file DefnExtensions? IMO it can just sit in ExtractExtensions. Only Defn's have mods, but we do not actual require this. All we care about is that `Extract[A, Mod]` is defined.

xavier-fernandez: Thank you, will fix now :)

xavier-fernandez: The IDE autocomplete will work in the same way if moving to `ExtractExtensions`?

Issue #777, opened 2017-03-30
Semantic DB is not portable


stuhood:
The semantic db currently stores absolute filenames: 
```
$ cat semanticdb
file:/Users/example-user/Projects/Meta1x/sandbox/Library.scala
[7..14): Library => _empty_.Library.
...
```
... which means that the db isn't portable across machines/directories.

Inferring or allowing for configuration of a "source root" or "sourcepath" at compile time for a particular compile unit would allow for relativizing these paths. One approach that I've seen before is to attempt to strip the package name from the source filename, and use the resulting directory as a root. But that won't be perfect, because scalac doesn't require that a source be located in a relevant package-named directory.

olafurpg:
One alternative approach would be to include the source file contents in semanticdb. We can use `Input.LabeledString` for that https://github.com/scalameta/scalameta/blob/13087aafb7e9db4d8926740fb86578bdd494ad6b/scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala#L50 The schema can accommodate this change https://github.com/scalameta/scalameta/blob/13087aafb7e9db4d8926740fb86578bdd494ad6b/scalameta/semantic/shared/src/main/protobuf/semanticdb.proto#L21-L22

Benefits to this approach
- solves portability
- semanticdb continues to work even if the file contents change on disk

Downsides
- bigger semanticdb file size, could be offset with compression
- ?

We might still want to make the paths relative in the `label` field.

olafurpg:
Related https://github.com/scalameta/scalameta/pull/808#pullrequestreview-33947891 I would be interested to hear your opinion @stuhood on what do with `Anchor` and adding a `contents` field to `AttributedSource`

stuhood:
@olafurpg : I don't really have an opinion on `Anchor` vs `Location` vs `Position`, as they all seem to be shaped approximately the same way, and I don't really know why they might have been split in the first place.

With regard to adding the complete contents of the file to the database: it seems more durable, and might make things easier in the context of build systems that don't have an accurate fingerprint to match a set of compile outputs (and thus `semanticdbs`) to a set of source files.

But I'm not sure it solves the problem of relativizing paths... for example, in the case of scalafix or scalafmt you might want to execute updates "in place", which means you need to be able to construct an absolute path to the output file, and correlate it to the contents of the database. This still feels like it ends up meaning that you need a root at "mirror use time" that matches the root from "compile time".

[...]

Issue #788, opened 2017-04-06
Incorrect dependencies of scalahost-nsc


xeno-by:
```
20:25 ~/Downloads$ cat scalahost-nsc_2.11.9-1.7.0-497-afcdd324.pom
<?xml version='1.0' encoding='UTF-8'?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0">
    <modelVersion>4.0.0</modelVersion>
    <groupId>org.scalameta</groupId>
    <artifactId>scalahost-nsc_2.11.9</artifactId>
    <packaging>jar</packaging>
    <description>Scala 2.x compiler plugin that persists the semantic DB on compile.</description>
    <version>1.7.0-497-afcdd324</version>
    <licenses>
        <license>
            <name>BSD</name>
            <url>https://github.com/scalameta/scalameta/blob/master/LICENSE.md</url>
            <distribution>repo</distribution>
        </license>
    </licenses>
    <name>scalahostNsc</name>
    <organization>
        <name>org.scalameta</name>
    </organization>
    <url>https://github.com/scalameta/scalameta</url>
    <inceptionYear>2014</inceptionYear>
    <scm>
        <url>git://github.com/scalameta/scalameta.git</url>
        <connection>scm:git:git://github.com/scalameta/scalameta.git</connection>
    </scm>
    <issueManagement>
        <system>GitHub</system>
        <url>https://github.com/scalameta/scalameta/issues</url>
    </issueManagement>
    <developers>
        <developer>
            <id>xeno-by</id>
            <name>Eugene Burmako</name>
            <url>http://xeno.by</url>
        </developer>
        <developer>
            <id>densh</id>
            <name>Denys Shabalin</name>
            <url>http://den.sh</url>
        </developer>
        <developer>
            <id>olafurpg</id>
            <name>Ólafur Páll Geirsson</name>
            <url>https://geirsson.com/</url>
        </developer>
    </developers>
    <dependencies>
        <dependency>
            <groupId>org.scala-lang</groupId>
            <artifactId>scala-library</artifactId>
            <version>2.11.9</version>
        </dependency>
        <dependency>
            <groupId>org.scalameta</groupId>
            <artifactId>scalahost_2.11.9</artifactId>
[...]

Issue #623, opened 2017-01-30
Unicorn semantic API


xeno-by:
Prerequisites:
  - [ ] Semantic API (#604)
  - [ ] Deprecate and remove symbols (#610)
  - [ ] Support for mixed unattributed/attributed trees (#621)

Our vision for semantic API, as designed ca. 2013, is significantly more ambitious than the vision that we currently plan to initially ship in #604. This ticket tracks the delta between these approaches.

xeno-by:
As described in #808, we're giving up on the unicorn semantic API. As a result, I'm closing this issue as won't fix.

PR #808, opened 2017-04-20
ScalaDays Chicago hacking spree

### LINKED ISSUE #610: Deprecate and remove symbols
Prerequisites:
  - [x] Online mirror with support for Ref.symbol (#606)
  - [ ] Rudimentary functionality to work with symbols (#609)

Expressing semantic APIs via symbols (#606, #607, #608, #609) has the benefit of engineering simplicity (both scalac and dotc use the notion of symbols internally). 

However, symbol-based semantic APIs also have known issues explored in our scala.reflect experiment:
  * Introduction of an additional view of source code (trees and quasiquotes are no longer enough, users need to understand symbols and symbol APIs)
  * Necessity to manually take prefixes into account (symbol signatures depend on prefixes from which symbols are selected, which means that just symbols are not enough for many popular semantic operations)

As a result, we'd like to ultimately migrate rudimentary symbol-based APIs introduced in #609 to use trees. For example, `Ref.symbol: Symbol` would be replaced with `Ref.defn: Member`, and so on. See section 8.5. "Operations" in my dissertation for more details: https://dl.dropboxusercontent.com/u/10497693/dissertation.pdf.

Tricky points:
  * Migrate all usages of `ref.symbol == otherSymbol` to `ref.defn === otherDefn`. That's because `Symbol.==` and `Tree.==` behave differently (structural vs reference equality).
  * classpath -> scala.meta converter (#433)
  * Denotations are too precise (#146)
  * TASTY -> scala.meta converter (#435)

### LINKED ISSUE #782: Improve autodetection of scalahost-nsc
In principle, we should be able to assemble the scalahost-nsc jar on the fly from scalahost-core and its dependencies.

### LINKED ISSUE #784: Semantic DB is not loaded from inside a jar on the classpath

xeno-by:
This pull request accumulates the results of about a week of intense hacking that brings significant changes to how we do semantic APIs in scala.meta. 

While at the moment of writing I don't have enough time to document the changes thoroughly, I'll provide a brief list of what's been done. If you'd like to learn more, please also take a look at commit messages - some of them are pretty big.

1) After many attempts to make it work, I've decided to give up on the unicorn semantic API. The rationale for that is described in https://github.com/scalameta/scalameta/commit/a4649557b38f101cfc97c3b6976a89912d0ea287 and https://github.com/scalameta/scalameta/commit/99eaf627ed47d950d86f52e804c25976b2c7cd40. This means that if this pull request gets in, we'll close #610, #621 and #623. The associated simplification of the internal infrastructure automatically fixes #782, #788 and #815.

2) Since we're no longer hoping for a future where symbols and trees are unified, we've doubled down on our efforts to improve the symbol API according to #609. This PR introduces `Symbol.flags` (and the accompanying `Symbol.isXXX` family of methods), as well as a very-very rough support for `Symbol.info`.

3) The semanticdb schema and storage format have been changed to better accommodate incremental compilation. The previously monstrous `semanticdb` file has been split into a collection of per-compilation-unit `.semanticdb` files that live in `META-INF/scalameta` according to #666. While implementing that, I've also fixed #784.

To try this pull request out, add the following incantations to your build:
```scala
resolvers += Resolver.bintrayRepo("scalameta", "maven")
libraryDependencies += "org.scalameta" %% "scalameta" % "1.8.0-567-8b1b7f17"
```


scalameta/semantic/shared/src/main/protobuf/semanticdb.proto
```diff
@@ -31,10 +31,9 @@ message Range {
 
 message Message {
   enum Severity {
-    UNKNOWN = 0;
-    INFO = 1;
-    WARNING = 2;
-    ERROR = 3;
+    INFO = 0;
```
olafurpg: The enum id should never change, for example by moving INFO from 1 to 0 then when we read an older semanticdb INFO messages become WARNING. It seems you can't reserve an enum id, but that's OK, you can remove the UNKNOWN field if you like and start counting at 1.

iirc, for protobuf enums it's conventional to have an UNKNOWN value with tag 0. This is because enums are implemented as int32 and 0 is the default value for such fields.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Mirror.scala
```diff
@@ -33,24 +32,17 @@ object Mirror {
     *
     *                  Example:
     *                      "src/main/scala:build/Bar.scala".
-    * @param scalahostNscPluginPath File path to the scalahost-nsc compiler plugin fatjar,
```
olafurpg: This is an amazing simplification. The complexity of this parameter was leaked onto all downstream users. I won't miss `q"Int".symbol`, however cute it was.

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -300,19 +300,6 @@ object Pat {
 }
 object Lit {
   def unapply(arg: Lit): Option[scala.Any] = Some(arg.value)
-  @deprecated("Use Lit.Null instead", "1.7.0")    def apply(x: scala.Null): Lit = Lit.Null(x)
```
olafurpg: Good. 👍 These are only needed for a single release.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Anchor.scala
```diff
@@ -0,0 +1,15 @@
...
+package semantic
+
+import org.scalameta.data._
+import scala.meta.io._
+import scala.meta.prettyprinters._
+
+// NOTE: `start` and `end` are String.substring-style,
+// i.e. `start` is inclusive and `end` is not.
+// Therefore Anchor.end can point to the last character plus one.
+@data class Anchor(path: AbsolutePath, start: Int, end: Int) {
```
olafurpg: How is Anchor different from a `Position.Range`? One issue I have with `Anchor` is that it cannot represent `Input.LabeledString`, which I think would be nice to support in the semanticdb, as I mention in semanticdb.proto.

olafurpg:
Some questions that just came up during lunch:

- Q: what is the difference between `Signature` and `Denotation.info`? It seems that they're closely related, example

```
symbol: _root_.foo.C1#m1(I)I.
denotation: def m1: [T](x: Int)Int
```

My attempt at an answer: a signature is part of a symbol, which is a unique **identifier** of a single definition. Denotation is the shape of that symbol at some phase in the compiler.

cc/ @ShaneDelmore you might want to wait with the pretty-printer for symbol signatures (for example `(I)` => `Int`) since it seems denotations partly address the same problem.

- Q:  can a symbol can have multiple denotations? Should the schema support that?

ShaneDelmore:
@olafurpg thanks for the heads up. The denotations look great 👍

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Denotation.scala
```diff
@@ -0,0 +1,110 @@
+package scala.meta
+package semantic
+
+import org.scalameta.adt._
+import org.scalameta.data._
+
+@data class Denotation(flags: Long, name: String, info: String) extends HasFlags {
```
DarkDimius: I believe that this a very bad misuse of the term. This should be called Definition instead.

DarkDimius: Or, even better, Declaration. This will match naming inside compilers.

xeno-by:
I've just added support for Database.sugars (see 8b1b7f1), which is available for experimentation in `1.8.0-567-8b1b7f17`. Will address the pending comments in the next couple days.

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -147,15 +147,12 @@ package object dialects {
   )
 
   implicit val Scala212 = Scala211.copy(
-    allowLiteralTypes = false, // Scheduled to be included in 2.12.2
-    allowTrailingCommas = false // Scheduled to be included in 2.12.2
```
dwijnand: :+1: I was just about to PR this :)

scalahost/nsc/src/main/scala/scala/meta/internal/databases/AnchorOps.scala
```diff
@@ -0,0 +1,30 @@
...
+  implicit class XtensionGFileAddr(gfile: GFile) {
+    def toAbsolutePath: m.AbsolutePath = gfile match {
+      case gfile: GPlainFile => m.AbsolutePath(gfile.file)
+      case other => sys.error(s"unsupported file " + other)
+    }
+  }
+
+  implicit class XtensionGPositionAnchor(pos: GPosition) {
+    def toAnchor: m.Anchor = {
+      assert(pos.isRange)
```
olafurpg: This line is problematic, I hit on an error (with stacktrace https://gist.github.com/olafurpg/aa8a42c186bce4e51d79e637420b60bf) which I suspect was caused by a for-comprehension. The culprit code looked something like 
```scala
    val patches = for {
      s <- mirror.database.sources
      (Anchor(_, start, _), sugar) <- s.sugars
      if !sugar.startsWith("[")
      patch <- tokensByOffset.get(start).toList.map {
        case tok @ Trivia() =>
          ctx.addLeft(tok, sugar.trimSugar)
        case tok =>
          ctx.addRight(tok, sugar.trimSugar)
      }
    } yield patch
```

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/codecs/package.scala
```diff
@@ -0,0 +1,102 @@
...
+    def toMeta[B](implicit ev: ProtoDecoder[B, A]): B = ev.fromProto(a)
+    def toProto[B](implicit ev: ProtoEncoder[A, B]): B = ev.toProto(a)
+    def toProtoOpt[B](implicit ev: ProtoEncoder[A, B]): Option[B] = Option(toProto[B])
+  }
+
+  implicit val AttributedSourceCodec: ProtoCodec[m.AttributedSource, p.AttributedSource] =
+    new ProtoCodec[m.AttributedSource, p.AttributedSource] {
+      override def toProto(e: m.AttributedSource): p.AttributedSource = e match {
+        case m.AttributedSource(path, names, messages, denotations, sugars) =>
+          implicit val resolvedNameEncoder = new ProtoEncoder[(m.Anchor, m.Symbol), p.ResolvedName] {
```
olafurpg: Can we move all of these nested encoders into the top-level? It's a bit difficult to follow the code for AttributedSourceCodec because there are 4 encoders defined the case body. We should try to keep each encoder/decoder as small as possible.

scalameta/semantic/shared/src/main/protobuf/semanticdb.proto
```diff
@@ -1,25 +1,22 @@
...
-  repeated ResolvedName symbols = 2;
-  repeated CompilerMessage messages = 3;
-}
-
-message Address {
-  string path = 1; // Address.file
-  string contents = 2; // Address.snippet
+message AttributedSource {
+  reserved 1;
+  string path = 5;
```
olafurpg: Following discussions from https://github.com/scalameta/scalameta/issues/777 .

I propose that we name this field `relative_path` and add a comment explaining that an absolute "source root" path is required at mirror construction time to turn `relative_path` into an absolute path.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/codecs/package.scala
```diff
@@ -0,0 +1,102 @@
...
+import scala.meta.internal.semantic.{proto => p}
+
+package object codecs {
+  implicit class XtensionProtoSerializable[A](val a: A) extends AnyVal {
+    def toMeta[B](implicit ev: ProtoDecoder[B, A]): B = ev.fromProto(a)
+    def toProto[B](implicit ev: ProtoEncoder[A, B]): B = ev.toProto(a)
+    def toProtoOpt[B](implicit ev: ProtoEncoder[A, B]): Option[B] = Option(toProto[B])
+  }
+
+  implicit val AttributedSourceCodec: ProtoCodec[m.AttributedSource, p.AttributedSource] =
```
olafurpg: Following discussions in https://github.com/scalameta/scalameta/issues/777 

We will need to have a a `sourceRoot: AbsolutePath` available in scope in order to turn `p.AttributedSource.path/relative_path: String` to construct an AbsolutePath.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  @deprecated("Use names instead", "1.8.0") def symbols: Map[Anchor, Symbol] = names
+  def names: Map[Anchor, Symbol] = sources.foldLeft(Map[Anchor, Symbol]())(_ ++ _.names)
+  def messages: Seq[Message] = sources.flatMap(_.messages)
+  def denotations: Map[Symbol, Denotation] = sources.foldLeft(Map[Symbol, Denotation]())(_ ++ _.denotations)
+  def sugars: Map[Anchor, String] = sources.foldLeft(Map[Anchor, String]())(_ ++ _.sugars)
+
+  override def toString: String = sources.mkString(EOL + EOL)
+  def toBinary: Array[Byte] = this.toProto[p.Database].toByteArray
```
olafurpg: Following discussions in https://github.com/scalameta/scalameta/issues/777 

I believe we need to pass a `sourceRoot: AbsolutePath` here in order to relativize any `AbsolutePath` inside this Database.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/codecs/package.scala
```diff
@@ -0,0 +1,102 @@
...
+          }
+          m.AttributedSource(AbsolutePath(path),
+                             names.map(_.toMeta[(m.Anchor, m.Symbol)]).toMap,
+                             messages.map(_.toMeta[m.Message]).toList,
+                             denotations.map(_.toMeta[(m.Symbol, m.Denotation)]).toMap,
+                             sugars.map(_.toMeta[(m.Anchor, String)]).toMap)
+      }
+    }
+
+  implicit val DatabaseCodec: ProtoCodec[m.Database, p.Database] =
```
olafurpg: I think this needs to become a 

```scala
def DatabaseCodec(sourceRoot: AbsolutePath): ProtoCodec[m.Database, p.Database]
```

olafurpg: It doesn't have to be implicit, we can construct it manually by hand since it's only used in two places.

olafurpg: Alternatively, we could wrap all of the encoders/decoders into a class

```scala
class DatabaseCodecs(sourceRoot: AbsolutePath) {
  implicit val DatabaseCodec: ProtoCodec[m.Database, p.Database]
}
```

This would avoid having to pass around the sourceRoot between codecs.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/mirrors/OfflineMirror.scala
```diff
@@ -0,0 +1,56 @@
...
+import java.io._
+import org.scalameta.data._
+import scala.{Seq => _}
+import scala.collection.immutable.Seq
+import scala.collection.mutable
+import scala.meta.internal.io.PlatformIO
+import scala.meta.parsers._
+import scala.meta.semantic._
+
+@data class OfflineMirror(classpath: String, sourcepath: String) extends CommonMirror {
```
olafurpg: Following discussions in #777 

I propose that this definition becomes
```scala
@data class OfflineMirror(database: Database, sources: Seq[Source], dialect: Dialect = Scala212) extends CommonMirror // no body implementation

object OfflineMirror {
  def apply(classpath: String, sourcepath: String): Try[OfflineMirror]
}
```
This way an OfflineMirror becomes a true `@data` class. 

In sbt-scalafix, I will probably want to construct an OfflineMirror from a single `.semanticdb` file + `sourceRoot: AbsolutePath`. The current `classpath: String, sourcepath: String` primary constructor is not suitable for that use-case.

Note. My current workaround is to do 
```scala
new CommonMirror {
  override dialect ...
  override sources
  override database
}
```

build.sbt
```diff
@@ -209,26 +209,15 @@ lazy val scalameta = crossProject
...
-    hasLargeIntegrationTests,
-    isFullCrossVersion,
-    libraryDependencies += "org.scala-lang" % "scala-compiler" % scalaVersion.value
-  )
-  .dependsOn(scalametaJVM)
-
 lazy val scalahostNsc = project
   .in(file("scalahost/nsc"))
   .settings(
-    moduleName := "scalahost-nsc",
```
xeno-by: Let's see what happens to paradise in the future. It may very well be that we manage to integrate it directly into scalac. So far, both paradise 3.x and scalahost are sufficiently esoteric to not worry about this imo.

xeno-by: Also, even though both paradise and scalahost will bring their own copies of scalameta, that's not going to be a problem, because every compiler plugin has its own classloader.

scalameta/semantic/shared/src/main/protobuf/semanticdb.proto
xeno-by: Protobuf doesn't allow counting from 1 for enums. I'll put back UNKNOWN and fail it during the deserialization step.

xeno-by:
> Regarding c10e9d8
> 
> Do you plan to move the converter to paradise? There are a lot of libraries that use the newstyle macro annotations, I wouldn't be so quick to throw this away. We could move it into a scalahost-converter module for time being, if you prefer.

@olafurpg Yes, I do. Once this PR is merged, I'll submit a follow-up PR to paradise.

scalameta/io/shared/src/main/scala/scala/meta/io/AbsolutePath.scala
```diff
@@ -1,29 +1,27 @@
 package scala.meta.io
 
-import scala.meta.internal.io.PlatformIO._
+import java.io._
+import org.scalameta.data._
+import scala.meta.internal.io.PlatformIO
 
-// sealed abstract to ensure that AbsolutePath.apply constructor is used.
-sealed abstract class AbsolutePath(val absolute: String) extends Serializable {
```
xeno-by: Good point. I will see what we can do in `@data` to make the constructor private.

scalameta/io/shared/src/main/scala/scala/meta/io/AbsolutePath.scala
```diff
@@ -1,29 +1,27 @@
...
-  @deprecated("Use .absolute instead", "1.8.0") // kept for source compabilitity with old Input.File/Address.File
-  def getAbsolutePath: String = absolute
-  def /(other: String): AbsolutePath =
-    new AbsolutePath(absolute + fileSeparator + other.stripPrefix(fileSeparator)) {}
+@data class AbsolutePath(underlying: String) {
+  override def toString: String = underlying
+
+  def toFile: File = new File(underlying)
+
+  @deprecated("Use .toString instead", "1.7.0") // kept for source compabilitity with old Input.File/Address.File
```
xeno-by: Yep, it is removed in a later commit.

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/PublicSuite.scala
```diff
@@ -139,7 +140,7 @@ class PublicSuite extends FunSuite {
   test("scala.meta.inputs.Input.File.toString") {
     import java.io._
     import java.nio.charset.Charset
-    val path = AbsolutePath.fromRelative("hello.scala")
+    val path = RelativePath("hello.scala").absolutize(PlatformIO.workingDirectory)
```
xeno-by: I think it's cleaner to have two separate concepts of `AbsolutePath` and `RelativePath`. After that, we can have converters between them.

xeno-by: What do you dislike in the current scheme of things? Would renaming the methods to `toAbsolute` and `toRelative` work for you? Or is it something else?

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/PublicSuite.scala
```diff
@@ -165,7 +168,7 @@ class PublicSuite extends FunSuite {
 
   test("scala.meta.inputs.Input.LabeledString.toString") {
     val input = Input.LabeledString("foo.scala", "foo")
-    org.scalameta.logger.elem(input.toString)
+    // org.scalameta.logger.elem(input.toString)
```
xeno-by: Allright.

scalameta/semantic/shared/src/main/protobuf/semanticdb.proto
```diff
@@ -3,14 +3,15 @@ syntax = "proto3";
...
+  // A database is treated as a sequence of attributed source files.
+  // This accommodates incremental compilation, because database entries
+  // map one-to-one to compilation units.
+  // This also makes merging databases trivial, as the bytes of two databases
+  // can be simply concatenated to produce a new database.
+  repeated AttributedSource sources = 1;
 }
 
-message DatabaseFile {
+message AttributedSource {
```
xeno-by: Yes! I really appreciated that.

scalameta/semantic/shared/src/main/protobuf/semanticdb.proto
xeno-by: I suggest that we discuss this separately from this pull request. Initially, I planned to handle #777 here, but I quickly found that it's hairy. Later today, I will add my thoughts to that ticket.

scalameta/io/js/src/main/scala/scala/meta/internal/io/PlatformIO.scala
```diff
@@ -4,10 +4,37 @@ import java.nio.charset.Charset
...
+    path.startsWith("/")
+
+  def isRelativePath(path: String): Boolean =
+    !path.startsWith("/")
+
+  def absolutize(parent: AbsolutePath, path: RelativePath): AbsolutePath = {
+    AbsolutePath(parent.toString + fileSeparator + path.toString)
+  }
+
+  def relativize(parent: AbsolutePath, path: AbsolutePath): RelativePath = {
```
xeno-by: Yes, there's quite some duplication between js/jvm code in PlatformIO. You should've seen me when I hacked that functionality at 3am. That's clearly unacceptable, and I'll look into that soon.

scalameta/io/shared/src/main/scala/scala/meta/io/AbsolutePath.scala
```diff
@@ -1,29 +1,24 @@
...
-  def fromRelative(path: String): AbsolutePath = workingDirectory / path
-  def fromAbsoluteOrRelative(path: String): AbsolutePath =
-    apply(path).getOrElse(fromRelative(path))
+  def apply(file: File): AbsolutePath = {
+    AbsolutePath(file.getAbsolutePath)
+  }
+
+  def apply(path: String): AbsolutePath = {
+    if (PlatformIO.isAbsolutePath(path)) new AbsolutePath(path)
+    else sys.error(s"not an absolute path: $path")
```
xeno-by: Later, I will write one megacomment about my thoughts wrt exceptions/monads for error handling.

scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformIO.scala
```diff
@@ -1,15 +1,41 @@
...
+  def relativize(parent: AbsolutePath, path: AbsolutePath): RelativePath = {
+    if (parent == path) return RelativePath(".")
+    var s_parent = parent.toString
+    if (!s_parent.endsWith(fileSeparator)) s_parent += fileSeparator
+    val s_path = path.toString
+    if (s_path.startsWith(s_parent)) RelativePath(s_path.substring(s_parent.length))
+    else throw new IllegalArgumentException(s"$path is not relative to $parent")
+  }
+
   def slurp(path: AbsolutePath, charset: Charset): String =
```
xeno-by: Later, I will write one megacomment about my thoughts wrt exceptions/monads for error handling.

scalameta/scalameta/jvm/src/test/scala/scala/meta/tests/api/SurfaceSuite.scala
```diff
@@ -28,7 +28,9 @@ class SurfaceSuite extends scala.meta.tests.ast.AstSuite {
       s"$fullName$suffix"
     }).mkString(EOL)
     // println(diagnostic)
+    // TODO: lol at com.trueaccord.scalapb.GeneratedEnumCompanion.A *
     assert(diagnostic === """
+      |com.trueaccord.scalapb.GeneratedEnumCompanion.A *
```
xeno-by: Yet another part of my hacking spree. Gonna investigate this later today.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/codecs/package.scala
```diff
@@ -0,0 +1,86 @@
+package scala.meta.internal.semantic
+
+import org.scalameta.unreachable
+import scala.meta.io._
+import scala.meta.{semantic => m}
+import scala.meta.internal.semantic.{proto => p}
+
+package object codecs {
```
xeno-by: Will do.

xeno-by: We'll have to discuss #799 separately and come up with a fix in a different PR. In this pull request, I only plan to rename the package, nothing beyond that.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
```diff
@@ -0,0 +1,181 @@
...
+    if (classpath.isDirectory) {
+      new File(new File(classpath, "META-INF"), "semanticdb").toURI
+    } else if (classpath.getAbsolutePath.endsWith(".jar")) {
+      new URI("jar:" + classpath.toURI.toURL + "!META-INF/semanticdb/")
+    } else {
+      sys.error(s"unsupported classpath entry: $classpath")
+    }
+  }
+
+  def fromBinary(bytes: Array[Byte]): Database = p.Database.parseFrom(bytes).toMeta[Database]
```
xeno-by: Later, I will write one megacomment about my thoughts wrt exceptions/monads for error handling.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
```diff
@@ -0,0 +1,181 @@
...
+  def fromFile(file: File): Database = {
```
xeno-by: Same here.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
```diff
@@ -0,0 +1,181 @@
...
+    Database(databaseEntries.map(AttributedSource.fromURI).toList)
+  }
+  def fromClasspath(classpath: String): Database = {
+    val fragments = classpath.split(File.pathSeparatorChar).toList
+    Database(fragments.map(s => new File(s)).map(fromFile).flatMap(_.sources))
+  }
+}
+
+@data class AttributedSource(
+  path: AbsolutePath,
```
xeno-by: See my reply there as well.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/mirrors/OfflineMirror.scala
```diff
@@ -0,0 +1,56 @@
...
+  private def failEmpty(what: String) =
+    sys.error(
+      s"$what must be non-empty. " +
+        s"This may indicate that Mirror is badly configured. " +
+        s"If you use sbt-scalahost, make sure your project defines " +
+        s"`dependsOn(<projectname> % Scalameta)` for at least one <projectname>.")
+  if (classpath == null || classpath == "") failEmpty("classpath")
```
xeno-by: That's an interesting way of looking at this. I think I agree that involved classes like that shouldn't be `@data`.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Api.scala
```diff
@@ -1,6 +1,64 @@
...
+    def toAnchor: Anchor = pos.input match {
+      case scala.meta.inputs.Input.File(path, _) =>
+        Anchor(path, pos.start.offset, pos.end.offset)
+      case other =>
+        sys.error(s"unsupported input " + other)
+    }
+  }
+
+  implicit class XtensionSemanticEquality(tree1: Tree)(implicit m: Mirror) {
+    def ===(tree2: Tree): Boolean = scala.meta.internal.semantic.equality.equals(tree1, tree2)
```
xeno-by: My only problem with these names is indeed verbosity. I think there's a reason why methods like `<:<` and `=:=` are called like that. Less frequently used methods like semantic/structural/syntactic equality we can probably keep in their long form.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Denotation.scala
xeno-by: To quote `Denotations.scala` from Dotty (https://github.com/lampepfl/dotty/blob/fa13f8d772cc53a8dce5e6854309f00288876d0d/compiler/src/dotty/tools/dotc/core/Denotations.scala#L56-L58):

>  *  Symbol                   A label for a definition or declaration in one compiler run
> *  Denotation              The meaning of a named type or symbol during a period

This corresponds exactly to what we're doing in scalameta. A symbol is just a lightweight wrapper of a string id, and a denotation represents its meaning within a given mirror (just like in Dotty, a symbol can have multiple denotations depending on the current context).

Even the fields of the `Denotation` class in scalameta are inspired by the fields of a similar class in Dotty: `flags`, `name`, `info`. In the future, we plan to add `privateWithin` and most likely `owner`.

All in all, I'm confident that separating `Symbol.id` from `Symbol.info` and friends is a very important insight. In scalac, these concepts are mixed in a fat `Symbol` object, which makes it hard to reliably metaprogram between compiler runs. @densh and I started realizing this in late 2012 when working on hygiene from scala.reflect (back then scalameta didn't exist yet), and I'm happy that in 2017 this is something that's employed in multiple metaprogramming frameworks.

DarkDimius: I still believe that this overloads a term is a very misleading way.

In your case every denotation has a corresponding symbol to start with. The symbol is the original abstraction, it's just that it makes practical sense to serialize\pass unique identifier of the symbol instead of serializing ther entire information about it. I totally agree that separating `Symbol.id` and `Symbol.info` makes sense, similarly how introduction of primary keys in databases makes sense.

In Dotty's case, the Denotation is the original abstraction, there are a lot of denotations that do not correspond to any specific symbols or definitions. This is a lot more than simple separation of a primary key that I see here.

I stand unconvinced and believe that the terms `Declaration` \ `Definiton` represent it better. Otherwise we can have a big long-term problem in terminology between Dotty and scalameta. As compilers and meta-programming are close to each-other I believe there is a huge value to keep terminology consistent.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Message.scala
```diff
@@ -0,0 +1,33 @@
+package scala.meta
+package semantic
+
+import org.scalameta.adt._
+import org.scalameta.data._
+import org.scalameta.unreachable
+
+@data class Message(anchor: Anchor, severity: Severity, message: String) {
```
xeno-by: The problem with `Position.Range` is that it is tied to an `Input`, and that's a bit more than just `AbsolutePath`.

upd. Oh wow, actually `Input` fits perfectly here. It doesn't have to be `Input.File` that simply wraps an absolute path, but can also be `Input.LabeledString` in which case it will embed the contents. While the usefulness of the latter is debatable (see #777), the flexibility of `Position.Range` is clear.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Message.scala
```diff
@@ -0,0 +1,33 @@
...
+  override def toString = syntax
+  def syntax = s"[${severity.toString.toLowerCase}] ${anchor.syntax}: $message"
+  def structure = s"""Message(${anchor.structure}, Severity.$severity, "$message")"""
+}
+
+@root trait Severity {
```
xeno-by: Agreed.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Mirror.scala
xeno-by: \o/

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Anchor.scala
xeno-by: See https://github.com/scalameta/scalameta/pull/808#discussion_r113827526

xeno-by:
> Q: what is the difference between Signature and Denotation.info? It seems that they're closely related.
They seem similar, but they serve different purposes. 

A Signature is a unique id of a symbol (unique as understood by the JVM module infrastructure). Because of how SLS works, method signatures have to include fragments of info, but that's only method signatures.

Denotation.info is full type information of a given symbol. For example, for methods it included non-erased versions of parameter types and return type, as well as parameter names.

scalahost/nsc/src/main/scala/scala/meta/internal/databases/DenotationOps.scala
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    private def name: String = {
+      gsym.decodedName.toString
+    }
+
+    private def info: String = {
+      if (gsym.isClass || gsym.isModule) ""
+      else {
+        val sig = gsym.info.toString
```
xeno-by: Yes, it is definitely possible. But for that we, of course, can't rely on toString :)

scalahost/nsc/src/main/scala/scala/meta/internal/databases/DenotationOps.scala
```diff
@@ -0,0 +1,101 @@
...
+        if (sig.startsWith("=> ")) sig.substring("=> ".length)
```
xeno-by: Indeed. Thanks!

scalameta/io/shared/src/main/scala/scala/meta/io/AbsolutePath.scala
```diff
@@ -1,29 +1,24 @@
...
-  override def equals(obj: scala.Any): Boolean = obj match {
-    case p: AbsolutePath => absolute == p.absolute
-    case _ => false
-  }
-  override def toString: String = s"AbsolutePath($absolute)"
-  @deprecated("Use .absolute instead", "1.8.0") // kept for source compabilitity with old Input.File/Address.File
-  def getAbsolutePath: String = absolute
-  def /(other: String): AbsolutePath =
-    new AbsolutePath(absolute + fileSeparator + other.stripPrefix(fileSeparator)) {}
+@data class AbsolutePath(underlying: String) {
```
xeno-by: Noted. Let's discuss:
  * There's going to be a separate point about exceptions.
  * You're right. `underlying` is an unnecessary implementation detail. Let's hide it altogether. I used this name to be consistent with the newly introduced `RelativePath`.
  * I was also annoyed by how much I needed to type without this helper.
  * Generally I'm not a fan of `/` in similar contexts. Would you be fine with something that does the same, but is named differently?

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/mirrors/OfflineMirror.scala
```diff
@@ -0,0 +1,56 @@
...
+          files.filter(_.isFile).foreach(addFile)
+          files.filter(_.isDirectory).foreach(explore)
+        }
+      } else {
+        addFile(file)
+      }
+    }
+    val fragments = sourcepath.split(PlatformIO.pathSeparator).toList
+    fragments.foreach(fragment => explore(new File(fragment)))
+    scalaFiles.toList.map(_.parse[Source].get)
```
xeno-by: See #814.

scalahost/nsc/src/main/scala/scala/meta/internal/databases/AnchorOps.scala
xeno-by: This is a bug in scalahost. We shouldn't get here with a non-range position.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/codecs/package.scala
xeno-by: I don't think we can, because they require `path` from `m.AttributedSource` to be in scope. I learned this the hard way by implementing them at the top level :)

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/codecs/package.scala
xeno-by: Probably not `AbsolutePath` but `Multipath`, i.e. a sourcepath.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
xeno-by: I suggest that instead we construct Databases in a relativized form already. More details in my upcoming comment to #777.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/mirrors/OfflineMirror.scala
xeno-by: Looks plausible. I don't think that we need `dialect` though, because it can be inferred from `sources`. If it can't (because sources are empty), then we can just pick whatever.

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -147,15 +147,12 @@ package object dialects {
   )
 
   implicit val Scala212 = Scala211.copy(
-    allowLiteralTypes = false, // Scheduled to be included in 2.12.2
```
dwijnand: Literal types didn't ship in 2.12.2. Looks like they're scheduled for 2.12.3

dwijnand: Forgot to link: https://github.com/scala/scala/pull/5310

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Api.scala
olafurpg: I think the difference between semantic and syntactic equality is abstract enough for most people. I'm not convinced symbolic operators like `<:<` help. The symbolic operators should be opt-in in my opinion through some `scala.meta.syntax._` import. One  benefit of alphanumerical names is that they are easier to search for on google and discuss about on gitter/SO.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Message.scala
olafurpg: The fact that it's tied to an `Input` is exactly why I think it's a good idea. For example, `Anchor` does not include the file encoding like `Input.File`, which could be problematic in some cases.

`Input.LabeledString` would be a perfect fit for a `metadoc` tool I want to write, which generates a static site from a Mirror and uses Scala.js for interactive code browsing and symbol search.

scalameta/io/shared/src/main/scala/scala/meta/io/AbsolutePath.scala
olafurpg: I'm fine with a non-symbolic method name like `resolve` instead of `/`, that would be consistent with `nio.Path` and if someone really wants `/` then it's possible to with extensions (maybe via contrib).

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
olafurpg: This is unrelated to this PR, we can merge this change immediately if you open a separate PR @dwijnand :)

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
```diff
@@ -0,0 +1,187 @@
...
+      val duplicates = occs.filter(_._2 > 1).map(_._1)
+      sys.error("duplicate paths for " + occs.mkString(", "))
+    }
+  }
+
+  @deprecated("Use names instead", "1.8.0") def symbols: Map[Anchor, Symbol] = names
+  def names: Map[Anchor, Symbol] = sources.foldLeft(Map[Anchor, Symbol]())(_ ++ _.names)
+  def messages: Seq[Message] = sources.flatMap(_.messages)
+  def denotations: Map[Symbol, Denotation] = sources.foldLeft(Map[Symbol, Denotation]())(_ ++ _.denotations)
+  def sugars: Map[Anchor, String] = sources.foldLeft(Map[Anchor, String]())(_ ++ _.sugars)
```
olafurpg: Do we keep sugars as `String` in this PR or do want to add a richer structure to them? I'm fine with leaving it to a separate PR.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
```diff
@@ -0,0 +1,181 @@
...
+    file.getParentFile.mkdirs()
+    val fos = new FileOutputStream(file)
+    try fos.write(this.toProto[p.AttributedSource].toByteArray)
+    finally { fos.close() }
+  }
+
+  private[meta] def locateInDatabase(databaseRoot: URI): URI = {
+    var result = databaseRoot.toString
+    if (!result.endsWith("/")) result += "/"
+    // TODO: This relativization is unsound and should be removed.
```
olafurpg: whoops, I misunderstood what's happening here, nevermind my comment!

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/codecs/package.scala
olafurpg: Let's put this discussion on hold. I'm happy to open a followup PR showing what reorganization I'd like to make.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Api.scala
xeno-by: I'll just remove this from the API for the time being.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/codecs/package.scala
xeno-by: Allright. Thanks a lot for the input here and in general in the review!

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
xeno-by: Or alternatively I get this PR in a bit later today (in PST, meaning late night CEST) as we just discussed with @olafurpg.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/codecs/package.scala
xeno-by: Agreed and agreed. As discussed via hangouts, I'll soon submit a follow-up commit that fixes #777 and features some additional good stuff.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
xeno-by: That will be possible.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Message.scala
xeno-by: Agreed.

scalameta/io/shared/src/main/scala/scala/meta/io/AbsolutePath.scala
xeno-by: Allright.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
xeno-by: I'll open a ticket for that. Adding a richer structure to sugars and infos will take a while, probably on a scale of days, not hours.

xeno-by:
@olafurpg I've started a discussion about error handling in https://github.com/scalameta/scalameta/issues/821.

xeno-by:
@olafurpg See #822 wrt your question on semantic information for infos and sugars.

xeno-by:
As discussed with @olafurpg today, I'm merging this PR to unblock his flatMap presentation. I've addressed almost all feedback, and the remaining items aren't so urgent to warrant further delays.

xeno-by:
I thought that today I'll be able to finish the migration of converters from meta to paradise. However, I'm too exhausted now, so I'll have to leave that for tomorrow.

This is suboptimal (I wouldn't have merged this PR if I realized that the migration has to be held off), but not tragic (because paradise depends on a concrete version of meta, not on meta's HEAD).

xeno-by:
@olafurpg Scalameta `1.8.0-585-e8245aa7` with the changes from #808 has just been published for both 2.11.11 and 2.12.2.

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -147,13 +147,14 @@ package object dialects {
...
-    allowTrailingCommas = false // Scheduled to be included in 2.12.2
+    // NOTE: support for literal types is tentatively scheduled for 2.12.3
+    // https://github.com/scala/scala/pull/5310#issuecomment-290617202
+    allowLiteralTypes = false,
```
dwijnand: Thanks, @xeno-by.

PR #818, opened 2017-04-29
Add support for file IO inside nodejs.

olafurpg:
This commits implements the full PlatformIO api on nodejs. This opens
possibilities to

- build command-line tools with scalameta on JS
- run file IO in tests, which will come in handy in the scalafmt test
  suite where almost all unit tests are written in external files
  using an external DSL.

This implementation uses
- the native `fs` module to read files
- the native `path` module to get file/path separators and check for
  absolute paths.
- uses npm package `shelljs` module to get cross-platform workingDirectory.
  It's a shame this is not provided by default in nodejs, but I'm not
  motivated nor competent enough to roll my own native implementation
  using the `process` module.


xeno-by:
@olafurpg Is this a spurious failure?

olafurpg:
It's caused by a misconfiguration in the build. I'll try to take a look tomorrow or Monday. Don't let this block thr mega pr.

olafurpg:
Closing for now, will reopen when I have time again.

PR #820, opened 2017-04-30
Contrib replace

DavidDudson:
I'll tackle update later.

@ShaneDelmore you may also be interested.

### DIFF
--- a/scalameta/contrib/shared/src/main/scala/scala/meta/contrib/implicits/ReplaceExtensions.scala
+++ b/scalameta/contrib/shared/src/main/scala/scala/meta/contrib/implicits/ReplaceExtensions.scala
@@ -1,0 +1,14 @@
+package scala.meta.contrib.implicits
+
+import scala.collection.immutable.Seq
+import scala.meta._
+import scala.meta.contrib._
+
+trait ReplaceExtensions {
+  implicit class ExtensionReplacers[A](a: A) {
+    def withStats(bs: Seq[Stat])(implicit ev: Replace[A, Stat]): A = ev.replace(a, bs)
+    def withMods(bs: Seq[Mod])(implicit ev: Replace[A, Mod]): A = ev.replace(a, bs)
+  }
+}
+
+object ReplaceExtensions extends ReplaceExtensions
--- a/scalameta/contrib/shared/src/main/scala/scala/meta/contrib/package.scala
+++ b/scalameta/contrib/shared/src/main/scala/scala/meta/contrib/package.scala
@@ -1,21 +1,10 @@
 package scala.meta
 
-import scala.meta.contrib.implicits._
-import scala.meta.contrib.instances._
+import scala.meta.contrib.implicits.implicits
+import scala.meta.contrib.instances.instances
 
 package object contrib
-  extends TreeExtensions
+  extends implicits
+  with instances
   with TreeExtractors
-  with SetExtensions
-  with CommentExtensions
-  with Equality
-  with Converters
-  with ExtractStatInstances
-  with ExtractStatSubtypeInstances
-  with ExtractModsInstances
-  with ExtractAnnotationInstances
-  with ExtractExtensions
-  with ReplaceExtensions
-  with ReplaceModsInstances
-  with ReplaceStatInstances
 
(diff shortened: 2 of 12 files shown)


DavidDudson:
As a side note. this should (along with update) will actually allow you to not require the type cast in macro annotations (As `A` can be anything)...

The below should work for all Defn's (provided an implicit can be found)
```scala
inline def apply(a: Any): Any = meta {
  a.replace[Stat](q"def foo = 2" :: Nil)
}
```

DavidDudson:
@olafurpg @xeno-by 
I take it the drone failure is unrelated looking at other PR's?

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/package.scala
```diff
@@ -4,15 +4,6 @@ import scala.meta.contrib.implicits._
...
-  with SetExtensions
-  with CommentExtensions
-  with Equality
-  with Converters
-  with ExtractStatInstances
-  with ExtractStatSubtypeInstances
-  with ExtractModsInstances
-  with ExtractAnnotationInstances
-  with ExtractExtensions
+  extends implicits
```
olafurpg: Nice 👍

olafurpg:
Great work btw! I am excited to use this  😄

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/implicits/ReplaceExtensions.scala
```diff
@@ -0,0 +1,12 @@
+package scala.meta.contrib.implicits
+
+import scala.collection.immutable.Seq
+import scala.meta.contrib._
+
+trait ReplaceExtensions {
+  implicit class ExtensionReplacers[A](a: A) {
+    def replace[B](bs: Seq[B])(implicit ev: Replace[A, B]): A = ev.replace(a, bs)
```
ShaneDelmore: I'm ok with replace but names are important so I'll throw out a few more. Here are my thoughts on various words.

Set: () => B. I don't care what (if anything) comes in, give me a B.  
Replace:  _: A => B. Take A and Replace with B without inspecting the value A.  B is decided upon without referring to the specific A, but I am only going to put a B in if an A was there before.
Map/Transform: A => B: Take A and based on it's contents, create a B.

DavidDudson: @ShaneDelmore @olafurpg I mostly agree with you both.

Heres my take. (Probably because I write 90% Java code at $JOB)

x.setY = mutate the Y property of X
x.withY = return a new X with the property Y changed
x.replace = replace without caring about current value (as @ShaneDelmore said)
x.map = apply a function A => B. (For each element A, give me a B)
x.transform = map for tree's (Possibly????)
x.update = same as map.

In this PR, I will add `withStats` and `withMods` as @olafurpg recommended. If you disagree with the term `with`, I'm open for discussion.

scalameta/contrib/shared/src/test/scala/scala/meta/contrib/StatReplacementTest.scala
```diff
@@ -0,0 +1,66 @@
...
+    val newClass = q"class Foo".replace[Stat](methodDef :: Nil)
+    assert(newClass.extract[Stat].head isEqual methodDef)
+  }
+
+  // This is the weird behaviour we cannot avoid,
+  // unless we edge case this.
+  test("Test replace empty class stats with Nil") {
+    val newClass = q"class Foo".replace[Stat](Nil)
+    assert(newClass.extract[Stat] == Nil)
+    assert(newClass isEqual q"class Foo {}")
```
DavidDudson: I changed the comment slightly

DavidDudson:
@olafurpg Ah the failing tests were from me forgetting a couple of with calls during the `implicit` trait creation.

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/implicits/ReplaceExtensions.scala
```diff
@@ -0,0 +1,14 @@
+package scala.meta.contrib.implicits
+
+import scala.collection.immutable.Seq
+import scala.meta._
+import scala.meta.contrib._
+
+trait ReplaceExtensions {
+  implicit class ExtensionReplacers[A](a: A) {
+    def withStats(bs: Seq[Stat])(implicit ev: Replace[A, Stat]): A = ev.replace(a, bs)
```
olafurpg: withStats and withMods is nice 👍

PR #826, opened 2017-05-01
Don't synthesize a block in a case expression if there is only one ex…

oyvindberg:
The motivating thing we wanted to fix was this:
```scala
foo match{ case _ => _ => false}
```
Which would come back as 
```scala
foo match{ case _ => _ => {false}}
```
After a roundtrip through parser/pretty printer

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -356,6 +356,12 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
     if (token.is[LeftBrace]) inBraces(body)
     else body
 
+  def dropTrivialBlock(term: Term): Term =
+    term match {
+      case Term.Block((stat: Term) :: Nil) => stat
+      case _ => term
+    }
+
   @inline final def inBrackets[T](body: => T): T = {
     accept[LeftBracket]
     val ret = body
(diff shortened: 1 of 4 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1759,7 +1759,15 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
         }
         if (looksLikeLambda) {
           next()
-          t = atPos(t, auto)(Term.Function(convertToParams(t), if (location != InBlock) expr() else block()))
+          t = atPos(t, auto)({
+
+            val term = (if (location != InBlock) expr() else block()) match {
+              // Don't synthesize a block in a case expression if there is only one expression
```
olafurpg: This is a tricky one to explain. 

> NOTE: This case unwraps a single expression where we expect a BlockStatSeq. Without this case, `false` in `case _ => _ => false` would get wrapped in a Term.Block and therefore get pretty-printed as `case _ => _ => {false}`.

olafurpg: cc/ @xeno-by I'm not really sure if this is the right solution, but it seems to do the right thing. I'm sure you know this better than me.

xeno-by: This looks legit. Thank you, @oyvindberg!

The only suggestion would be to reuse the `dropTrivialBlock` method that we already have elsewhere in the parser.

oyvindberg: Great! Pushed an updated commit now

olafurpg:
🎉

PR #828, opened 2017-05-01
refactor semanticdb

xeno-by:
This sets the stage for #777 without actually doing anything to implement that functionality.


scalameta/io/shared/src/main/scala/scala/meta/io/AbsolutePath.scala
```diff
@@ -1,12 +1,18 @@
...
 import java.io._
+import java.net._
 import org.scalameta.data._
 import scala.meta.internal.io.{FileIO, PathIO}
 
 @data class AbsolutePath private (value: String) {
-  override def toString: String = value
+  def syntax: String = value
+  def structure: String = s"""AbsolutePath("$value")"""
+  override def toString: String = syntax
```
xeno-by: I think that all our data structures should have pairs of `syntax` and `structure` methods.

In this pull request, this is partially implemented in a really adhoc manner. In the future, I suggest that we enforce this with `@data` and `@leaf` annotations. Moreover, we can use these annotations to easily derive `def structure: String`. That'll save quite some boilerplate.

scalameta/io/shared/src/main/scala/scala/meta/io/Multipath.scala
```diff
@@ -0,0 +1,82 @@
...
+}
+
+@leaf class Sourcepath(value: String) extends Multipath {
+  def syntax: String = value
+  def structure: String = s"""Sourcepath("$value")"""
+  override def toString: String = syntax
+}
+object Sourcepath {
+  implicit def stringToSourcepath(value: String): Sourcepath = Sourcepath(value)
+}
```
xeno-by: I'm not a fan of implicit conversions here, but I've no idea how to express this better without having to introduce `SourcepathLike` and the like.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Schema.scala
```diff
@@ -0,0 +1,97 @@
...
+        val scalaNameOpt = sourcepath.relativize(AbsolutePath(entry.path).toURI)
+        val scalaName = scalaNameOpt.getOrElse(sys.error(s"can't find ${entry.path} in $sourcepath"))
+        val semanticdbName = v.Paths.scalaToSemanticdb(scalaName)
+        val fragment = Fragment(base, semanticdbName)
+        v.Entry.InMemory(fragment, entry.toByteArray)
+      })
+      v.Database(ventries)
+    }
+
+    def toMeta: m.Database = {
```
xeno-by: I've rewritten serialization/deserialization logic from typeclasses to plain old Scala code, because I found that type signatures of typeclass instances really got in the way.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Vfs.scala
```diff
@@ -0,0 +1,82 @@
...
+import scala.meta.internal.io.PathIO.fileSeparator
+import scala.meta.internal.semantic.{vfs => v}
+import scala.meta.internal.semantic.{schema => s}
+
+package object vfs {
+  object Database {
+    def load(classpath: Classpath): v.Database = {
+      val fentries = classpath.deep.filter(e => v.Paths.isSemanticdb(e.name))
+      val ventries = fentries.map(fentry => v.Entry.OnDisk(fentry))
+      v.Database(ventries)
```
xeno-by: Oh boy, you can't even imagine how ugly and entangled this logic was before.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Vfs.scala
```diff
@@ -0,0 +1,82 @@
...
+
+    def isScala(path: RelativePath): Boolean = {
+      path.toString.startsWith(scalaPrefix) && path.toString.endsWith(scalaSuffix)
+    }
+
+    def scalaToSemanticdb(path: RelativePath): RelativePath = {
+      require(isScala(path))
+      RelativePath(semanticDbPrefix + path.toString.stripPrefix(scalaPrefix).stripSuffix(scalaSuffix) + semanticDbSuffix)
+    }
+  }
```
xeno-by: This part needs work, but I haven't found a way to make it look better.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Schema.scala
```diff
@@ -0,0 +1,97 @@
...
+import scala.{Seq => _}
+import scala.collection.immutable.Seq
+import scala.meta.io._
+import scala.meta.internal.semantic.{vfs => v}
+import scala.meta.internal.semantic.{schema => s}
+import scala.meta.{semantic => m}
+import scala.meta.inputs.{Input => mInput, Position => mPosition, Point => mPoint}
+
+package object schema {
+  @data class Database(entries: Seq[Attributes]) {
```
xeno-by: For #777, this should read `entries: Seq[(RelativePath, Attributes)]`. Relative paths are already encoded in `v.Entry.name`, so we don't need to store them in Attributes at all.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Schema.scala
```diff
@@ -0,0 +1,97 @@
...
+    def toVfs(classpath: Classpath, sourcepath: Sourcepath): v.Database = {
+      if (classpath.shallow.isEmpty) sys.error("can't save semanticdb to an empty classpath")
+      val ventries = entries.map(entry => {
+        // TODO: Would it make sense to support multiclasspaths?
+        // One use-case for this would be in-place updates of semanticdb files.
+        val base = AbsolutePath(classpath.shallow.head)
+        val scalaNameOpt = sourcepath.relativize(AbsolutePath(entry.path).toURI)
```
xeno-by: For #777, we will no longer need `sourcepath` in this method, because `entries` will already include a relative path for `sentry`.

scalameta/semantic/shared/src/main/protobuf/semanticdb.proto
```diff
@@ -1,19 +1,11 @@
...
-  // This also makes merging databases trivial, as the bytes of two databases
-  // can be simply concatenated to produce a new database.
-  repeated AttributedSource sources = 1;
-}
+package scala.meta.internal.semantic.schema;
 
-message AttributedSource {
+message Attributes {
   reserved 1;
   string path = 5;
```
xeno-by: For #777, this field will be removed. Relative paths are already encoded in `v.Entry.name`, so we don't need to duplicate them in `v.bytes` (i.e. `s.Attributes`).

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Schema.scala
xeno-by: For #777, we'll need a `sourcepath` parameter here. We'll no longer have the luxury of the absolute `s.Attributes.path`, just a relative path accompanying attributes, so we'll need something to absolutize.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
```diff
@@ -1,186 +1,39 @@
...
-  }
-
-  def fromFile(file: File): AttributedSource = {
-    p.AttributedSource.parseFrom(new FileInputStream(file)).toMeta[AttributedSource]
-  }
-
-  def fromURI(uri: URI): AttributedSource = {
-    p.AttributedSource.parseFrom(uri.toURL.openStream).toMeta[AttributedSource]
+object Database {
+  def load(classpath: Classpath): Database = {
```
xeno-by: For #777, this method will need to take a `sourcepath`, because `toSchema` will now take a sourcepath.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Mirror.scala
```diff
@@ -1,48 +1,35 @@
...
-    *                  source files. For example
-    *
-    *                  Example:
-    *                      "src/main/scala:build/Bar.scala".
-    */
-  def apply(classpath: String, sourcepath: String): Mirror =
-    new OfflineMirror(classpath, sourcepath)
+  def apply(classpath: Classpath, sourcepath: Sourcepath): Mirror = {
+    // TODO: sourcepath is unused (that's intentional until #777 is fixed)
+    Database.load(classpath)
```
xeno-by: For #777, we'll need to pass `sourcepath` to the subsequent call chain.

scalameta/io/shared/src/main/scala/scala/meta/io/Fragment.scala
```diff
@@ -0,0 +1,16 @@
+package scala.meta
+package io
+
+import java.net._
+import java.io._
+import org.scalameta.data._
+
+@data class Fragment(base: AbsolutePath, name: RelativePath) {
+  def uri: URI = {
+    val suri = if (base.toString.endsWith(".jar")) s"jar:file:$base!/$name" else s"file:$base/$name"
```
olafurpg: Does `/` work on Windows? Should it be `File.separator`?

scalameta/io/shared/src/main/scala/scala/meta/io/Multipath.scala
olafurpg: I'm OK with removing implicit conversion in favor of explicit `Sourcepath(...)` calls.

olafurpg: I don't like having an implicit conversions from String to both classpath and sourcepath, I'd prefer to be explicit here.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Vfs.scala
olafurpg: Oh I think I know :)

scalameta/io/shared/src/main/scala/scala/meta/io/Fragment.scala
xeno-by: Yes, it does. However, this code contained other problems that should be fixed now.

scalameta/io/shared/src/main/scala/scala/meta/io/Multipath.scala
xeno-by: Done.

xeno-by:
Superseded by https://github.com/scalameta/scalameta/pull/841

PR #832, opened 2017-05-03
Parses empty unquote Xml expression

allanrenucci:
Parses "unquote Xml expression" similarly to "unquote expression".
This fixes a bug with empty expression (e.g. <a>{}</a>).

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1591,7 +1591,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
 
   def unquoteXmlExpr(): Term = {
     // TODO: verify this
-    dropAnyBraces(expr())
+    unquoteExpr()
   }
 
   // TODO: when parsing `(2 + 3)`, do we want the ApplyInfix's position to include parentheses?
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1590,7 +1590,6 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
   }
 
   def unquoteXmlExpr(): Term = {
-    // TODO: verify this
     unquoteExpr()
   }
 
@@ -2342,7 +2341,6 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
     }
 
     def unquoteXmlPattern(): Pat = {
-      // TODO: verify this
       dropAnyBraces(pattern().require[Pat])
     }
 
--- a/scalameta/scalameta/shared/src/test/scala/scala/meta/tests/parsers/XmlSuite.scala
+++ b/scalameta/scalameta/shared/src/test/scala/scala/meta/tests/parsers/XmlSuite.scala
@@ -288,7 +288,7 @@ class XmlSuite extends ParseSuite {
     """.trim.stripMargin)
 
 
-  //checkOK("<a>{}</a>") // FIXME: Should parse
+  checkOK("<a>{}</a>")
   //checkError("<a></b>") // FIXME: Should not parse
 
   // Matches bugs in scalac
--- a/scalameta/scalameta/shared/src/test/scala/scala/meta/tests/parsers/XmlSuite.scala
+++ b/scalameta/scalameta/shared/src/test/scala/scala/meta/tests/parsers/XmlSuite.scala
@@ -289,6 +289,7 @@ class XmlSuite extends ParseSuite {
 
 
   checkOK("<a>{}</a>")
+  checkError("e match { case <a>{}</a> => ??? }")
   //checkError("<a></b>") // FIXME: Should not parse
 
   // Matches bugs in scalac
--- a/scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
+++ b/scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
@@ -528,6 +528,18 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
     assert(tree.show[Syntax] === "<foo>{bar}</foo>")
   }
 
+  test("xml literals unit") {
+    val tree = term("<foo>{}</foo>")
+    assert(tree.show[Structure] == """Term.Xml(Seq(Lit.String("<foo>"), Lit.String("</foo>")), Seq(Term.Block(Nil)))""")
+    assert(tree.show[Syntax] == "<foo>{{}}</foo>")
+  }
+
+  test("interpolator unit") {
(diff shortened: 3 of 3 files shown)


scalameta/scalameta/shared/src/test/scala/scala/meta/tests/parsers/XmlSuite.scala
```diff
@@ -288,7 +288,7 @@ class XmlSuite extends ParseSuite {
     """.trim.stripMargin)
 
 
-  //checkOK("<a>{}</a>") // FIXME: Should parse
+  checkOK("<a>{}</a>")
```
xeno-by: What does this code mean?

allanrenucci: scalac insert `()`: `<a>{(): Unit}</a>`. Same for string interpolation `s"${}"`

xeno-by: Wow this is crazy 😮

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1591,7 +1591,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
 
   def unquoteXmlExpr(): Term = {
     // TODO: verify this
-    dropAnyBraces(expr())
+    unquoteExpr()
   }
```
xeno-by: What about other usages of `dropAnyBraces`, like in `unquoteXmlPattern`?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1591,7 +1591,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
 
   def unquoteXmlExpr(): Term = {
     // TODO: verify this
```
xeno-by: I guess this comment can be deleted now.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
allanrenucci: I haven't had a look to xml in pattern position. Seems to be very similar to interpolator in pattern position. Will need to investigate more.

allanrenucci:
@xeno-by @olafurpg How should `s"${}"` be parsed?
scalac doesn't differentiate between `s"${}"` and`s"${()}"` but scalameta does:

scalareflect:
```scala
> showRaw(q""" s"$${}" """) 
Apply(Select(Apply(Ident(TermName("StringContext")), List(Literal(Constant("")), Literal(Constant("")))), TermName("s")), List(Literal(Constant(()))))
> showRaw(q""" s"$${()}" """) 
Apply(Select(Apply(Ident(TermName("StringContext")), List(Literal(Constant("")), Literal(Constant("")))), TermName("s")), List(Literal(Constant(()))))
```
scalameta:
```scala
> s"""s"Hello $${}"""".parse[Stat].get.structure
Term.Interpolate(Term.Name("s"), Seq(Lit.String(""), Lit.String("")), Seq(Term.Block(Nil)))
> s"""s"Hello $${()}"""".parse[Stat].get.structure
Term.Interpolate(Term.Name("s"), Seq(Lit.String(""), Lit.String("")), Seq(Lit.Unit(())))
```

I think we should parse xml literals the same way we parse string interpolators (modulo mandatory `{}` for xml literals). This is what scalac does.

olafurpg:
I propose that we do the same thing in xml literals as we do in string interpolators in scalameta, so the parser should emit `Term.Block(Nil)`. Either that, or we insert `Lit.Unit(())` in both to be the same as scalac. I agree that xml and regular interpolators should behave the same in this case. However, the scalac behavior of inserting a `Lit.Unit(())` is just baffling to me.

allanrenucci:
@olafurpg This is what happen in the scalac parser:
```scala
def block(): Tree = makeBlock(blockStatSeq())

def makeBlock(stats: List[Tree]): Tree = gen.mkBlock(stats)

def mkBlock(stats: List[Tree], doFlatten: Boolean = true): Tree =
  if (stats.isEmpty) mkSyntheticUnit()
  else if (!stats.last.isTerm) Block(stats, mkSyntheticUnit())
  else if (stats.length == 1 && doFlatten) stats.head
  else Block(stats.init, stats.last)
```
So when it parses a an empty block, it inserts a synthetic unit.

olafurpg:
I'm OK with reproducing what the scalac parser does, wdyt @xeno-by ?

xeno-by:
I think we should do `Term.Block(Nil)`. Scalac does a lot of desugarings, but we're not obliged to, and to me `{}` looks like an empty block, hence `Term.Block(Nil)`. Wdyt?

olafurpg:
OK, `Term.Block(Nil)` it is then!

allanrenucci:
Then I guess it is good to go

allanrenucci:
>t's interesting to note that the Syntax is {{}}, since the difference between {{}} and {} is now considered token trivia.

@olafurpg I have:
```scala
tree.show[Syntax] == "<foo>{}</foo>"
```

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/parsers/XmlSuite.scala
```diff
@@ -198,6 +198,12 @@ class XmlSuite extends ParseSuite {
     // format: on
   }
 
+  test("xml literals unit") {
+    val tree = term("<foo>{}</foo>")
+    assert(tree.show[Structure] == """Term.Xml(Seq(Lit.String("<foo>"), Lit.String("</foo>")), Seq(Term.Block(Nil)))""")
+    assert(tree.show[Syntax] == "<foo>{}</foo>")
```
olafurpg: We need to remove the original tokens as is done here https://github.com/scalameta/scalameta/blob/2128ff7b233aafb797c8c1b9bc747a67e10efab4/scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala#L22 I propose we move this test case to SyntacticSuite right under the `xml literals` test.

allanrenucci: Done

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
xeno-by: Did you have a chance to look into this further?

allanrenucci: If I remove `dropAnyBraces` in `unquoteXmlPattern` then it fails to parse xml literals in pattern position.
Also I added a test to make sure that empty splices in pattern position fail to parse (like in scalac):
```scala
checkError("e match { case <a>{}</a> => ??? }")
```

xeno-by: Thanks a lot for your investigation!

Issue #839, opened 2017-05-08
Imprecise warning messages for unused members of selective imports


gabro:
With reference to https://github.com/scalacenter/scalafix/pull/152#issuecomment-299727017

Given this code:

```scala
object imp {
  import scala.collection.mutable.{ Map, Set, ListBuffer }

  val l = ListBuffer.empty[Int]
}
```

scalameta produces the following semantic db:

```
Names:
[7..10): imp => _empty_.imp.
[22..27): scala => _root_.scala.
[28..38): collection => _root_.scala.collection.
[39..46): mutable => _root_.scala.collection.mutable.
[49..52): Map => _root_.scala.collection.mutable.Map.;_root_.scala.collection.mutable.Map#
[54..57): Set => _root_.scala.collection.mutable.Set.;_root_.scala.collection.mutable.Set#
[59..69): ListBuffer => _root_.scala.collection.mutable.ListBuffer.;_root_.scala.collection.mutable.ListBuffer#
[79..80): l => _empty_.imp.l.
[83..93): ListBuffer => _root_.scala.collection.mutable.ListBuffer.
[94..99): empty => _root_.scala.collection.generic.GenericCompanion#empty()Lscala/collection/GenTraversable;.
[100..103): Int => _root_.scala.Int#

Messages:
[15..71): [warning] Unused import

Denotations:
_empty_.imp. => final object imp
_empty_.imp.l. => val l: scala.collection.mutable.ListBuffer[Int]
```

Notice the single warning message referring to the whole import statement. The expected result would instead be multiple error messages referring to the separate members that are unused.

olafurpg:
Thank you for reporting! I tracked the error down to this line https://github.com/scalameta/scalameta/blob/d085c4c22282a809c7e2ea91669a2838592016b6/scalahost/nsc/src/main/scala/scala/meta/internal/databases/AnchorOps.scala#L26 it should be `pos.point` instead of `pos.start`. `Anchor` is going away in https://github.com/scalameta/scalameta/pull/828 but the fix for this bug will still be the same. I will open a PR to fix this as soon as #828 is merged

gabro:
Thanks for the fix! I've left a comment on the commit, take a look when you have a minute :)

olafurpg:
The commit that closed this issue got reverted because the change affected positions of sugars. I am investigating what we can do to support both sugars and compiler messages.

olafurpg:
[...]

PR #841, opened 2017-05-09
Refactor semanticdb

olafurpg:
Supersedes #828


scalameta/io/shared/src/main/scala/scala/meta/io/Multipath.scala
```diff
@@ -0,0 +1,78 @@
...
+    deep.find(_.name == name).map(_.uri)
+  }
+}
+
+@leaf class Classpath(value: String) extends Multipath {
+  def syntax: String = value
+  def structure: String = s"""Classpath("$value")"""
+  override def toString: String = syntax
+}
+object Classpath
```
olafurpg: I added this to get rid of the classnotfound error. I really want to get this PR in asap to use in scalafix.

scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
```diff
@@ -196,8 +196,8 @@ class SemanticSuite extends DatabaseSuite {
       |}
   """.trim.stripMargin,
     """
-      |[333..333) [commandeer.Foo, commandeer.FooDSL]
-      |[357..357) (commandeer.this.Foo.fooDSL)
+      |[320..333) [commandeer.Foo, commandeer.FooDSL]
+      |[333..357) (commandeer.this.Foo.fooDSL)
```
xeno-by: These positions don't seem right. They are supposed to be empty.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/InputOps.scala
```diff
@@ -19,7 +19,7 @@ trait InputOps { self: DatabaseOps =>
     def toMeta: m.Position = {
       assert(pos.isRange)
       val input = m.Input.File(pos.source.toAbsolutePath)
-      m.Position.Range(input, m.Point.Offset(input, pos.start), m.Point.Offset(input, pos.end))
+      m.Position.Range(input, m.Point.Offset(input, pos.point), m.Point.Offset(input, pos.end))
```
xeno-by: I suggest we reconsider this change, since it seems to break some positions in our tests.

scalahost/nsc/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
```diff
@@ -34,7 +34,10 @@ trait ScalahostPipeline extends DatabaseOps { self: ScalahostPlugin =>
         global.settings.outputDirs.getSingleOutput
           .map(_.file.getAbsolutePath)
           .getOrElse(global.settings.d.value))
-      val assumedSourcepath = Sourcepath(".")
+      // NOTE: user.dir is not a great default for sourcepaths since it will include
+      // irrelevant directories such as target. This parameter should ideall be passed by
+      // the build integration as a compiler flag -Xplugin:scalahost:xxx.
+      val assumedSourcepath = Sourcepath(sys.props("user.dir"))
```
xeno-by: What's the difference between `user.dir` and `.`?

scalameta/io/shared/src/main/scala/scala/meta/io/Fragment.scala
```diff
@@ -9,7 +9,7 @@ import org.scalameta.data._
   def uri: URI = {
     val baseuri = base.toURI.normalize
     if (baseuri.toString.endsWith(".jar")) new URI(s"jar:$baseuri!/$name")
-    else baseuri
+    else base.resolve(name).toURI
```
xeno-by: Shouldn't we normalize here, too?

scalahost/nsc/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
olafurpg: One is AbsolutePath and the other is RelativePath. `.` can evaluate to different paths at different times, which is not fun to debug. Sourcepath should IMO accept a `Seq[AbsolutePath]` instead of `String`. 

I didn't dig exactly into why `.` failed in the scripted tests, but I suspect it was related to how scripted tests are run from a temporary directory.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/InputOps.scala
olafurpg: I'll revert this change. We can discuss it better in #839

olafurpg:
Let's continue the discussion on positions in #839 and merge this PR

scalahost/nsc/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
xeno-by: Allright, let's replace the dot with user.dir.

Issue #816, opened 2017-04-27
Reorganize docs


olafurpg:
Currently there's

- http://scalameta.org/
- http://scalameta.org/tutorial
- https://github.com/scalameta/sbt-semantic-example (see readme)
- https://github.com/scalameta/scalameta/blob/master/notes/quasiquotes.md
- https://github.com/scalameta/sbt-macro-example

It would be nice if these documents were available from a single page.

I propose that we 

1. rename tutorial to `docs` so that it will have the url http://scalameta.org/docs It's not really a tutorial anymore since I recently removed the exercise sessions because they were outdated and were based on an ancient fork of scalafix. 
1. inline the quasiquote table into the tutorial around trees http://scalameta.org/tutorial/#Trees 
1. add a section in the tutorial on sbt-scalahost and move the contents from the sbt-semantic-example readme to that new section
1. move sbt-semantic-example to `sbt-scalahost.g8` and make it a giter8 template
1. move sbt-macro-example to `paradise.g8` and make it a giter8 template 

- [ ] Update readme to link to user docs instead of scalameta.org
- [x] Mention `.structure => copy-paste` trick, https://gitter.im/scalameta/scalameta?at=592473b15dc6925a7a44c48c
- [ ] crank up stack size to avoid stack overflow when parsing. Scalameta codebase uses lots of extension methods which quickly adds up.
- [ ] clarify installation instructions. Library vs. macro vs. intellij vs. sbt vs. stable release vs. pre-release
- [ ] link scaladoc or unidoc or Trees.scala
- [ ] provide example mini application that reads files and parses them
- [ ] clearly state current known limitations

ShaneDelmore:
I find myself constantly referring to the quasi quote table. I like the idea of putting it right in the main doc.

xeno-by:
Great points! Thanks for assembling them, they will be very useful.

olafurpg:
I think it's best to push this issue back another release. The milestone has a lot of open issues considering the deadline https://github.com/scalameta/scalameta/milestone/3

xeno-by:
@olafurpg Let's merge this ticket into https://github.com/scalameta/scalameta/issues/1265?

olafurpg:
Closing as fixed, we have a new documentation site that aggregates our current documentation https://scalameta.org/ the website can still be improved by adding more content, but those changes can be handled by separate tickets.

PR #842, opened 2017-05-10
Add version badges

DavidDudson:
There's currently no easy way to find out the latest versions. Hopefully these badges help.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,11 @@
 [![Build Status](https://platform-ci.scala-lang.org/api/badges/scalameta/scalameta/status.svg)](https://platform-ci.scala-lang.org/scalameta/scalameta)
 [![Join the chat at https://gitter.im/scalameta/scalameta](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/scalameta/scalameta?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
+## Versions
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,6 +2,11 @@
 [![Build Status](https://platform-ci.scala-lang.org/api/badges/scalameta/scalameta/status.svg)](https://platform-ci.scala-lang.org/scalameta/scalameta)
 [![Join the chat at https://gitter.im/scalameta/scalameta](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/scalameta/scalameta?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
+## Versions
+Release: ![Release](https://img.shields.io/maven-central/v/org.scalameta/scalameta_2.12.svg)
```
olafurpg: How about?

> See http://scalameta.org/#Gettingstarted for the latest stable and pre-release versions.

Those version numbers are copy-pasteable. The pre-release version is updated on every merge into master.

DavidDudson: Oh, I have not looked at the website for ages. I did not realise that was autogenerated. Ill close this PR.

olafurpg: I still think it's a good idea to link it from the readme, because I suspect that's where many people will look for the version number.

DavidDudson: @olafurpg It also confuses me that the meta page does not have paradise versions, you have to go to the tutorial site for that IIRC.

olafurpg: Same here, has confused me too. With #816, it should be possible to have a matrix with all versions in one place.

PR #845, opened 2017-05-10
Add support for file IO inside nodejs.

olafurpg:
This commits implements PathIO and FileIO for nodejs. This opens
possibilities to

- build command-line tools with scalameta on JS
- run file IO in tests, which will come in handy in the scalafmt test
  suite where almost all unit tests are written in external files
  using an external DSL.

This implementation uses
- the native `fs` module to read files
- the native `path` module to get file/path separators and check for
  absolute paths.
- uses npm package `shelljs` module to get cross-platform workingDirectory.
  It's a shame this is not provided by default in nodejs, but I'm not
  motivated nor competent enough to roll my own native implementation
  using the `process` module.


scalameta/io/js/src/main/scala/scala/meta/internal/io/PlatformPathIO.scala
```diff
@@ -2,28 +2,28 @@ package scala.meta.internal.io
 
 import scala.meta.io._
 
-object PathIO extends CommonPathIO {
+object PlatformPathIO {
+  private [io] def isNode = JSFs != null
```
xeno-by: How about we move this to a separate object?

olafurpg: This object is internal and in a Platform* file, what object do you think would be better?

xeno-by: scala.meta.internal/Platform.scala

olafurpg: I moved it to internal.io.PlatformIO. I can't move this to the scalameta project because it's used in the `io` module. This is really an internal implementation detail of ioJS and it should not leak into the shared project.

Issue #822, opened 2017-04-30
Semantic information for Database.sugars


xeno-by:
#808 introduces a prototype implementation of infos and sugars that represents these introspection artifacts as strings. It turns out it's not that outrageous, because even strings can be extremely useful for code analysis tools.

However, we clearly need to provide more information to our users. Representing both infos and sugars as full-fledged trees sounds like a good idea.

olafurpg:
I will take this on since I have an idea I'd like to try out.

I propose to

- Change `Sugar` message in schema
```proto
message Sugar {
  Range range = 1;
-  string syntax = 2;
+  Attributes attributes = 3;
}
``` 
- `Sugar.syntax` becomes `Sugar.attributes.contents`
- `m.Database.sugar(pos)` returns `Stat` instead of `String`.
  - sugars at the same position are concatenated, example `[1..1) [T]` and `[1..1) (blah)` becomes `[1..1) [T](blah)`. The ordering used for concatenation is 1) .apply 2) type params 3) implicit args.
  - Database.names includes names from nested sugars, this makes it possible to do `name.symbol` on names inside the returned Stat using the root Database.
  - If `Sugar.attributes.contents` does not parse as a Stat, then we prefix it with `*`. Example, `.apply => *.apply`/ `[T] => *[T]`/`implicitConf(*)(args)` is unchanged. The `*` identifier resolves to the symbol `_root_.scala.meta.sugar`.
  - other magic terms like `*` in `conversion(*)(arg)` also resolve to the symbol `_root_.scala.meta.sugar`.

olafurpg:
#1000 took the first steps towards fixing this problem. The PR added two data structures `Input.Sugar` and `semantic.Sugar`, which can be used to attribute "invisible code". Now that this model is merged, I propose we 
- close this issue
- delegate Database.denotations to #1015
- open separate issues for bugs/rfe on the new model to attribute sugar

PR #848, opened 2017-05-10
Make sugars and denotations safer to insert into code.

olafurpg:
Until #822 is implemented, this PR is just a sanity check that the strings in Denotation.info and Database.sugars are usable for scalafix with the v1.8 release.

### DIFF
--- a/scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
+++ b/scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
@@ -235,4 +235,50 @@ class SemanticSuite extends DatabaseSuite {
       |[294..294) [C[Int]]
   """.trim.stripMargin
   )
+
+  denotations(
+    """
+      |import scala.collection.mutable.ListBuffer
+      |import scala.collection.mutable.{HashSet => HS}
+      |trait B {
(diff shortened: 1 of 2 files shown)


scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
```diff
@@ -229,10 +229,61 @@ class SemanticSuite extends DatabaseSuite {
...
+      |    val result = b.x
+      |    result
+      |  }
+      |}
+    """.stripMargin,
+    """
+      |<...>@340..356 => val result: b.X
+      |_empty_.B# => trait B
+      |_empty_.B#X# => abstract type X
+      |_empty_.B#x()Ljava/lang/Object;. => abstract def x: B.this.X
```
xeno-by: This is still `B.this.X`. Are we happy with that?

olafurpg: That's fine, it typechecks which is the most important part.

olafurpg: Maybe we can use g.showCode to improve the pretty printing, but I wouldn't block a release for this.

Issue #821, opened 2017-04-30
Monads vs exceptions for error handling


xeno-by:
This is a long-standing debate between @densh (monads) and myself (exceptions). Some time ago, @olafurpg joined the fun on Denys's side.

The debate basically boils down to: "for which APIs do we represent errors in type signatures?". A related question is: "if we use types, then which ones?".

DavidDudson:
My opinion is all should use Monads.

I quite like macro failures throwing exceptions, but only because it's compile time only.

Do we have a list of Api's and what exceptions the can throw?

xeno-by:
The immediate motivation for filing this ticket is the discussion in https://github.com/scalameta/scalameta/pull/808. 

As mentioned in one of the reviews of the pull request, invoking `AbsolutePath.apply(path: String)` can result in an error if `path` is not absolute. How do we represent this error?

In the pull request, I specified the return type as plain `AbsolutePath` and then had the body of `apply` to throw `sys.error` if the argument is invalid. Below you can find my motivation.

1) When deciding between monads and exceptions, I typically ask myself a question: "can future callers of a given API reasonably handle the error?". If the answer is "no", then I use exceptions. 

For example, I wouldn't expect anyone except for hardcode VM/framework programmers to handle things like NPE and SO, so I don't express those erroneous conditions in signatures.

2) If it is imaginable for the error to be handled, I ask myself another question: "are API users likely to want to handle the error?". If the answer is "not really", then I use exceptions. 

For example, let's take `Mirror.apply(classpath, sourcepath)` that is supposed to be the foundation of adhoc semantic API. It clearly can fail, because the underlying semanticdb may be corrupt, may be in an incompatible format, etc. So should it return `Mirror` or `Try[Mirror]`?

I think that `Mirror.apply` should return `Mirror`, because I can hardly imagine API users wanting to handle the aforementioned errors.

If the semanticdb is corrupt, what is a user likely to do? Sure, sometimes they may want to take this error and convert it into something domain-specific, but which use case is going to be more often: 1) the happy `implicit val mirror = Mirror(...)`, or 2) the overly cautious `implicit val mirror = Mirror(...).orElse(...)`? I think that 1 is much more frequent, so I believe we should optimize the API towards that.

3) Otherwise, I guess it may make sense to use monads if the end result is not too heavyweight.

[...]

PR #849, opened 2017-05-10
Add Position.RangeWithPoint, fixes #383.

### LINKED ISSUE #383: scalac compatibility for error messages
This is something that we inherited from scalac. Since scalac runs in -Yrangepos only occasionally, its error reporting is offset-based. Therefore, all positions (both offset and range ones) have `Position.point` in scalac.

In our case, everything had range positions from the very beginning. Therefore, we should consider dropping `Position.point` altogether, in favor of Clang-style/Elm-style range errors.

olafurpg:
Position.point is necessary in some cases where start != point != end,
for example unused import warnings. The original Position.Range did
contain a point field that got removed in https://github.com/olafurpg/scalameta/commit/918e2ecab7a0dd95c6ff769f4fd76fb2e5d097f4
Instead of adding a new `point` field to Position, this commit
introduces a new Position.RangeWithPoint type. I opted for a new type
because

- it's relatively rare when start != point != end
- it's possible to provide a custom Range.unapply extractor that works
  to extract start/end from RangeWithPoint.

The hope is that RangeWithPoint will be mostly invisible unless you
know that you need the point field, for example in the
RemoveUnusedImports rewrite in scalafix.

### DIFF
--- a/scalahost/nsc/src/main/scala/scala/meta/internal/semantic/InputOps.scala
+++ b/scalahost/nsc/src/main/scala/scala/meta/internal/semantic/InputOps.scala
@@ -19,7 +19,14 @@ trait InputOps { self: DatabaseOps =>
     def toMeta: m.Position = {
       assert(pos.isRange)
       val input = m.Input.File(pos.source.toAbsolutePath)
-      m.Position.Range(input, m.Point.Offset(input, pos.start), m.Point.Offset(input, pos.end))
+      val mstart = m.Point.Offset(input, pos.start)
+      val mend = m.Point.Offset(input, pos.end)
+      if (pos.point > pos.start && pos.point < pos.end) {
+        val mpoint = m.Point.Offset(input, pos.point)
+        m.Position.RangeWithPoint(input, mstart, mpoint, mend)
+      } else {
+        m.Position.Range(input, mstart, mend)
+      }
     }
   }
 }
--- a/scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
+++ b/scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
@@ -286,4 +286,15 @@ class SemanticSuite extends DatabaseSuite {
     "class F[T: Manifest] { val arr = Array.empty[T] }",
     "[47..47) (F.this.evidence$1)".trim
   )
+
+  messages(
+    """
+      |import scala.collection.mutable.{ Map, Set, ListBuffer }
+      |object a { ListBuffer.empty[Int] }
+    """.stripMargin.trim,
+    """
+      |[0..34..56): [warning] Unused import
+      |[0..39..56): [warning] Unused import
+    """.stripMargin.trim
+  )
 }
--- a/scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
+++ b/scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
@@ -2,33 +2,43 @@ package scala.meta
 package inputs
 
 import java.nio.charset.Charset
+
 import org.scalameta.adt.{Liftables => AdtLiftables}
 import org.scalameta.data._
 import org.scalameta.invariants._
 import scala.meta.common._
 import scala.meta.internal.inputs._
+import scala.meta.internal.io.PathIO
 import scala.meta.io._
 
 trait Input extends Optional with Product with Serializable with InternalInput {
   def chars: Array[Char]
   def text: String = new String(chars)
+  def syntax: String
+  def structure: String
 }
 
 object Input {
   @none object None extends Input {
     lazy val chars = new Array[Char](0)
-    override def toString = "Input.None"
+    def syntax = "<none>"
+    def structure = "Input.None"
+    override def toString = structure
   }
 
   @data class String(s: scala.Predef.String) extends Input {
     lazy val chars = s.toArray
-    override def toString = "Input.String(\"" + s + "\")"
+    def syntax = "<input>"
+    def structure = "Input.String(\"" + s + "\")"
+    override def toString = structure
   }
 
   @data class Stream(stream: java.io.InputStream, charset: Charset) extends Input {
     lazy val chars = new scala.Predef.String(scala.meta.internal.io.InputStreamIO.readBytes(stream), charset).toArray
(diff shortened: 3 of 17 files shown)


scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
```diff
@@ -235,4 +235,14 @@ class SemanticSuite extends DatabaseSuite {
       |[294..294) [C[Int]]
   """.trim.stripMargin
   )
+  messages(
+    """
+      |import scala.collection.mutable.{ Map, Set, ListBuffer }
+      |object a { ListBuffer.empty[Int] }
+    """.stripMargin.trim,
+    """
+      |[0..34..56): [warning] Unused import
```
olafurpg: cc/ @gabro this is how Position.RangeWithPoint looks like, the middle number is the `.point` that is where the caret `^` appears in scalac error messages.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/InputOps.scala
```diff
@@ -19,7 +19,14 @@ trait InputOps { self: DatabaseOps =>
     def toMeta: m.Position = {
       assert(pos.isRange)
       val input = m.Input.File(pos.source.toAbsolutePath)
-      m.Position.Range(input, m.Point.Offset(input, pos.start), m.Point.Offset(input, pos.end))
+      val mstart = m.Point.Offset(input, pos.start)
+      val mend = m.Point.Offset(input, pos.end)
+      if (pos.point > pos.start && pos.point < pos.end) {
```
gabro: being paranoid here, but can `pos.point` be `< pos.start` or `> pos.end` under any circumstance?
Would `if (point.point != pos.start && pos.point != pos.end)` be more defensive towards compiler's quirks?

scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
gabro: looking good!

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/InputOps.scala
olafurpg: wdyt @xeno-by ? I don't know if/when point can be outside the range.

xeno-by: It can't, I believe.

scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
```diff
@@ -2,33 +2,43 @@ package scala.meta
...
 import scala.meta.common._
 import scala.meta.internal.inputs._
+import scala.meta.internal.io.PathIO
 import scala.meta.io._
 
 trait Input extends Optional with Product with Serializable with InternalInput {
   def chars: Array[Char]
   def text: String = new String(chars)
+  def syntax: String
+  def structure: String
```
xeno-by: How about we defer this change to the upcoming 2.x PR? There are some things to decide (whether `toString` should return syntax or structure, how to get rid of the `syntax/structure` boilerplate) that I wouldn't want to rush for 1.8.0.

scalameta/inputs/shared/src/main/scala/scala/meta/internal/inputs/package.scala
```diff
@@ -6,23 +6,19 @@ import scala.meta.inputs._
...
     def formatMessage(severity: String, message: String): String = {
       if (point != Point.None) {
         val input = point.input
-        val shortContent = input match {
-          case Input.File(path, _) => path.toString
-          case Input.LabeledString(label, _) => label
-          case _ => "<input>"
-        }
-        val header = s"$shortContent:${point.line + 1}: $severity: $message"
+        val header = s"${input.syntax}:${point.line + 1}: $severity: $message"
```
xeno-by: This discovers a really great opportunity for code reuse!

scalameta/io/shared/src/main/scala/scala/meta/io/AbsolutePath.scala
```diff
@@ -16,7 +16,11 @@ import scala.meta.internal.io.{FileIO, PathIO}
   @deprecated("Use toString() instead", "1.8")
   def absolute: String = toString()
 
-  def toRelative: RelativePath = toRelative(PathIO.workingDirectory)
+  // NOTE: Anticipating exceptions in this method is just plain terrible.
+  // It's not exceptional at all that this AbsolutePath is not relative to the
+  // working directory. Fix this once https://github.com/scalameta/scalameta/issues/821
+  // is settled.
+  def toRelative: Option[RelativePath] = scala.util.Try(toRelative(PathIO.workingDirectory)).toOption
```
xeno-by: I suggest we remove this method altogether until we work out #821. Since that issue has been brought up multiple times over the last couple weeks, I also suggest we bump its priority to critical, lower than only the issues that block def macros.

scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
xeno-by: Given the amount of hacky changes necessary to accommodate this use case, I suggest we instead hack at the source - by hijacking the positions of `unused import` warnings.

scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
olafurpg: I agree, let's defer this for another release.

scalameta/inputs/shared/src/main/scala/scala/meta/internal/inputs/package.scala
olafurpg: I have written that pattern match at least 4-5 times in metaconfig/scalafix/scalafmt 😄  This change will be welcome once we clean up pretty-printing in a future release.

scalameta/io/shared/src/main/scala/scala/meta/io/AbsolutePath.scala
olafurpg: I will open a separate PR addressing what I'd like to see changed for #821

Issue #827, opened 2017-05-01
Tests for standalone reading/writing semantic dbs under JVM/JS


xeno-by:
Creating a note for myself in order not to forget to do this.

olafurpg:
We have such tests now in https://github.com/scalameta/scalameta/blob/f7463f78227e092b5dab0a05667733b0a2b23952/scalameta/tests/shared/src/test/scala/scala/meta/tests/CrossPlatformSemanticSuite.scala writing + reading in JVM and reading only in JS

PR #854, opened 2017-05-11
fix #777

### LINKED ISSUE #777: Semantic DB is not portable

xeno-by:
/cc @stuhood

### DIFF
--- a/scalahost/nsc/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
+++ b/scalahost/nsc/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
@@ -4,8 +4,10 @@ package scalahost
 import java.io._
 import java.net.URI
 import scala.collection.mutable
+import scala.compat.Platform.EOL
 import scala.tools.nsc.Phase
 import scala.tools.nsc.plugins.PluginComponent
+import scala.util.control.NonFatal
 import scala.{meta => m}
 import scala.meta.io._
 import scala.meta.internal.semantic.DatabaseOps
@@ -41,9 +43,18 @@ trait ScalahostPipeline extends DatabaseOps { self: ScalahostPlugin =>
       implicit class XtensionURI(uri: URI) { def toFile: File = new File(uri) }
 
       override def apply(unit: g.CompilationUnit): Unit = {
-        if (!unit.source.file.name.endsWith(".scala")) return
-        val mminidb = m.Database(List(unit.source.toInput -> unit.toAttributes))
-        mminidb.save(outputClasspath, assumedSourcepath)
+        try {
+          if (!unit.source.file.name.endsWith(".scala")) return
+          val mminidb = m.Database(List(unit.source.toInput -> unit.toAttributes))
+          mminidb.save(outputClasspath, assumedSourcepath)
+        } catch {
+          case NonFatal(ex) =>
+            val msg = new StringWriter()
+            val path = unit.source.file.path
+            msg.write(s"failed to generate semanticdb for $path:$EOL")
+            ex.printStackTrace(new PrintWriter(msg))
+            global.reporter.error(g.NoPosition, msg.toString)
+        }
       }
 
       override def run(): Unit = {
--- a/scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
+++ b/scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
@@ -60,10 +60,7 @@ object ScalahostSbtPlugin extends AutoPlugin {
     javaOptions ++= {
       if (scalametaDependencies.value.isEmpty) Nil
       else {
-        val sourcepath =
-          scalahostSourcepath.value
-            .flatMap(_.map(_.getAbsolutePath))
-            .mkString(java.io.File.pathSeparator)
+        val sourcepath = sys.props("user.dir")
         val classpath =
           scalahostClasspath.value
             .flatMap(_.files.map(_.getAbsolutePath))
@@ -136,8 +133,6 @@ object ScalahostSbtPlugin extends AutoPlugin {
     ScopeFilter(inProjects(scalametaDependencies.value.map(x => LocalProject(x.project)): _*),
                 inConfigurations(Compile, Test, IntegrationTest))
   }
-  private val scalahostSourcepath: Def.Initialize[Seq[Seq[File]]] =
-    Def.settingDyn(sourceDirectories.all(scalahostAggregateFilter.value))
   private val scalahostClasspath: Def.Initialize[Task[Seq[Classpath]]] =
     Def.taskDyn(fullClasspath.all(scalahostAggregateFilter.value))
 }
--- a/scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
+++ b/scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
@@ -14,7 +14,7 @@ import scala.meta.inputs.{Input => mInput, Position => mPosition, Point => mPoin
 
 package object meta {
   implicit class XtensionDatabaseSchema(mdatabase: m.Database) {
-    def toSchema: s.Database = {
+    def toSchema(sourcepath: Sourcepath): s.Database = {
       val sentries = mdatabase.entries.map {
         case (minput, m.Attributes(mdialect, mnames, mmessages, mdenots, msugars)) =>
           object mDialect {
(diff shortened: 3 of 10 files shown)


xeno-by:
@olafurpg I'll finish with fat semantic databases tomorrow.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
```diff
@@ -0,0 +1,82 @@
...
+package internal
+package semantic
+
+import java.nio.charset.Charset
+import org.scalameta.data._
+import scala.{Seq => _}
+import scala.collection.immutable.Seq
+import scala.meta.io._
+import scala.meta.internal.semantic.{schema => s}
+import scala.meta.{semantic => m}
```
olafurpg: Can we try to use consistent names? I thought `m` was `meta`.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
```diff
@@ -0,0 +1,82 @@
...
+import scala.meta.{Dialect => mDialect}
```
olafurpg: What's the motivation for this rename? This gets shadowed by the `mDialect` extractor below.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
```diff
@@ -0,0 +1,82 @@
...
+        case (minput, m.Attributes(mdialect, mnames, mmessages, mdenots, msugars)) =>
+          object mDialect {
+            def unapply(mdialect: mDialect): Option[String] = {
+              val isStandard = scala.meta.Dialect.standards.exists(_._2 == mdialect)
+              if (isStandard) Some(mdialect.toString) else None
+            }
+          }
+          object mPosition {
+            def unapply(mpos: mPosition): Option[s.Range] = mpos match {
+              case scala.meta.inputs.Position.Range(`minput`, mPoint.Offset(_, sstart), mPoint.Offset(_, send)) => Some(s.Range(sstart, send))
```
olafurpg: Position is fully qualified here but imported as `mPosition`. Why the difference?

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
```diff
@@ -0,0 +1,82 @@
...
+import scala.meta.inputs.{Input => mInput, Position => mPosition, Point => mPoint}
+
+package object meta {
+  implicit class XtensionDatabaseSchema(mdatabase: m.Database) {
+    def toSchema(sourcepath: Sourcepath): s.Database = {
+      val sentries = mdatabase.entries.map {
+        case (minput, m.Attributes(mdialect, mnames, mmessages, mdenots, msugars)) =>
+          object mDialect {
+            def unapply(mdialect: mDialect): Option[String] = {
+              val isStandard = scala.meta.Dialect.standards.exists(_._2 == mdialect)
```
olafurpg: I'm not sure I like this `mdialect` vs. `mDialect` trick, took me a while to spot.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
```diff
@@ -0,0 +1,82 @@
...
+              case m.Denotation(sflags, sname, sinfo) => Some(s.Denotation(sflags, sname, sinfo))
+              case _ => None
+            }
+          }
+          val spath = minput match {
+            case mInput.File(path, charset) if charset == Charset.forName("UTF-8") => sourcepath.relativize(path.toURI).getOrElse(sys.error(s"can't find $path in $sourcepath"))
+            case other => sys.error(s"bad database: unsupported input $other")
+          }
+          val sdialect = mdialect match {
+            case mDialect(sdialect) => sdialect
```
olafurpg: Could we do something like `m.Dialect.standards.find(_.name == mdialect).getOrElse(sys.error)`? This whole extractor business is not helping readability IMO.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
olafurpg: PS, here is how this line looks in the review UI

screen shot 2017-05-11 at 10 40 45

Please try to keep lines ~100 chars

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
olafurpg: Aah, is it to get the `object mDialect { def unapply(dialect: mDialect) }` sync between the extractor and the parameter type names? If so, then I'm not sure it's worth it.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
xeno-by: It is. But in this project, we don't yet have an umbrella `package object meta extends Everything`.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
xeno-by: Fixed.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
xeno-by: Fixed.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
xeno-by: Fixed.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
xeno-by: Fixed.

xeno-by:
@olafurpg The tests are now passing. `scalahostSbt` again ended up being super useful thanks to its integration tests of the semanticdb workflow. I suggest we postpone non-integration tests till https://github.com/scalameta/scalameta/issues/827.

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
```diff
@@ -60,10 +60,7 @@ object ScalahostSbtPlugin extends AutoPlugin {
     javaOptions ++= {
       if (scalametaDependencies.value.isEmpty) Nil
       else {
-        val sourcepath =
-          scalahostSourcepath.value
-            .flatMap(_.map(_.getAbsolutePath))
-            .mkString(java.io.File.pathSeparator)
+        val sourcepath = sys.props("user.dir")
```
olafurpg: We may hit on performance issues in `Fragment.deep/find` with this sourcepath. For example, in the scalafmt/scalameta repos `target/` often contains (tens of) thousands of files from testkit.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
olafurpg: 👍

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Meta.scala
olafurpg: 👍

scalahost/nsc/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
```diff
@@ -41,9 +43,18 @@ trait ScalahostPipeline extends DatabaseOps { self: ScalahostPlugin =>
...
-        mminidb.save(outputClasspath, assumedSourcepath)
+        try {
+          if (!unit.source.file.name.endsWith(".scala")) return
+          val mminidb = m.Database(List(unit.source.toInput -> unit.toAttributes))
+          mminidb.save(outputClasspath, assumedSourcepath)
+        } catch {
+          case NonFatal(ex) =>
+            val msg = new StringWriter()
+            val path = unit.source.file.path
+            msg.write(s"failed to generate semanticdb for $path:$EOL")
```
olafurpg: Nice!

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
xeno-by: We can definitely improve the implementations on deep and find if the need arises. At the moment, they are intentionally very naive in order to focus the development effort.

PR #858, opened 2017-05-12
support for fat semantic databases

xeno-by:


scalahost/nsc/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
```diff
@@ -15,6 +15,20 @@ import scala.meta.internal.semantic.{vfs => v}
...
+  //   * Working directory represents both previous and current compilation roots
+  //     (that's the protocol that we assume that Scala build tools support)
+  //   * All uris related to the previous semantic db are based on the file: protocol
+  //     (that follows from the fact that the output directory is based on the file: protocol)
+  lazy val scalametaClasspath = Classpath(
+    global.settings.outputDirs.getSingleOutput
+      .map(_.file.getAbsolutePath)
+      .getOrElse(global.settings.d.value))
+  lazy val scalametaSourcepath = Sourcepath(
+    sys.props.getOrElse("scalameta.sourcepath", sys.props("user.dir")))
```
olafurpg: Configuration 👏 How about we move this to `ConfigOps` so there's one place to look these things up?

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/ConfigOps.scala
```diff
@@ -0,0 +1,12 @@
+package scala.meta.internal
+package semantic
+
+trait ConfigOps { self: DatabaseOps =>
+  object config {
```
olafurpg: There is an official way to pass custom flags to compiler plugins, I think we should eventually move to that system because sys.props has problems like

- hard to use different configuration  for different projects (at least in sbt)
- jvm forking becomes problematic
- global mutable state (boo!)

Can we add a note here saying that we should ideally change to compiler -X flags in the future?

scalahost/sbt/src/sbt-test/sbt-scalahost/semantic-example-fat/build.sbt
```diff
@@ -1,6 +1,9 @@
 scalaVersion in ThisBuild := "2.11.11"
 scalacOptions in ThisBuild += "-deprecation"
 scalacOptions in ThisBuild += "-Ywarn-unused-import"
+compile in Compile <<= (compile in Compile).dependsOn(Def.task {
+  sys.props("scalameta.semanticdb") = "fat"
```
olafurpg: The idiomatic way to do this however is to provide a `scalametaSemanticdb := fat/slim/none` settingKey. I imagine that the `none` will be very useful. I can do that change in sbt-scalahost in a followup PR if you like.

xeno-by: That'd be great!

scalahost/sbt/src/sbt-test/sbt-scalahost/semantic-example-slim/test
```diff
@@ -0,0 +1 @@
+> app/run
```
xeno-by: It tests that the roundtrip produces a db that looks reasonable and has reasonable number of entries. Can you follow-up on this in a subsequent PR?

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Symbol.scala
```diff
@@ -35,10 +36,10 @@ object Symbol {
     override def fullName: Ref = unsupported(this, "fullName")
   }
 
-  @leaf class Local(position: Position) extends Symbol {
```
xeno-by: `Input.LabeledString` would be nice, but you need `contents` to create it, and we don't have that when parsing a local symbol.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Symbol.scala
```diff
@@ -35,10 +36,10 @@ object Symbol {
...
+  @leaf class Local(id: String) extends Symbol {
```
xeno-by: See the comment above.

scalahost/nsc/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
xeno-by: Good idea

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/ConfigOps.scala
xeno-by: A counterargument here would be that compiler plugins work differently across different implementation platforms, and we'd like a uniform way to configure semantic dbs regardless of the platform. Wdyt?

scalameta/io/shared/src/main/scala/scala/meta/internal/io/FileIO.scala
```diff
@@ -9,5 +9,5 @@ object FileIO {
     PlatformFileIO.slurp(path, charset)
 
   def slurp(path: AbsolutePath): String =
-    slurp(path, Charset.forName("UTF8"))
+    slurp(path, Charset.forName("UTF-8"))
```
xeno-by: I wonder why it's not a constant in the JDK, given its importance.

scalahost/sbt/src/sbt-test/sbt-scalahost/semantic-example-slim/test
olafurpg: I can do that.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Symbol.scala
olafurpg: I see. Makes sense.

xeno-by:
Thanks!

PR #861, opened 2017-05-14
Improve Xml parser performances

allanrenucci:
I believe this can improve the performances of the xml parser.
We instantiate a new parser for every single xml literal we parse. I believe
most of the time is spent during initialization since our xml parsers are
short lived and not reusable. These changes reduce initialization time
and memory consumption as it avoids pre-computing the lookup tables.

A better solution would be to have a reusable parser. But we would need
to remove the parser dependency on `Dialect`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -136,7 +136,7 @@ lazy val tokenizers = crossProject
   .settings(
     publishableSettings,
     description := "Scalameta APIs for tokenization and their baseline implementation",
-    libraryDependencies += "com.lihaoyi" %%% "fastparse" % "0.4.2",
+    libraryDependencies += "com.lihaoyi" %%% "fastparse" % "0.4.3",
     enableMacros
   )
   .dependsOn(common, dialects, inputs, tokens)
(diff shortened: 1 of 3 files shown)


allanrenucci:
I observe a 2x speedup when running the XmlSuite with this change

build.sbt
```diff
@@ -136,7 +136,7 @@ lazy val tokenizers = crossProject
   .settings(
     publishableSettings,
     description := "Scalameta APIs for tokenization and their baseline implementation",
-    libraryDependencies += "com.lihaoyi" %%% "fastparse" % "0.4.2",
+    libraryDependencies += "com.lihaoyi" %%% "fastparse" % "0.4.3",
```
allanrenucci: @olafurpg This breaks scalatex. Do you have any idea why? To reproduce:
```
sbt readme/doc
```

olafurpg: I opened https://github.com/lihaoyi/Scalatex/pull/52 and published a pre-release of scalatex to our bintray repo with that fix. We can use the pre-release until the official scalatex is out

```scala
// project/plugins.sbt
resolvers += Resolver.bintrayIvyRepo("scalameta", "sbt-plugins") // only needed for scalatex 0.3.8-pre
addSbtPlugin("com.lihaoyi" % "scalatex-sbt-plugin" % "0.3.8-pre")
// build.sbt
resolvers += Resolver.bintrayRepo("scalameta", "maven") // under sharedSettings
```

allanrenucci:
If we can get rid of the dependency on dialect in `LegacyScanner`, I can optimise further (since the xml parser depends on `LegacyScanner`). What's your thoughts about this?

PR #862, opened 2017-05-14
Publish sbt-scalahost ivy style.

olafurpg:
Currently, it's not possible to depend on sbt-scalahost pre-releases
with addSbtPlugin or libraryDependencies because it's published
maven-style to bintray.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -255,6 +255,7 @@ lazy val scalahostSbt = project
     buildInfoSettings,
     sbt.ScriptedPlugin.scriptedSettings,
     sbtPlugin := true,
+    publishMavenStyle := false, // necessary for pre-releases to work with addSbtPlugin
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -255,6 +255,7 @@ lazy val scalahostSbt = project
     buildInfoSettings,
     sbt.ScriptedPlugin.scriptedSettings,
     sbtPlugin := true,
+    publishMavenStyle := false, // necessary for pre-releases to work with addSbtPlugin
```
olafurpg: @xeno-by I couldn't find where publishTo is set to sonatype for stable releases, this field should only be false when publishing to Bintray.

xeno-by: I never got to reinstating this part of our build. Currently doing this manually (ugh!).

xeno-by:
LGTM

PR #864, opened 2017-05-14
Refactor internal.semantic module

olafurpg:


scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/Entry.scala
```diff
@@ -0,0 +1,53 @@
...
+  def bytes: Array[Byte]
+}
+
+object Entry {
+  private def readBytes(fragment: Fragment) = InputStreamIO.readBytes(fragment.uri.toURL.openStream)
+  @leaf class OnDisk(fragment: Fragment) extends Entry { lazy val bytes = readBytes(fragment) }
+  @leaf class InMemory(fragment: Fragment, bytes: Array[Byte]) extends Entry
+}
+
+object Paths {
```
xeno-by: We can move this to a separate file, too.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/Entry.scala
```diff
@@ -0,0 +1,29 @@
...
+import org.scalameta.adt._
+import org.scalameta.data._
+import org.scalameta.invariants._
+import scala.{Seq => _}
+import scala.collection.immutable.Seq
+import scala.meta.io._
+import scala.meta.internal.io.InputStreamIO
+import scala.meta.internal.io.PathIO.fileSeparator
+import scala.meta.internal.semantic.{vfs => v}
+import scala.meta.internal.semantic.{schema => s}
```
xeno-by: I think most of these imports can be removed now.

olafurpg: God I need to implement a remove unused imports rewrite in scalafix 😄

PR #866, opened 2017-05-15
Use scalacOptions for configuration instead of sys.props.

olafurpg:
System properties make it difficult/hacky to
- customize configuration per project
- dynamically update the settings in a sbt shell (set scalametaSourcepath := blah)
- fork the JVM
- debug when something goes wrong
- test scalahost-nsc (requires try/finally)

This commit makes scalahost use the default `init` mechanism provided by
scala-compiler, which makes it possible to configure scalahost via
scalacOptions. sbt-scalahost has been updated to take advantage of this
new feature. This change opens possibilities to build custom semanticdb
files, for example only emit "Messages".

Review @jvican


scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
```diff
@@ -22,15 +37,21 @@ object ScalahostSbtPlugin extends AutoPlugin {
 
   lazy val scalahostAllSettings =
     scalahostBaseSettings ++
-      scalahostInjectCompilerPluginSettings ++
+      inTestAndCompile(scalahostConfigSettings) ++
```
olafurpg: Is this an idiomatic way to allow users to configure the semanticdb mode by project/config? See semantic-example `test` how to disable/enable custom semantic dbs by config.

build.sbt
```diff
@@ -618,4 +620,7 @@ def CiCommand(name: String)(commands: List[String]): Command = Command.command(n
   }
 }
 def ci(command: String) = s"plz $ciScalaVersion $command"
+lazy val scala210 = "2.10.6"
+lazy val scala211 = "2.11.11"
+lazy val scala212 = "2.12.2"
```
xeno-by: How about we move these numbers to the top of the build? To me they seem pretty important.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/ConfigOps.scala
```diff
@@ -2,17 +2,38 @@ package scala.meta.internal
 package semantic
 
 import scala.meta.io._
+import scala.meta.internal.scalahost.ScalahostPlugin.name
 
+case class ScalahostConfig(sourcepath: Sourcepath, semanticdb: SemanticdbMode) {
+  def syntax = s"-P:$name:sourcepath:$sourcepath -P:$name:semanticdb:$sourcepath"
```
xeno-by: Not sure that I understand this part: `-P:$name:semanticdb:$sourcepath`.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/ConfigOps.scala
```diff
@@ -2,17 +2,38 @@ package scala.meta.internal
 package semantic
```
xeno-by: Shall we split this file into several and then create a dedicated package for all this functionality, like we did with `Vfs.scala` and `Schema.scala`?

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/ConfigOps.scala
```diff
@@ -2,17 +2,38 @@ package scala.meta.internal
...
-    def disabled: Boolean = style == "none"
-    def slim: Boolean = style == "slim"
-    def fat: Boolean = style == "fat"
+  val SetSemanticdb = "semanticdb:(.*)".r
+  val SetSourcepath = "sourcepath:(.*)".r
 
-    def sourcepath: Sourcepath =
-      Sourcepath(sys.props.getOrElse("scalameta.sourcepath", sys.props("user.dir")))
+  var config: ScalahostConfig = ScalahostConfig.default
+  // Hack to support config.setSourcepath().
```
xeno-by: Wdym by "hack"?

scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/DatabaseSuite.scala
```diff
@@ -40,49 +41,43 @@ abstract class DatabaseSuite extends FunSuite { self =>
     val global: self.g.type = self.g
   }
   import databaseOps._
+  databaseOps.config.setSemanticdbMode(SemanticdbMode.Slim)
```
xeno-by: Isn't `databaseOps` already imported here?

scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/DatabaseSuite.scala
```diff
@@ -40,49 +41,43 @@ abstract class DatabaseSuite extends FunSuite { self =>
...
-        errors.foreach(error => fail(s"scalac ${phase.name} error: ${error.msg} at ${error.pos}"))
-      })
-      g.phase = run.phaseNamed("patmat")
-      g.globalPhase = run.phaseNamed("patmat")
-
-      m.Database(List(unit.source.toInput -> unit.toAttributes))
-    } finally {
-      if (oldSourcepath != null) sys.props("scalameta.sourcepath") = oldSourcepath
-      else sys.props.remove("scalameta.sourcepath")
-    }
```
xeno-by: Not needing this is definitely an improvement! 👍

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
```diff
@@ -22,15 +40,21 @@ object ScalahostSbtPlugin extends AutoPlugin {
...
+      scalahostInjectDependency ++
       scalahostHasMirrorSettings
 
   lazy val scalahostBaseSettings: Seq[Def.Setting[_]] = Def.settings(
     ivyConfigurations += Scalameta,
+    scalametaSourcepath := baseDirectory.in(ThisBuild).value,
     resolvers += Resolver.bintrayRepo("scalameta", "maven")
   )
+  lazy val scalahostConfigSettings: Seq[Def.Setting[_]] = Def.settings(
+    scalametaSemanticdb := ScalametaSemanticdb.Fat
```
xeno-by: I don't like the duplication in saying `foobarbaz := Foobarbaz.qux`. Can we do better here?

scalahost/sbt/src/sbt-test/sbt-scalahost/semantic-example/test
```diff
@@ -0,0 +1,11 @@
+> app/testOnly Fat
```
xeno-by: How about we say `set every scalametaSemanticdb := ScalametaSemanticdb.Fat` before this line?

build.sbt
xeno-by: Allright, an alternative suggestion. Have file called `Versions.scala` in `build/`. And then call these guys `LatestScala210`, `LatestScala211` and `LatestScala212`.

xeno-by: If we need to support 2.11.8, we can add something `ExceptionallyPreviousScala211`.

build.sbt
```diff
@@ -250,26 +251,28 @@ lazy val scalahostNsc = project
...
   .settings(
     publishableSettings,
     buildInfoSettings,
+    Defaults.itSettings,
     sbt.ScriptedPlugin.scriptedSettings,
     sbtPlugin := true,
     publishMavenStyle := false, // necessary for pre-releases to work with addSbtPlugin
     bintrayRepository := "maven", // sbtPlugin overrides this to sbt-plugins
-    testQuick := {
+    testQuick.in(IntegrationTest) := {
```
jvican: Note: will hack incremental testing this week.

project/Versions.scala
```diff
@@ -0,0 +1,5 @@
+object Versions {
```
jvican: 👍 This is one of those things I'd like to see in a sbt common package that every build depens on.

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
jvican: AFAIK, this is very idiomatic. Make sure you document it clearly -- tasks are scoped in the `Compile` and `Test` configurations.

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
```diff
@@ -22,15 +40,21 @@ object ScalahostSbtPlugin extends AutoPlugin {
...
+      scalahostInjectScalacOptions ++
+      scalahostInjectDependency ++
       scalahostHasMirrorSettings
 
   lazy val scalahostBaseSettings: Seq[Def.Setting[_]] = Def.settings(
     ivyConfigurations += Scalameta,
+    scalametaSourcepath := baseDirectory.in(ThisBuild).value,
     resolvers += Resolver.bintrayRepo("scalameta", "maven")
   )
+  lazy val scalahostConfigSettings: Seq[Def.Setting[_]] = Def.settings(
```
jvican: You don't need `lazy` here.

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
jvican: What duplication do you mean Eugene?

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
```diff
@@ -39,14 +63,21 @@ object ScalahostSbtPlugin extends AutoPlugin {
           )
         case _ => Nil
       }
-    },
+    }
+  )
+  lazy val scalahostInjectScalacOptions: Seq[Def.Setting[_]] = Def.settings(
```
jvican: You can probably remove all these lazy's.

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
```diff
@@ -22,15 +40,21 @@ object ScalahostSbtPlugin extends AutoPlugin {
 
   lazy val scalahostAllSettings =
     scalahostBaseSettings ++
-      scalahostInjectCompilerPluginSettings ++
+      inTestAndCompile(scalahostConfigSettings) ++
+      scalahostInjectScalacOptions ++
```
jvican: I would group the settings based on the place they are scoped. Following this criterion, these two can be grouped together.

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
```diff
@@ -90,6 +121,9 @@ object ScalahostSbtPlugin extends AutoPlugin {
     }
   )
 
+  private def inTestAndCompile(settings: Seq[Def.Setting[_]]) =
```
jvican: I've used this so many times that i think it's about time we add it to sbt/sbt.

scalahost/sbt/src/sbt-test/sbt-scalahost/semantic-example/app/src/test/scala/MainTest.scala
```diff
@@ -0,0 +1,74 @@
...
+    test("Mirror.sources") {
+      assert(mirror.sources.length == 3)
+    }
+    val paths = List(
+      "library1/target/scala-2.11/classes/META-INF/semanticdb/library1/src/main/scala-2.11/Test.semanticdb",
+      "library2/target/scala-2.11/classes/META-INF/semanticdb/library2/src/main/scala/Test.semanticdb",
+      "library2/target/scala-2.11/test-classes/META-INF/semanticdb/library2/src/test/scala/TestMain.semanticdb"
+    )
+    // parent because fork := true
+    val workingDir = Paths.get(sys.props("user.dir")).getParent
```
jvican: Question: have you experienced a performance hit because of this forking? In theory, it should be slower.

scalahost/sbt/src/sbt-test/sbt-scalahost/semantic-example/test
```diff
@@ -0,0 +1,11 @@
...
+> set every scalametaSemanticdb := ScalametaSemanticdb.Slim
+> clean
+> app/testOnly Slim
+> clean
+> set every scalametaSemanticdb := ScalametaSemanticdb.Fat
```
jvican: `every` works here, but you need to be careful because it overrides all the keys in every scope. `set scalametaSemanticdb in Compile := ScalametaSemanticdb.Fat` and its counterpart for `Test` should work. I would prefer the last ones since they are more precise.
  
See `help set`:
```
set [every] <setting-expression>

	Applies the given setting to the current project:
	  1) Constructs the expression provided as an argument by compiling and loading it.
	  2) Appends the new setting to the current project's settings.
	  3) Re-evaluates the build's settings.

	This command does not rebuild the build definitions, plugins, or configurations.
	It does not automatically persist the setting(s) either.
	To persist the setting(s), run 'session save' or 'session save-all'.

	If 'every' is specified, the setting is evaluated in the current context
	and the resulting value is used in every scope.  This overrides the value
	bound to the key everywhere.

```

project/Versions.scala
olafurpg: I totally agree, I do this in almost every build. I'm not sure if we would use it in this build however since the scala version is very important for us.

olafurpg: Although, I can imagine that might make sbt version upgrades harder since upgrading sbt would mean upgrading to latest Scala. Not sure if it's a good thing or bad.

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
olafurpg: Good point. Docs are totally behind atm. cc/ https://github.com/scalameta/scalameta/issues/816

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
olafurpg: The `lazy` is actually intentional, but maybe not for the best reason. ScalahostSbtPlugin is an object so removing the lazy would make initialization errors more cryptic compared to typical stack traces.

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
olafurpg: I think he's referring to `scalametaX := scalametaY`

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
olafurpg: See comment above.

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
olafurpg: 👏

scalahost/sbt/src/sbt-test/sbt-scalahost/semantic-example/app/src/test/scala/MainTest.scala
olafurpg: I haven't. But these tests run via scripted so the overhead of forking here is negligible compared to scripted slowness.

scalahost/sbt/src/sbt-test/sbt-scalahost/semantic-example/test
olafurpg: I need every here since I want to define it for all projects. When I tried without `every` it only changed the setting in the root project.

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
xeno-by: Yes.

xeno-by: I suggest we at least drop the `scalameta` prefix. There are no other semantic dbs around.

`semanticDb := SemanticDb.Fat` looks more or less allright.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/ConfigOps.scala
xeno-by: Has anything changed in this regard? Can't seem to find the changes.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/ConfigOps.scala
xeno-by: I don't mind either way, just wanted to figure out your preferences wrt this aspect of code organization.

scalahost/sbt/src/main/scala/scala/meta/scalahost/sbt/ScalahostSbtPlugin.scala
olafurpg: It's conventional to prefix all plugin keys by the plugin name, this makes it easier to discover what settings are provided by a plugin, for example via `scalameta<TAB>` in the sbt shell. I'm open to make the syntax `scalametaSemanticdb := Semanticdb.Fat`, would that be better?

olafurpg: Relevant http://www.lihaoyi.com/post/StrategicScalaStyleConcisenessNames.html#long-names-vs-short-names

Considering that users will typically write this setting few times, I believe a longer/verbose name that follows sbt plugin naming conventions is fine.

xeno-by: Well, then I guess it should be `scalahostSemanticDb := ScalahostSemanticDb.Fat`.

olafurpg: Or rather, the plugin should be `ScalametaPlugin` named instead of `ScalahostSbtPlugin`?

This also opens the question if we should rename `sbt-scalahost` to `sbt-scalameta`. Having explained what were doing to several people, I have noticed that the word "scalahost" is confusing to many.

xeno-by: Good point. How about we explore this in 2.x? I'm afraid big refactorings like this will interfere with the 2.x branch, which will have to coexist with 1.x for quite a while.

olafurpg: So keep things as is for now? So far the plugin exposes only `scalameta*` keys to the build.sbt scope so I think renaming the plugin to sbt-scalameta would be a logical step, but I'm fine with waiting with that change.

xeno-by: Yes, let's keep this as is for now.

Issue #867, opened 2017-05-15
Semanticdb bytes are not concatenable


olafurpg:
Semanticdb [Attributes have no `path` field](https://github.com/scalameta/scalameta/blob/16fda03f70671db1c822b329beef2ac0dbb0dbdb/scalameta/semantic/shared/src/main/protobuf/semanticdb.proto#L7). The path name to the original source is inferred from the path to the `.semanticdb` file. This makes it problematic to combine multiple `.semanticdb` files into a single fat `.semanticdb` file.

I propose that we add a `path` field to Attributes and stop giving meaning to the filename of the `.semanticdb` files.

xeno-by:
I don't think we can stop giving meaning to relative paths of `.semanticdb` files, because of how incremental compilation works. 

However, I agree with adding `s.Attributes.name` in order to better support fat dbs. Then, I guess, we should also change `s.Database` to say `entries: Seq[Attributes]` instead of `entries: Seq[(RelativePath, Attributes)]`.

olafurpg:
> then, I guess, we should also change s.Database to say entries: Seq[Attributes] instead of entries: Seq[(RelativePath, Attributes)].

Correct. Does this merit postponing v1.8 a wee bit longer? I can fix this tomorrow morning.

xeno-by:
I think we can postpone, yeah

PR #868, opened 2017-05-15
Add support for Database.load(Array[Byte]) constructor

olafurpg:
This PR adds better support to load a Mirror from a single fat semanticdb, for example in Scala.js. While implementing this feature, I stumbled on several issues in the current implementation which I took liberty to fix along the way

- Multpath.value is now List[AbsolutePath]
- File IO like jar in Multipath has been refactored out into Platform specific code or is guarded against a `isJVM` boolean literal to please the Scala.js dead code eliminator.
- Support to unit tests against `Mirror` in both Scala.js and JVM


olafurpg:
This PR is almost ready for review. Still need to fix the JS tests, which I will try look at tomorrow along with #867 .

olafurpg:
This PR is ready for review now @xeno-by

olafurpg:
This PR is ready for review now :) 

The s/Sourcepath/AbsolutePath/ change is probably the most controversial. If you don't like that change, then I would like to see a test case where we need Sourcepath. Relativization of filenames becomes much simpler when using a single `sourceroot: AbsolutePath` instead of `Sourcepath`.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/package.scala
```diff
@@ -44,15 +44,13 @@ package object semantic {
           }
           val (spath, scontents) = minput match {
             case mInput.File(path, charset) if charset == Charset.forName("UTF-8") =>
-              val spath = sourcepath.relativize(path.toURI)
-              spath.getOrElse(sys.error(s"bad database: can't find $path in $sourcepath")) -> ""
+             path.toRelative(sourceroot).toString -> ""
```
xeno-by: The indentation is a bit weird

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/Entry.scala
```diff
@@ -3,10 +3,13 @@ package scala.meta.internal.semantic.vfs
...
 import org.scalameta.adt._
 import scala.collection.immutable.Seq
+import scala.meta.internal.io.FileIO
 import scala.meta.io._
 import scala.meta.internal.io.InputStreamIO
+import scala.meta.internal.platform
 
-@root trait Entry {
+@root
+trait Entry {
```
xeno-by: I'd prefer to have this and the like on a single line

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
```diff
@@ -12,6 +12,8 @@ import scala.meta.internal.semantic.{vfs => v}
 import scala.meta.internal.semantic.{schema => s}
 import scala.meta.{semantic => m}
 
+import org.scalameta.logger
```
xeno-by: Let's remove this?

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
```diff
@@ -36,4 +38,10 @@ object Database {
   def load(classpath: Classpath, sourcepath: Sourcepath): Database = {
```
xeno-by: Why does this compile?

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
```diff
@@ -36,4 +38,10 @@ object Database {
...
     v.Database.load(classpath).toSchema.toMeta(sourcepath)
   }
+  def load(bytes: Array[Byte]): Database = {
+    val sattrs = s.Attributes.parseFrom(bytes)
+    val sdb = new s.Database(Seq(sattrs))
+    val mdb = sdb.toMeta(Sourcepath.workingDirectory)
```
xeno-by: I don't like this. How about we just crash if we encounter a slim database?

olafurpg: Changed to throw an exception if trying to load a slimdb from Array[Byte]

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
olafurpg: I have removed Sourcepath.

xeno-by:
Do we consider that #867 is fixed by this pull request?

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
```diff
@@ -20,8 +20,8 @@ import scala.meta.{semantic => m}
   lazy val denotations: Map[Symbol, Denotation] = entries.flatMap(_._2.denotations).toMap
   lazy val sugars: Map[Position, String] = entries.flatMap(_._2.sugars).toMap
 
-  def save(classpath: Classpath, sourcepath: Sourcepath): Unit = {
-    this.toSchema(sourcepath).toVfs(classpath).save()
+  def save(classpath: Classpath, sourceroot: AbsolutePath): Unit = {
```
xeno-by: `classpath` here should be `AbsolutePath` instead. This fixes another ugliness in the design that I disliked. Thank you!

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Database.scala
```diff
@@ -34,6 +34,12 @@ import scala.meta.{semantic => m}
 
 object Database {
   def load(classpath: Classpath, sourcepath: Sourcepath): Database = {
-    v.Database.load(classpath).toSchema.toMeta(sourcepath)
+    v.Database.load(classpath).toSchema.toMeta(Some(sourcepath))
+  }
```
xeno-by: How about we add another overload that takes `(classpath: Classpath)`, i.e operates exactly like `Array[Byte]`, but without its users having to know things like `META-INF`?

olafurpg:
Seems v1.8 is ready for release then 🎉

Issue #873, opened 2017-05-19
Scala syntax warts/bugs


xeno-by:
### List of Scala syntax / Scala parser warts discovered so far

  1. Function parameters can not be annotated
  1. `try expr catch expr` syntax
  1. Package objects can't have annotations
  1. Lambdas can not be disambiguated from self types in templates
  1. Patterns in vals are parsed inconsistently
  1. Awkward meaning of infix patterns: `a infix (b, c)` means `infix(a, b, c)`
  1. `pq"_: F[_]"` is something completely different from `pq"_: F[_ >: lo <: hi]"`
  1. `pq"a[b]"` is a legal pattern
  1. Guess what `pq"a -1"` means
  1. No way to fully qualify things that are in empty package
  1. Vars with default values may not contain patterns
  1. Constr block
  1. `q"def x { case 1 => 2 }"`
  1. `q"trait F[T] <: Sobaka[T] with Balabaka { def bazinga } "`
  1. `*` is ambigious (ident vs seq wildcard) in most contexts
  1. Floating scala doc
  1. `(x = 2) + (y = 3)` here x = 2 is assign but y = 3 is named argument
  1. `x + ()` is `x.+()` but not `x.+(())`
  1. `A.super[B]` could very well be allow to be an expression on its own
  1. Anonymous top-level type parameters
  1. `import M.{_ => _}`

## Syntax spec bugs

  1. Annotations can have multiple argument lists
  1. Infix expressions can have type arguments
  1. Right hand side of infix expression should be ::= SimpleExpr | ArgumentExprs
  1. Whitespace separators aren't documented well. Sometimes you can have things separated by newlines, sometimes you can't.

xeno-by:
This is non-actionable, since we don't have control over Scala syntax.

PR #907, opened 2017-05-31
Massive cleanup of trees

xeno-by:
We've been holding off refactorings of our trees for years, and finally the time has come. Inspired by [liufengyun/gestalt](https://github.com/liufengyun/gestalt) and based on my work on [new-style def macros](https://github.com/scalamacros/scalamacros/pull/1), I propose that we change our trees and supporting data structures to make user experience better.


xeno-by:
@olafurpg If we're doing all these breaking changes, how about we get rid of `Term.Placeholder`? We'll just silently create synthetic `Term.Function` nodes. I've given it a bit of thought now, and everything seems to work out name-wise and token-wise.

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -409,9 +409,16 @@ object Enumerator {
 @branch trait Importee extends Tree with Ref
 object Importee {
   @ast class Wildcard() extends Importee
-  @ast class Name(name: scala.meta.Name.Indeterminate) extends Importee
-  @ast class Rename(name: scala.meta.Name.Indeterminate, rename: scala.meta.Name.Indeterminate) extends Importee
-  @ast class Unimport(name: scala.meta.Name.Indeterminate) extends Importee
+  @ast class Name(name: scala.meta.Name) extends Importee {
```
olafurpg: Additionally, can we expose a `Name(): Name = Name.Anonymous()` constructor? Structure should then also print out `Name()` instead of `Name.Anonymous()`.

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -96,7 +92,8 @@ object Term {
     checkFields(enums.head.is[Enumerator.Generator])
   }
   @ast class ForYield(enums: List[Enumerator] @nonEmpty, body: Term) extends Term
-  @ast class New(templ: Template) extends Term
+  @ast class New(init: Init) extends Term
+  @ast class NewAnonymous(templ: Template) extends Term
```
olafurpg: How about `NewRefinement`? That would be consistent with the language spec https://www.scala-lang.org/files/archive/spec/2.11/03-types.html#compound-types

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -329,53 +323,38 @@ object Pkg {
...
-  }
-  def fresh(): Ctor.Name = fresh("fresh")
-  def fresh(prefix: String): Ctor.Name = Ctor.Name(prefix + Fresh.nextId())
+                       init: Init,
+                       stats: List[Stat]) extends Ctor with Stat {
+    checkFields(stats.forall(_.isBlockStat))
+  }
+}
+
+@ast class Init(tpe: Type, name: Name, argss: List[List[Term]]) extends Ref {
```
olafurpg: Amazing, I like `Init` a lot.

Can we expose a custom constructor `Init(Type) = Init(Type, Name(), Nil)`? Init.structure should likewise print that syntax if name is anonymous and argss.isEmpty

olafurpg: Additionally, it might be worth adding a comment highlighting difference between Init and Term.Apply. Why are both `tpe: Type` and `name: Name` necessary? According to my best understanding, it's necessary for the semantic api so that Ref.symbol references the correct type + ctor symbols, unlike regular Term.Apply

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -403,11 +403,6 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
   def acceptOpt[T <: Token : TokenInfo]: Unit =
     if (token.is[T]) next()
 
-  /** {{{
```
olafurpg: Thanks for removing these, esp. if they're out of sync with the actual spec.

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/XmlParser.scala
```diff
@@ -123,15 +123,13 @@ class XmlParser(Block: P0,
   }
 }
 
-case class RangePosition(from: Int, to: Int)
```
olafurpg: I personally dislike `(Int, Int)` tuples, it's not clear what each field does. I would prefer to give it a more precise name like `XmlTokenRange`, or reuse Position.Range. It adds mental cycles to see what `List.newBuilder[(Int, Int)]` means.

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -408,7 +408,7 @@ object Importee {
   // checkFields(stats.forall(_.isTopLevelStat))
 }
 
-package internal.ast {
+package internal.trees {
```
olafurpg: 👍 I've bumped into this inconsistency before.

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -348,7 +348,7 @@ object Ctor {
 
 @branch trait Mod extends Tree
 object Mod {
-  @ast class Annot(body: Init) extends Mod
+  @ast class Annot(init: Init) extends Mod
```
olafurpg: Can we add a `@deprecated def body = init`? I'd like to migrate this automatically with scalafix.

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -333,9 +333,9 @@ object Ctor {
 @ast class Template(early: List[Stat],
                     inits: List[Init],
                     self: Term.Param,
-                    stats: Option[List[Stat]]) extends Tree {
+                    stats: List[Stat]) extends Tree {
```
olafurpg: On one hand, I am excited to see this complication gone. With Init and Term.NewAnonymous (Term.NewRefinement is better btw!), it is no longer necessary to distinguish between `object a {}` and `object a`. cc/ @DavidDudson

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -330,9 +330,11 @@ object Ctor {
   checkParent(ParentChecks.Init)
 }
 
+@ast class Self(name: Term.Name, decltpe: Option[Type]) extends Member
+
 @ast class Template(early: List[Stat],
                     inits: List[Init],
-                    self: Term.Param,
+                    self: Self,
```
olafurpg: Interestingly, the old design didn't guard against non-sensical trees like 

```scala
@ Template(Nil, Nil, Term.Param(List(mod"lazy"), Term.Name("b"), Some(Type.Name("X")), Some(q"2")), Some(Nil))
res7: Template = { lazy b: X = 2 => }
```

This is a nice improvement.

olafurpg:
## s/immutable.Seq/List/

I agree immutable.Seq has always been annoying, and I've seen many beginners
hit on this. However, I feel like we are rushing too quickly to use List. I'd
like to see a closer comparison why `List` is better than `scala.Seq`, if only
just to document why we don't want scala.Seq at all. 

- scala.Seq remains source and binary compatible with immutable.Seq. It solves
  the only sore point in the current design, which is the need for `import
  s.c.i.Seq`
  - switching to List is quite a breaking change, although if we go that route
    I'm sure we can lazily fixable with with a deprecated implicit conversion +
    scalafix rewrite
- scala.Seq retains flexibility to plug in more memory/cpu efficient
  *immutable* collections, if necessary. We don't know what applications
  scalameta will be used for in the future, just like it wasn't known 2 years
  ago how Scalameta would look like today.
- scala.Seq retains ability to implement Tree with List by default (quasiquotes/parser)
- trees built with immutable collections cannot become mutable
- `Tree.==` is by reference anyways

By using scala.Seq instead of List we risk some problems if someone explicitly
takes advantage of `mutable.Seq <: scala.Seq`

- invariant errors, for example
  - `Tree.parent` will return bogus results if we append mutable seqs 
  - `Tree.tokens` may behave weirdly
- defensive copying in utility methods, we can't be sure whether a tree is
  immutable or not. Maybe we can expose an `.immutable: Tree` method that creates
  a new immutable tree to prevent that (please don't shoot me for suggesting
  this!)
- fun debugging sessions with mutable trees

Observe that the downsides cannot happen to trees generated by quasiquotes or
the parser because those will always be backed by immutable collections
anyways.

olafurpg:
## Unify Term.Assign and Term.Update

I am not a fan of this change. Term.Assign is used for reassigning vars while
the other is syntax sugar for calling `.update`. I like Term.Update because it
matches `def update`. I have not experienced confusion between the two, can you
please motivate this better?

olafurpg:
## s/Pat.Var.Type/Type.Var/

I'm not sold on the `Type.Var` name. Why?
- there's not Term.Var
- it doesn't seem to be related to `Defn.Var/Decl.Var`

If the spec calls the "Type variable", how about we call them `Type.Variable`?
It's more verbose, but type variables are not so common in my experience.

olafurpg:
## s/Pat.Var.Term/Pat or Pat.Var?/

In some places, Pat.Var.Term is replaced with Pat.Var and in others it's Pat.
If scalafix encounters a Pat.Var.Term (in term/type/pat position), what should it rewrit it to?

olafurpg:
## s/RangePosition/XmlTokenRange/

Please lets try to avoid tuples when we can give things reasonable names.
I agree RangePosition was a bit generic, how about XmlTokenRange?

olafurpg:
## s/Point//

I am not sold by this change, and it feels out of place in this PR.  I can't
say that I have experienced much pain from Point besides the verbose interface
to manually construct a Position.Range (which I only done once, I believe).  I
like the fact that `Position.start` forces me to think if I mean the
column/line/offset, even if it's unwieldy. If you are not feel strongly about this
change I propose we move the discussion to an issue or separate PR.

olafurpg:
Adding @DavidDudson and @mutcianm to the reviewer list, they may have more insights on pain points in the current AST design.

olafurpg:
While reviewing, I took note of most of the changes I believe we can automate with scalafix

```
s/TryWithCases/Try/
s/TryWithTerm/TryWithHandler/
s/Term.Arg.Named/Term.Assign/
s/Term.Arg/Term/
s/Type.Arg.ByName/Type.ByName/
s/Type.Arg.Repeated/Type.Repeated/
s/Pat.Arg.SeqWildcard/Pat.SeqWildcard/
s/Type.Arg#/Type/
s/Param.Name/Name/
s/Pat.Type/Type.Var/
s/internal.ast/internal.trees/
s/Mod.Annot.body/Mod.Annot.init/
s/Template.stats.*/Template.stats/
s/Lit.Unit(())/Lit.Unit()/
s/Lit.Null(null)/Lit.Null()/
s/Template(_, _, Term.Param(...))/Template(_, _, Self(...))/
```
Adding them here for documentation purposes. Ideally, we can host this rewrite in the scalameta repo and support the instructions "run `scalafix scalameta/scalameta/v1.9`"

One question, when encountering `Pat.Var.Term`, should it be rewritten to `Pat` or `Pat.Var`?

notes/syntaxwarts.md
```diff
@@ -1,30 +0,0 @@
-### List of Scala syntax / Scala parser warts discovered so far
```
DavidDudson: Is this file reporduced somewhere? Is it not useful to keep it around?

DavidDudson:
@olafurpg I personally prefer list, at $Job we made a similar desicion, except out language is an antlr4 grammar that is then converted into a scala ast.

Heres my opinion.

List is extremely efficient when constructing and deconstructing from the head, or traversing.

99.99% of the time those are the operations you perform on a the lists in an AST.

Exceptions may be...

- Partitioning lists to parallize work
- Accessing via index

Although I've never had to do either of these things.

Also `List` is very efficient for a small number of elements. 
`Vector` on the other hand is not.

I would say most lists we create in Tree's are sub 16 elements.
I doubt many go over 100.
I always reference http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#lists-vs-vectors.

There are times, like return a flattened list of all descendants, where we should be returning Vector and not List. But I think List is more beneficial as a general collection type.

However, I would like to see some benchmarks run before this commit is merged, both with memory and performance.

The other benefit of `List` is that you can copy paste `.syntax` without issue

notes/syntaxwarts.md
olafurpg: Here https://github.com/scalameta/scalameta/issues/873

olafurpg:
@xeno-by I merged the latest changes in master with your PR here https://github.com/olafurpg/scalameta/commit/b4fa6ab5932673bd7a6ca5436bcdfee6441da241 you can fast-forward olafurpg/1.9

- the EqualProps tests in contribJVM/test are failing because of `val Name.Term("_") = Name.Anonymous()` matches and `Term.Name("_")` throws error. How about making `Term.Name("_")` instantiate `Name.Anonmous()`?
- I'm getting assertion errors when running `benchmarks/test`. see stack trace https://gist.github.com/olafurpg/15ef751aa62ac09d056ee1fef0b75bb3 It seems something is off in the parser/tokenizer

olafurpg:
I used my super-power merge rights to push my 1.9 branch onto your fork. Let me know if I should not do that, I just wanted to check if it was possible and thought it would be good to update the CI logs.

Please take a look at the failing test cases in EqualProps and BenchmarkOK

olafurpg:
Weird, the CI is unable to reproduce the errors in EqualProps. The errors I get locally are 

```scala
Error: java.lang.UnsupportedOperationException java.lang.UnsupportedOperationException: Invalid transformation of Self.name: Name.Anonymous -> Type.Name. From: Name.Anonymous(), to: Type.Name("_")
scala.meta.transversers.Transformer.fail(Transformer.scala:4)
scala.meta.transversers.Transformer.apply(Transformer.scala:4)
scala.meta.transversers.Api$XtensionCollectionLikeUI$transformer$2$.apply(Api.scala:10)
scala.meta.transversers.Transformer.apply(Transformer.scala:4)
scala.meta.transversers.Api$XtensionCollectionLikeUI$transformer$2$.apply(Api.scala:10)
scala.meta.transversers.Transformer.apply(Transformer.scala:4)
```

xeno-by:
Hi guys! Thanks a lot for your thoughtful feedback. Reviewing this pull request must've been a lot of time, and I really appreciate that you've taken this time. I'm currently preparing for my talk tomorrow and catching up at work, so I'll be addressing your comments later this week.

@olafurpg Special thanks for doing immediate fixups!

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/XmlParser.scala
ShaneDelmore: Enough programmers treat ranges as inclusive -> inclusive that I also prefer Range so I don't get surprised on Range.Exclusive vs Range.Inclusive.

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
xeno-by: Done

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
xeno-by: I like this idea. However, instead of `Name()`, I'll do `Name("_")` in order to be symmetric with pattern matching.

xeno-by: Actually, I have an even better idea imo. How about we merge `Name.Anonymous` and `Name.Indeterminate` into just `Name`?! I'll prototype this today or tomorrow and will report back with results.

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
xeno-by: Is there "on the other hand"?

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -75,31 +75,31 @@ object Term {
   @ast class Tuple(args: Seq[Term] @nonEmpty) extends Term {
     // tuple must have more than one element
     // however, this element may be Quasi with "hidden" list of elements inside
-    require(args.length > 1 || (args.length == 1 && args.head.is[scala.meta.internal.ast.Quasi]))
+    checkFields(args.length > 1 || (args.length == 1 && args.head.is[scala.meta.internal.ast.Quasi]))
```
xeno-by: Done.

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
xeno-by: I'd be in favor of `NewAnonymous`, since [SLS 6.10](https://www.scala-lang.org/files/archive/spec/2.11/06-expressions.html#instance-creation-expressions) says that this is instance creation of an anonymous class. 

I don't think that [SLS 3.2.7](https://www.scala-lang.org/files/archive/spec/2.11/03-types.html#compound-types) is applicable here, because refinements are a type-level concept, and `new t` from SLS 6.10 is operating not on types, but on templates.

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
xeno-by: 1) See https://github.com/scalameta/scalameta/issues/972.

2) Your understanding is correct.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -18,7 +18,7 @@ import org.scalameta.{unreachable, debug}
 import scala.compat.Platform.EOL
 
 object TreeSyntax {
```
xeno-by: Our current prettyprinter is outrageously inadequate - it is both slow and buggy - so I agree that it should be rewritten from scratch. However, I'm not sure about the exact strategy.

My concern about using `typelevel/paiges` is that the 3-5x slowdown advertised in the readme is not really competitive (even though the readme claims the opposite). It would be a shame to invest significant time into a rewrite only to discover that the new version is inadequate, too.

@olafurpg @DavidDudson What do you think?

/cc @densh, since he seems to have experimented with prettyprinters a lot with native. Denys, I see that you're currently using a dumb imperative prettyprinter. Are you happy with it? Do you plan to change it in the future?

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/XmlParser.scala
xeno-by: Done.

xeno-by:
> s/immutable.Seq/List/

Having read your arguments, I think I still prefer `List`, since it's the most prescriptive option. `scala.Seq` is also a good one, because it provides flexibility, but until this flexibility is actually needed, I'd say we stick with `List`. In the past, most of our future-proofing attempts have failed (hence my spree of "kill wishful thinking" commits), so I'd be wary of applying this kind of design reasoning. /cc @densh 

> Unify Term.Assign and Term.Update

@olafurpg I unified these nodes simply because it seemed like a way to get rid of code - no other hidden agenda. In our chat in Copenhagen after the review of this PR, you mentioned that you had changed your mind and were fine with this unification. What do you think now?

> s/Pat.Var.Type/Type.Var/

SLS 8.2 says that those guys are indeed type variables. It is unsatisfying that there are no term variables (well, actually, there are, but they are hidden in typer), but I'd say that this irregularity is on SLS, not on us.

While I agree that `Var` can be confused with `Decl/Defn.Var`, consider that there's also `Pat.Var`. If we were to rename `Type.Var` to `Type.Variable`, we would also have to rename `Pat.Var` to `Pat.Variable` to stay consistent. However, `Pat.Var` is a very frequent AST node, so there is a non-trivial downside to this renaming. What do you think about this trade-off?

> s/Pat.Var.Term/Pat or Pat.Var?/

There is no easy answer here. On the one hand, `Pat.Var` means exactly the same as what `Pat.Var.Term` meant previously. On the other hand, some uncommon type signatures were relaxed, so some uncommon code that previously compiled with `Pat.Var.Term` won't compile now. I'd say we go for `Pat.Var` and ask users to fix their code if it doesn't work.

> s/RangePosition/XmlTokenRange/

Fixed.

> s/Point//

I feel this is a significant improvement. I think that the new API is easier to construct and to deconstruct, as well as to explain to newcomers (now I can just say use `pos.start` or `pos.end`). I suggest we keep this.

xeno-by:
@olafurpg @DavidDudson Here's the current status of the pull request:
  * [x] Rebased against the current master.
  * [ ] `Name.Anonymous` is still broken.
  * [x] All other code comments are addressed.
  * [x] All regular comments are addressed.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
olafurpg: I think you are misunderstanding the 3-5x slowdown as being caused by non-imperativeness. The [underlying implementation](https://github.com/typelevel/paiges/blob/78777cbc6d7f95fd909141d7fe7a7346bd2a19fb/core/src/main/scala/org/typelevel/paiges/Chunk.scala) is imperative. It's ~4x slower than `.mkString` because it implements line wrapping, which is non-trivial and highly valuable IMO for pretty-printers.

The complexity in TreeSyntax is partly caused it trying to implement line-wrapping in some places. For example, see `s(stats.head).toString.contains(EOL)` [here](https://github.com/scalameta/scalameta/blob/d0804413ab1845fc9603ef73282bed1c0616970d/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala#L492). I suspect a paiges based line wrapper might be an order of magnitude faster than the current TreeSyntax, and even if TreeSyntax is terribly slow I think performance is not its biggest issue.

olafurpg: btw, this output here https://gist.github.com/olafurpg/44ee2d0a8c5b8330702cf00d23d0978a which was pretty-printing by `.structure` output rendered in ~300ms. It took me 10 mins to implement that pretty printer. Scalafmt would choke on that input, despite months of optimization efforts.

xeno-by:
Finally, all green! Merging.

PR #916, opened 2017-06-04
Add new logo to website.

olafurpg:
I made some unsuccessful attempts to put the logo next to the
"Scalameta" top header. Instead, the logo is now center before the
first paragraph. Ideally, the default scalatex page should support
some "logo" setting.

screen shot 2017-06-04 at 18 51 59


xeno-by:
@lihaoyi Could you provide advice about this? It would be great to have the logo in the header.

lihaoyi:
@xeno-by Scalatex is just HTML, so you can do whatever you want with it. Here's what I use to center some images in the Ammonite readme

```scala
@img(src:="Editing.gif", display.block, marginLeft.auto, marginRight.auto, height := 172)
```

I don't think it makes sense for Scalatex to support "logo" by default; that sounds variable enough that different people will want different things, and anyway is basically trivial to add yourself (as above) unless there's some special handling or behavior it needs that I'm not seeing

olafurpg:
@lihaoyi Thank you for the help. I was trying to accomplish something like this

screen shot 2017-06-05 at 09 16 23

but I struggled to get it to work with the top-level `@sect(Scalameta, version)` without customizing scalatex.Main.

lihaoyi:
@olafurpg one option is to `position: absolute` it, the way I tack on the "fork me on github" ribbons e.g. here http://www.lihaoyi.com/Ammonite/

olafurpg:
@lihaoyi 
I tried `position: absolute` but it doesn't seem to play nicely with different screen sizes

screen shot 2017-06-05 at 09 28 12


I'm now experimenting with extending `Section.headers` https://github.com/lihaoyi/Scalatex/blob/abf077fab654e2954997aed86c7cf75c75629320/site/src/main/scala/scalatex/site/Section.scala#L80

lihaoyi:
Hmm yeah inheriting from `Site` and overriding `Site#section#headers` seems like the correct thing to do https://github.com/lihaoyi/Scalatex/blob/master/site/src/main/scala/scalatex/site/Section.scala#L80-L88

xeno-by:
Moved to https://github.com/scalameta/scalameta/issues/969.

PR #917, opened 2017-06-04
Upgrade to Scala.js 0.6.17 for faster link times.

olafurpg:
My experiments show a drop from ~17s to ~3s for scalametaJS/test, see
https://github.com/scala-js/scala-js/issues/2827#issuecomment-306063974

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,4 +1,4 @@
-addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.16")
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.17")
 
 addSbtPlugin("me.lessis" % "bintray-sbt" % "0.3.0")
 


xeno-by:
@jvican @olafurpg Looks like another spurious error caused by SBT resolution (https://platform-ci.scala-lang.org/scalameta/scalameta/341/2). What can we do about this?

project/plugins.sbt
```diff
@@ -1,4 +1,4 @@
-addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.16")
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.17")
```
xeno-by: LGTM

olafurpg:
I think the server's bandwidth is being limited. Fabien setup an artifactory recently that we can proxy through to limit bandwith. I'll try to configure our docker image to use that this week.

PS. I restarted the build but if failed again.

olafurpg:
I opened a PR to setup travis CI https://github.com/scalameta/scalameta/pull/920 I don't want to waste more time on CI resolution/caching problems.

xeno-by:
@olafurpg I suggest we merge this as is.

olafurpg:
Agreed, I have changed the CI to Travis in #921

PR #920, opened 2017-06-05
Setup Travis CI.

olafurpg:
We are experiencing resolution errors with the platform-ci, which is
slowing us down. My hypothesis is that our bandwith is being limited,
partly due to bad caching setup. The dotty-ci is apparently downloading
50gb a day. Instead of spending more days on the drone CI, I'm setting
up travis again. I disable the ci-slow tests since those are too heavy
for the free-tier Travis servers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,36 @@
+language: scala
+script:
+  - ./bin/runci.sh $CI_TEST
+jdk:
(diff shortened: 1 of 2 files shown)


olafurpg:
To unblock other PRs, I'm gonna go ahead and merge this and disable the drone CI.

PR #933, opened 2017-06-09
minor changes to publishing

xeno-by:


xeno-by:
@olafurpg We need this internally at Twitter, so please take a quick look on Monday. Otherwise, I'll just merge, and we'll figure out the details later.

PR #939, opened 2017-06-12
Emit denotations for all used symbols, fixes #930.

### LINKED ISSUE #930: Emit denotations for all used symbols
Currently, Scalahost only emits denotations for symbols defined in the compilation unit. However, in some cases it might be useful to have denotations for all symbols in the Names section. For example, for the NoAutoTupling rewrite https://scalacenter.github.io/scalafix/#NoAutoTupling (see https://github.com/scalacenter/scalafix/issues/147) would benefit from this as well as the new eta expansion changes in Dotty that require explicit parentheses for nullary method application.

Question is whether this should be enabled by default or not, it does add some overhead. However, I think we should not worry about the overhead until performance becomes an issue, and then we cut out the bottleneck first.

olafurpg:


scalahost/nsc/src/main/scala/scala/meta/internal/semantic/AttributesOps.scala
```diff
@@ -147,7 +147,7 @@ trait AttributesOps { self: DatabaseOps =>
                 if (symbol == m.Symbol.None) return
 
                 names(mtree.pos) = symbol
-                if (mtree.isBinder) {
+                if (config.semanticdb.isFat || mtree.isBinder) {
```
xeno-by: Why restrict this just to fat semantic databases? I can imagine slim databases also wanting this functionality.

olafurpg: Fair enough, let's do this for all dbs.

xeno-by:
LGTM

PR #940, opened 2017-06-12
Add Message.kind field to label message categories.

### LINKED ISSUE #924: Classify/label reported messages with unique ID
We currently match against the "Unused import" string in several places

- https://github.com/scalacenter/scalafix/blob/8641796376b924d2d8719cfdc9c7c4975719569e/scalafix-core/src/main/scala/scalafix/rewrite/RemoveUnusedImports.scala#L11
- https://github.com/scalameta/scalameta/blob/a3b8b1f8c7fc7b75d6ea0663e5f3a39555fa574e/scalahost/nsc/src/main/scala/scala/meta/internal/semantic/AttributesOps.scala#L341

This approach has some problems
- not portable, what if the message is updated on the next Scala version? Is it "Unused import" in Dotty?
- it makes it hard to customize scalahost's behavior towards particular messages. For example, fatal warnings will all warnings into errors, even warnings that we can automatically with scalafix (like unused imports). It would be really nice if users could configure scalahost to collect unused import messages without emitting it to the console. Maybe https://github.com/scala/scala-dev/issues/333 can improve the situation.

Ideally, this should be baked into the Scala compiler. However, even if it does get into the compiler then we won't have it until 2.12.3 or 2.13 earliest and we are still not guaranteed that the IDs remain consistent across scalac/dotc.

I propose that we add a new `enum Id` field to `Message` in the semanticdb schema https://github.com/scalameta/scalameta/blob/master/scalameta/semantic/shared/src/main/protobuf/semanticdb.proto#L28 and then parse messages inside ScalahostReporter.

olafurpg:
Fixes #924.

### DIFF
--- a/scalahost/nsc/src/main/scala/scala/meta/internal/semantic/AttributesOps.scala
+++ b/scalahost/nsc/src/main/scala/scala/meta/internal/semantic/AttributesOps.scala
@@ -334,11 +334,15 @@ trait AttributesOps { self: DatabaseOps =>
 
         val messages = unit.hijackedMessages.map {
           case (gpos, gseverity, msg) =>
+            val mkind = msg match {
+              case "Unused import" => m.Message.Kind.UnusedImport
+              case _ => m.Message.Kind.None
+            }
             val mpos = {
               // NOTE: The caret in unused import warnings points to Importee.pos, but
               // the message position start/end point to the enclosing Import.pos.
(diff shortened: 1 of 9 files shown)


olafurpg:
I'm not too happy with the boilerplate needed to add a new category here. Also,

- the protobuf schema will quickly explode with every new message kind
- the scalameta public API will grow with every new message kind

I think this issue requires more thinking.

PR #942, opened 2017-06-14
Fix #941.

### LINKED ISSUE #941: Parser error: duplicate access qualifier with "private[foo] protected var bar = false"

usrinivasan:


scalameta/scalameta/shared/src/test/scala/scala/meta/tests/parsers/ModSuite.scala
```diff
@@ -367,14 +367,12 @@ class ModSuite extends ParseSuite {
 
   test("invalid private and protected") {
     interceptParseErrors(
-      "private protected class A",
-      "protected private class A",
-      "private[something] protected class A",
-      "protected private[something] class A",
-      "protected[something] private class A",
-      "private protected[something] class A",
```
xeno-by: Turns out that `private[foo] protected` and friends are actually valid Scala code. We actually encountered that in our codebase.

xeno-by:
See https://github.com/usrinivasan/scalameta/pull/1 for my feedback.

xeno-by:
LGTM

Issue #945, opened 2017-06-14
Scalameta does not properly tokenise XML literal when followed by `{`


allanrenucci:
```scala
@ "<a/>{}".tokenize.get 
res2: Tokens = Tokens(
  BOF,
  Xml.Start(0, 0),
  Xml.Part(0, 4, "<a/>"),
  Xml.SpliceStart(4, 4), // Should be Xml.End
  LeftBrace(4),
  RightBrace(5),
  Xml.SpliceEnd(6, 6),
  End(6, 6),
  EOF
)
```

PR #946, opened 2017-06-14
Expose issue #945

allanrenucci:
This is not a fix for #945. I don't see an easy solution.
Now Scalameta will crash instead of emitting wrong tokens


xeno-by:
I'd say crashing is definitely better.

allanrenucci:
The issue is that in the case of `<a/>{0}` when we see the opening brace, we don't know if it is a legitimate splice or if we reached the end of the xml literal.
A solution would be to introduce new tokens in the `LegacyScanner` (i.e. tokens for splice start and splice end)

xeno-by:
LGTM

PR #947, opened 2017-06-15
Wrap Pat.Var.Term names in backticks, fixes #931.

### LINKED ISSUE #931: .syntax method returns unparseable code
```
Welcome to the Ammonite Repl 0.9.5
(Scala 2.12.2 Java 1.8.0_131)
If you like Ammonite, please support our development at www.patreon.com/lihaoyi
@ import $ivy.`org.scalameta::scalameta:1.8.0` 
import $ivy.$                               
@ import scala.meta._  
import scala.meta._ 
@ q"val `a b` = 1".syntax.parse[Stat]  
res2: Parsed[Stat] = Error(
  Range(String("val a b = 1"), Offset(String("val a b = 1"), 8), Offset(String("val a b = 1"), 9)),
  "illegal start of simple pattern",
  ParseException(Range(String("val a b = 1"), Offset(String("val a b = 1"), 8), Offset(String("val a b = 1"), 9)), "illegal start of simple pattern")
)
```

This bug also breaks inline/meta macro annotations if there are encoded names in the returned `Tree`.

olafurpg:


xeno-by:
LGTM

Issue #948, opened 2017-06-16
Add productFields on Trees


gabro:
As discussed on gitter, it would be really cool to access all the fields of a tree.

We already have `productPrefix` which enumerates the values, but there's no way of enumerating the names of the fields.

Ultimately the idea would be to have enough pieces to expose a `toMap` that returns a `Map[String, Tree | Seq[Tree]]` (yeah, I know, no unions in Scala yet, but clarity). For example, given

```scala
1 + 2
```

the resulting tree would be

```scala
Term.ApplyInfix(
  lhs = Lit.Int(
    value = 1
  ),
  op = Term.Name(
    name = "+"
  ),
  targs = Seq(),
  args = Seq(
    Lit.Int(
      value = 2
    )
  )
)
```

then the `toMap` function applied to the top-level Tree would return

```scala
Map(
  "lhs" -> Lit.Int(1),
  "op" -> Term.Name("+"),
  "targs" -> Seq(),
  "args" -> Seq(Lit.Int(2))
)
```

With that, one could recursively apply `toMap` to the structure and add more metadata (e.g. the `type`), ending up with a serializable representation of a tree.

```scala
Map(
  "type" -> "Term.ApplyInfix",
  "lhs" -> Map(
     "type" -> "Lit.Int",
    "value" -> 1
  ),
  "op" -> Map(
    "type" -> "Term.Name",
    "name" -> "+"
  ),
  "targs" -> Seq(),
  "args" -> Seq(
    Map(
      "type": "Lit.Int",
      "value": 2
    )
  )
)
```

As I'm writing this, I think it would probably be cool to have a `serialize` method that does this, perhaps provided by `contrib`.

Thoughts?

DavidDudson:
We could also use this to generate prettyStructure from this.

xeno-by:
@gabro How has your prototype evolved? Have your requirements changed in the last few days?

gabro:
no, they're essentially the same. It would still be very useful to access the names of the fields of a Tree, so to zip them with the values.

xeno-by:
Would you consider contributing this functionality? I'm quite overwhelmed at the moment by the end-of-quarter stuff.

Here's where I'd add synthesis of the `productFields` method: https://github.com/scalameta/scalameta/blob/master/scalameta/common/shared/src/main/scala/scala/meta/internal/ast/ast.scala#L168. I'd obtain the labels for `productFields` from `fieldParamss`.

gabro:
[...]

PR #949, opened 2017-06-16
Add parsers JS facade

gabro:
This adds ScalaJS Facade to the parsers API, in order to produce a common js module that can parse Scala files.

The final polished output (without the `children` nesting) is still depending on #948.
Also there's currently no way of configuring the parser (I'll work on it later on) but a basic working infrastructure is already in place.

As discussed on gitter, having a JS parser enables integrations with browser-based solutions such as astexplorer.net

Here's a POC!

A few things to notice about this PR:

- since this facade won't be published together with scalameta, I've kept it as a separate project, so tests live in the same subprojects, as opposed to scalameta/tests

- these are my very first lines of ScalaJS, so I may be doing something stupid with the types. I fiddled with them quite a bit, and using native js types seemed to produce the best output, but if anyone wants to chime in, be my guest

- the js module needs to published on npm, but I've no clue on whether a blessed ScalaJS solution exists for that purpose, or if we have to do it manually (also, under which account?)


scalameta/parsersJSFacade/src/main/scala/scala/meta/parsersjsfacade/Api.scala
```diff
@@ -0,0 +1,59 @@
+package scala.meta
+package parsersjsfacade
+
+import scala.scalajs.js
+import js.JSConverters._
+import js.annotation._
+
+import parsers._
+
+object ScalaParser {
```
olafurpg: What do you think about putting this in the `parsers` module? We can keep it in the `js` directory.

scalameta/parsersJSFacade/src/test/scala/scala/meta/parsersjsfacade/ApiSuite.scala
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  test("Api") {
+    val code =
+      s"""|object Main {
+          |  def main(args: Array[String]): Unit =
+          |    println("Hello, World")
+          |}""".stripMargin
+    val parsed = ScalaParser.parse(code)
+    val expected = d(
```
olafurpg: Cool test!

scalameta/parsersJSFacade/src/test/scala/scala/meta/parsersjsfacade/ApiSuite.scala
```diff
@@ -0,0 +1,136 @@
...
+      "type" -> "Source",
```
olafurpg: This would definitely benefit from #948

scalameta/parsersJSFacade/src/main/scala/scala/meta/parsersjsfacade/Api.scala
```diff
@@ -0,0 +1,59 @@
...
+    val base = js.Dynamic.literal(
+      "type" -> t.productPrefix,
+      "children" -> t.children.map(toNode).toJSArray,
+      "pos" -> js.Dynamic.literal(
+        "start" -> t.pos.start.offset,
+        "end" -> t.pos.end.offset
+      )
+    )
+
+    // FIXME: this is really just for personal gratification
```
olafurpg: Let's link #948 and say this is a temporary measure to make this module immediately useful.

olafurpg: If there was no `.value` then it would be hard to make sense of the tree.

scalameta/parsersJSFacade/src/main/scala/scala/meta/parsersjsfacade/Api.scala
```diff
@@ -0,0 +1,59 @@
...
+    // ideally `.toMap` on Tree will take care of adding these
+    // attributes
+    val specific = t match {
+      case t: Lit => js.Dynamic.literal(
+        "value" -> t.value.toString
```
olafurpg: We may want to special case float and double since those store the value as a string https://github.com/scalameta/scalameta/blob/33c47b6fc9be8fb0bce60c17e4a4933379c23870/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala#L311-L313 and calling `.value` on them will actually loose precision in JS runtimes. This was the biggest breaking change necessary to port scalameta to Scala.js, before there was a single `class Lit(value: Any)`

gabro: I'm aware of it, I was admittedly lazy and I wanted to open this PR asap. I'll add the remaining cases tomorrow ;)

gabro:
@olafurpg I initially opened this PR thinking we would wait for #948, but I think we can merge this and then refactor the output. 

However (apart from addressing your review comments) I think we should at least decide how to publish on npm, so that I can start a PR to astexplorer too :)

scalameta/parsersJSFacade/src/main/scala/scala/meta/parsersjsfacade/Api.scala
gabro: intersting idea, do you think the module option of scalajs won't conflict with the existing project?

olafurpg: Could you explain the difference between the module kinds? What would it mean if we change the parsersJS module to be common-js?

olafurpg:
If we can reuse the parsers project and avoid the new project then that would be great. Then once Lit.Float/Double are handled then I think we are good to go. I have created a scalameta org on npm https://www.npmjs.com/org/scalameta/ I have never published to npm, I hope it's easier than publishing to sonatype :)

scalameta/parsersJSFacade/src/main/scala/scala/meta/parsersjsfacade/Api.scala
gabro: on a second thought, it shouldn't make a difference, since`@JSExport` is not used anywhere else in scalameta. The main difference is whether the export becomes a global variable or it is exported with the commonjs format, but again this shouldn't break anything.

scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
```diff
@@ -0,0 +1,80 @@
...
+
+    def v[A](a: A): js.Dynamic =
+      js.Dynamic.literal("value" -> a.asInstanceOf[js.Any])
+
+    // FIXME(gabro)
+    // This is to make the facade immediately useful by showing the trees' values
+    // Ideally `.toMap` on Tree will take care of adding the '.value' attribute.
+    // See https://github.com/scalameta/scalameta/issues/948
+    val value = t match {
+      case t: Lit.Int => v(t.value)
```
olafurpg: This was actually only necessary for Lit.Float and Lit.Double, the other ones can use `.value.toString`. However, I just realized we can drop all of these matches and use `.syntax` instead.

```scala
val value = t match { case _: Name | _: Lit => v(t.syntax); case _ => ...}
```

That should do the right thing for Lit.Float and Lit.Double.

scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
```diff
@@ -0,0 +1,80 @@
...
+      case t: Lit.Symbol => v(t.value.name)
+      case t: Name => v(t.value)
+      case _ => js.Dynamic.literal()
+    }
+
+    mergeJSObjects(base, value)
+  }
+
+  private[this] def parse[A <: Tree: Parse](code: String): js.Dictionary[Any] =
+    code.parse[A] match {
```
olafurpg: Note. This will Scala211 by default, which does not support trailing commas and some other features. We are actually planning to drop dialects for parsing/pretty-printing so this may not be a problem soon.

gabro: I'm aware of the default, and I'm planning to add some configuration soon.
I think it'd be interesting to switch from a dialect to another while exploring the AST.

scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
gabro: Would syntax be of the right type? I thought syntax always returned a string.

Also, in the match I'm handling some edge cases of the Scala <-> JS conversion, like Long and Symbol

scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
olafurpg: Different dialects will always produce the same AST, the difference lies in guards against usage of different features.

gabro:
I've added a basic `npm` project setup, including the instructions on how to publish (it's a manual process for now, lots of room for improvements there!).

We also have our first scalameta npm package online! https://www.npmjs.com/package/scalameta-parsers

scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
```diff
@@ -59,7 +61,12 @@ object JSFacade {
       case _ => js.Dynamic.literal()
     }
```
olafurpg: Actually, I just realized that `.productFields` will only solve what *keys* to put into the dictionary, not the values. 
I think we can remove the FIXME comment and claim this as a final solution. It might be good to handle the custom cases at the top and then fallback with a `case Lit(value) => value`, which will do the right thing for int/double/float/...

scalameta/parsers/js/src/test/scala/scala/meta/parsers/JSFacadeSuite.scala
```diff
@@ -36,7 +36,8 @@ class JSFacadeSuite extends FunSuite {
               "type" -> "Term.Name",
               "children" -> a(),
               "pos" -> pos(7, 11),
-              "value" -> "Main"
+              "value" -> "Main",
```
olafurpg: Does it make sense to have value+syntax for names? I feel like it only makes sense for literals.

scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
gabro: sure. By the way, I think we should ultimately have a "Tree.serialize" method in `contrib` that does all of these things.

scalameta/parsers/js/src/test/scala/scala/meta/parsers/JSFacadeSuite.scala
gabro: right, probably names should be spared.

gabro:
@olafurpg done!

scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
```diff
@@ -0,0 +1,83 @@
...
+    def v[A](a: A): js.Dynamic =
+      js.Dynamic.literal("value" -> a.asInstanceOf[js.Any])
+
+    val value = t match {
+      case t: Lit.Int => v(t.value)
+      case t: Lit.Double => v(t.value)
+      case t: Lit.Float => v(t.value)
+      case t: Lit.Byte => v(t.value)
+      case t: Lit.Short => v(t.value)
+      case t: Lit.Char => v(t.value.toString)
```
olafurpg: Can we move these special cases to the top and fallback with `case Lit(value) => v(value)`? That would make the three special cases stand out.

gabro: I can move the "special" cases at the top, but I don't think matching on `Lit(value)` will provide a reasonable fallback.

In all other cases (despite the code looking the same) matching on the specific node retains the precise type.

For instances matching on `Lit.Short` will make `t.value` have type `scala.Short` which in turn will result in `js.Number`.

If we treat them all the same, we'll end up with a value of `Any`, which is (rightfully) not a valid type for a `js.Dynamic` literal.

scalameta/parsers/js/npm/README.md
```diff
@@ -0,0 +1,9 @@
+# scalameta-parsers
+
+A JS API for the `parsers` module of [`scalameta`](https://github.com/scalameta/scalameta).
+
+It exposes two functions:
```
gabro: Since we're exporting a commonjs module, it needs a `require`, I'll add an example.

scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
gabro: nvm, it works as you suggested!

gabro:
@olafurpg 🎉 

Last-minute update, I've simplified the npm publishing process! :)

gabro:
rebased (it got out of sync with master)

xeno-by:
LGTM

olafurpg:
🎉

gabro:
🙌

Issue #885, opened 2017-05-23
Path needs more work


olafurpg:
```scala
@ internal.io.FileIO.listFiles(RelativePath("build.sbt").toAbsolute).toList
res3: List[AbsolutePath] = List()
@ internal.io.FileIO.listAllFilesRecursively(RelativePath("build.sbt").toAbsolute).toList
res3: List[AbsolutePath] = List(/Users/ollie/dev/scalafix/build.sbt/.)
@ AbsolutePath("/a") == AbsolutePath("/a/.")
res1: Boolean = false // missing `.normalize` method?
@ RelativePath("build.sbt").toAbsolute.exists // missing .exists method?
cmd5.sc:1: value exists is not a member of scala.meta.io.AbsolutePath
val res5 = RelativePath("build.sbt").toAbsolute.exists
                                                ^
Compilation Failed
```

olafurpg:
In #953, I refactored Absolute and Relative path to become thin wrappers around nio.file.Path. This means that all io operations on nio.file.Path are available with no conversion overhead.

PR #953, opened 2017-06-19
Refactor io module for proper cross-platform support

olafurpg:
The io module in 1.8 has a lot of problems. It does now work properly on Windows, AbsolutePath has very limited methods (e.g. no .resolveSibling, .getParent, .endsWith, .startsWith), it's very inefficient (no prefix sharing for AbsolutePaths in same directory). This PR refactors AbsolutePath and RelativePath to become small wrappers around java.nio.file.Path, addressing all of the limitations of the old Absolute/RelativePath. This PR also implements nio.Path and io.File in Scala.js using Node.js APIs.


olafurpg:
The tests seems to pass now on windows 👏 This PR still needs more work to clean up before code review, but in general I'm very happy with the changes and it addresses most of the issues I had with the current io module, for example #885

olafurpg:
This PR is ready for review @xeno-by

olafurpg:
Nice to see that the test from ae4cbd5 that was intended to fail on Windows did indeed fail on Windows
```scala
[info] - s.Attributes.filename has no Windows slash (\) *** FAILED ***
[info]   "scalahost\integration\src\main\scala\example\Example.scala" contained '\' (CrossPlatformSemanticSuite.scala:38)
```
I don't have a local windows computer to reproduce. b2be956 should fix the issue.

scalameta/io/js/src/main/scala/java/io/File.scala
```diff
@@ -0,0 +1,55 @@
...
+
+import java.net.URI
+import java.nio.file.Path
+import scala.meta.internal.io.FileIO
+import scala.meta.internal.io.JSFs
+import scala.meta.internal.io.JSPath
+import scala.meta.internal.io.JSStats
+import scala.meta.internal.io.NodeNIOPath
+import scala.meta.internal.io.PathIO
+import scala.meta.internal.io.PlatformFileIO
```
xeno-by: Why not simply say `import scala.meta.internal.io._`

scalameta/io/js/src/main/scala/java/io/File.scala
```diff
@@ -0,0 +1,55 @@
...
+
+// obtained implementation by experimentation on the JDK.
```
xeno-by: @sjrd @densh What are the legal implications of doing something like this?

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/package.scala
```diff
@@ -42,14 +42,15 @@ package object semantic {
...
-          val (spath, scontents) = minput match {
+          val (splatformpath, scontents) = minput match {
             case mInput.File(path, charset) if charset == Charset.forName("UTF-8") =>
               path.toRelative(sourceroot).toString -> ""
             case mInput.LabeledString(label, contents) =>
               label -> contents
             case other =>
               sys.error(s"bad database: unsupported input $other")
           }
+          val spath = PathIO.toUnix(splatformpath)
```
xeno-by: What will this return on Windows?

scalameta/io/js/src/main/scala/java/io/File.scala
sjrd: That's perfectly fine. It means someone reverse-engineered the spec by observing the behavior, and then reimplemented that spec. That's the very best incarnation of a clean room implementation.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/package.scala
olafurpg: A unix filename would have forward slashes /

xeno-by: How would `c:\foo.scala` be translated? Can we really say that `c:/foo.scala` is a unix path?

olafurpg: The path is relative at this point so it should not be a problem.

xeno-by:
LGTM

PR #954, opened 2017-06-20
Minor tweaks in testkit syntax analysis.

olafurpg:
I was running a few experiments on infix operators and hit on a few
bumps.


xeno-by:
LGTM

PR #955, opened 2017-06-21
Issue-#896 - Fix ScaladocParser not parsing multilevel headings

xavier-fernandez:
This commit changes the 'DocToken' AST, including the heading level on `Heading` directly. I should use another approach, or I need to add the change to the changelog?

### DIFF
--- a/scalameta/contrib/shared/src/main/scala/scala/meta/contrib/DocToken.scala
+++ b/scalameta/contrib/shared/src/main/scala/scala/meta/contrib/DocToken.scala
@@ -264,11 +264,6 @@ object DocToken {
   /**
     * Documents a Scaladoc Heading.
     */
-  case object Heading extends Kind
-
-  /**
-    * Documents a Scaladoc sub-heading.
-    */
-  case object SubHeading extends Kind
+  case class Heading(level: Int) extends Kind
 
 }
--- a/scalameta/contrib/shared/src/main/scala/scala/meta/contrib/DocToken.scala
+++ b/scalameta/contrib/shared/src/main/scala/scala/meta/contrib/DocToken.scala
@@ -32,6 +32,7 @@ case class DocToken(kind: DocToken.Kind, name: Option[String], body: Option[Stri
             case _ => Seq[DocToken.Reference]()
           }
         }
+
         parseBodyFrom(0)
       }
       .getOrElse(Nil)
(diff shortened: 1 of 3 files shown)


xavier-fernandez:
Working in improving this PR solution, will have in an hour from now...

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/DocToken.scala
```diff
@@ -264,11 +264,6 @@ object DocToken {
   /**
     * Documents a Scaladoc Heading.
     */
-  case object Heading extends Kind
```
olafurpg: Since the number of headings is limited, can we use multiple case objects?

```scala
abstract class Heading(val level: Int) extends Kind
case object Heading1 extends Heading(1)
case object Heading2 extends Heading(2)
...
```
Sure, it a bit of boilerplate but it avoids validating false input like `Heading(-1)`.

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/ScaladocParser.scala
```diff
@@ -43,6 +45,18 @@ object ScaladocParser {
     comment.content.map(parseRec)
   }
 
+  private[this] def generateHeadingParser(level: Int, f: (String) => DocToken): Parser[DocToken] = {
+    val headingSimbols = "=" * level
+    P(
+      // Code block start
+      headingSimbols
+      // Heading description
+        ~ ((AnyChar ~ !"=").rep ~ AnyChar).!.map(f(_))
```
olafurpg: I think we can get away with a single header parser if we cut `"=".rep` and inside the map we construct the correct header by the length of the `=`.

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/DocToken.scala
```diff
@@ -264,11 +265,48 @@ object DocToken {
...
+  case object Heading5 extends Heading(5)
 
   /**
-    * Documents a Scaladoc sub-heading.
+    * Represents a sixth level heading:
+    *
+    * i.e: '======HEADING======'
     */
-  case object SubHeading extends Kind
+  case object Heading6 extends Heading(6)
```
olafurpg:
```scala
def allHeadings = List(Heading1, Heading2, ...)
def headingForLevel(level: Int): Option[Header] = allHeading.find(_.level == level)
```

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/ScaladocParser.scala
```diff
@@ -43,6 +45,18 @@ object ScaladocParser {
     comment.content.map(parseRec)
   }
 
+  private[this] def generateHeadingParser(level: Int, f: (String) => DocToken): Parser[DocToken] = {
+    val headingSimbols = "=" * level
```
olafurpg: s/simbol/symbol/

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/ScaladocParser.scala
xavier-fernandez: Not sure on how to do that. if it is a wronly formatted header, how I can cause the parser fail in the map? 

With an exception?

olafurpg: Did you try `flatMap`? http://www.lihaoyi.com/fastparse/#FlatMap We should avoid exceptions,  parsers can report errors at positions.

xavier-fernandez: No, didn't knew that method, will check for sure :)

Thank you

xeno-by:
LGTM

PR #956, opened 2017-06-21
Add productFields

### LINKED ISSUE #948: Add productFields on Trees

gabro:
Closes #948

### DIFF
--- a/scalameta/common/shared/src/main/scala/scala/meta/internal/ast/ast.scala
+++ b/scalameta/common/shared/src/main/scala/scala/meta/internal/ast/ast.scala
@@ -172,6 +172,8 @@ class AstNamerMacros(val c: Context) extends AstReflection with CommonNamerMacro
       pelClauses += cq"_ => throw new $IndexOutOfBoundsException(n.toString)"
       stats1 += q"override def productElement(n: $IntClass): Any = n match { case ..$pelClauses }"
       stats1 += q"override def productIterator: $IteratorClass[$AnyClass] = $ScalaRunTimeModule.typedProductIterator(this)"
+      val productFields = productParamss.head.map(_.name.toString)
+      stats1 += q"override def productFields: $SeqClass[$StringClass] = Seq(..$productFields)"
 
       // step 12: generate serialization logic
       val fieldInits = fieldParamss.flatten.map(p => q"$CommonTyperMacrosModule.loadField(this.${internalize(p.name)})")
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/ast/InternalTrees.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/ast/InternalTrees.scala
@@ -44,6 +44,8 @@ trait InternalTree {
     if (privateParent != null) scala.Some(privateParent) else None
   }
 
+  def productFields: Seq[String]
+
   private[meta] def origin: Origin = {
     if (privateOrigin != null) privateOrigin else Origin.None
   }
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/ast/InternalTrees.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/ast/InternalTrees.scala
@@ -17,7 +17,7 @@ import scala.meta.internal.tokens._
 
 // NOTE: Methods that start with "private" are NOT intended to be called outside scala.meta.
 // Calling these methods from hosts will compile (because hosts are in meta), but is strongly discouraged.
-trait InternalTree {
+trait InternalTree extends Product {
   self: Tree =>
 
   // =============================================================================================
@@ -44,6 +44,12 @@ trait InternalTree {
     if (privateParent != null) scala.Some(privateParent) else None
   }
 
+  // NOTE: InternalTree inherits traditional productXXX methods from Product
+  // and also adds a new method called productFields.
+  // def productPrefix: String
+  // def productArity: Int
+  // def productElement(n: Int): Any
+  // def productIterator: Iterator[Any]
   def productFields: Seq[String]
 
   private[meta] def origin: Origin = {


scalameta/common/shared/src/main/scala/scala/meta/internal/ast/ast.scala
```diff
@@ -172,6 +172,8 @@ class AstNamerMacros(val c: Context) extends AstReflection with CommonNamerMacro
       pelClauses += cq"_ => throw new $IndexOutOfBoundsException(n.toString)"
       stats1 += q"override def productElement(n: $IntClass): Any = n match { case ..$pelClauses }"
       stats1 += q"override def productIterator: $IteratorClass[$AnyClass] = $ScalaRunTimeModule.typedProductIterator(this)"
+      val productFields = productParamss.head.map(_.name.toString)
+      stats1 += q"override def productFields: $ListClass[$StringClass] = List(..$productFields)"
```
olafurpg: Can we add a test for something like `q"class Foo".productFields == List("mods", "name", ...)`?

scalameta/common/shared/src/main/scala/org/scalameta/internal/MacroHelpers.scala
```diff
@@ -70,6 +70,7 @@ trait MacroHelpers extends DebugFinder
   lazy val IndexOutOfBoundsException = hygienicRef[IndexOutOfBoundsException]
   lazy val IteratorClass = tq"_root_.scala.collection.Iterator"
   lazy val SeqClass = tq"_root_.scala.collection.immutable.Seq"
```
olafurpg: How about we return an `immutable.Seq` for consistency with `children`? It'll likely change to List with the re-organization PR.

scalameta/common/shared/src/main/scala/scala/meta/internal/ast/ast.scala
gabro: yep, is there any existing test suite I can add this to?

scalameta/common/shared/src/main/scala/org/scalameta/internal/MacroHelpers.scala
gabro: I had that PR in mind when I chose List, but if you prefer I can use `Seq` and let you guys handle the refactor.

olafurpg: I think `Seq` makes more sense now.

scalameta/common/shared/src/main/scala/scala/meta/internal/ast/ast.scala
olafurpg: I don't have a test suite at the top of my head. You can try to use it from the js parsers facade  ;)

gabro: you mean something like this? 😄

scalameta/common/shared/src/main/scala/org/scalameta/internal/MacroHelpers.scala
gabro: done!

scalameta/common/shared/src/main/scala/scala/meta/internal/ast/ast.scala
gabro: @olafurpg take a look at https://github.com/scalameta/scalameta/pull/957 which uses `productFields`. It's implicitly tested by the parser facade test.

scalameta/trees/shared/src/main/scala/scala/meta/internal/ast/InternalTrees.scala
```diff
@@ -44,6 +44,8 @@ trait InternalTree {
     if (privateParent != null) scala.Some(privateParent) else None
   }
 
+  def productFields: Seq[String]
```
xeno-by: Let's inherit `InternalTree` from `Product` and then here say something like:

```
// NOTE: This and related methods are inherited from Product
// def productIterator: Iterator[Any]
```

gabro: Not sure I follow what you mean

xeno-by: I'd like to highlight the fact that this method is accompanied by the more traditional `Product` methods, so that `InternalTrees.scala` looks like:

```scala
trait InternalTree extends Product {
  ...
  // NOTE: InternalTree inherits traditional productXXX methods from Product
  // and also adds a new method called productFields.
  // def productPrefix: String
  // def productArity: Int
  // def productElement(n: Int): Any
  // def productIterator: Iterator[Any]
  def productFields: Seq[String]
  ...
}
```

gabro: ah sure, that makes sense, how does `Product` end up in `Tree` currently?

xeno-by: The `@ast` annotation adds it to the parent list.

gabro: ok, so should that stay?

xeno-by: Let's keep it. I'd like to say `extends Product` here just for the sake of documentation.

gabro: 👍 I pushed the change you suggested

xeno-by:
LGTM

PR #957, opened 2017-06-21
Use productFields in parsersJS

gabro:
This allows for a nicer output that enables this:

~Also this is based on #956, I'll rebase once that one gets merged.~

### DIFF
--- a/scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
+++ b/scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
@@ -29,15 +29,25 @@ object JSFacade {
     }.asInstanceOf[JSLong]
   }
 
+  private[this] def toNode(t: Any): js.Any = t match {
+    case t: Tree => toNode(t)
+    case tt: Seq[_] => tt.map(toNode).toJSArray
+    case t: Option[_] => t.map(toNode).orUndefined
+    case _ => ()
+  }
+
   private[this] def toNode(t: Tree): js.Dynamic = {
     val base = js.Dynamic.literal(
       "type" -> t.productPrefix,
-      "children" -> t.children.map(toNode).toJSArray,
       "pos" -> js.Dynamic.literal(
         "start" -> t.pos.start.offset,
         "end" -> t.pos.end.offset
       )
     )
+
+    val fields = js.Dictionary(t.productFields.zip(t.productIterator.toList).collect {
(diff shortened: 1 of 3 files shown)


gabro:
Mmm, I would've expected https://github.com/scalameta/scalameta/pull/957/commits/c5155a39bdde5883507c6bebc1c9b98553f64095 to fail on the CI, maybe parsersJS test are not running? /cc @xeno-by

gabro:
Rebased on master now that #956 has been merged.

scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
```diff
@@ -29,16 +29,26 @@ object JSFacade {
...
+    case t: Tree => toNode(t)
+    case tt: Seq[_] => tt.map(toNode).toJSArray
+    case t: Option[_] => t.map(toNode).orUndefined
+    case _ => ()
+  }
+
   private[this] def toNode(t: Tree): js.Dynamic = {
     val base = js.Dynamic.literal(
       "type" -> t.productPrefix,
-      "children" -> t.children.map(toNode).toJSArray,
```
olafurpg: Would it be possible to expose a method to replicate `.children`?

gabro: you mean producing a tree with the `children` node?

olafurpg: `.children` is all the fields concatenated with lists and options flattened.

gabro: Yep, so you want to expose a JS for getting the children of a tree?

olafurpg: It could be something as a method that takes a json tree and filters out the `type` and `pos` fields and flattens lists. You think that would be useful? We can skip it otherwise.

gabro: Ah, got it.
It would be an easy addition, but I don't have a use case in mind for it atm.

I would keep it out of the scope of this PR and add it on demand in case it's deemed useful.

olafurpg: Sounds good 👍

PR #958, opened 2017-06-22
Allow specifying dialect through parsers facade

gabro:
This PR allows specifying a dialect through the JS api.

Here's an example of usage from ASTExplorer:

This PR is based on #957 (which in turn is based on #956). I'll rebase as soon as the parent PRs get merged.

### DIFF
--- a/scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
+++ b/scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
@@ -6,6 +6,7 @@ import js.JSConverters._
 import js.annotation._
 
 import prettyprinters._
+import inputs.Position
 
 object JSFacade {
 
@@ -36,13 +37,16 @@ object JSFacade {
     case _ => ()
   }
 
+  private[this] def toPosition(p: Position): js.Dynamic =
+    js.Dynamic.literal(
+      "start" -> p.start.offset,
+      "end" -> p.end.offset
+    )
+
   private[this] def toNode(t: Tree): js.Dynamic = {
     val base = js.Dynamic.literal(
       "type" -> t.productPrefix,
-      "pos" -> js.Dynamic.literal(
-        "start" -> t.pos.start.offset,
-        "end" -> t.pos.end.offset
-      )
+      "pos" -> toPosition(t.pos)
     )
 
     val fields = js.Dictionary(t.productFields.zip(t.productIterator.toList).collect {
@@ -84,8 +88,9 @@ object JSFacade {
 
     dialect(code).parse[A] match {
       case Parsed.Success(t) => toNode(t).asInstanceOf[js.Dictionary[Any]]
-      case Parsed.Error(_, message, _) => js.Dictionary(
-        "error" -> message
+      case Parsed.Error(pos, message, _) => js.Dictionary(
+        "error" -> message,
+        "pos" -> toPosition(pos)
       )
     }
   }
--- a/scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
+++ b/scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
@@ -76,24 +76,31 @@ object JSFacade {
   private[this] type Settings = js.UndefOr[js.Dictionary[String]]
   private[this] val defaultSettings = None.orUndefined
 
+  private[this] def extractDialect(s: Settings): Either[String, Dialect] = {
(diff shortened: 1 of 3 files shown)


scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
```diff
@@ -56,22 +66,41 @@ object JSFacade {
...
+
+  private[this] def parse[A <: Tree: Parse](
+    code: String,
+    settings: Settings
+  ): js.Dictionary[Any] = {
+    val dialect = (for {
+      settings <- settings.toOption
+      dialectStr <- settings.get("dialect")
+      dialect <- Dialect.standards.get(dialectStr)
+    } yield dialect).getOrElse(dialects.Scala211)
```
olafurpg: Default to 2.12?

gabro: I wanted to keep it consistent with the current meta behavior. Is that going to change?

olafurpg: Good point, let's keep it then. The default dialect is planned to change soon, either to something lenient like https://github.com/scalameta/scalafmt/blob/d204ae1437247aae0f0b8a9901e433368057e678/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala#L43 or maybe even go away for parsing/pretty-printing.

gabro: Cool, I'll be glad to drop the option completely in the future.

gabro:
As per the rebase, I'd wait for #957 to get merged then I'll rebased on master.

gabro:
@olafurpg rebased

scalameta/parsers/js/src/test/scala/scala/meta/parsers/JSFacadeSuite.scala
```diff
@@ -193,5 +193,54 @@ class JSFacadeSuite extends FunSuite {
...
 
+  test("default dialect is Scala 2.11") {
+    val code =
+      """|List(
+         |  1,
+         |  2,
+         |)""".stripMargin
+    val parsedDefaultDialect = JSFacade.parseStat(code)
+    val expected = d(
+      "error" -> "illegal start of simple expression"
```
olafurpg: I just noticed the position is not included, do you think it would be useful to have it? It can be a separate PR or part of this one.

gabro: Yes, I've been thinking about it.
I've added it to this PR for simplicity.

gabro:
woops, the latest commit broke some tests, it should be ok now

olafurpg:
There seems to be an error in an unrelated JS parsers test

```scala
[info] - parse Lit.Double *** FAILED ***
[info]   "...2,
[info]     "syntax": "42.2[d]"
[info]   }" did not equal "...2,
[info]     "syntax": "42.2[]"
[info]   }" (JSFacadeSuite.scala:19)
```

gabro:
Odd, it fails only with Scala 2.12. It seems like `syntax` acts differently across version of Scala for Double :O

gabro:
I can reproduce this on my machine, by running `parsersJS/test` after setting the version to Scala 2.12.2

olafurpg:
Interesting @gabro, the bug happens only in 2.12 JS. I opened https://github.com/scalameta/scalameta/issues/961, I propose we remove that test and I can follow up on that issue in a separate PR.

gabro:
I agree, I've ignored the test and added a reference to #961

scalameta/parsers/js/src/test/scala/scala/meta/parsers/JSFacadeSuite.scala
```diff
@@ -193,5 +195,56 @@ class JSFacadeSuite extends FunSuite {
...
+    check(parsedDefaultDialect, expected)
+  }
+
+  test("inexisting dialects fallback to Scala 2.11") {
+    val code =
+      """|List(
+         |  1,
+         |  2,
+         |)""".stripMargin
+    val parsedDefaultDialect = JSFacade.parseStat(code, js.Dictionary("dialect" -> "wrong"))
```
gabro: Right, probably if you pass a wrong one you want to know. It makes sense I'll report an error (instead of throwing an exception)

gabro:
@olafurpg done, take a look

PR #960, opened 2017-06-22
Return Symbol.None instead of Symbol.Local for no position, fixes #959.

### LINKED ISSUE #959: Symbol <none> for local symbols
Reported in https://github.com/scalacenter/scalafix/issues/222, 
To reproduce

```
git clone git@github.com:twitter/scalding.git
cd scalding
git checkout 73c2e694bf235c398aaf61f879ceefa1fbd4f944
echo 'addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.4.2")' >> project/plugins.sbt
sbt "scalafix ExplicitReturnTypes"
```

triggers the following error

```
java.lang.RuntimeException: bad protobuf: unsupported name range {
  start: 925
  end: 932
}
symbol: "<none>"
```

The reported positions point to `flatMap` and `mapStream` here
https://github.com/twitter/scalding/blob/7615dbf6237657a5e2b8778e97b9bc2da353adf6/scalding-core/src/test/scala/com/twitter/scalding/SideEffectTest.scala#L33

I tracked down the root cause to Symbol.syntax for Symbol.Local, which uses Position.syntax here https://github.com/scalameta/scalameta/blob/5684acc8dbb5c311a5b1d54123298f828e3743b0/scalahost/nsc/src/main/scala/scala/meta/internal/semantic/SymbolOps.scala#L26 which can be `<none>` here https://github.com/scalameta/scalameta/blob/5684acc8dbb5c311a5b1d54123298f828e3743b0/scalameta/inputs/shared/src/main/scala/scala/meta/internal/inputs/package.scala#L53

I made several failed attempts to minimize the error with something like
```scala
  names(
    """
      |object B {
      |  class Bar {
      |    def using(x: Any): Bar = ???
      |    def flatMap[A, B](t: (Symbol, (Symbol, Symbol)))(f: (Any, Any) => Any): Unit = ()
      |  }
      |  new Bar()
      |    .using { 1 }
      |    .flatMap[String, (String, String)]('line -> ('l1, 'l2)) { case (a, b) => }
      |}
    """.stripMargin,
    """
      |
    """.stripMargin
  )
```

I propose we guard against `Position.None` when producing Symbol.Local, which should avoid the error.

olafurpg:
I struggled to find a minimal repro, but the fix seems logical to me. I
tested the fix manually by re-compiling the original reported issue from
twitter/scalding and confirmed that the ResolvedName with symbol
"<none>" no longer appears in the .semanticdb file.

### DIFF
--- a/scalahost/nsc/src/main/scala/scala/meta/internal/semantic/SymbolOps.scala
+++ b/scalahost/nsc/src/main/scala/scala/meta/internal/semantic/SymbolOps.scala
@@ -23,7 +23,13 @@ trait SymbolOps { self: DatabaseOps =>
             ((sym.owner.isAliasType || sym.owner.isAbstractType) && !sym.isParameter)
         !definitelyGlobal && (definitelyLocal || isLocal(sym.owner))
       }
-      if (isLocal(sym)) return m.Symbol.Local(sym.pos.toMeta.syntax)
+      if (isLocal(sym)) {
+        val mpos = sym.pos.toMeta
+        return {
+          if (mpos == m.Position.None) m.Symbol.None
+          else m.Symbol.Local(mpos.syntax)
+        }
+      }
 
       val owner = sym.owner.toSemantic
       val signature = {


scalahost/nsc/src/main/scala/scala/meta/internal/semantic/SymbolOps.scala
```diff
@@ -23,7 +23,13 @@ trait SymbolOps { self: DatabaseOps =>
...
         !definitelyGlobal && (definitelyLocal || isLocal(sym.owner))
       }
-      if (isLocal(sym)) return m.Symbol.Local(sym.pos.toMeta.syntax)
+      if (isLocal(sym)) {
+        val mpos = sym.pos.toMeta
+        return {
+          if (mpos == m.Position.None) m.Symbol.None
+          else m.Symbol.Local(mpos.syntax)
+        }
+      }
```
xeno-by: Looks reasonable.

PR #962, opened 2017-06-23
Add lineNumber and columnNumber to parser errors

gabro:
These are useful for highlighting parse errors in AST explorer. Here's an example:

image


scalameta/parsers/js/npm/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "scalameta-parsers",
-  "version": "1.9.0-beta.1",
+  "version": "1.9.0-beta.3",
```
gabro: 3 because I've already published 2 and I forgot to commit this file.

PR #966, opened 2017-06-26
Fix #918

### LINKED ISSUE #918: Add support for Dotty implicit by-name arguments
See http://dotty.epfl.ch/docs/reference/implicit-by-name-parameters.html and http://dotty.epfl.ch/docs/internals/syntax.html#types

allanrenucci:
Add support for Dotty implicit by-name arguments


scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -68,6 +68,10 @@ import scala.compat.Platform.EOL
   // They are in Dotty, but not in Scala 2.12 or older.
   allowTraitParameters: Boolean,
 
+  // Are implicit by name parameters supported?
+  // They are in DOtty, but not in Scala 2.12 or older.
```
xeno-by: Typo

Issue #973, opened 2017-06-27
InvariantFailedException when tokenising keyword after $ in interpolated string


allanrenucci:
```scala
scala> "s\"$class\"".tokenize
org.scalameta.invariants.InvariantFailedException: invariant failed:
when verifying curr.token.==(10).||(curr.token.==(21))
found that curr.token is not equal to 10
and also curr.token is not equal to 21()
  at org.scalameta.invariants.InvariantFailedException$.raise(Exceptions.scala:15)
  at scala.meta.internal.tokenizers.ScalametaTokenizer.emitContents$1(ScalametaTokenizer.scala:217)
  at scala.meta.internal.tokenizers.ScalametaTokenizer.loop$1(ScalametaTokenizer.scala:235)
  at scala.meta.internal.tokenizers.ScalametaTokenizer.uncachedTokenize(ScalametaTokenizer.scala:280)
  at scala.meta.internal.tokenizers.ScalametaTokenizer.tokenize(ScalametaTokenizer.scala:35)
  at scala.meta.internal.tokenizers.ScalametaTokenizer$$anon$2.apply(ScalametaTokenizer.scala:291)
  at scala.meta.tokenizers.Api$XtensionTokenizeDialectInput.tokenize(Api.scala:23)
  at scala.meta.tokenizers.Api$XtensionTokenizeInputLike.tokenize(Api.scala:12)
```
I am not sure what is the expected behaviour. Should it returns `Tokenized.Success` and then be rejected by the parser, or `Tokenized.Error`. I don't think it should throw an exception though.

Note that it works for `this` and identifiers.

xeno-by:
Crashing is definitely a wrong thing to do.

I'd say tokenization failure is allright. We fail tokenization if string literals are malformed, so it's kinda logical to also fail on malformed interpolations.

xeno-by:
@allanrenucci Have we fixed this in master?

allanrenucci:
It is fixed in #974 which isn't merged

PR #974, opened 2017-06-27
Fix bug in enum tokenization and #973

allanrenucci:


xeno-by:
Let's wait. I'll do my best to fix names tomorrow, and then we can merge
#907.

Issue #830, opened 2017-05-03
Missing denotation for case class unapply


olafurpg:
This code here

```scala
object A {
  def unapply(e: Int): Option[String] = Some(e.toString)
}
case class Foo(a: Int)
object ya {
  A.unapply(2)
  Foo.unapply(Foo(2))
}
```

has database

```

Names:
[7..8): A => _empty_.A.
[17..24): unapply => _empty_.A.unapply(I)Lscala/Option;.
[25..26): e => _empty_.A.unapply(I)Lscala/Option;.(e)
[28..31): Int => _root_.scala.Int#
[34..40): Option => _root_.scala.Option#
[41..47): String => _root_.scala.Predef.String#
[51..55): Some => _root_.scala.Some.apply(Ljava/lang/Object;)Lscala/Some;.
[56..57): e => _empty_.A.unapply(I)Lscala/Option;.(e)
[58..66): toString => _root_.scala.Any#toString()Ljava/lang/String;.
[81..84): Foo => _empty_.Foo#
[85..86): a => _empty_.Foo#(a)
[88..91): Int => _root_.scala.Int#
[100..102): ya => _empty_.ya.
[107..108): A => _empty_.A.
[109..116): unapply => _empty_.A.unapply(I)Lscala/Option;.
[122..125): Foo => _empty_.Foo.
[126..133): unapply => _empty_.Foo.unapply(LFoo;)Lscala/Option;.
[134..137): Foo => _empty_.Foo.apply(I)LFoo;.

Denotations:
_empty_.A. => final object A
_empty_.A.unapply(I)Lscala/Option;. => def unapply: (e: Int)Option[String]
_empty_.A.unapply(I)Lscala/Option;.(e) => param e: Int
_empty_.Foo# => case class Foo
_empty_.Foo#(a) => val param a: Int
_empty_.Foo#`<init>`(I)V. => primaryctor <init>: (a: Int)Foo
_empty_.ya. => final object ya

Sugars:
[55..55) [String]
```

Observe that `A.unapply` has a denotation but `Foo.unapply` does not. It would be nice if we could get the denotation for the compiler generated `Foo.unapply`

olafurpg:
This opens up a question whether we should persist denotations of synthetic definitions. The members of the companion module are quite many
```scala
gsym.companionModule.toType.members]:
Scope{
  private def readResolve(): Object;
  case def unapply(x$0: Foo): Option[Int];
  case def apply(a: Int): Foo;
[...]

PR #983, opened 2017-07-03
Add test to show #830 is fixed.

olafurpg:


scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
```diff
@@ -365,4 +365,11 @@ class SemanticSuite extends DatabaseSuite(SemanticdbMode.Slim) {
       |_root_.scala.collection.mutable.Builder#result()Ljava/lang/Object;. => abstract def result: ()To
     """.stripMargin.trim
   )
+
+  denotationsContain(
```
xeno-by: How about making this a targeted test?

olafurpg: Good idea, updated.

PR #986, opened 2017-07-03
Add configurable error handling, fixes #982.

### LINKED ISSUE #982: Configurable error handling in scalahost
Currently, scalahost reports a compilation error when it for example fails to parse a file. It would be nice if it was possible to configure the failure more in Scalahost via `-P:scalahost:failures:info/warn/error/ignore`. During development, I want errors but the default setting should be `warning` in my opinion.

olafurpg:
Previously, it was impossible to ignore a single parse error when
compiling a large project with scalahost. This resulted in a bad user
experience with no workaround but to disable scalahost. This commit adds
the ability to configure at which level the errors are reported,
defaulting to a warning.

### DIFF
--- a/scalahost/nsc/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
+++ b/scalahost/nsc/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
@@ -48,11 +48,12 @@ trait ScalahostPipeline extends DatabaseOps { self: ScalahostPlugin =>
             writer.write(s"failed to generate semanticdb for $path:$EOL")
             ex.printStackTrace(new PrintWriter(writer))
             val msg = writer.toString
+            import scala.meta.internal.semantic.FailureMode._
             config.failures match {
-              case Severity.ERROR => global.reporter.error(g.NoPosition, msg)
-              case Severity.WARNING => global.reporter.warning(g.NoPosition, msg)
-              case Severity.INFO => global.reporter.info(g.NoPosition, msg, force = true)
-              case _ =>
+              case Error => global.reporter.error(g.NoPosition, msg)
+              case Warning => global.reporter.warning(g.NoPosition, msg)
+              case Info => global.reporter.info(g.NoPosition, msg, force = true)
+              case Ignore => // do nothing.
             }
         }
       }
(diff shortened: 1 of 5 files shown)


olafurpg:
It's up for debate if the default should be warning. I chose warning because that's what I suspect most beginners/end-users will want it to be and the warning will get promoted to an error with -Xfatal-warnings for the users who really want fail fast.

scalahost/nsc/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
```diff
@@ -42,11 +43,17 @@ trait ScalahostPipeline extends DatabaseOps { self: ScalahostPlugin =>
...
-            msg.write(s"failed to generate semanticdb for $path:$EOL")
-            ex.printStackTrace(new PrintWriter(msg))
-            global.reporter.error(g.NoPosition, msg.toString)
+            writer.write(s"failed to generate semanticdb for $path:$EOL")
+            ex.printStackTrace(new PrintWriter(writer))
+            val msg = writer.toString
+            config.failures match {
+              case Severity.ERROR => global.reporter.error(g.NoPosition, msg)
+              case Severity.WARNING => global.reporter.warning(g.NoPosition, msg)
+              case Severity.INFO => global.reporter.info(g.NoPosition, msg, force = true)
```
xeno-by: What's the practical difference between `force = true` and `force = false`?

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/ConfigOps.scala
```diff
@@ -4,14 +4,18 @@ package semantic
 import scala.meta.internal.io.PathIO
 import scala.meta.internal.scalahost.ScalahostPlugin
 import scala.meta.io._
+import scala.meta.internal.semantic.schema.Message.Severity
```
xeno-by: Why not define a custom data structure here? Using an internal class from a different module feels wrong.

scalahost/nsc/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
olafurpg: No idea, you have to choose however. I suspect `force = false` may mean that the message is not flushed immediately?

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/ConfigOps.scala
olafurpg: I opted to re-use the existing data structure since I'm lazy and writing enums in Scala is annoying and boilerplatey. I'll change.

olafurpg:
Superseded by #988

PR #985, opened 2017-07-03
Handle unapply and unapplySeq, fixes #977.

olafurpg:

PR #984, opened 2017-07-03
Emit correct position for unused Importee.Wildcard(), fixes #899.

olafurpg:

PR #988, opened 2017-07-04
Several PRs combined into a single PR.

olafurpg:
This is the combination of

- #986 
- #984 
- #985 
- #989 
- #992
- this original PR, which fixed several issues with sugars.

All of the PRs did something small that caused merge conflicts with each other. It's much simpler to manage them together.


scalahost/nsc/src/main/scala/scala/meta/internal/semantic/AttributesOps.scala
```diff
@@ -264,13 +265,15 @@ trait AttributesOps { self: DatabaseOps =>
               gtree match {
                 case gview: g.ApplyImplicitView =>
                   val pos = gtree.pos.toMeta
-                  val syntax = gview.fun + "(*)"
+                  val syntax = g.showCode(gview.fun) + "(*)"
```
xeno-by: What's the difference between `toString` and `showCode` here? I'm very wary of `showCode`, because it's known to sometimes swallow snippets of code that it considers irrelevant.

scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
```diff
@@ -365,4 +365,43 @@ class SemanticSuite extends DatabaseSuite(SemanticdbMode.Slim) {
...
+  sugars(
+    """|object b {
+       |  class F
+       |  implicit val ordering: Ordering[F] = ???
+       |  val x: Ordered[F] = new F
+       |}
+    """.stripMargin,
+    """|[86..91) scala.math.Ordered.orderingToOrdered[b.F](*)(b.this.ordering)
+       |""".stripMargin
+  )
```
xeno-by: Please rename the enclosing objects in the tests once we merge #907.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/AttributesOps.scala
olafurpg: Here's an example failing test from rebase on master

```
[info]   -[273..275) h.X.cvt[Int](*)(h.C.int)
[info]   +[273..275) h.this.X.cvt[Int](*)(h.this.C.int)
```
The second one (.toString) would not compile if inserted into a source file while the first one (showCode) would compile.

olafurpg:
CI on appveyor is already proving itself valuable, this surfaced in the new expect tests

```
[info]   =======
[info]   => Diff
[info]   =======
[info]   -scalahost\integration\src\main\scala\example\Example.scala
[info]   +scalahost/integration/src/main/scala/example/Example.scala
[info]    ---------------------------------------------------------- (SemanticdbExpectSuite.scala:14)
[info] ScalaTest
```

olafurpg:
No new features or changes were introduced from yesterday. I'm gonna merge this now to unblock further changes today.

scalahost/nsc/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
```diff
@@ -14,6 +14,7 @@ import scala.meta.internal.semantic.DatabaseOps
 import scala.meta.internal.semantic.{vfs => v}
 import scala.meta.internal.semantic.{schema => s}
 import scala.tools.nsc.doc.ScaladocGlobal
+import scala.meta.internal.semantic.schema.Message.Severity
```
xeno-by: Looks like an unused import.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/AttributesOps.scala
```diff
@@ -325,12 +328,16 @@ trait AttributesOps { self: DatabaseOps =>
                   val morePrecisePos = fun.pos.withStart(fun.pos.end).toMeta
                   val syntax = "[" + targs.mkString(", ") + "]"
                   success(morePrecisePos, syntax)
+                case g.Apply(select @ g.Select(qual, nme), _) if isSyntheticName(select) =>
+                  val pos = qual.pos.withStart(qual.pos.end).toMeta
+                  success(pos, s".${nme.decoded}")
```
xeno-by: Why do we need `decoded` here?

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/ConfigOps.scala
```diff
@@ -5,13 +5,32 @@ import scala.meta.internal.io.PathIO
...
+  def default = ScalahostConfig(PathIO.workingDirectory, SemanticdbMode.Fat, FailureMode.Warning)
+}
+
+sealed abstract class FailureMode {
+  def name: String = toString.toLowerCase
+}
+object FailureMode {
+  def unapply(arg: String): Option[FailureMode] = {
+    val lower = arg.toLowerCase()
+    all.find(_.name == lower)
```
xeno-by: This implementation looks inconsistent with `SemanticdbMode.unapply`.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/package.scala
```diff
@@ -2,18 +2,102 @@ package scala.meta
...
+        val abspath = targetroot.resolve(relpath).toNIO
+        Files.createDirectories(abspath.getParent)
+        val fos = Files.newOutputStream(abspath)
+        try {
+          s.Database(List(entry)).writeTo(fos)
+        } finally {
+          fos.close()
+        }
+      }
+    }
```
xeno-by: I think we should either consistently get rid of vfs or consistently use it in both directions.

xeno-by: Let's first discuss the idea behind merging `toVfs` and `save`. Is that to save memory spent on precaching arrays of bytes? If yes, why not make `bytes` lazy?

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/Database.scala
```diff
@@ -19,17 +18,7 @@ object Database {
 @data
 class Database(entries: List[Entry]) {
   def toSchema: s.Database = {
-    val sentries = entries.map(ventry => s.Attributes.parseFrom(ventry.bytes))
+    val sentries = entries.flatMap(ventry => s.Database.parseFrom(ventry.inputStream).entries)
```
xeno-by: Am I right in thinking that the change from `.bytes` to `.inputStream` here is done for optimization purposes?

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/Paths.scala
```diff
@@ -1,28 +1,26 @@
 package scala.meta.internal.semantic.vfs
 
 import scala.meta.internal.io.PathIO.fileSeparator
+import scala.meta.io.AbsolutePath
```
xeno-by: Looks like an unused import.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/Paths.scala
```diff
@@ -1,28 +1,26 @@
...
+  private val semanticDbPrefix: RelativePath = RelativePath("META-INF").resolve("semanticdb")
   private val semanticDbSuffix = ".semanticdb"
   private val scalaPrefix = ""
   private val scalaSuffix = ".scala"
 
   def isSemanticdb(path: RelativePath): Boolean = {
-    path.toString.startsWith(semanticDbPrefix) &&
-    path.toString.endsWith(semanticDbSuffix)
+    path.toNIO.startsWith(semanticDbPrefix.toNIO) &&
+    path.path.getFileName.toString.endsWith(semanticDbSuffix)
```
xeno-by: `path.path` reads weirdly. Can we change it to `path.value`?

xeno-by: Also, I'm not sure whether `path.path.getFileName.toString` constitutes an improvement with respect to `path.toString`. Can you comment on this?

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/Paths.scala
```diff
@@ -1,28 +1,26 @@
...
   def semanticdbToScala(path: RelativePath): RelativePath = {
     require(isSemanticdb(path))
-    RelativePath(
-      scalaPrefix +
-        path.toString
-          .stripPrefix(semanticDbPrefix)
-          .stripSuffix(semanticDbSuffix) +
-        scalaSuffix)
+    RelativePath(path.toNIO
+      .relativize(path.toNIO)
```
xeno-by: `path.relativize(path)` reads like a typo. Have I misunderstood something?

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/Paths.scala
```diff
@@ -32,11 +30,6 @@ object Paths {
 
   def scalaToSemanticdb(path: RelativePath): RelativePath = {
     require(isScala(path))
-    RelativePath(
-      semanticDbPrefix +
-        path.toString
-          .stripPrefix(scalaPrefix)
-          .stripSuffix(scalaSuffix) +
-        semanticDbSuffix)
+    semanticDbPrefix.resolve(path.resolveSibling(_.stripSuffix(scalaSuffix) + semanticDbSuffix))
```
xeno-by: I find the refactoring of logic in `Paths` hard to follow. Let's discuss a formulation that would work for both of us. What did you find unnice about the previous implementation?

scalameta/tests/jvm/src/test/resources/semanticdb.expect
```diff
@@ -0,0 +1,26 @@
+scalahost/integration/src/main/scala/example/Example.scala
+----------------------------------------------------------
+Dialect:
+Scala212
```
xeno-by: How does this work? Do we force expect tests to Scala 2.12 only?

scalahost/nsc/src/main/scala/scala/meta/internal/scalahost/ScalahostPipeline.scala
olafurpg: I've removed all unused imports in the build https://github.com/scalameta/scalameta/pull/1004

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/AttributesOps.scala
olafurpg: I thought that was the way to get the name of a symbol, is there a better method?

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/package.scala
olafurpg: With the previous vfs design, we had `Array[Byte]` + `s.Database` + `m.Database` loaded into memory for all .semanticdb files at some point. This is prohibitively expensive when running large-scale analysis. We want to avoid materializing an `Array[Byte]` entirely. Protos can be loaded in a streaming fashion, tag by tag, and that's what this refactoring did.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/Database.scala
olafurpg: Correct.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/Paths.scala
olafurpg: Removed in https://github.com/scalameta/scalameta/pull/1004

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/Paths.scala
olafurpg: I've refactored the logic a bit to make it cleaner. I agree my refactoring was a bit messy. I did not like the previous implementation since it was purely string based. A path is more a `List[String]` than a `String`, so when changing the extension we should try to use `resolveSibling` and when checking the extension we should check the extension against the filename instead of `Path.toString.endsWith`.

scalameta/tests/jvm/src/test/resources/semanticdb.expect
olafurpg: That's fine for me by now. We can normalize the dialect to something else if you prefer. You won't have to fix this file manually though, just run the SaveExpect... main and commit.

olafurpg:
Thank you for the review @xeno-by. I apologize for merging so quickly. I wanted to avoid merge conflicts with all the recent activity. I've addressed your feedback in #1005

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/Paths.scala
ShaneDelmore: Can we just enable -Ywarn-unused:import to prevent humans having to catch this? If it makes the local development cycle awkward you can either disable that flag locally, or we could modify build to enable the flag in CI runs.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/AttributesOps.scala
xeno-by: I mixed this up with the macro API, where you have to do `decodedName.toString`.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/Paths.scala
xeno-by: @olafurpg Can we implement an unused import check on CI via scalafix?

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/Paths.scala
xeno-by: On a number of occasions, we have seen that moving from `String` to `*Path` brings correctness benefits. In general, I'm in favor of refactorings like that, as long as they are reasonably concise. I had problems with the first take on this, but I do like the new version. Thank you!

scalameta/tests/jvm/src/test/resources/semanticdb.expect
xeno-by: 2.12 was released quite a while ago, so it's only fair that we use it as default. Also see https://github.com/scalameta/scalameta/issues/1011.

xeno-by:
@olafurpg It's alright. I should've communicated better by explicitly requesting changes on this PR, and I appreciate that you followed up with a new pull request so quickly. 

I suggest that in the future we don't merge without a green review. I haven't been rigorous about this in the past, but I will do better in the future. 

I really like the fact that in Scalameta we don't have processes that prevent us from moving fast. In my opinion, occasional breakages are fine, as long as putting out fires doesn't become our lifestyle. I think that the new policy is lightweight enough for us to remain comfortably productive. Wdyt?

PR #997, opened 2017-07-06
Clean up Database interface.

olafurpg:


olafurpg:
I'm convinced #998 is a good idea, so I'm closing this PR in favor of that one.

PR #998, opened 2017-07-06
Make Database a Seq[Attributes]

olafurpg:
This change may be controversial so I kept it out of #997. Wdyt @xeno-by ?


olafurpg:
This is a bad idea.

PR #1000, opened 2017-07-06
First step towards semantic information for sugars.

olafurpg:
This implements the breaking changes necessary to use the scheme explained here: https://github.com/scalameta/scalameta/issues/822#issuecomment-313368110

Sugars are still not attributed with resolved named, that can be left for a future PR even for a future release. The nice thing is that the schema will not have to change.


scalahost/nsc/src/main/scala/scala/meta/internal/semantic/AttributesOps.scala
```diff
@@ -404,13 +405,24 @@ trait AttributesOps { self: DatabaseOps =>
             }
             m.Message(mpos, mseverity, msg)
         }
+        val sugars = inferred.toIterator.map {
+          case (pos, syntax) =>
+            pos -> m.Attributes(
+              Input.String(syntax),
```
xeno-by: `m.Input.String`?

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Attributes.scala
```diff
@@ -4,15 +4,18 @@ package semantic
...
   input: Input,
   dialect: Dialect,
   names: List[(Position, Symbol)],
   messages: List[Message],
   denotations: List[(Symbol, Denotation)],
-  sugars: List[(Position, String)]
+  sugars: List[(Position, Attributes)]
 ) {
+  def sugarStats: List[(Position, Stat)] = sugars.map { case (pos, attrs) => pos -> attrs.stat }
+  def stat: Stat = dialect(input).parse[Stat].get
```
xeno-by: Why do we need these two methods?

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Conversion.scala
```diff
@@ -0,0 +1,143 @@
+package scala.meta
+package internal.semantic
```
xeno-by: I don't mind refactoring the structure of `scala/meta/internal/semantic`, but I find somewhat counterproductive to discuss incremental steps in the refactoring split over several pull requests. Let's have a dedicated chat about where we want to bring this module.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Conversion.scala
```diff
@@ -0,0 +1,143 @@
...
+import scala.meta.inputs.{Position => mPosition}
+import scala.meta.internal.io._
+import scala.meta.internal.semantic.vfs.Paths
+import scala.meta.internal.semantic.{schema => s}
+import scala.meta.internal.semantic.{vfs => v}
+import scala.meta.io._
+import scala.meta.{Dialect => mDialect}
+import scala.meta.{semantic => m}
+
+object SchemaToMetaConversion {
```
xeno-by: Does this commit simply move the logic around, or it moves the logic + applies some changes? If it's the latter, I would find it useful to have this commit split into several commits.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/Inferred.scala
```diff
@@ -0,0 +1,36 @@
+package scala.meta.internal.semantic
+
+// data structure to manage multiple inferred sugars at the same position.
+case class Inferred(
+    select: Option[String] = None,
+    types: Option[String] = None,
```
xeno-by: I suggest `targs`.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/Inferred.scala
```diff
@@ -0,0 +1,36 @@
...
+  )
+
+  private def onlyConversionIsDefined =
+    conversion.isDefined &&
+      select.isEmpty &&
+      types.isEmpty &&
+      args.isEmpty
+
+  private def needsPrefix: Boolean =
+    !onlyConversionIsDefined
```
xeno-by: Am I right in thinking that these two methods are only needed in `syntax`?

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/Inferred.scala
```diff
@@ -0,0 +1,36 @@
+package scala.meta.internal.semantic
+
+// data structure to manage multiple inferred sugars at the same position.
```
xeno-by: I have a feeling that conversion should be represented by a different data structure. Or maybe it doesn't need any data structure at all? After all, it has a unique range position.

scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
```diff
@@ -205,11 +205,11 @@ class SemanticSuite extends DatabaseSuite(SemanticdbMode.Slim) {
       |object RunMe {
       |  CommandeerDSL(null.asInstanceOf[Foo])
       |}
-  """.trim.stripMargin,
+    """.trim.stripMargin,
     """
-      |[324..324) [g.Foo, g.FooDSL]
-      |[348..348) (g.Foo.fooDSL)
-  """.trim.stripMargin
+      |[324..324) (*).apply[g.Foo, g.FooDSL]
```
xeno-by: Why isn't this `*.apply[...]`?

xeno-by: Also, how about positioning this sugar at `CommandeerDSL`?

scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
```diff
@@ -205,11 +205,11 @@ class SemanticSuite extends DatabaseSuite(SemanticdbMode.Slim) {
...
+      |[348..348) (*)(g.Foo.fooDSL)
```
xeno-by: Same question about `*(...)` and positioning the sugar at `CommandeerDSL(null.asInstanceOf[Foo])`.

xeno-by:
All in all, I like the idea that sugars must be parseable as stats. This enables attribution without introducing additional mechanisms. This is definitely a big win.

I think that before becoming mergeable, this idea needs further refinement. Let's discuss exact syntax for sugars and potential changes in positioning. Please see my comments that Github "helpfully" hides under "Show outdated".

Moreover, I have questions about the ongoing refactoring of `scala/meta/internal/semantic`. How do you see this module after the refactoring is completed? What workflow of saving/loading semantic databases do you see as desirable? I asked some of the questions in https://github.com/scalameta/scalameta/pull/988#discussion_r126063349, but I think this merits a conceptual discussion first, with going into details later.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/Inferred.scala
olafurpg: Agreed. Done.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/Inferred.scala
olafurpg: Correct, I've inlined them.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/Conversion.scala
olafurpg: I have reverted these changes and kept the conversion in package.scala. I'd like to move them out of the package object eventually, nesting has gotten out of hand

-  package object
- Xtension
- def
- .map
- object Extractor
- .unapply
- match
- .map 
- match

The quickest win would be to refactor

```scala
// before
def convert(s.Db): m.Db = {
  val mdb = sdb.entries.map {
     // ... hundreds of liens
  }
  m.Db(mdb)
}

// after
def convert(s.Db): m.Db = {
  val mdb = sdb.entries.map(convertEntry)
}
def convertEntry(s.Attributes): m.Attributes = {
  // ....
}
```

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Attributes.scala
olafurpg: Turns out we don't! I've replaced `String` with a custom class called `Sugar` (instead of `Attributes`).

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/AttributesOps.scala
olafurpg: Fixed.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/Inferred.scala
olafurpg: That's an interesting thought, it would greatly simplify `Inferred.syntax` which is a win. The latest changes added a new `AttributedSugar` class which seems to be enough to handle conversions. However, what should the signature of `inferred: Map[Position, Inferred]` be then?

olafurpg:
Let me try to summary the new design that I believe models attributed sugars quite nicely

- Add new field `s.Sugar.names: List[s.ResolvedName]`. The offsets in the resolved names start at 0 and match the names in the `s.Sugar.syntax`.
- all sugars can be parsed as terms. A magic name `*` is inserted for missing "holes" in the sugar. The `*` name resolves to the symbol `_star_`. A hole is either on of two
 1. a prefix to make unparsable sugars parseable, example: `.apply => *.apply`, `[T] => *[T]`
 2. an argument to an implicit conversion, `intArrayOps(*)`
- We use an adapted g.CodePrinter that emits positions for resolved name in sugar.

Rooms for improvement:
- resolve the magic `*` into different symbols depending on if it's filling an implicit conversion or not.
- Move conversions out of `Inferred` to avoid complicated logic to insert `*`

scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/DatabaseSuite.scala
```diff
@@ -87,7 +87,7 @@ abstract class DatabaseSuite(mode: SemanticdbMode) extends FunSuite with DiffAss
     val path = g.currentRun.units.toList.last.source.file.file.getAbsolutePath
     val payload = database.toString.split(EOL)
     val section = payload.dropWhile(_ != sectionName + ":").drop(1).takeWhile(_ != "")
-    // println(section.mkString(EOL).replace(path, "<...>"))
+//     println(section.mkString(EOL).replace(path, "<...>"))
```
xeno-by: Do we treat this PR as something that we plan to discuss or something that we plan to merge? If that's the latter, I suggest we don't leave comments like that in the diff.

scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
```diff
@@ -240,11 +244,21 @@ class SemanticSuite extends DatabaseSuite(SemanticdbMode.Slim) {
       |}
   """.trim.stripMargin,
     """
-      |[201..201) [Int, List[Int]]
-      |[209..209) (scala.collection.immutable.List.canBuildFrom[Int])
-      |[247..247) (h.C.list[Int](h.C.int))
-      |[273..275) h.X.cvt[Int](*)(h.C.int)
-      |[304..304) [h.C[Int]]
+      |[201..201): *[Int, List[Int]]
+      |  [0..1): * => _star_.
```
xeno-by: I like the new name!

scalameta/scalameta/jvm/src/test/scala/scala/meta/tests/api/SurfaceSuite.scala
```diff
@@ -64,6 +64,7 @@ class SurfaceSuite extends FunSuite {
       |scala.meta.inputs.Input.Slice
       |scala.meta.inputs.Input.Stream
       |scala.meta.inputs.Input.String
+      |scala.meta.inputs.Input.Sugar
```
xeno-by: Can we reuse one of the existing inputs for this?

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/LengthWriter.scala
```diff
@@ -0,0 +1,17 @@
+package scala.meta.internal.semantic
+
+import java.io.Writer
+
+// A writer that keeps track of the current length.
```
xeno-by: @sjrd Have you encountered anything like that in the JDK?

scalameta/semantic/shared/src/main/protobuf/semanticdb.proto
```diff
@@ -54,4 +54,5 @@ message Denotation {
 message Sugar {
   Range range = 1;
   string syntax = 2;
+  repeated ResolvedName names = 3;
```
xeno-by: I really like the simplification here.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Sugar.scala
```diff
@@ -0,0 +1,13 @@
+package scala.meta
+package semantic
+
+import org.scalameta.data._
+import scala.meta.inputs._
+
+@data class Sugar(
+    input: Input.Sugar,
+    names: List[(Position, Symbol)]
```
xeno-by: Shall we have a map here to be consistent with Database?

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/LengthWriter.scala
sjrd: No, I haven't. AFAIK, the only `Writer` that gives you a `size()`-like method is `CharArrayWriter`, but that's a final sink.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/LengthWriter.scala
```diff
@@ -0,0 +1,17 @@
...
+class LengthWriter(delegate: Writer, start: Int) extends Writer {
+  var length: Int = start
+  override def flush(): Unit = {
+    length = 0
+    delegate.flush()
```
sjrd: `length = 0` does not make any sense here. It should stay the same. `flush()` does not "clear" the contents of the writer; it only ensures all written data are pushed down to the final sink, without buffering.

olafurpg: You are totally right! Wrote that without thinking, thanks.

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Sugar.scala
olafurpg: I agree 👍

scalameta/scalameta/jvm/src/test/scala/scala/meta/tests/api/SurfaceSuite.scala
xeno-by: I agree.

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/Inferred.scala
```diff
@@ -0,0 +1,68 @@
+package scala.meta
+package internal.semantic
+
+import scala.meta.inputs.Position.Range
+
+case class SugarRange(start: Int, end: Int, symbol: Symbol) {
+  def withOffset(offset: Int) = SugarRange(start + offset, end + offset, symbol)
```
xeno-by: `addOffset`

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/Inferred.scala
```diff
@@ -0,0 +1,68 @@
...
+  def +(other: String) = AttributedSugar(syntax + other, names)
+  def +(other: AttributedSugar) =
+    AttributedSugar(syntax + other.syntax, names ++ other.names.map(_.withOffset(syntax.length)))
+}
+
+object AttributedSugar {
+  val empty = AttributedSugar("", Nil)
+  val star = AttributedSugar("*", List(SugarRange(0, 1, Symbol("_star_."))))
+  def apply(syntax: String): AttributedSugar = AttributedSugar(syntax, Nil)
+  def mkString(sugars: List[AttributedSugar], sep: String): AttributedSugar = sugars match {
```
xeno-by: `sugars.map(_.toString).mkString(sep)`?

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/Inferred.scala
```diff
@@ -0,0 +1,68 @@
...
+        if (needsPrefix) AttributedSugar.star
+        else AttributedSugar.empty
+      all.foldLeft(start)(_ + _)
+    }
+    val sugarInput = Input.Sugar(sugar.syntax, input, pos.start, pos.end)
+    new Sugar(sugarInput, sugar.names.map(_.toMeta(sugarInput)))
+  }
+
+  def withConversion(syntax: AttributedSugar): Inferred =
+    if (conversion.isDefined) this
```
xeno-by: So this swallows the input argument?

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Api.scala
```diff
@@ -9,7 +9,21 @@ import scala.meta.parsers.{XtensionParsersDialectInput, XtensionParseDialectInpu
...
-    def sources: Seq[Source] = mirror.database.entries.map { attrs => attrs.dialect(attrs.input).parse[Source].get }
+    def sources: Seq[Source] = mirror.database.entries.map(_.source)
+  }
+
+  implicit class XtensionAttributesSource(attributes: Attributes) {
+    def source: Source = attributes.dialect(attributes.input).parse[Source].get
+  }
+
+  implicit class XtensionTermSugar(term: Term)(implicit m: Mirror) {
+    def sugar: Option[Stat] =
```
xeno-by: `Option[Term]`

scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Sugar.scala
olafurpg: Whoops, this never became a Map, I changed `Attributes.names` into a Map. Will update Sugar accordingly.

PR #1005, opened 2017-07-07
Address review feedback in #988

olafurpg:


scalameta/tests/shared/src/test/scala/scala/meta/tests/CrossPlatformSemanticSuite.scala
```diff
@@ -10,7 +10,7 @@ class CrossPlatformSemanticSuite extends FunSuite {
   val classpath = Classpath(BuildInfo.mirrorClasspath)
 
   def semanticdbs: List[AbsolutePath] = classpath.deep.collect {
-    case path if v.Paths.isSemanticdb(path.name) =>
+    case path if v.SemanticdbPaths.isSemanticdb(path.name) =>
```
olafurpg: I admit this is a bit awkward, any suggestion that doesn't go back to `Paths`?

ShaneDelmore: These are always a little awkward due to the fact the meaning changes before and after the =>
Before the => this is a classpath, but after, if the check is correct, it is a semanticDbPath.  Even though it reads oddly, since the if is so close to the naming I tend to name it in the case that the off is correct, which makes readability of the code after the => better.
How about 
```
Case semanticDbPath if v.SemanticdbPaths.isSemanticdb(semanticDbPath.name) =>
```
At the point of checking in isSemanticdb(...) it is really candidatePath, it may or may not be true, but if the check succeeds this is definitely a semanticDbPath, not just any path so I prefer to at least carry that information along in the name since it is not in the type.

scalameta/io/js/src/main/scala/scala/meta/internal/io/NodeNIOPath.scala
```diff
@@ -29,7 +29,7 @@ case class NodeNIOPath(filename: String) extends Path {
     NodeNIOPath(JSPath.dirname(filename))
   override def toAbsolutePath: Path =
     if (JSPath.isAbsolute(filename)) this
-    else PathIO.workingDirectory.path.resolve(this)
+    else PathIO.workingDirectory.toNIO.resolve(this)
```
xeno-by: 👍

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/SemanticdbPaths.scala
```diff
@@ -0,0 +1,34 @@
+package scala.meta.internal.semantic.vfs
+
+import scala.meta.internal.io.PathIO
+import scala.meta.io.RelativePath
+import org.scalameta.invariants.require
+
+object SemanticdbPaths {
+  private val semanticdbPrefix: RelativePath = RelativePath("META-INF").resolve("semanticdb")
```
xeno-by: Let's remove the explicit type annotation.

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/SemanticdbPaths.scala
```diff
@@ -0,0 +1,34 @@
...
+  private val semanticdbExtension = "semanticdb"
+  private val scalaExtension = "scala"
+
+  def isSemanticdb(path: RelativePath): Boolean = {
+    path.toNIO.startsWith(semanticdbPrefix.toNIO) &&
+    PathIO.extension(path.toNIO).contains(semanticdbExtension)
```
xeno-by: `== semanticdbExtension`?

scalameta/semantic/shared/src/main/scala/scala/meta/internal/semantic/vfs/SemanticdbPaths.scala
```diff
@@ -0,0 +1,34 @@
...
+    val scalaSibling =
+      path.resolveSibling(_.stripSuffix(semanticdbExtension) + scalaExtension)
+    semanticdbPrefix.relativize(scalaSibling)
+  }
+
+  def isScala(path: RelativePath): Boolean = {
+    PathIO.extension(path.toNIO).contains(scalaExtension)
+  }
+
+  def fromScala(path: RelativePath): RelativePath = {
```
xeno-by: `toSemanticdb`?

xeno-by:
Since we now have a new pull request, let's continue the discussion from https://github.com/scalameta/scalameta/pull/988 here:

> With the previous vfs design, we had Array[Byte] + s.Database + m.Database loaded into memory for all .semanticdb files at some point. This is prohibitively expensive when running large-scale analysis. We want to avoid materializing an Array[Byte] entirely. Protos can be loaded in a streaming fashion, tag by tag, and that's what this refactoring did.

I suggest we don't make claims about performance without benchmarks. Our execution environments are very complex, to the degree when a human has almost zero chance to accurately estimate performance effects of even simplest changes. Therefore, I don't think it's a good idea to change otherwise adequate code on a hunch that it will improve performance.

Now, there can definitely be other reasons for changing code. Fixing bugs, adding new features, making code more maintainable, etc - there is a multitude of valid reasons for change. Let's discuss the proposed refactoring of the semanticdb engine within that framework. 

Could you describe what makes you unhappy about the old implementation of the engine? Is that a lot of ceremony associated with going through multiple layers of the data model? Is that deeply nested code in converters? Something else entirely? Please share your thoughts.

xeno-by:
LGTM

PR #1007, opened 2017-07-07
Make nio.Path and AbsolutePath parseable, fixes #995.

### LINKED ISSUE #995: java.nio.file.Path should be parseable
@olafurpg What other JDK 1.7+ objects would we want to be parseable?

olafurpg:


scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
```diff
@@ -61,6 +62,7 @@ object Input {
     def apply(path: AbsolutePath): Input.File = apply(path, Charset.forName("UTF-8"))
     def apply(file: java.io.File, charset: Charset): Input.File = apply(AbsolutePath(file), charset)
     def apply(file: java.io.File): Input.File = apply(file, Charset.forName("UTF-8"))
+    def apply(path: nio.Path): Input.File = apply(AbsolutePath(path))
```
xeno-by: How about a charset-based overload?

PR #1010, opened 2017-07-09
Upgrade sbt plugins.

olafurpg:
Most notably, use the latest sbt-scalatex unblocking us from wiping out
the scalameta/maven bintray repository.


xeno-by:
LGTM

PR #1017, opened 2017-07-11
Skip bintrayRelease when custom repo is defined.

olafurpg:


build.sbt
```diff
@@ -555,6 +555,7 @@ lazy val publishableSettings = Def.settings(
   publishArtifact.in(Compile) := true,
   publishArtifact.in(Test) := false,
   publishMavenStyle := true,
+  bintrayReleaseOnPublish := !isCustomRepository,
```
olafurpg: Can you validate this fixes the custom repository setup @xeno-by ?

xeno-by: Yes, this works for us at Twitter. 

I'm getting a new warning saying `[warn] You must run bintrayRelease once all artifacts are staged`, but I can live with it.

PR #1021, opened 2017-07-12
Update parser property tests to latest changes in trees.

olafurpg:
I was nervous to find out if the recent fixes in the parser introduced regression. We stopped running these tests on every PR after moving to Travis CI. Turns out there are no regressions, and some bugs have been fixed along the way 🎉

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -71,6 +71,7 @@ testOnlyJVM := {
   val runBenchmarkTests = test.in(benchmarks, Test).value
   val runContribTests = test.in(contribJVM, Test).value
   val runTests = test.in(testsJVM, Test).value
+  val propertyTests = compile.in(testkit, Test).value
   val runDocs = test.in(readme).value
 }
 testOnlyJS := {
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -71,6 +71,7 @@ testOnlyJVM := {
   val runBenchmarkTests = test.in(benchmarks, Test).value
   val runContribTests = test.in(contribJVM, Test).value
   val runTests = test.in(testsJVM, Test).value
+  val propertyTests = compile.in(testkit, Test).value
```
olafurpg: Just checking that these compile, as long as they compile we should be able to use git bisect to track down regressions in case something goes wrong.

olafurpg:
The test failure is bogus, I've noticed that the --validate-links check is failing more often than usual lately.

scalameta/testkit/src/test/scala/scala/meta/testkit/ScalametaParserProperties.scala
```diff
@@ -78,7 +78,7 @@ object ScalametaParserPropertyTest extends FunSuiteLike {
     val prettyPrinterBroken = result.count(_._2.kind == PrettyPrinterBroken)
     println(s"""Parser broken: $parserProken
                |Pretty printer broken: $prettyPrinterBroken""".stripMargin)
-    assert(parserProken <= 7)
-    assert(prettyPrinterBroken <= 1922)
+    assert(parserProken <= 4)
```
xeno-by: `Proken`? :)

olafurpg: Fixed, although this change is unrelated to the PR.

Issue #677, opened 2017-02-11
automatic release on tag


xeno-by:
@olafurpg says that this is already a thing in drone: `event: when: [tag]`, and with our recent build refactoring we're almost there infrastructurally. Looks legit.

olafurpg:
This may give hints how to publish to sonatype from travis

- https://github.com/lucidsoftware/neo-sbt-scalafmt/blob/7321eeddbf52f2a9e3ffa44e061eebeb3d1febb4/build.sbt#L18
- https://github.com/lucidsoftware/neo-sbt-scalafmt/blob/7321eeddbf52f2a9e3ffa44e061eebeb3d1febb4/.travis.yml#L22-L29

olafurpg:
Removing this from milestone, we can release stable releases manually for now.

olafurpg:
Fixed in #1029, 2.0.0-M1 was released with tag push 👏

PR #1029, opened 2017-07-14
Setup build to publish to sonatype on tag push #677

olafurpg:
If this turns out to work well, then I propose we ditch the bintray
pre-releases altogether and more regularly publish milestone releases
by demand.


olafurpg:
Gonna merge since the real test happens on tag push.

PR #1030, opened 2017-07-14
Fix publish on tag push, try 2.

olafurpg:
- Use custom pgp passphrase.
- Handle the case when TRAVIS_TAG exists sys.env but is empty.
- Only run pgp setup for sonatype release.


olafurpg:
The only way to test this is to merge. Let's first double check that sbt shell fires up OK.

PR #1032, opened 2017-07-14
Remove os.version in favor of sbt-dynver.

olafurpg:
The first attempt to publish on push tag failed due to a bug in
os.version. This commit removes os.version in favor of using
sbt-dynver, shrinking the scalameta build.

### DIFF
--- a/readme/Readme.scala
+++ b/readme/Readme.scala
@@ -5,6 +5,7 @@ import java.util.Calendar
 import scala.util.Try
 import scalatags.Text.all._
 import org.scalameta.os._
+import org.scalameta.BuildInfo
 import scala.compat.Platform.EOL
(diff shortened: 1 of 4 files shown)


readme/Readme.scala
```diff
@@ -86,11 +87,17 @@ object Readme {
     pairs(List(left, right).map(x => half(hl.scala(x))): _*)
 
   def stableVersionString = {
-    version.stable()
+    val stdout = shell.check_output("git tag -l v*")
+    val StableVersion = """v(\d+.\d+.\d+)""".r
+    stdout
```
olafurpg: I verified locally that this resolves to 1.8.0 when the latest tag is v2.0.0-M1

PR #1034, opened 2017-07-15
Implement PlatformPathIO for the browser, fixes #1031.

### LINKED ISSUE #1031: PlatformPathIO.js should work in the browser
I upgraded to meta v2 in metadoc and hit on this issue

```
TypeError: Cannot read property 'sep' of undefined
    at $c_s_meta_internal_io_PlatformPathIO$.fileSeparatorChar__C (metadoc-js-fastopt-bundle.js:21618)
    at $c_s_meta_internal_io_PathIO$.fromUnix__T__T (metadoc-js-fastopt-bundle.js:21533)
```

The problem is that PlatformPathIO assumes a Node.js environment. In this case, we should fallback to `/`, but more generally we should maybe add some tests reading semanticdb bytes in a non-nodejs environment if that's possible

olafurpg:
@jonas any suggestion on how to fix the "module not found error"?

### DIFF
--- a/scalameta/io/js/src/main/scala/scala/meta/internal/io/JSIO.scala
+++ b/scalameta/io/js/src/main/scala/scala/meta/internal/io/JSIO.scala
@@ -78,7 +78,25 @@ object JSPath extends js.Any {
 }
 
 object JSIO {
-  private[io] def isNode = JSFs != null
+  def sep: String =
+    if (isNode) JSPath.sep
+    else "/"
+  def delimiter: String =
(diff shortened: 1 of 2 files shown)


jonas:
The only idea I can think of is to load the modules lazily with require if `isNode`. So instead of using`@JSImport` we could use `js.Dynamic.global.require(...)`.

However, maybe Scala.js has some linker option that allows to replace modules.

olafurpg:
Closing this in favor of dynamically loading the modules.

Issue #1039, opened 2017-07-17
Enable scalafix RemoveUnusedImports in the build


olafurpg:
This is currently blocked by https://github.com/scalameta/scalameta/issues/924 in addition to ability to demote/promote particular messages to info/warn/error. Currently, -Xfatal-warnings promotes unused import warnings to errors which makes it impossible to use scalafix to remove the unused imports.

Another issue that needs to be solved is how to configure the demotion/promotion of particular messages. Ideally, scalahost and sbt-messagehost should be able to use the same classification/configuration.

xeno-by:
As discussed with @olafurpg, we wouldn't like to introduce a circular dependency between Scalameta and Scalafix builds. As a result, I'll be closing this issue.

PR #1040, opened 2017-07-17
Remove unused imports.

olafurpg:
Was gonna work on #1039 but hit on several problems that prevent us from
enabling this by default.

### DIFF
--- a/scalahost/nsc/src/main/scala/scala/meta/internal/semantic/AttributesOps.scala
+++ b/scalahost/nsc/src/main/scala/scala/meta/internal/semantic/AttributesOps.scala
@@ -1,8 +1,6 @@
 package scala.meta.internal
 package semantic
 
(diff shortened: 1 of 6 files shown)


xeno-by:
Is the failure spurious?

olafurpg:
Yes the error is spurious, I restarted the job but we should be safe to merge.

olafurpg:
Restarting still reproduces

> There is insufficient memory for the Java Runtime Environment to continue.

I wonder why this starts now, maybe Travis CI is under heavy usage. I propose we let this wait a bit.

olafurpg:
The failures consistently happen for 2.12.2 with JS=true. Interesting, we may want to delete that entry of the build in case the failures continue. We should then investigate the option of getting more powerful machines to run our CI.

PR #1043, opened 2017-07-18
Add Denotation.isJavaDefined, fixes #1042.

### LINKED ISSUE #1042: Denotation.isJavaDefined
See http://dotty.epfl.ch/docs/reference/dropped/auto-apply.html

> But where Scala allows that freedom for all method references, Dotty restricts it to references of external methods that are not defined themselves in Dotty.

In order to know whether to auto-apply a nullary method call in scalafix we need to be able to query if a denotation was defined by javac or not. There is a JAVA flag in the reflect api for this https://github.com/scala/scala/blob/ba9dacc2066dfbffdffc4dd89d512859b23db5a4/src/reflect/scala/reflect/internal/Flags.scala#L108

olafurpg:


scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Flags.scala
```diff
@@ -27,6 +27,7 @@ private[semantic] trait Flags {
   final val COVARIANT: Long = 1 << 22
   final val CONTRAVARIANT: Long = 1 << 23
   final val INLINE: Long = 1 << 24
+  final val JAVA: Long = 1 << 25
```
xeno-by: `JAVADEFINED`?

scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
```diff
@@ -528,4 +528,16 @@ class SemanticSuite extends DatabaseSuite(SemanticdbMode.Slim) {
     }
   )
 
+  targeted(
+    """
+    object isJavaDefined {
```
xeno-by: We really need to come up with a way to generate unique ids for top-level names in these tests.

olafurpg:
The failure on 2.12.2 with JS=true is caused by the low memory constraints on Travis and unrelated to this PR.

PR #1044, opened 2017-07-18
Remove sbt-scalahost.

olafurpg:
The motivation for this change is that the plugin is no longer holding
its weight.

- the automatic javaOptions setup was not useful
- the automatic -Xplugin:scalahost.jar setup was flaky
- the scalametaSemanticdb/scalametaSourceroot settings
  were useful but not providing enough value to justify
  the added complexity to the build.

Previously, users did this to use scalahost

```scala
// project/plugins.sbt
addSbtPlugin("org.scalameta" % "sbt-scalahost" % "2.0")
// build.sbt (optional)
scalametaSemanticdb := ScalametaSemanticdb.Fat
scalametaSourceroot := sourceDirectory.in(Compile).value
```

Now they can do instead
```scala
// build.sbt
addCompilerPlugin("org.scalameta" % "scalahost" % "2.0" cross CrossVersion.full)
scalacOptions ++= List(
  "-Yrangepos", // (required)
  // (optional)
  "-Xplugin-require:scalahost",
  "-P:scalahost:semanticdb:fat",
  s"-P:scalahost:sourceroot:${sourceDirectory.in(Compile).value}"
)
```

Since most builds have a specific place where scalacOptions and compiler
plugins are added, this feels more idiomatic.


.travis.yml
```diff
@@ -25,9 +25,6 @@ env:
     - CI_TEST: ci-fast
       CI_SCALA_VERSION: 2.12.2
       CI_SCALA_JS: true
-    - CI_TEST: ci-sbt-scalahost
-      CI_SCALA_VERSION: 2.10.6
-      CI_PUBLISH: true
```
xeno-by: 👍

olafurpg:
CI is green :shipit:

PR #1052, opened 2017-07-26
Adding denotation position

adam-singer:
Including position to the `Denotation` object

### DIFF
--- a/scalahost/nsc/src/main/scala/scala/meta/internal/semantic/DenotationOps.scala
+++ b/scalahost/nsc/src/main/scala/scala/meta/internal/semantic/DenotationOps.scala
@@ -92,7 +92,7 @@ trait DenotationOps { self: DatabaseOps =>
     }
 
     def toDenotation: m.Denotation = {
-      m.Denotation(flags, name, info)
+      m.Denotation(flags, name, info, gsym.pos.toMeta)
     }
   }
 }
--- a/scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Denotation.scala
+++ b/scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Denotation.scala
@@ -2,9 +2,11 @@ package scala.meta
 package semantic
 
 import org.scalameta.data._
+import scala.meta.inputs._
+import scala.meta.internal.inputs._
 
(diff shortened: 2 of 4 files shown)


scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Denotation.scala
```diff
@@ -2,9 +2,11 @@ package scala.meta
...
 
 import org.scalameta.data._
+import scala.meta.inputs._
+import scala.meta.internal.inputs._
 
-@data class Denotation(flags: Long, name: String, info: String) extends HasFlags {
-  def syntax = s"$flagSyntax $name" + (if (info != "") ": " + info else "")
-  def structure = s"""Denotation($flagStructure, "$name", "$info")"""
+@data class Denotation(flags: Long, name: String, info: String, position: Position) extends HasFlags {
+  def syntax = s"$flagSyntax $name ${position.syntax}" + (if (info != "") ": " + info else "")
```
xeno-by: I saw some printouts of `.syntax` in https://travis-ci.org/scalameta/scalameta/jobs/257894597, and I'd like to propose changes to this method:
  1. How about we don't print `position` at all if it's `None`?
  1. How about we add `position` to the very end of `syntax`, not in the middle?

scalahost/nsc/src/main/scala/scala/meta/internal/semantic/DenotationOps.scala
```diff
@@ -92,7 +92,7 @@ trait DenotationOps { self: DatabaseOps =>
     }
 
     def toDenotation: m.Denotation = {
-      m.Denotation(flags, name, info)
+      m.Denotation(flags, name, info, gsym.pos.toMeta)
```
xeno-by: Judging by the output of https://travis-ci.org/scalameta/scalameta/jobs/257894597, `Symbol.pos` is not quite what we need. Looks like in the Scala compiler, this position is the range of the entire definition, whereas in Scalameta we want only the range of the name of the definition.

xeno-by: This seems to be a blocker for us, so I'm closing this pull request.

Issue #991, opened 2017-07-05
Crosscompile semantic to 2.10/2.11/2.12


olafurpg:
I've made successful experiments with generating semanticdb files from 2.10 in https://github.com/scalameta/sbthost and messagehost (unpublished sbt plugin that emits only messages from compilerReporter). Currently, I copy-paste semanticdb.proto around and setup sbt-scalapb in all projects. It would be nice if the proto schema was cross-released in a `semantic-schema` package for 2.10-2.12. In the future, it may even be possible to publish this module for dotty when https://github.com/lampepfl/dotty/issues/2782 is fixed.

olafurpg:
Fixed in #1065

PR #1055, opened 2017-07-28
some thoughts after hacking on metaide

xeno-by:
Almost solves #991.


scalameta/semantic/shared/src/main/scala/scala/meta/semantic/Api.scala
```diff
@@ -1,119 +1,9 @@
-package scala.meta
-package semantic
-
-import org.scalameta.unreachable
-import org.scalameta.debug
-import scala.meta.inputs._
-import scala.meta.prettyprinters._
-import scala.meta.parsers.{XtensionParsersDialectInput, XtensionParseDialectInput}
-
-private[meta] trait Api extends Flags {
```
olafurpg: Can we keep this in scalameta/semantic? I find them useful.

xeno-by: Everything that was in scalameta/semantic is still there.

```
package star.meta.internal.semanticdb {
  trait Api extends star.meta.semanticdb.Api
  trait Aliases extends star.meta.semanticdb.Aliases
}

package scala.meta.semantic {
  private[meta] trait Api extends star.meta.internal.semanticdb.Api
  private[meta] trait Aliases extends star.meta.internal.semanticdb.Aliases
}
```

starmeta/semanticdb/shared/src/main/protobuf/semanticdb.proto
```diff
@@ -11,19 +11,20 @@ message Attributes {
...
+  repeated ResolvedSymbol symbols = 4;
   repeated Sugar sugars = 6;
 }
 
 message ResolvedName {
-  Range range = 1;
-  string symbol = 2;
+  Position pos = 1;
+  string sym = 2;
+  bool is_binder = 3;
```
xeno-by: This simplifies metaide very very much and is almost mandatory when semanticdb becomes polyglot (because then it is not easy to have a parser at hand for everything that you read).

scalahost/nsc/src/test/scala/scala/meta/tests/scalahost/mirrors/SemanticSuite.scala
```diff
@@ -534,9 +535,11 @@ class SemanticSuite extends DatabaseSuite(SemanticdbMode.Slim) {
       new Object().<<toString>>
       List(1).<<toString>>
     }
-    """, { implicit db => (objectToString, listToString) =>
-      assert(objectToString.symbol.denot.isJavaDefined)
-      assert(!listToString.symbol.denot.isJavaDefined)
+    """, { (db, objectToString, listToString) =>
+      val denot1 = db.symbols.find(_.sym == objectToString).get.denot
```
xeno-by: Let's discuss a good way to expose this API. I'd probably like to do without implicits this time.

xeno-by:
@olafurpg Alternative names that have come to mind now: `polyglotmeta`, `polymeta`, `coremeta`.

ShaneDelmore:
I’m thinking on naming as well. I’m not liking star (although I see the intent).  Poly, core, or generic feel better to me.  The other idea is pushing meta up to top level instead of language.

Instead of star.meta.semantic what about:
  - meta.printers //no need for poly or core, just leave out language
  - meta.scala.semantic
  - meta.otherlang.semantic
  - meta.core.foo //alternatively replace lang with core/generic/poly
  - meta.generic.bar

ShaneDelmore:
👍 on the improved naming

PR #1065, opened 2017-08-14
Cross-build langmeta to 2.10.

olafurpg:
This allows semanticdb-sbt (previously sbthost) to depend on it directly
instead of duplicating protobuf generation.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -35,11 +35,11 @@ commands += CiCommand("ci-slow")(
   "testkit/test:runMain scala.meta.testkit.ScalametaParserPropertyTest" ::
   Nil
 )
-commands += CiCommand("ci-publish")(
-  if (isCiPublish && isTagPush) s"publishSigned" :: Nil
-  else if (isCiPublish) s"publish" :: Nil
-  else Nil
-)
+commands += Command.command("ci-publish") { s =>
+  if (isCiPublish && isTagPush) s"very publishSigned" :: s
+  else if (isCiPublish) s"very publish" :: s
+  else s
+}
 // NOTE: These tasks are aliased here in order to support running "tests/test"
 // from a command. Running "test" inside a command will trigger the `test` task
 // to run in all defined modules, including ones like inputs/io/dialects which
--- a/build.sbt
+++ b/build.sbt
@@ -84,10 +84,15 @@ console := console.in(scalametaJVM, Compile).value
 
 /** ======================== LANGMETA ======================== **/
 
+lazy val langmetaSettings = List(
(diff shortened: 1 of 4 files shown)


olafurpg:
I took the liberty to publish 2.0.0-M2 for langmeta_2.10 to unblock my work on sbtfix. We currently don't run tests for langmeta on 2.10, which we should try to do before merging this PR.

build.sbt
```diff
@@ -137,6 +145,11 @@ lazy val langmeta = crossProject
     langmetaIo,
     langmetaSemanticdb
   )
+  .aggregate(
+    langmetaInputs,
+    langmetaIo,
+    langmetaSemanticdb
+  )
```
xeno-by: Why do we need to aggregate here?

langmeta/semanticdb/shared/src/main/scala/star/meta/semanticdb/Symbol.scala
```diff
@@ -11,7 +11,7 @@ sealed trait Symbol extends Product {
 object Symbol {
   case object None extends Symbol {
     override def toString = syntax
-    override def syntax = s""
+    override def syntax = ""
```
xeno-by: :D

build.sbt
```diff
@@ -84,10 +84,15 @@ console := console.in(scalametaJVM, Compile).value
 
 /** ======================== LANGMETA ======================== **/
 
+lazy val langmetaSettings = List(
+  crossScalaVersions := List(LatestScala210, LatestScala211, LatestScala212)
+)
```
xeno-by: I'm very surprised it's so easy!!

martijnhoekstra:
The appveyor build failures can be worked around with the approach in https://github.com/scalameta/scalameta/pull/1069

The cause seems to be that for some reason the output to stdout of the powershell command is taken as input for powershell. That fails. As to *why* that happens, I have no idea, but eating the error and continuing makes the build succeed with no signficicant issues other than a spurious error output.

If the command would actually fail, it would also try to continue, but fail on the next line, so the error pops up in the wrong place, but not somewhere completely unrelated; they will be right next to each other in the logs.

it's probably easiest to just manually bring those changes to this PR.

olafurpg:
Thank you so much @martijnhoekstra I responded in #1069. We can rebase this PR on master once it's been merged :)

build.sbt
olafurpg: I aggregated here so I can do `langmeta/publish`, but it's inconsistent with the rest of the build so I'll remove it. I feel like the build could benefit from sprinkling aggregate in some other projects too like `scalameta`. but it's a lot of boilerplate.

PR #1074, opened 2017-08-21
Hack to fully qualify magic names.

olafurpg:
The scalac pretty printer shortens the names for several symbols by
stripping out the prefix. This causes bugs like
https://github.com/scalacenter/scalafix/issues/254

This commit adds a special case in DenotationOps to add the prefix when
the printed name contains no `.`. I spent a while trying to fix this at
the root by extending the pretty printer, but gave up eventually in
favor of this hack.


olafurpg:
A proper solution would be to 
- roll our own pretty printer for types (scalac code printer uses Type.toString, which is not extendable)
- Make Denotation.info parseable
- emit AttributedSugars for resolved names in Denotation.info
- add Input.Denotation to position the resolved names of printed Denotation.info

olafurpg:
I've started working on the correct approach which solves the hack in this PR.

PR #1076, opened 2017-08-22
Add semantic information for denotations

### LINKED ISSUE #1015: Semantic information for Database.denotations
Follow up of #822. Currently it's possible to resolve a few names in Database.sugars but not in Database.denotations. It would be nice if for example it was possible to attribute references in Denotation.info

olafurpg:
Fixes #1015. Note, this PR does not add Type.Method since it's not necessary for me at the moment in scalafix. Type.Method can be added separately to make Denotation.info parseable.

### DIFF
--- a/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/AttributesOps.scala
+++ b/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/AttributesOps.scala
@@ -165,11 +165,10 @@ trait AttributesOps { self: DatabaseOps =>
 
                 names(mtree.pos) = symbol
                 if (mtree.isBinder) binders += mtree.pos
-                denotations(symbol) = gsym.toDenotation(symbol)
+                denotations(symbol) = gsym.toDenotation
                 if (gsym.isClass && !gsym.isTrait) {
                   val gprim = gsym.primaryConstructor
-                  val symbol = gprim.toSemantic
-                  denotations(symbol) = gprim.toDenotation(symbol)
+                  denotations(gprim.toSemantic) = gprim.toDenotation
                 }
                 todo -= mtree
 
--- a/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/AttributesOps.scala
+++ b/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/AttributesOps.scala
@@ -165,10 +165,11 @@ trait AttributesOps { self: DatabaseOps =>
 
                 names(mtree.pos) = symbol
                 if (mtree.isBinder) binders += mtree.pos
-                denotations(symbol) = gsym.toDenotation
+                denotations(symbol) = gsym.toDenotation(symbol)
                 if (gsym.isClass && !gsym.isTrait) {
                   val gprim = gsym.primaryConstructor
-                  denotations(gprim.toSemantic) = gprim.toDenotation
+                  val symbol = gprim.toSemantic
+                  denotations(symbol) = gprim.toDenotation(symbol)
                 }
                 todo -= mtree
 
--- a/scalameta/tests/jvm/src/test/resources/semanticdb.expect
+++ b/scalameta/tests/jvm/src/test/resources/semanticdb.expect
@@ -29,6 +29,7 @@ Messages:
 Symbols:
 _root_.example. => package example
 _root_.example.Example. => final object Example
+<<<<<<< 9dd45b2c41b09f1d3d5fa79bb7aa959242bf107b
 _root_.example.Example.main([Ljava/lang/String;)V. => def main: (args: Array[String])Unit
 _root_.example.Example.main([Ljava/lang/String;)V.(args) => param args: Array[String]
 _root_.scala. => package scala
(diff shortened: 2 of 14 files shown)


olafurpg:
I don't like Input.Denotation. We could use Input.VirtualFile since it has identical fields but it's a bit awkward to overload it's meaning. The original idea with LabeledString was to not make assumptions about the key/label.

Another options is to remove sealed from `Input`.

xeno-by:
Can we have `Input.Invisible` shared with sugars?

scalameta/tests/jvm/src/test/scala/scala/meta/tests/SemanticdbExpectSuite.scala
```diff
@@ -16,7 +17,7 @@ class SemanticdbExpectSuite extends FunSuite with DiffAssertions {
       // later down the road if that turns out to be useful.
       case "2" :: "12" :: Nil =>
         val obtained = SemanticdbExpectSuite.loadDatabase.toString
-        val expected = new String(Files.readAllBytes(SemanticdbExpectSuite.expectPath))
```
xeno-by: What encoding does this code use?

scalameta/tests/jvm/src/test/resources/semanticdb.expect
```diff
@@ -16,14 +16,30 @@ Names:
...
 _root_.example. => package example
 _root_.example.Example. => final object Example
-_root_.example.Example.main([Ljava/lang/String;)V. => def main: (args: Array[String])Unit
-_root_.example.Example.main([Ljava/lang/String;)V.(args) => param args: scala.Array[String]
+_root_.example.Example.main([Ljava/lang/String;)V. => def main: (args: scala.Array[scala.Predef.String])scala.Unit
+  [46..50): Unit => _root_.scala.Unit#
+  [13..18): Array => _root_.scala.Array#
+  [19..24): scala => _root_.scala.
+  [25..31): Predef => _root_.scala.Predef.
+  [32..38): String => _root_.scala.Predef.String#
```
xeno-by: Looks like not all names from info syntax are not covered here. Is this intentional?

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/AttributesOps.scala
```diff
@@ -165,10 +165,11 @@ trait AttributesOps { self: DatabaseOps =>
 
                 names(mtree.pos) = symbol
                 if (mtree.isBinder) binders += mtree.pos
-                denotations(symbol) = gsym.toDenotation
+                denotations(symbol) = gsym.toDenotation(symbol)
```
xeno-by: Can't we just recompute `symbol` on the fly inside `toDenotation`?

scalameta/tests/jvm/src/test/scala/scala/meta/tests/SemanticdbExpectSuite.scala
olafurpg: The default encoding from sys.props. FileIO.slurp uses utf-8

scalameta/tests/jvm/src/test/resources/semanticdb.expect
olafurpg: Yes, I intentionally left out for example the symbols of parameter names since I'm not convinced they're useful. We can add more resolved names later after demand.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/AttributesOps.scala
olafurpg: I don't see a reason to compute `Symbol.toSemantic` multiple times. This is not because of performance reasons, it pollutes the console when println debugging and the same method is invoked twice with the same arguments.

olafurpg:
I agree it would be nice to have a shared model for sugars and denotation inputs. However, their shapes/keys are totally different

- Input.Sugar is keyed by a Position.Range
- Input.Denotation is keyed by a Symbol

We could add an `Input.Hidden[T](key: T, value: String)`, but we might as well unseal Input at that point instead.

I'm leaning towards one of

- unchanged, but I don't like the fact that Input.Denotation.symbol is a String and not Symbol. I can also imagine that we will continue adding more specific Inputs, for example Input.Macro.
- unseal, but that makes `Input` more complicated to reason about since it can be anything at runtime.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/AttributesOps.scala
xeno-by: I'm not sure it's worth it to add complications to the signature of `toDenotation` just because we want to simplify debugging. Once I saw this change, I had to stop myself and see why you need to pass the symbol explicitly, and the answer was non-obvious. I wouldn't want to impose this comprehension cost on others reading this code (including us in the future).

How about we cache the results of `toSemantic` if you feel strongly about performing the computations only once?

xeno-by:
Alright, let's keep Input.Denotation unchanged.

olafurpg:
Are we good to merge this?

PR #1085, opened 2017-08-24
remove UTF-8 only overload of slurp

martijnhoekstra:
Assuming UTF-8 is a dangerous game that often depends on local setup.
The convenience it brings rarely outweighs the bugs that go along with
it


olafurpg:
Travis is green but check is still yellow. Merging!

PR #1088, opened 2017-08-24
Make tests succeed on Windows

martijnhoekstra:
Mostly newline funny business. Multiline strings are as-is and don't
aquire different line endings on Windows.

It also seems like the platform-specific line endings in
Position.format aren't actually being expanded to platform specific line
endings.

That shouldn't be causing any actual harm, so the tests just plaster
over it in an "whichever way makes it succeed" kind of way - (almost?)
all Windows tooling doesn't care much about whether newlines are \r\n or
\n

Other than that, "\foo\bar" is not an absolute path on Windows

Leaves two tests failing, of which I'm not actually sure the failures
are spurious.

### DIFF
--- a/langmeta/io/js/src/main/scala/java/io/File.scala
+++ b/langmeta/io/js/src/main/scala/java/io/File.scala
@@ -44,6 +44,10 @@ class File(path: String) {
 }
 
 object File {
+  def listRoots() = 
+    if (JSIO.isNode) List(JSIO.path.parse(JSIO.path.resolve()).root)
+    else List(new File("/"))
+
   def separatorChar: Char =
     separator.charAt(0)
 
--- a/langmeta/io/js/src/main/scala/java/io/File.scala
+++ b/langmeta/io/js/src/main/scala/java/io/File.scala
@@ -44,9 +44,12 @@ class File(path: String) {
 }
 
 object File {
-  def listRoots() = 
-    if (JSIO.isNode) List(JSIO.path.parse(JSIO.path.resolve()).root)
-    else List(new File("/"))
+  def listRoots(): Array[File] = Array(
+    new File(
+      if (JSIO.isNode) JSIO.path.parse(JSIO.path.resolve()).root
+      else "/"
+    )
+  )
 
   def separatorChar: Char =
     separator.charAt(0)
--- a/scalameta/scalameta/jvm/src/test/scala/scala/meta/tests/io/MultipathSuite.scala
+++ b/scalameta/scalameta/jvm/src/test/scala/scala/meta/tests/io/MultipathSuite.scala
@@ -40,7 +40,6 @@ class MultipathSuite extends FunSuite {
   }
 
   test("Classpath.syntax") {
-    assert(Classpath("/Foo/Bar.jar").syntax == "/Foo/Bar.jar")
     val fromPath = Classpath(List(AbsolutePath(tmp)))
     val fromSyntax = Classpath(tmp.getAbsolutePath)
     assert(fromPath == fromSyntax)
--- a/scalameta/scalameta/shared/src/test/scala/scala/meta/tests/inputs/FormatMessageSuite.scala
+++ b/scalameta/scalameta/shared/src/test/scala/scala/meta/tests/inputs/FormatMessageSuite.scala
@@ -22,7 +22,7 @@ class FormatMessageSuite extends FunSuite {
     |<input>:1: error: foo
     |
     |^
-  """.trim.stripMargin)
+  """.trim.stripMargin.split('\n').mkString(EOL))
 
   test("\n")("""
     |<input>:1: error: foo
@@ -31,7 +31,7 @@ class FormatMessageSuite extends FunSuite {
     |<input>:2: error: foo
     |
     |^
-  """.trim.stripMargin)
+  """.trim.stripMargin.split('\n').mkString(EOL))
 
   test("foo")("""
     |<input>:1: error: foo
@@ -46,7 +46,7 @@ class FormatMessageSuite extends FunSuite {
     |<input>:1: error: foo
     |foo
     |   ^
-  """.trim.stripMargin)
+  """.trim.stripMargin.split('\n').mkString(EOL))
 
   test("foo\n")("""
     |<input>:1: error: foo
@@ -64,7 +64,7 @@ class FormatMessageSuite extends FunSuite {
     |<input>:2: error: foo
     |
     |^
-  """.trim.stripMargin)
+  """.trim.stripMargin.split('\n').mkString(EOL))
 
   test("foo\nbar")("""
     |<input>:1: error: foo
@@ -91,5 +91,5 @@ class FormatMessageSuite extends FunSuite {
     |<input>:2: error: foo
     |bar
     |   ^
-  """.trim.stripMargin)
+  """.trim.stripMargin.split('\n').mkString(EOL))
 }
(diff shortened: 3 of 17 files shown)


martijnhoekstra:
the scala-js failures don't seem spurious. I'll see if I can find out what's wrong/fix it

martijnhoekstra:
I'm afraid I'm going to need a hand here - I don't understand why `File.listRoots` isn't found - and if it is, whether my approach is feasible.

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/inputs/FormatMessageSuite.scala
```diff
@@ -22,7 +22,7 @@ class FormatMessageSuite extends FunSuite {
     |<input>:1: error: foo
     |
     |^
-  """.trim.stripMargin)
+  """.trim.stripMargin.split('\n').mkString(EOL))
```
olafurpg: Would it make sense to move this to `.multiline` extension method that can be reused for all tests?

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
```diff
@@ -436,7 +436,7 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
   test("package foo; class C; package baz { class D }") {
     val tree = source("package foo; class C; package baz { class D }")
     assert(tree.structure === "Source(List(Pkg(Term.Name(\"foo\"), List(Defn.Class(Nil, Type.Name(\"C\"), Nil, Ctor.Primary(Nil, Name(\"\"), Nil), Template(Nil, Nil, Self(Name(\"\"), None), Nil)), Pkg(Term.Name(\"baz\"), List(Defn.Class(Nil, Type.Name(\"D\"), Nil, Ctor.Primary(Nil, Name(\"\"), Nil), Template(Nil, Nil, Self(Name(\"\"), None), Nil))))))))")
-    assert(tree.syntax === "package foo\nclass C\npackage baz {\n  class D\n}")
+    assert(tree.syntax === s"package foo${EOL}class C${EOL}package baz {$EOL  class D${EOL}}")
```
olafurpg: I find it a bit unsettling that `tree.syntax` is OS dependent, would it make sense if TreeSyntax emitted \n instead of EOL?

scalameta/scalameta/jvm/src/test/scala/scala/meta/tests/io/MultipathSuite.scala
```diff
@@ -40,7 +40,6 @@ class MultipathSuite extends FunSuite {
   }
 
   test("Classpath.syntax") {
-    assert(Classpath("/Foo/Bar.jar").syntax == "/Foo/Bar.jar")
```
olafurpg: 👍 bad test

langmeta/io/js/src/main/scala/java/io/File.scala
```diff
@@ -44,6 +44,10 @@ class File(path: String) {
 }
 
 object File {
+  def listRoots() =
```
olafurpg: Try to add an explicit `: Array[File]` return type, it's probably getting inferred to `List[File]` which produces a different jvm signature that fails to link. The reason it still compiles fine is because the JVM boot classpath is available during compilation, but not linking.

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/inputs/FormatMessageSuite.scala
martijnhoekstra: Maybe factor out the entire comparison to something like `sameLines`?

olafurpg: Agreed, that is even better ^^ 👍

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
martijnhoekstra: I know so little about the actual library that I don't really have an opinion on that in this specific case.

In the general case (perhaps surprisingly), things tend to get more rather then less portable if `\n` is consistently used for everything, as long as notepad isn't the tool of choice of the Windows user (and it never is).

Unfortunately things like `println` are platform specific, making you in some cases end up with mixed line endings if you attempt to do that.

I could experiment with globally replacing EOL with \n and see where that brings us - maybe it removes the need for this PR almost entirely.

One notable issue is git. In some configurations, git on windows will replace \n line endings with \r\n. This is problematic for scala in general, since line endings may in some cases (for example in multiline strings) not be interchangeable.

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/quasiquotes/ErrorSuite.scala
```diff
@@ -4,14 +4,15 @@
...
+import compat.Platform.EOL
 
 class ErrorSuite extends FunSuite {
   test("val q\"type $name[$A] = $B\"") {
     assert(typecheckError("""
       import scala.meta._
       import scala.meta.dialects.Scala211
       val q"type $name[$X] = $Y" = q"type List[+A] = List[A]"
-    """) === """
+    """).replace("\r", "") === """
```
martijnhoekstra: @olafurpg maybe you can shed some light on what's happening here: typecheckError seems to emit mixed line endings. My best guess is that line endings are getting inlined from a different compile-time Platform in XtensionPositionFormatMessage but by best guess isn't very good.

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
olafurpg: Thanks for sharing your view. I propose we don't change the pretty-printing in this PR and focus on getting tests passing in Windows, as you've done. Globally EOL with \n  might be worth it IMO, I'm happy to explore that path separately later.

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/quasiquotes/ErrorSuite.scala
olafurpg: Hmm, I am not sure what is happening here. If typecheckError is defined in scalameta, then I'm happy to move the replace there instead of at every callsite.

olafurpg:
I have an upcoming PR that renames a significant part of the semantic api. See https://github.com/scalameta/scalameta/issues/1091. Do you think we can get the CI green soon? I'm happy to fix the remaining issues if you don't have time now. I would like to get this merged asap to possible rebase conflicts, although I suspect they won't conflict too much.

martijnhoekstra:
I should be able to have the CI green tonight - It's blocked on the JS File stuff, and making workarounds until it passes is a strategy that will get it green eventually, possibly at the cost of sloppyness in the platform-independentness of the JS File API, which could be addressed separately.

olafurpg:
Would you mind if I rebased #1094 on top of this PR and cleaned up the remaining JS issues? I'm preparing today for the upcoming v2.0 release. Your commits would stay unchanged.

martijnhoekstra:
By all means go for it.

martijnhoekstra:
Just to note: the two tests that are still failing on Windows are not the JS tests, they are separate tests that I'm not sure are actually bad tests, or might actually be problems in windows with EOL not being escaped properly in some situations. I didn't properly look in to that yet, and it might not be possible to get appveyor actually run the tests and succeed while remaining sure that the tests test what they should with the looming 2.0 release.

martijnhoekstra:
No worries, just wanted to manage expectations here.

olafurpg:
I'll wait before I advertise a fully green build on appveyor :)

PR #1096, opened 2017-08-28
Show triple-quoted string on all line endings

martijnhoekstra:
This is an "actual" change (even if it pertains only to pretty-printing) rather than a test change, bringing the windows implementation in line with the *nix implementation.

Also expanding the appveyor tests - all tests should now pass.

Review requested on the actual bug fix, whether assertSameLines is in the right place, and whether cleaning up the rest of the tests in terms of assertSameLines should be in this PR or in a separate one.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,7 @@ environment:
 build_script:
   - sbt version # dummy step, a build_script step is required by appveyor.
 test_script:
-  - sbt testsJVM/test
+  - sbt testAll
 cache:
   - C:\Users\appveyor\.m2
   - C:\Users\appveyor\.ivy2
--- a/scalameta/scalameta/shared/src/test/scala/scala/meta/tests/parsers/ParseSuite.scala
+++ b/scalameta/scalameta/shared/src/test/scala/scala/meta/tests/parsers/ParseSuite.scala
@@ -12,6 +12,14 @@ import org.scalameta.logger
 class ParseSuite extends FunSuite with CommonTrees {
   val EOL = scala.compat.Platform.EOL
   val escapedEOL = if (EOL == "\n") """\n""" else """\r\n"""
+
+
+  def assertSameLines(actual: String, expected: String) = {
+    val actualLines = actual.lines.toList
+    val expectedLines = expected.lines.toList
+    assert(actualLines === expectedLines)
+  }
+
   def stat(code: String)(implicit dialect: Dialect) = code.applyRule(_.parseStat())
   def term(code: String)(implicit dialect: Dialect) = code.parseRule(_.expr())
   def pat(code: String)(implicit dialect: Dialect) = code.parseRule(_.pattern())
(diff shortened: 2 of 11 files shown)


appveyor.yml
```diff
@@ -12,7 +12,7 @@ environment:
 build_script:
   - sbt version # dummy step, a build_script step is required by appveyor.
 test_script:
-  - sbt testsJVM/test
+  - sbt testOnlyJVM
```
olafurpg: Do the JS tests also pass on Windows? We can run `testAll`  on Appveyor, it would make sure the Node.js IO facades function as expected on Windows.

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/prettyprinters/PublicSuite.scala
```diff
@@ -292,7 +292,7 @@ class PublicSuite extends FunSuite {
     val defIdentity = Denotation(DEF | FINAL, "identity", info, names)
     assert(defIdentity.toString ===
       """final def identity: [T](e: E)T
-        |  [7..8): E => _root_.E#""".stripMargin
+        |  [7..8): E => _root_.E#""".stripMargin.split('\n').mkString(EOL)
```
olafurpg: My bad, that was a new test I added.

scalameta/scalameta/shared/src/test/scala/scala/meta/tests/parsers/ParseSuite.scala
```diff
@@ -12,6 +12,14 @@ import org.scalameta.logger
 class ParseSuite extends FunSuite with CommonTrees {
   val EOL = scala.compat.Platform.EOL
   val escapedEOL = if (EOL == "\n") """\n""" else """\r\n"""
+
+
+  def assertSameLines(actual: String, expected: String) = {
```
olafurpg: Maybe add a note saying this should eventually be replaced with `DiffAssertions.assertNoDiff`, which is available in scalameta-testkit. It works similarly but produces nice unified diffs on failures. I'm planning to move `scalametaJVM/test` to `testsJVM`, but haven't gotten around to it yet.

appveyor.yml
martijnhoekstra: Alas, 9 failures still. In scope of this PR or a separate one?

olafurpg: This PR already greatly improves the state of the art so we can separate it if you prefer, up to you.

martijnhoekstra: We already split once - If we can have the full test suite pass on Appveyor rather than most of it we can have a real base line and regression prevention. That's probably worth getting it in here.

martijnhoekstra:
This should work for most happy paths.

For pathalogical cases, things might start falling apart. It turns out that writing a robust cross-platform file library that is compatible with java (n)IO, URI's, and js is not trivial.

appveyor.yml
```diff
@@ -12,7 +12,7 @@ environment:
 build_script:
   - sbt version # dummy step, a build_script step is required by appveyor.
 test_script:
-  - sbt testsJVM/test
+  - sbt testAll
```
olafurpg: 🎉

PR #1102, opened 2017-09-01
Adapt semanticdb-scalac to new Type.Lambda/Type.Method syntax.

olafurpg:
- `def x[T](e: Int): T`.signature is `[T] => (e: Int): T`
- Denotation.signature no longer include redundant fully qualified paths,
  to fully qualify use the resolved symbols instead.

### DIFF
--- a/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/PrinterOps.scala
+++ b/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/PrinterOps.scala
@@ -10,8 +10,10 @@ import scala.reflect.internal.ModifierFlags._
 trait PrinterOps { self: DatabaseOps =>
   import g._
 
-  def showSynthetic(tpe: g.Type): AttributedSynthetic = {
-    showSynthetic(g.TypeTree(tpe))
+  def showSynthetic(tpe: g.Type): AttributedSynthetic = tpe match {
+    // Skip signature for abstract type members, e.g. type T <: Int
+    case g.TypeBounds(_, _) => AttributedSynthetic.empty
+    case _ => showSynthetic(g.TypeTree(tpe))
   }
   def showSynthetic(what: g.Tree): AttributedSynthetic = {
     val out = new StringWriter()
@@ -319,17 +321,14 @@ trait PrinterOps { self: DatabaseOps =>
 
     // + scalac deviation
     def printType(tpe: Type): Unit = {
-      def wrapped[T](
-          ts: List[T],
-          open: String = "[",
-          close: String = "]",
-          forceOpen: Boolean = false)(f: T => Unit): Unit = {
+      def wrapped[T](ts: List[T], open: String, close: String, forceOpen: Boolean = false)(
+          f: T => Unit): Unit = {
         if (ts.nonEmpty || forceOpen) {
           this.print(open)
           var first = true
           ts.foreach { t =>
             if (!first) {
-              this.print(",")
+              this.print(", ")
             }
             f(t)
             first = false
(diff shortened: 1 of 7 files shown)


olafurpg:
I restarted appveyor due to a OOM error on Node.js, I hope this was a spurious error.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/PrinterOps.scala
```diff
@@ -10,8 +10,10 @@ import scala.reflect.internal.ModifierFlags._
 trait PrinterOps { self: DatabaseOps =>
   import g._
 
-  def showSynthetic(tpe: g.Type): AttributedSynthetic = {
-    showSynthetic(g.TypeTree(tpe))
+  def showSynthetic(tpe: g.Type): AttributedSynthetic = tpe match {
+    // Skip signature for abstract type members, e.g. type T <: Int
```
xeno-by: Why's that? We can use Type.Bounds here, I think, wrapped in a Type.Lambda if necessary.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/PrinterOps.scala
```diff
@@ -374,10 +362,28 @@ trait PrinterOps { self: DatabaseOps =>
             this.printType(s.typeSignature)
             this.print(s.flagsExplanationString)
           }
+          if (!result.isInstanceOf[MethodType]) {
+            // Only print `: ` before last curried parameter list.
+            this.print(": ")
+          }
           this.printType(result)
+        case NullaryMethodType(result) =>
+          printType(result)
```
xeno-by: How about `: $result`?

scalameta/tests/jvm/src/test/resources/semanticdb.expect
```diff
@@ -34,62 +34,58 @@ Messages:
...
-  [46..50): Unit => _root_.scala.Unit#
-_root_.example.Example.main([Ljava/lang/String;)V.(args) => param args: scala.Array[scala.Predef.String]
-  [6..11): Array => _root_.scala.Array#
-  [12..17): scala => _root_.scala.
-  [18..24): Predef => _root_.scala.Predef.
-  [25..31): String => _root_.scala.Predef.String#
-_root_.example.Example.x. => val x: scala.reflect.ClassTag[scala.Int]
-  [14..22): ClassTag => _root_.scala.reflect.ClassTag#
-  [29..32): Int => _root_.scala.Int#
+_root_.example.Example.main([Ljava/lang/String;)V. => def main: (args: Array[String]): Unit
```
xeno-by: This so sooooooo good.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/PrinterOps.scala
olafurpg: Good idea, but it seems to trigger for `val`s too. I propose we keep it unchanged since this is enough to get ExplicitReturnTypes working for defs, but consider adding this later for `def`s.

scalameta/tests/jvm/src/test/resources/semanticdb.expect
olafurpg: 😄

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/PrinterOps.scala
olafurpg: I thought it would be best to skip these since I have no need for them at the moment, we can handle them later if they turn out to be useful.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/PrinterOps.scala
olafurpg: Should I open an issue for this? I agree the denotation.signature for nullary methods should be `Type.Method(Nil, _)`

PR #1103, opened 2017-09-02
Fix compile errors in 2.10 Scala.js langmeta.

olafurpg:
To prevent further issues,
- travis runs tests on 2.10.6
- no need to cross-build langmeta for JS 2.0.6, only JVM


olafurpg:
FYI @martijnhoekstra dce1a452923bbcb9e87f6fecb0b063d7e22b26f4 moves the io tests from scalametaJVM to langmetaJVM, this should make it nicer to work with in the future :)

olafurpg:
Gonna go ahead and merge this to unblock release. The 2.10 compile errors got caught on tag push.

PR #1107, opened 2017-09-04
Fix #1106.

### LINKED ISSUE #1106: Path dependant types are lost in denotation signatures
```scala
object a {
  class Path {
    class B { class C }
    implicit val x = new B
    implicit val y = new x.C
  }
  implicit val b = new Path().x
}
// obtained
_empty_.a.Path#y. => implicit val y: C
  [0..1): C => _empty_.a.Path#B#C#
// expected
_empty_.a.Path#y. => implicit val y: Path.this.x.C
  [10..11): x => _empty_.a.Path#x.
  [12..13): C => _empty_.a.Path#B#C#
```

olafurpg:
I'm not sure if this solution is acceptable, but it seems to handle the problematic case I could come up with.

### DIFF
--- a/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/PrinterOps.scala
+++ b/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/PrinterOps.scala
@@ -320,6 +320,13 @@ trait PrinterOps { self: DatabaseOps =>
     }
 
     // + scalac deviation
+    object PathDependentType {
+      def unapply(arg: g.Type): Option[g.Symbol] = arg match {
+        case SingleType(ThisType(_), sym) if sym.isTerm && !sym.isModule =>
+          Some(sym)
+        case _ => None
+      }
+    }
     def printType(tpe: Type): Unit = {
       def wrapped[T](ts: List[T], open: String, close: String, forceOpen: Boolean = false)(
           f: T => Unit): Unit = {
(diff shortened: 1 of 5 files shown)


scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/PrinterOps.scala
```diff
@@ -320,6 +320,13 @@ trait PrinterOps { self: DatabaseOps =>
     }
 
     // + scalac deviation
+    object PathDependentType {
+      def unapply(arg: g.Type): Option[g.Symbol] = arg match {
+        case SingleType(ThisType(_), sym) if sym.isTerm && !sym.isModule =>
```
xeno-by: `SingleType(_, sym)`

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/PrinterOps.scala
```diff
@@ -320,6 +320,13 @@ trait PrinterOps { self: DatabaseOps =>
     }
 
     // + scalac deviation
+    object PathDependentType {
```
xeno-by: `PathDependentPrefix`

olafurpg: Done.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/PrinterOps.scala
olafurpg: Done.

PR #1109, opened 2017-09-05
Several type printer fixes

olafurpg:
I'm still hitting on some problems with singleton types on companion objects. Will investigate tomorrow. This PR is ready for merge, regardless.

### DIFF
--- a/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/SymbolOps.scala
+++ b/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/SymbolOps.scala
@@ -18,7 +18,6 @@ trait SymbolOps { self: DatabaseOps =>
         def definitelyGlobal = sym.hasPackageFlag
         def definitelyLocal =
           sym.name.decoded.startsWith(g.nme.LOCALDUMMY_PREFIX) ||
-            sym.name.decoded == g.tpnme.REFINE_CLASS_NAME ||
             (sym.owner.isMethod && !sym.isParameter) ||
             ((sym.owner.isAliasType || sym.owner.isAbstractType) && !sym.isParameter)
         !definitelyGlobal && (definitelyLocal || isLocal(sym.owner))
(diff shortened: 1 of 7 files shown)


scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/SymbolOps.scala
```diff
@@ -18,7 +18,6 @@ trait SymbolOps { self: DatabaseOps =>
         def definitelyGlobal = sym.hasPackageFlag
         def definitelyLocal =
           sym.name.decoded.startsWith(g.nme.LOCALDUMMY_PREFIX) ||
-            sym.name.decoded == g.tpnme.REFINE_CLASS_NAME ||
```
xeno-by: Is this right?

olafurpg:
Appveyor has started to regularly run into OOM errors on Node.js. We should try to change the build script to run `; testOnlyJVM ; testOnlyJS`

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/SymbolOps.scala
olafurpg: Yes, it was a fruitless test comparing Name with String. If we treat refinement symbols as local, then they don't have any denotations. This PR makes the symbols inside the refinement types global so that they match correctly with references to those symbols. Example

```scala
object ab {
  trait Foo
  val x = new Foo {
    val y = 2
    def z[T](e: T) = e
  }
}

object c {
  val y = ab.x.y // resolved name to `y` matches symbol of `y` in denotation.signature inside refinement type
  val z = ab.x.z(2)
}
```

The encoding in this PR is not perfect, it does result in conflicts, see the `y: Int` and `y: Any` test. But it's still a major improvement over the previous encoding.

One motivation for this PR is to be able to narrow refinement types to their parents alone. If no one references the custom decls, then there's no need to keep track of them in the type. I've seen cases in codebases like scalding where tons of public members are defined in a refinement types

```scala
val x = new Foo {
  val blah = ...
 // ...
}
```
that leak into the public signature. It would be nice if scalafix could determine that those symbols are never invoked so we the type of x can be narrowed to `Foo`.

scalameta/semanticdb-scalac/src/test/scala/scala/meta/tests/SemanticSuite.scala
```diff
@@ -771,4 +768,115 @@ class SemanticSuite extends DatabaseSuite(SemanticdbMode.Slim) {
...
+      |  [4..7): Int => _root_.scala.Int#
+    """.stripMargin
+  )
+
+  targeted(
+    """
+      |object ab {
+      |  trait Foo
+      |  val x = new Foo {
+      |    val <<y>> = 2
```
olafurpg: Kudos @xeno-by for `targeted`, it's incredibly powerful in cases like this.

PR #1123, opened 2017-09-20
Fix incorrect names in semanticdb for for-comprehensions

cb372:
Fixes #1037

Adds a special case for the `foreach`/`map`/`flatMap`/`withFilter` symbols that are generated by desugaring of for-comprehensions.

We still add the synthetic symbol to the cache, but later we come along and replace it with the correct symbol.

PR #1125, opened 2017-09-21
Fix #1037.

### LINKED ISSUE #1037: Names are off inside for comprehensions
```scala
    object flatmap {
      for (x <- 1 to 10; y <- 0 until 10) println(x -> x)
      for (i <- 1 to 10; j <- 0 until 10) yield (i, j)
    }

// semanticdb names
[info] - object flatmap {       for (x <- 1 to 10; y <- 0 u... *** FAILED ***
[info]   ===========
[info]   => Obtained
[info]   ===========
[info]   [12..19): flatmap => _empty_.flatmap.
[info]   [33..34): x => <...>@33..34
[info]   [40..42): to => _root_.scala.collection.immutable.Range#foreach(Lscala/Function1;)V.
[info]   [47..48): y => <...>@47..48
[info]   [54..59): until => _root_.scala.collection.immutable.Range#foreach(Lscala/Function1;)V.
[info]   [64..71): println => _root_.scala.Predef.println(Ljava/lang/Object;)V.
[info]   [72..73): x => <...>@33..34
[info]   [74..76): -> => _root_.scala.Predef.ArrowAssoc#`->`(Ljava/lang/Object;)Lscala/Tuple2;.
[info]   [77..78): x => <...>@33..34
[info]   [91..92): i => <...>@91..92
[info]   [98..100): to => _root_.scala.collection.TraversableLike#flatMap(Lscala/Function1;Lscala/collection/generic/CanBuildFrom;)Ljava/lang/Object;.
[info]   [105..106): j => <...>@105..106
[info]   [112..117): until => _root_.scala.collection.TraversableLike#map(Lscala/Function1;Lscala/collection/generic/CanBuildFrom;)Ljava/lang/Object;.
[info]   [129..130): i => <...>@91..92
[info]   [132..133): j => <...>@105..106
```
I expected the names `to` and `until` to match Range.to/until instead of foreach/flatmap/map

olafurpg:
It turns out that we can use the same approach as with synthetic
.apply/.unapply as with .foreach/.flatMap in for comprehensions.

Supersedes #1123 

Review @cb372 @yilinwei

### DIFF
--- a/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/DocumentOps.scala
+++ b/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/DocumentOps.scala
@@ -424,6 +424,10 @@ trait DocumentOps { self: DatabaseOps =>
   private def isSyntheticName(select: g.Select): Boolean =
     select.pos == select.qualifier.pos &&
       (select.name == g.nme.apply ||
+        select.name == g.nme.foreach ||
+        select.name == g.nme.withFilter ||
+        select.name == g.nme.flatMap ||
+        select.name == g.nme.map ||
         select.name == g.nme.unapplySeq ||
         select.name == g.nme.unapply)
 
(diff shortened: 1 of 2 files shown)


cb372:
👍  for reusing an existing approach/hack instead of introducing a new one. LGTM

olafurpg:
The same approach is used in dotty lsp to detect synthetic trees (that is, `b` is synthetic if `a.b` has same position as `a`) so at least we're not the only ones using the same trick.

PR #1121, opened 2017-09-17
Add .fold, .toOption, and .toEither to Parsed and Tokenized ADTs

andyscott:
This adds `fold`, `toOption`, and `toEither` methods to `Parsed`.

Additionally, all methods were updated to be implemented in terms of `fold`.

I haven't added any new tests for `toOption` and `toEither` as I'm still getting familiar with the codebase's organization.

PR #1126, opened 2017-09-21
Enable MiMa checks on every PR.

olafurpg:
The changes in #1121 broke binary compatibility in 2.11 for potential
users who implement the sealed Tokenized/Parsed traits. Those should not
be extended, so we ignore the errors.

### DIFF
--- a/project/Mima.scala
+++ b/project/Mima.scala
@@ -1,0 +1,13 @@
+import com.typesafe.tools.mima.core._
+
+// More details about Mim:
+// https://github.com/typesafehub/migration-manager/wiki/sbt-plugin#basic-usage
+object Mima {
+  val ignoredABIProblems: Seq[ProblemFilter] = {
+    Seq(
+      ProblemFilters.exclude[Problem]("scala.meta.internal.*"),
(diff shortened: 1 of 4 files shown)


project/Mima.scala
```diff
@@ -0,0 +1,13 @@
+import com.typesafe.tools.mima.core._
+
+// More details about Mim:
+// https://github.com/typesafehub/migration-manager/wiki/sbt-plugin#basic-usage
+object Mima {
+  val ignoredABIProblems: Seq[ProblemFilter] = {
+    Seq(
+      ProblemFilters.exclude[Problem]("scala.meta.internal.*"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("scala.meta.parsers.Parsed.*"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("scala.meta.tokenizers.Tokenized.*")
```
xeno-by: How do we deal with these issues in the future? Under what conditions do we allow adding new rules here?

olafurpg: We use our common sense to determine if these reported ABI issues affect downstream users. In this case I think it's pretty clear that no regular user will be affected.

Issue #1133, opened 2017-10-11
Semanticsdb is missing synthetics information about implicits in for comprehensions


fikovnik:
I have the following file (`Input2.scala`):

```scala
import scala.concurrent.Future

object Input2 {
  import scala.concurrent.ExecutionContext.Implicits.global

  val f1 = Future { 1 }
  val f2 = Future { 2 }

  f1.flatMap(v1 => f2.map(v2 => v1 + v2))
}
```
The synthetics part of the corresponding semanticsdb correctly shows 4 occurrences of the global execution context being inserted at the right places:

```
scala/fix/Input2.scala
195 -- *(scala.concurrent.ExecutionContext.Implicits.global)
269 -- *(scala.concurrent.ExecutionContext.Implicits.global)
214 -- *.apply[Int]
268 -- *(scala.concurrent.ExecutionContext.Implicits.global)
226 -- *(scala.concurrent.ExecutionContext.Implicits.global)
183 -- *.apply[Int]
253 -- *[Int]
240 -- *[Int]
```
However, if I rewrite the code above to (`Input1.scala`):
```scala
import scala.concurrent.Future

object Input2 {
  import scala.concurrent.ExecutionContext.Implicits.global

  val f1 = Future { 1 }
  val f2 = Future { 2 }

  for {
   v1 <- f1
   v2 <- f2
  } yield v1 + v2
}
```
I get only 3 occurrences:
```
scala/fix/Input1.scala
183 -- *.apply[Int]
277 -- *(scala.concurrent.ExecutionContext.Implicits.global)
214 -- *.apply[Int]
261 -- *.map[Int]
195 -- *(scala.concurrent.ExecutionContext.Implicits.global)
248 -- *.flatMap[Int]
226 -- *(scala.concurrent.ExecutionContext.Implicits.global)
```
In the desugaring of for comprehensions, one use of implicit is missing.

An example project (using scalafix to setup things quickly) is here:
https://www.dropbox.com/s/fsg8z86p9spe4n9/semdb-forcomp.tar.bz2?dl=0

Using scalameta 2.0.1, scala 2.12.3

fikovnik:
I had a quick look at the scalameta sources and the problem is that for both `flatMap` and `map` the `ApplyToImplicitArgs` has the same end position. Since the storage is a map, they get overwritten.

olafurpg:
[...]

PR #1142, opened 2017-11-01
Merging inferred args with calls to synthetics

fikovnik:
This is a possible fix to #1133. 

For
```scala
import scala.concurrent.ExecutionContext.Implicits.global
val f1 = scala.concurrent.Future {1}
val f2 = scala.concurrent.Future {2}

for (v1 <- f1; v2 <- f2 if v1 > v2) yield v1+v2
```
it generates:
```
[169..169): *.flatMap[Int](*)(scala.concurrent.ExecutionContext.Implicits.global)
...
[179..179): *.withFilter(*)(scala.concurrent.ExecutionContext.Implicits.global)
...
[190..190): *.map[Int](*)(scala.concurrent.ExecutionContext.Implicits.global)
...
```

### DIFF
--- a/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/DocumentOps.scala
+++ b/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/DocumentOps.scala
@@ -320,7 +320,8 @@ trait DocumentOps { self: DatabaseOps =>
                 private def isForComprehensionSyntheticName(select: g.Select): Boolean = {
                   select.pos == select.qualifier.pos && (select.name == g.nme.map ||
                   select.name == g.nme.withFilter ||
-                  select.name == g.nme.flatMap)
+                  select.name == g.nme.flatMap ||
+                  select.name == g.nme.foreach)
                 }
 
                 private def findSelect(t: g.Tree): Option[g.Tree] = t match {
--- a/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/DocumentOps.scala
+++ b/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/DocumentOps.scala
@@ -316,6 +316,17 @@ trait DocumentOps { self: DatabaseOps =>
                 }
               }
 
+              object SyntheticApplyToImplicitArgs {
+                private def findSelect(t: g.Tree): Option[g.Tree] = t match {
+                  case g.Apply(fun, _) => findSelect(fun)
+                  case g.TypeApply(fun, _) => findSelect(fun)
+                  case s@g.Select(qual, _) if isSyntheticName(s) => Some(qual)
+                  case _ => None
+                }
+
+                def unapply(gfun: g.Apply): Option[g.Tree] = findSelect(gfun)
+              }
+
               gtree match {
                 case gview: g.ApplyImplicitView =>
                   val pos = gtree.pos.toMeta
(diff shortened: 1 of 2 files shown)


scalameta/semanticdb-scalac/src/test/scala/scala/meta/tests/SemanticSuite.scala
```diff
@@ -248,14 +248,13 @@ class SemanticSuite extends DatabaseSuite(SemanticdbMode.Slim) {
       |  CommandeerDSL(null.asInstanceOf[Foo])
       |}
     """.trim.stripMargin,
-    """|[324..324): *.apply[Foo, FooDSL]
+    """|[324..324): *.apply[Foo, FooDSL](*)(g.Foo.fooDSL)
```
olafurpg: How about we limit this change only to `.foreach/.flatMap/.map`? I think we should try to limit the number of `*` placeholder when possible since multiple `*` make it harder to disambiguate what they are placeholding.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/DocumentOps.scala
```diff
@@ -316,24 +316,38 @@ trait DocumentOps { self: DatabaseOps =>
                 }
               }
 
+              object SyntheticApplyToImplicitArgs {
```
olafurpg: How about `ForComprehensionImplicitArg`?

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/DocumentOps.scala
```diff
@@ -316,24 +316,44 @@ trait DocumentOps { self: DatabaseOps =>
                 }
               }
 
+              object ForComprehensionImplicitArg {
+                private def isForComprehensionSyntheticName(select: g.Select): Boolean = {
+                  select.pos == select.qualifier.pos && (select.name == g.nme.map ||
```
olafurpg: We should also include `foreach` and add a test for it with a `for (...) y` (no yield).

scalameta/semanticdb-scalac/src/test/scala/scala/meta/tests/SemanticSuite.scala
```diff
@@ -1007,12 +1016,96 @@ class SemanticSuite extends DatabaseSuite(SemanticdbMode.Slim) {
...
+    |  [2..5): map => _root_.scala.concurrent.Future#map(Lscala/Function1;Lscala/concurrent/ExecutionContext;)Lscala/concurrent/Future;.
+    |  [6..9): Int => _root_.scala.Int#
+    |  [11..12): * => _star_.
+    |  [58..64): global => _root_.scala.concurrent.ExecutionContext.Implicits.global.
+   """.trim.stripMargin
+  )
+
+  synthetics(
+    """
+      |object ForeachFuture {
```
olafurpg: We should try to keep the tests as narrow as possible, there is a lot of duplication from above. A single test like this should be enough to cover all cases

```scala
import scala.concurrent.ExecutionContext.Implicits.global
for (a <- scala.concurrent.Future.successful(1)) println(a)
for {
  b <- scala.concurrent.Future.successful(2)
  if a < b
} yield a
```

fikovnik: Shouldn't we do:

```scala
for {
  b <- scala.concurrent.Future.successful(2)
} println(a)
```
to make sure we see two distinct injection of implicits for foreach?

olafurpg: The first case covers foreach and the second case covers flatmap/map/withFilter, I don't think we can cover map and foreach in the same for comprehension.

olafurpg: Aah, I see what you mean. Sorry! Yes we can put two generators to introduce a conflict in positions.

xeno-by:
LGTM. Thanks for taking care of for comprehensions - this was a significant omission in our functionality :)

olafurpg:
Awesome work @fikovnik !

PR #1144, opened 2017-11-03
Upgrade semanticdb-scalac to 2.12.4

olafurpg:
There were a lot of changes in typer between 2.12.3 and 2.12.4, which made this upgrade quite painful. Thanks to @jvican 's work in https://github.com/scala/scala/pull/6013/files we should consider contributing the attachments we use in semanticdb-scalac to scala/scala to avoid hijacking analyzer and maintaining this large chunk of copy-pasted code from typer. Going through such a painful upgrade on every patch upgrade is not sustainable for semanticdb-scalac.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -14,8 +14,7 @@ project.excludeFilters = [
   sbt-test
   contrib/package.scala
   scalameta/semanticdb-scalac/src/main/resources/scalac-plugin.xml
-  scalameta/semanticdb-scalac/src/main/scala-2.11.11/scala/meta/internal/SemanticdbAnalyzer.scala
-  scalameta/semanticdb-scalac/src/main/scala-2.12.3/scala/meta/internal/SemanticdbAnalyzer.scala
+  scala/meta/internal/SemanticdbAnalyzer.scala
 ]
 runner.dialect = Scala211
 runner.optimizer.forceConfigStyleOnOffset = -1
--- a/scalameta/semanticdb-scalac/src/main/scala-2.12.4/scala/meta/internal/SemanticdbAnalyzer.scala
+++ b/scalameta/semanticdb-scalac/src/main/scala-2.12.4/scala/meta/internal/SemanticdbAnalyzer.scala
@@ -15,7 +15,7 @@ import scala.meta.internal.ReflectionToolkit
 trait SemanticdbAnalyzer extends NscAnalyzer with ReflectionToolkit {
   import global._
   import definitions._
-  import TypersStats._
+  import statistics._
 
   override def newNamer(context: Context) = new SemanticdbNamer(context)
   class SemanticdbNamer(context0: Context) extends Namer(context0) {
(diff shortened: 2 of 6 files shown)


.scalafmt.conf
```diff
@@ -14,8 +14,7 @@ project.excludeFilters = [
   sbt-test
   contrib/package.scala
   scalameta/semanticdb-scalac/src/main/resources/scalac-plugin.xml
-  scalameta/semanticdb-scalac/src/main/scala-2.11.11/scala/meta/internal/SemanticdbAnalyzer.scala
-  scalameta/semanticdb-scalac/src/main/scala-2.12.3/scala/meta/internal/SemanticdbAnalyzer.scala
```
xeno-by: Did I use the exclude syntax incorrectly here? Is scalafmt skipping these files anyway?

scalameta/semanticdb-scalac/src/main/scala-2.12.4/scala/meta/internal/SemanticdbAnalyzer.scala
```diff
@@ -0,0 +1,1151 @@
+// NOTE: has to be this package or otherwise we won't be able to access private[typechecker] methods
```
xeno-by: Is there a diff of what changed in comparison with https://github.com/scalameta/scalameta/pull/1144/commits/b62db4f2e9ca0a468dd21f8589959f19156f8d06? I'm a bit confused that we have two commits that add the same file. Do they come from different branches?

xeno-by:
I agree that it would be great to use attachments. That will not only help us with migration, but will also potentiall benefit other tools. If @adriaanm is on board, we could look into that.

adriaanm:
I am very supportive of a mechanism that moves the scalac upgrade burden out of external repos. As long as it doesn’t impact performance and has a reasonable maintenance story for scala/scala, I’m happy :-)

scalameta/semanticdb-scalac/src/main/scala-2.12.4/scala/meta/internal/SemanticdbAnalyzer.scala
olafurpg: I rebased the commits so the history is cleaner now https://github.com/scalameta/scalameta/commit/7acf345c3d00706dcabdd947a67476f4a7bbffe4 I recommend you keep the full 2.12.3..2.12.4 diff alongside to double-check it's correct.

olafurpg:
@adriaanm Thank you! I am eager to take you up on that offer :) It would mostly be adding attachments in a few places in typer.

.scalafmt.conf
xeno-by: Oooooh, this explains a lot.

scalameta/semanticdb-scalac/src/main/scala-2.12.4/scala/meta/internal/SemanticdbAnalyzer.scala
xeno-by: Very nice! Thank you, that's exactly what I wanted. I'll review the diff today.

scalameta/semanticdb-scalac/src/main/scala-2.12.4/scala/meta/internal/SemanticdbAnalyzer.scala
```diff
@@ -393,150 +389,162 @@ trait SemanticdbAnalyzer extends NscAnalyzer with ReflectionToolkit {
...
-                }
-            case _ =>
-              result
+                  asDynamicCall getOrElse {
+                    context.issue(accessibleError.get)
+                    setError(tree)
+                  }
+              case _ =>
+                result
+            }
```
xeno-by: Wow these changes were massive.

scalameta/semanticdb-scalac/src/main/scala-2.12.4/scala/meta/internal/SemanticdbAnalyzer.scala
```diff
@@ -393,150 +389,162 @@ trait SemanticdbAnalyzer extends NscAnalyzer with ReflectionToolkit {
...
+            val sym = tree1.symbol
+            if (sym != null && sym.isOnlyRefinementMember && !sym.isMacro)
+              checkFeature(tree1.pos, ReflectiveCallsFeature, sym.toString)
+
+            qualTyped.symbol match {
+              //+scalac deviation
+              case s: Symbol if s.isRootPackage => treeCopy.Ident(tree1, name).rememberSelectOf(tree1)
+              //-scalac deviation
+              case _ => tree1
+            }
```
xeno-by: Omg the party continues.

PR #1149, opened 2017-11-06
Assorted Fixes

ShaneDelmore:


scalameta/tests/jvm/src/test/resources/semanticdb.expect
```diff
@@ -88,6 +84,16 @@ Synthetics:
   [13..16): Int => _root_.scala.reflect.ClassTag.Int.
 
 
+scalameta/semanticdb-integration/src/main/scala/example/Example.scala
```
ShaneDelmore: We are not merging documents as that would require reading them from disk first, instead we just append documents which means they will get output as a separate document when calling syntax on the database.

xeno-by:
Superseded by https://github.com/xeno-by/scalameta/commits/hackweek.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/SemanticdbPipeline.scala
```diff
@@ -49,20 +49,8 @@ trait SemanticdbPipeline extends DatabaseOps { self: SemanticdbPlugin =>
           val fullName = unit.source.file.file.getAbsolutePath
           if (!fullName.matches(config.include) || fullName.matches(config.exclude)) return
           val mdoc = unit.toDocument
-          if (config.messages.saveMessages) {
```
xeno-by: If `saveMessages` is no longer used, it needs to be removed.

PR #1150, opened 2017-11-07
Assorted Fixes from Hackweek

ShaneDelmore:


scalameta/tests/jvm/src/test/resources/semanticdb.expect
```diff
@@ -27,10 +27,6 @@ Names:
 [195..203): classTag => _root_.scala.reflect.package.classTag(Lscala/reflect/ClassTag;)Lscala/reflect/ClassTag;.
 [204..207): Int => _root_.scala.Int#
 
-Messages:
```
olafurpg: Is this expected? I think the expect test should include messages.

scalameta/tests/jvm/src/test/resources/semanticdb.expect
```diff
@@ -88,6 +84,16 @@ Synthetics:
   [13..16): Int => _root_.scala.reflect.ClassTag.Int.
 
 
+scalameta/semanticdb-integration/src/main/scala/example/Example.scala
```
olafurpg: I think we should accommodate this case in `Database.load` (org.langmeta.internal.semanticdb) so that this document with only messages is merged with the one above. Something like this should do the trick
```scala
    private def mergedDatabase = {
      def isOnlyMessages(sdocument: s.Document) =
        sdocument.messages.nonEmpty &&
            sdocument.names.isEmpty &&
            sdocument.synthetics.isEmpty &&
            sdocument.contents.isEmpty &&
            sdocument.symbols.isEmpty
      sdatabase.documents match {
        case Seq(doc, messages) if isOnlyMessages(messages) =>
          doc.copy(messages = doc.messages ++ messages.messages)
        case _ => sdatabase
      }
    }
```

scalameta/tests/jvm/src/test/resources/semanticdb.expect
```diff
@@ -27,10 +27,6 @@ Names:
 [195..203): classTag => _root_.scala.reflect.package.classTag(Lscala/reflect/ClassTag;)Lscala/reflect/ClassTag;.
 [204..207): Int => _root_.scala.Int#
 
-Messages:
-[41..47): [warning] Unused import
```
olafurpg: I expected this message to be persisted with the first document along with all the names. Only messages that are reported after typer (like deprecation) should be in the second document. That would mean that a single document is the norm.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/ConfigOps.scala
```diff
@@ -5,48 +5,129 @@ import scala.meta.internal.io.PathIO
...
+  def isOff: Boolean = this == Off
+}
+object ProfilingMode {
+  def unapply(arg: String): Option[ProfilingMode] = all.find(_.toString.equalsIgnoreCase(arg))
+  def all = List(Console, Off)
+  case object Console extends ProfilingMode
+  case object Off extends ProfilingMode
+}
+
+sealed abstract class MessageMode {
```
olafurpg: Related https://github.com/scalameta/scalameta/issues/1151 I think we could actually avoid a ton of boilerplate here if we define these enums in semanticdb.proto and persist them in each Document or Database. We don't have to expose them in `scala.meta._`, they would only exists in the schema layer.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/SemanticdbPipeline.scala
```diff
@@ -98,7 +111,37 @@ trait SemanticdbPipeline extends DatabaseOps { self: SemanticdbPlugin =>
           cleanupUpwards(ve.uri.toFile)
         })
         super.run()
+        timestampPersistFinished = System.nanoTime()
+        reportSemanticdbSummary()
       }
     }
   }
+
+  private val timestampPluginCreated = System.nanoTime()
```
olafurpg: Did you consider using scalac Statistics instead of rolling it yourself? I'm fine either way, just a note.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/DocumentOps.scala
```diff
@@ -175,271 +177,274 @@ trait DocumentOps { self: DatabaseOps =>
...
+        traverser.traverse(unit.body)
+      }
 
-        val input = unit.source.toInput
+      val input = unit.source.toInput
 
-        val synthetics = inferred.toIterator.map {
-          case (pos, inferred) => inferred.toSynthetic(input, pos)
-        }
+      unit.body.appendMetadata("semanticdbMstarts" -> mstarts.mapValues(_.pos).toMap) // used in MessagesOps
```
olafurpg: `m.Position` holds a reference to `m.Input`. I think we can keep only the (start, end) ints to lower memory usage.

scalameta/tests/jvm/src/test/resources/semanticdb.expect
ShaneDelmore: This will slow down loading. Does it buy us much?

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/SemanticdbPipeline.scala
ShaneDelmore: These are changes from a PR I picked up and thought you had already approved.  I can’t comment on the reasoning behind them.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/DocumentOps.scala
ShaneDelmore: I can make this change.

scalameta/tests/jvm/src/test/resources/semanticdb.expect
ShaneDelmore: I was curious myself why messages weren’t persisted in the phase they were created but don’t understand enough about how semantic db works yet to answer the question.

scalameta/tests/jvm/src/test/resources/semanticdb.expect
xeno-by: One more design point. If we decide that we want to merge, we could merge all documents that have the same input. That looks like a less arbitrary criterion.

scalameta/semanticdb-scalac/src/main/scala-2.12.3/scala/meta/internal/SemanticdbAnalyzer.scala
```diff
@@ -1,1134 +0,0 @@
-// NOTE: has to be this package or otherwise we won't be able to access private[typechecker] methods
```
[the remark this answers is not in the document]
xeno-by: I'd prefer to delete currently unused files instead of keeping them around potentially creating confusion. We have git history for our backup needs.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/SemanticdbPipeline.scala
xeno-by: Didn't know much about `Statistics`, to be honest. I always thought it's something hacked up, but turns out it's not. Let's keep the existing implementation though and refactor if we add more performance counters.

ShaneDelmore:
Closing this PR to allow @olafurpg to cherry pick out the bits he wants to use.

PR #1154, opened 2017-11-08
Skip anonymous classes, fixes #1153

### LINKED ISSUE #1153: Term.NewAnonymous produces incorrect symbol
```scala
  targeted(
    """package oa
      |trait Foo
      |object Foo {
      |  new <<Foo>> {}
      |}
    """.stripMargin, { (db, Foo) =>
      assertNoDiff(Foo.syntax, "_root_.oa.Foo#")
    }
  )
[info] - package oa trait Foo object Foo {   new <<Foo>> {}... *** FAILED ***
[info]   ===========
[info]   => Obtained
[info]   ===========
[info]   /var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/paradise8570770741813349793.scala@40..46
[info]
[info]   =======
[info]   => Diff
[info]   =======
[info]   -/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/paradise8570770741813349793.scala@40..46
[info]   +_root_.oa.Foo# (XXX.scala:15)
```

The compiler expands the anonymous class into something like this

```scala
object Foo {
  {
    final class $anon extends Foo;
    new $anon()
  }
}
```

olafurpg:
Review @laughedelic

### DIFF
--- a/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/DocumentOps.scala
+++ b/scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/DocumentOps.scala
@@ -152,6 +152,8 @@ trait DocumentOps { self: DatabaseOps =>
                 // We cannot be guaranteed that all symbols have a position, see
                 // https://github.com/scalameta/scalameta/issues/665
                 // Instead of crashing with "unsupported file", we ignore these cases.
+                if (gsym0 == null) return
+                if (gsym0.isAnonymousClass) return
                 if (mtree.pos == m.Position.None) return
                 if (names.contains(mtree.pos)) return // NOTE: in the future, we may decide to preempt preexisting db entries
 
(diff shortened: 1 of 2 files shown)


laughedelic:
@olafurpg thanks for keeping me in the loop, but i'm not sure how I could be helpful here. To be honest I have almost no idea what you have changed in this PR 😅 
Can I help testing it or something else? I'd like to get a better ~~than no at all~~ understanding of how Scalameta works in general, but so far I didn't have enough time (and courage) to dive into the code..

olafurpg:
No worries @laughedelic there's a lot going on in Scalameta and it took me a long time to understand how the pieces fit together 😄 


`new Foo {}` gets desugared by the compiler into `{ class $anon extends Foo; new $anon}` and the go to definition bug was caused by the symbol of `Foo` getting resolved to `$anon` (which doesn't have a position in the source) when it should resolve to `Foo` in the `extends Foo` part. This commit just adds a guard that we don't resolve symbols to `$anon` classes.

laughedelic:
@olafurpg Thanks for the explanation! It makes sense to me now.

PR #1129, opened 2017-09-29
Assorted fixes

xeno-by:
Assorted fixes after the completion of the internal hackathon.

Issue #1151, opened 2017-11-07
Persist semanticdb flavor


olafurpg:
Currently, it's possible to build slim/fat semanticdbs. In #1150 and #1148, there will be 4-6 more options (denotations/messages/synthetics/members/...). Clients that read .semanticdb files don't know what settings where used to build the semanticdb. For settings like members, it's critical to be able to tell if `Denotation.members` is truly empty or if the setting was not enabled (protobuf is not able to represent `Option[List[Signature]]`. If we persist this information in `s.Document`, then we could let `scala.meta.Denotation.members` become an `Option[List[Signature]]`

olafurpg:
Related discussion in https://github.com/scalameta/scalameta/pull/1158/files#r150067558, we can include in the Document whether it's "message-only", meaning it has only filename + messages.

xeno-by:
@olafurpg Is this still relevant?

olafurpg:
No longer relevant, tools that would consume this information (like scalafix) should instead extract the scalac options from the build tool. 

If we are to implement this however, I believe a good place to put this information might be semanticidx.

PR #1158, opened 2017-11-09
Finish remaining pieces from hackweek

### LINKED ISSUE #1067: Errors in Database.load should include input filename in the error message
I'm currently debugging an issue in how sbthost emits .semanticdb files during incremental compilation. The stack traces look like this

```
[info]   com.google.protobuf.InvalidProtocolBufferException: While parsing a protocol message, the input ended unexpectedly in the middle of a field.  This could mean either that the input has been truncated or that an embedded message misreported its own length.
[info]   at com.google.protobuf.InvalidProtocolBufferException.truncatedMessage(InvalidProtocolBufferException.java:82)
[info]   at com.google.protobuf.CodedInputStream$ArrayDecoder.skipRawBytes(CodedInputStream.java:1200)
[info]   at com.google.protobuf.CodedInputStream$ArrayDecoder.skipField(CodedInputStream.java:575)
[info]   at lang.meta.internal.semanticdb.schema.Database.mergeFrom(Database.scala:45)
```

it would be nice if the message included the filename that was being loaded. That would help me inspect its contents and see why it's wrong.

olafurpg:
This PR supersedes  #1129. The biggest change is that semanticdbs are persisted immediately after typer, significantly reducing memory pressure for large projects. This change introduces a lot of complexity because some messages like deprecations are reported after typer. The solution in this PR is to append the messages reported after typer into the same .semanticdb file but as a separate m.Database and merge the two databases on m.Database.load. This is not an optimal solution, but it's the best we can do without compromising memory usage.

Fixes #1067

### DIFF
--- a/langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
+++ b/langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
@@ -170,6 +170,16 @@ package object semanticdb {
                 None
             }
           }
+          object dSeverity {
+            def unapply(dseverity: d.Severity): Option[s.Message.Severity] = {
+              dseverity match {
+                case d.Severity.Info => Some(s.Message.Severity.INFO)
+                case d.Severity.Warning => Some(s.Message.Severity.WARNING)
+                case d.Severity.Error => Some(s.Message.Severity.ERROR)
+                case _ => None
+              }
+            }
+          }
           object dDenotation {
             def unapply(ddefn: d.Denotation): Option[s.Denotation] = {
               import ddefn._
@@ -230,31 +240,4 @@ package object semanticdb {
     }
   }
 
-  implicit class XtensionMetaMessagesAppend(mmessages: List[d.Message]) {
-    private def toSchema: List[s.Message] = mmessages.collect {
-      case d.Message(pos, dSeverity(severity), msg) =>
-        s.Message(Some(s.Position(pos.start, pos.end)),severity, msg)
-    }
-    def append(filename: AbsolutePath, sourceroot: AbsolutePath, targetroot: AbsolutePath): Unit = {
-      val relpath = v.SemanticdbPaths.fromScala(filename.toRelative(sourceroot))
-      val semanticdbpath = targetroot.resolve(relpath)
-      val out = new RandomAccessFile(semanticdbpath.toFile, "rw")
-      try {
-        out.setLength(out.length() - 1)
-        s.Database().writeTo(???)
-        out.write(s.Database(s.Document(messages = toSchema) :: Nil).toByteArray)
-      } finally out.close()
-    }
-  }
-
-  private object dSeverity {
-    def unapply(dseverity: d.Severity): Option[s.Message.Severity] = {
-      dseverity match {
-        case d.Severity.Info => Some(s.Message.Severity.INFO)
-        case d.Severity.Warning => Some(s.Message.Severity.WARNING)
-        case d.Severity.Error => Some(s.Message.Severity.ERROR)
-        case _ => None
-      }
-    }
-  }
 }
--- a/langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
+++ b/langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
@@ -1,6 +1,11 @@
 package org.langmeta.internal
 
+import java.io.RandomAccessFile
 import java.nio.charset.Charset
+import java.nio.file.Files
+import java.nio.file.OpenOption
+import java.nio.file.StandardOpenOption
+import scala.util.control.NonFatal
 import org.langmeta.inputs.{Input => dInput}
 import org.langmeta.inputs.{Position => dPosition}
 import org.langmeta.semanticdb.{Synthetic => dSynthetic}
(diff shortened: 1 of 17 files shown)


olafurpg:
I kept the 2.11.11 sources of the analyzer, we can remove it in a couple of months once the community has had more time to upgrade to 2.11.12

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/SemanticdbPipeline.scala
```diff
@@ -72,10 +73,10 @@ trait SemanticdbPipeline extends DatabaseOps { self: SemanticdbPlugin =>
       override def apply(unit: g.CompilationUnit): Unit = {
         if (config.mode.isDisabled) return
         try {
-          unit.body.attachments.get[m.Document].foreach { mattrs =>
-            unit.body.removeAttachment[m.Document]
-            val messages = if (config.messages.saveMessages) unit.reportedMessages else Nil
-            val mminidb = m.Database(List(mattrs.copy(messages = messages)))
+          val messages = if (config.messages.saveMessages) unit.reportedMessages(Map.empty) else Nil
```
xeno-by: Does this only fetch additional messages?

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/SemanticdbPipeline.scala
```diff
@@ -72,10 +73,10 @@ trait SemanticdbPipeline extends DatabaseOps { self: SemanticdbPlugin =>
...
+          if (messages.nonEmpty) {
+            val mminidb = m.Database(List(m.Document(unit.source.toInput, language, Nil, messages, Nil, Nil)))
+            // TODO(olafur) append here
```
xeno-by: Does this todo mean that this is not appending but overwriting currently?

langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
```diff
@@ -25,94 +30,128 @@ package object semanticdb {
...
+    private def sdocumentsMerged: Seq[s.Document] = {
+      def isOnlyMessages(sdocument: s.Document) =
+        sdocument.contents.isEmpty &&
+          sdocument.messages.nonEmpty &&
+          sdocument.names.isEmpty &&
+          sdocument.synthetics.isEmpty &&
+          sdocument.symbols.isEmpty
+      if (sdatabase.documents.length <= 1) {
+        // NOTE(olafur) the most common case is that there is only a single database
+        // per document so we short-circuit here if that's the case.
```
xeno-by: How about merging the databases if the input is the same? That'll be more general than hardcoding this very specific condition.

langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
```diff
@@ -201,4 +230,32 @@ package object semanticdb {
...
+      case d.Message(pos, dSeverity(severity), msg) =>
+        s.Message(Some(s.Position(pos.start, pos.end)),severity, msg)
+    }
+    def append(filename: AbsolutePath, sourceroot: AbsolutePath, targetroot: AbsolutePath): Unit = {
+      val relpath = v.SemanticdbPaths.fromScala(filename.toRelative(sourceroot))
+      val semanticdbpath = targetroot.resolve(relpath)
+      val out = new RandomAccessFile(semanticdbpath.toFile, "rw")
+      try {
+        out.setLength(out.length() - 1)
+        s.Database().writeTo(???)
```
xeno-by: Triple question mark?

langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/vfs/Database.scala
```diff
@@ -20,16 +21,43 @@ final case class Database(entries: List[Entry]) {
   def toSchema: s.Database = {
     val sentries = entries.flatMap { ventry =>
       val sdb = s.Database.parseFrom(ventry.bytes)
-      sdb.documents
+      // returns true if this document contains only messages and nothing else.
+      // deprecation messages are reported in refchecks and get persisted
+      // as standalone documents that need to be merged with their typer-phase
+      // document during loading. It seems there's no way to merge the documents
+      // during compilation without introducing a lot of memory pressure.
+      def isOnlyMessages(sdocument: s.Document): Boolean =
```
xeno-by: Why do we have to duplicate this logic wrt `sdocumentsMerged`?

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/SemanticdbPipeline.scala
```diff
@@ -62,22 +69,32 @@ trait SemanticdbPipeline extends DatabaseOps { self: SemanticdbPlugin =>
...
-  object PersistSemanticdbComponent extends PluginComponent {
+  object SemanticdbJvmComponent extends PluginComponent {
     val global: SemanticdbPipeline.this.global.type = SemanticdbPipeline.this.global
     val runsAfter = List("jvm")
     override val runsRightAfter = Some("jvm")
-    val phaseName = "semanticdb-persist"
-    override val description = "persist semanticdb files"
+    val phaseName = "semanticdb-jvm"
+    override val description =
+      "compute and persist semanticdb information after typer (most commonly messages)"
```
xeno-by: `compute and persist additional semanticdb entries created after typer`

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/ConfigOps.scala
```diff
@@ -24,8 +24,8 @@ case class SemanticdbConfig(
       s"-P:${SemanticdbPlugin.name}:profiling:${profiling.name} " +
       s"-P:${SemanticdbPlugin.name}:include:${fileFilter.include} " +
       s"-P:${SemanticdbPlugin.name}:exclude:${fileFilter.exclude} "
-      s"-P:${SemanticdbPlugin.name}:messages:${messages.name} " +
-      s"-P:${SemanticdbPlugin.name}:synthetics:${synthetics.name} "
+  s"-P:${SemanticdbPlugin.name}:messages:${messages.name} " +
+    s"-P:${SemanticdbPlugin.name}:synthetics:${synthetics.name} "
```
xeno-by: Forgotten plus?

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/ReporterOps.scala
```diff
@@ -1,11 +1,14 @@
...
 
+import scala.collection.mutable
 import scala.tools.nsc.reporters.StoreReporter
 import scala.reflect.internal.util.{Position => gPosition}
 
 trait ReporterOps { self: DatabaseOps =>
-
+  // Hack, keep track of how many messages we have returns for each path to avoid
+  // duplicate messages. The key is System.identityHashCode to keep memory usage low.
+  private val returnedMessagesByPath = mutable.LongMap.empty[Int]
```
xeno-by: I think we can use `unit` as a key. Units stay in memory for the entire duration of the compilation pipeline.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/ParseOps.scala
```diff
@@ -9,7 +9,9 @@ trait ParseOps { self: DatabaseOps =>
     def toSource: m.Source = {
       val dialect =
         m.Dialect.standards.getOrElse(language, sys.error(s"unsupported dialect $language"))
-      unit.cache.getOrElse("source", dialect(unit.source.toInput).parse[m.Source].get)
+      // TODO(olafur) remove
+      // TODO(olafur) upgrade to 2.11.12
```
xeno-by: What do these todos mean?

langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
olafurpg: This method is a relic of a failed attempt, I've removed it.

> if the input is the same

Message-only documents have no .contents, and s.Database has no input.

langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
olafurpg: Removed, relic of failed attempt.

langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
```diff
@@ -201,4 +230,32 @@ package object semanticdb {
...
+        out.write(s.Database(s.Document(messages = toSchema) :: Nil).toByteArray)
+      } finally out.close()
+    }
+  }
+
+  private object dSeverity {
```
olafurpg: Moved this back to original location.

langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/vfs/Database.scala
olafurpg: I've consolidated the logic under the extension method `s.Database.mergeMessageOnlyDocuments: s.Database`.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/SemanticdbPipeline.scala
olafurpg: Done.

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/ParseOps.scala
olafurpg: Removed, I wrote those down during our call and committed them for some reason 😅

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/ReporterOps.scala
olafurpg: Done.

scalameta/semanticdb-integration/src/main/scala/example/Exclude.scala
```diff
@@ -0,0 +1,4 @@
+package exclude
+
+// This class should not be excluded by semanticdb.
```
xeno-by: Should or should not? The build configuration says `s"-P:semanticdb:exclude:Exclude.scala",`

langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
xeno-by: But it does have an input, doesn't it? `input = m.Input.File(unit.source.file.file)`. Can we use this to correlate documents?

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/ConfigOps.scala
```diff
@@ -4,52 +4,82 @@ package semanticdb
...
   def syntax: String =
     s"-P:${SemanticdbPlugin.name}:sourceroot:$sourceroot " +
-      s"-P:${SemanticdbPlugin.name}:mode:$mode" +
-      s"-P:${SemanticdbPlugin.name}:failures:${failures.name} "
+      s"-P:${SemanticdbPlugin.name}:mode:${mode.name}" +
+      s"-P:${SemanticdbPlugin.name}:failures:${failures.name} " +
+      s"-P:${SemanticdbPlugin.name}:denotations:${denotations.name} " +
+      s"-P:${SemanticdbPlugin.name}:profiling:${profiling.name} " +
+      s"-P:${SemanticdbPlugin.name}:include:${fileFilter.include} " +
+      s"-P:${SemanticdbPlugin.name}:exclude:${fileFilter.exclude} "
```
xeno-by: Still missing a plus here.

langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
xeno-by: If we don't have time to do this right now, let's merge but at least open an issue to make sure we follow through next time. This is some technical debt that we're taking on.

olafurpg: We don't have a source root so `m.Input.File()` won't work unless `sys.props("user.dir")` coincidentally matches --sourceroot. I think a step in the right direction to make this less hacky is https://github.com/scalameta/scalameta/issues/1151 we can persist the flavor of the persisted message in `s.Document`, one field could be to indicate that this is a message-only Document. Message-only are relevant outside of this use-case, I would use this also to implement something like https://github.com/olafurpg/sbt-messagehost

scalameta/semanticdb-integration/src/main/scala/example/Exclude.scala
olafurpg: Should be excluded, good catch!

scalameta/semanticdb-scalac/src/main/scala/scala/meta/internal/semanticdb/ConfigOps.scala
olafurpg: Whoa! Good eye 😉

PR #1159, opened 2017-11-09
Fix #1152, split semanticdb-scalac into library and plugin artifacts

### LINKED ISSUE #1152: Split semanticdb-scalac into plugin and library module
In https://github.com/scalameta/language-server we may want to build semanticbs on the fly with the presentation compiler on every key stroke. Currently, semanticdb-scalac is only published as a fatjar for compatibility with scalac's plugin infrastructure `-Xplugin:semanticdb.jar`, we should however also publish a non-fat artifact with only the compiler plugin classfiles.

olafurpg:
This PR is very important for scalameta/language-server where we want to try and build semanticdbs on the fly with the presentation compiler.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -257,16 +257,14 @@ lazy val scalameta = crossProject
 lazy val scalametaJVM = scalameta.jvm
 lazy val scalametaJS = scalameta.js
 
-lazy val semanticdbScalac = project
-  .in(file("scalameta/semanticdb-scalac"))
+lazy val semanticdbScalacPlugin = project
+  .in(file("scalameta/semanticdb-scalac-plugin"))
   .settings(
     moduleName := "semanticdb-scalac",
-    description := "Scala 2.x compiler plugin that generates semanticdb on compile",
+    description := "Fatjar of semanticdbScalac",
     publishableSettings,
     mergeSettings,
     isFullCrossVersion,
-    libraryDependencies += "org.scala-lang" % "scala-compiler" % scalaVersion.value,
-    exposePaths("semanticdb-scalac", Test),
     mimaPreviousArtifacts := Set.empty,
     pomPostProcess := { node =>
       new RuleTransformer(new RewriteRule {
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -257,16 +257,14 @@ lazy val scalameta = crossProject
...
 lazy val scalametaJS = scalameta.js
 
-lazy val semanticdbScalac = project
-  .in(file("scalameta/semanticdb-scalac"))
+lazy val semanticdbScalacPlugin = project
+  .in(file("scalameta/semanticdb-scalac-plugin"))
   .settings(
     moduleName := "semanticdb-scalac",
-    description := "Scala 2.x compiler plugin that generates semanticdb on compile",
+    description := "Fatjar of semanticdbScalac",
```
xeno-by: I suggest we keep an old description here.

build.sbt
```diff
@@ -285,6 +283,20 @@ lazy val semanticdbScalac = project
       }).transform(node).head
     }
   )
+  .dependsOn(semanticdbScalac)
+
+
+lazy val semanticdbScalac = project
```
xeno-by: How about `semanticdbScalacLibrary`?

build.sbt
```diff
@@ -285,6 +283,20 @@ lazy val semanticdbScalac = project
...
+  .in(file("scalameta/semanticdb-scalac"))
+  .settings(
+    moduleName := "semanticdb-scalac-core",
+    description := "Scala 2.x compiler plugin that generates semanticdb on compile",
```
xeno-by: `Library to generate semanticdb from Scala 2.x internal data structures`

build.sbt
olafurpg: Done.

build.sbt
olafurpg: semanticdbScalac is better IMO since it matches the directory name and I run `semanticdbScalac/test` from the console all the time (which is already too long IMO), writing `Library` on top is just annoying.

build.sbt
olafurpg: Done.

build.sbt
xeno-by: Fair enough. lgtm.

PR #1160, opened 2017-11-10
Exclude semanticdb-scalac-core from semanticdb-scalac.

olafurpg:
Hotfix, 2.1.0 is broken since `addCompilerPlugin(semanticdb-scalac)` causes "Ignoring duplicate plugin semanticdb (scala.meta.internal.SemanticdbPlugin)" warnings at best and InvocationTargetExceptions at work. The pomPostProcess no longer worked after #1159. This change was manually tested locally and the .pom now correctly excludes the semanticdb dependency.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -273,7 +273,7 @@ lazy val semanticdbScalacPlugin = project
             node.label == "artifactId" && fn(node.text)
           node.label == "dependency" && node.child.exists(child =>
             isArtifactId(child, _.startsWith("langmeta-")) ||
-            isArtifactId(child, _.startsWith("scalameta_")))
+            isArtifactId(child, _.startsWith("semanticdb")))
         }
         override def transform(node: XmlNode): XmlNodeSeq = node match {
           case e: Elem if isAbsorbedDependency(node) =>


marconilanna:
@xeno-by could you please publish 2.1.1 again? It looks like many artifacts are missing from Maven:

https://repo1.maven.org/maven2/org/scalameta/langmeta_2.12/2.1.1/

https://repo1.maven.org/maven2/org/scalameta/langmeta_2.12/2.1.0/

olafurpg:
@marconilanna Thank you for reporting! Langmeta 2.12 seems to have been only partially released and I can't trigger a full release since there are already some files for that release which would get overwritten. I have triggered a new release for v2.1.2.

olafurpg:
@marconilanna 2.1.2 is syncing to maven central, you can already use it with a resolver to sonatype releases. I've opened https://github.com/scalameta/scalameta/pull/1161 to prevent these problems in the future.

marconilanna:
@olafurpg Thank you very much for acting so quickly! The website still points to 2.0.1, if you would like to update it, too.

olafurpg:
I've triggered a rebuild of the website to update the version there. Good catch!

PR #1174, opened 2017-12-03
Relational schema for SemanticDB and a converter to SQLite

xeno-by:
Converting SemanticDB classpaths to SQLite databases takes reasonable time and requires reasonable disk space even for huge codebases. Even linear searches (like find all references) within these databases take acceptable time. When paired with appropriate indices, these searches become instantaneous. /cc @laughedelic @gabro @olafurpg @ShaneDelmore 

Here are the numbers from experimenting on the @olafurpg's legendary corpus of Scala projects (for full details see https://github.com/scalameta/scalameta/pull/1174#pullrequestreview-80697994):
```
14000 documents: 9575440 names, 1321 messages, 1037584 symbols, 650003 synthetics (~54s)
CPU time: 59.088s
Disk size: 764.2 MB (260MB compressed)
Performance: 250.254 documents/s
```

### DIFF
--- a/langmeta-sqlite/src/main/resources/semanticdb.ddl
+++ b/langmeta-sqlite/src/main/resources/semanticdb.ddl
@@ -1,7 +1,49 @@
-create table names(
+create table document(
   id int,
-  filename string,
+  filename text,
+  contents text,
+  language text,
+  primary key (id)
+);
+
+create table name(
+  id int,
+  document int,
   start int,
   end int,
-  symbol string,
-  is_definition boolean)
+  symbol int,
+  is_definition boolean,
+  primary key (id),
+  foreign key (document) references document (id),
+  foreign key (symbol) references symbol (id)
+);
+
+create table message(
+  id int,
+  document int,
+  start int,
+  end int,
+  severity int,
+  text text,
+  primary key (id),
+  foreign key (document) references document(id)
+);
+
+create table symbol(
+  id int,
+  symbol text,
+  flags int,
+  name text,
+  signature int,
+  primary key (id),
+  foreign key (signature) references document (id)
+);
+
+create table synthetic(
+  id int,
+  start int,
+  end int,
+  text int,
+  primary key (id),
+  foreign key (text) references document (id)
+);
--- a/langmeta-sqlite/src/main/resources/semanticdb.ddl
+++ b/langmeta-sqlite/src/main/resources/semanticdb.ddl
@@ -1,0 +1,7 @@
+create table names(
+  id int,
+  filename string,
+  start int,
+  end int,
+  symbol string,
+  is_definition boolean)
--- a/langmeta-sqlite/src/main/scala/org/langmeta/sqlite/Main.scala
+++ b/langmeta-sqlite/src/main/scala/org/langmeta/sqlite/Main.scala
@@ -1,5 +1,6 @@
 package org.langmeta.sqlite
 
+import java.io._
 import java.nio.file._
 import java.sql._
 import scala.Array
@@ -190,9 +191,11 @@ object Main {
         } finally {
           conn.commit()
           conn.close()
-          val appElapsed = (System.nanoTime() - appStart) * 1.0 / 1000000000
-          val appPerformance = genuineDocuments / appElapsed
-          println(s"Elapsed: ${"%.3f".format(appElapsed)}s")
+          val appCPU = (System.nanoTime() - appStart) * 1.0 / 1000000000
+          val appDisk = new File(sqliteFilename).length * 1.0 / 1024 / 1024
+          val appPerformance = genuineDocuments / appCPU
+          println(s"CPU time: ${"%.3f".format(appCPU)}s")
+          println(s"Disk size: ${"%.1f".format(appDisk)} MB")
           println(s"Performance: ${"%.3f".format(appPerformance)} documents/s")
         }
       case _ =>
--- a/langmeta-sqlite/src/main/scala/org/langmeta/sqlite/Main.scala
+++ b/langmeta-sqlite/src/main/scala/org/langmeta/sqlite/Main.scala
@@ -13,8 +13,8 @@ object Main {
     args match {
       case Array(sqliteFilename, semanticdbFilenames @ _*) =>
         val appStart = System.nanoTime()
+        var genuineDocuments = 0
         class Counter { var value = 0; def next() = { value += 1; value }; }
-        var documentId = new Counter()
         val sqlitePath = Paths.get(sqliteFilename)
         if (Files.exists(sqlitePath)) sys.error(s"$sqlitePath already exists")
         val conn = DriverManager.getConnection(s"jdbc:sqlite:$sqliteFilename")
@@ -39,15 +39,19 @@ object Main {
           }
           val documentStmt = tableInsertStmt("document")
           val nameStmt = tableInsertStmt("name")
+          val messageStmt = tableInsertStmt("message")
           val symbolStmt = tableInsertStmt("symbol")
+          val syntheticStmt = tableInsertStmt("synthetic")
 
           val semanticdbStart = System.nanoTime()
           val documentsToPaths = mutable.Map[String, Path]()
+          var documentId = new Counter()
           var nameId = new Counter()
           var messageId = new Counter()
           var _symbolId = new Counter()
           var symbolIds = mutable.Map[String, Int]()
           var symbolTodo = mutable.Map[Int, String]()
+          var symbolDone = mutable.Set[Int]()
           def symbolId(symbol: String): Int = {
             if (symbolIds.contains(symbol)) {
               symbolIds(symbol)
@@ -71,8 +75,11 @@ object Main {
                   val details = "both $existingPath and $path"
                   println(s"Duplicate document filename $what in $details")
                 case None =>
+                  genuineDocuments += 1
                   documentsToPaths(document.filename) = path
-                  documentStmt.setInt(1, documentId.next)
+
+                  val documentRef = documentId.next
+                  documentStmt.setInt(1, documentRef)
                   documentStmt.setString(2, document.filename)
                   documentStmt.setString(3, document.contents)
                   documentStmt.setString(4, document.language)
@@ -80,7 +87,7 @@ object Main {
 
                   document.names.foreach { name =>
                     nameStmt.setInt(1, nameId.next)
-                    nameStmt.setInt(2, documentId.value)
+                    nameStmt.setInt(2, documentRef)
                     nameStmt.setInt(3, name.position.get.start)
                     nameStmt.setInt(4, name.position.get.end)
                     nameStmt.setInt(5, symbolId(name.symbol))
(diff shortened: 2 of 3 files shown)


xeno-by:
@olafurpg Can you test the SQLite converter on your fabled corpus?

langmeta-sqlite/src/main/scala/org/langmeta/sqlite/Main.scala
```diff
@@ -67,113 +68,119 @@ object Main {
...
-            db.documents.foreach { document =>
-              documentsToPaths.get(document.filename) match {
-                case Some(existingPath) =>
-                  val what = document.filename
-                  val details = "both $existingPath and $path"
-                  println(s"Duplicate document filename $what in $details")
-                case None =>
-                  genuineDocuments += 1
-                  documentsToPaths(document.filename) = path
+            try {
```
ShaneDelmore: What sort of errors are you seeing?  I have no errors parsing semanticdbs, but I did find when I ran your dedup script on my laptop it did not filter down to semanticdb files specifically and threw errors on dsstore files.

langmeta-sqlite/src/main/resources/semanticdb.ddl
```diff
@@ -0,0 +1,49 @@
...
+  language text,
+  primary key (id)
+);
+
+create table name(
+  id int,
+  document int,
+  start int,
+  end int,
+  symbol int,
```
olafurpg: I am concerned about making this a foreign key on symbol since it complicates insertions. For references, we typically don't have the denotation.

langmeta-sqlite/src/main/resources/semanticdb.ddl
```diff
@@ -0,0 +1,49 @@
...
+  document int,
+  start int,
+  end int,
+  severity int,
+  text text,
+  primary key (id),
+  foreign key (document) references document(id)
+);
+
+create table symbol(
```
olafurpg: I propose we split this into two tables

```sql
create table symbol(
  id int,
  symbol text,
)

create table denotation -- same as current symbol with FK on symbol for text
```

We can easily insert a reference to a symbol without worrying if a denotation exists or not. We can insert a symbol or ignore it on conflict http://www.sqlite.org/lang_conflict.html

langmeta-sqlite/src/main/resources/semanticdb.ddl
```diff
@@ -0,0 +1,49 @@
...
+  id int,
+  symbol text,
+  flags int,
+  name text,
+  signature int,
+  primary key (id),
+  foreign key (signature) references document (id)
+);
+
+create table synthetic(
```
olafurpg: Are we missing a FK on the enclosing document here? It doesn't seem to be possible to get all synthetics of a given document currently.

langmeta-sqlite/src/main/scala/org/langmeta/sqlite/Main.scala
```diff
@@ -0,0 +1,214 @@
...
+                      messageStmt.setInt(1, messageId.next)
+                      messageStmt.setInt(2, documentRef)
+                      messageStmt.setInt(3, message.position.get.start)
+                      messageStmt.setInt(4, message.position.get.end)
+                      messageStmt.setInt(5, message.severity.value)
+                      messageStmt.setString(6, message.text)
+                      messageStmt.executeUpdate()
+                    }
+
+                    document.symbols.foreach { symbol =>
```
olafurpg: I propose we turn this into an `Iterator[schema.Document]` and move the logic out of the main function

olafurpg:
Another interesting topic I'm interested to hear you thoughts on is composability/reusability of these sqlite databases. 

For example, if we keep each db isolated to a library jar then we can quickly swap out one db with another when upgrading a dependency. Composing dbs could then be done by keeping connections to multiple databases and querying all of them for definition/references.

langmeta-sqlite/src/main/scala/org/langmeta/sqlite/Main.scala
xeno-by: If you accidentally pass a non-semanticdb file to the app, there's no need to crash it completely. That file should be skipped instead. That's the error that I saw - nothing too fancy :)

langmeta-sqlite/src/main/scala/org/langmeta/sqlite/Main.scala
```diff
@@ -0,0 +1,214 @@
...
+                        nameStmt.setInt(3, name.position.get.start)
+                        nameStmt.setInt(4, name.position.get.end)
+                        nameStmt.setInt(5, symbolId(name.symbol))
+                        nameStmt.setBoolean(6, name.isDefinition)
+                        nameStmt.executeUpdate()
+                      }
+                      syntheticStmt.executeUpdate()
+                    }
+
+                    if ((genuineDocuments % 1000) == 0) {
```
xeno-by: The logic in this block also needs to be called after the main loop exits, since we can't expect that the number of documents is going to always divide by 1000.

langmeta-sqlite/src/main/scala/org/langmeta/sqlite/Main.scala
xeno-by: This file definitely needs restructuring. I hacked it up in a couple of hours, so I optimized for writing, not for reading. The results were so exciting that I rushed to share them without any cleanup.

langmeta-sqlite/src/main/resources/semanticdb.ddl
xeno-by: Yep. Missed that one. Thanks!

langmeta-sqlite/src/main/resources/semanticdb.ddl
ShaneDelmore: How does it complicate insertions exactly?  FKs can be nullable so I am not understanding how it would matter whether we have a denotation or not.

olafurpg: I'd like to avoid updates altogether, so if you have a reference to a symbol but no flag/signature of that symbol then you're out of luck. See other comment proposing to split "symbol" table into two tables

langmeta-sqlite/src/main/resources/semanticdb.ddl
```diff
@@ -0,0 +1,49 @@
...
+  severity int,
+  text text,
+  primary key (id),
+  foreign key (document) references document(id)
+);
+
+create table symbol(
+  id int,
+  symbol text,
+  flags int,
```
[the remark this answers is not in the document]
xeno-by: I was thinking about flags when going to sleep yesterday - awesome that you've already thought about that!

In general, I would suggest that we try to make the sqlite database as useful in itself as possible, without requiring users to write code in a programming language other than SQL.

If we follow this principle, we either come up with new columns or with new [run-time loadable extensions](https://sqlite.org/loadext.html). SQLite seems doable, but is also clearly a lot of hassle. 

Luckily, new columns look quite alright. For the aforementioned corpus that you converted to SQLite, the additional payload will be 1 byte per column * 26 flag columns * 1037584 symbols = ~26 MB, which is not a big deal compared to ~764MB that the database currently takes. I'll experiment with this today.

langmeta-sqlite/src/main/resources/semanticdb.ddl
xeno-by: 👍

langmeta-sqlite/src/main/resources/semanticdb.ddl
xeno-by: This saves a lot of space, so I suggest we try to keep it as a FK. If we split symbol and denotation, this matter becomes moot anyway, right?

langmeta-sqlite/src/main/resources/semanticdb.ddl
```diff
@@ -0,0 +1,49 @@
+create table document(
+  id int,
+  filename text,
+  contents text,
```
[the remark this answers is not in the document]
xeno-by: Yes!! Another thing that I was thinking about when going to sleep ❤️

How about we first try the most straightforward thing - for every start or end, we try to store their: 1) offset, 2) line, 3) column? If this ends up being expensive, we can either remove offset and try again or try line_index instead.

langmeta-sqlite/src/main/resources/semanticdb.ddl
olafurpg: It can be a FK as long as we split symbol/denotation :)

langmeta-sqlite/src/main/resources/semanticdb.ddl
```diff
@@ -0,0 +1,49 @@
+create table document(
+  id int,
+  filename text,
```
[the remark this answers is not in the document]
xeno-by: The original idea was for the schema to be perfectly compatible with the protobuf file. I deviated from it anyway by inlining positions, so maybe we can start deviating further.

langmeta-sqlite/src/main/resources/semanticdb.ddl
olafurpg: was btw thinking it's probably possible to effificently load the symbol table into memory using a radix tree, at least global symbols. That should avoid doing a roundtrip for inserts only to get the FK id on existing symbols.

langmeta-sqlite/src/main/resources/semanticdb.ddl
xeno-by: In general, how open would you be to refactoring the protobuf schema? There are some things that we could do to it in order to ensure parity with the SQL schema:
  * Fix small inconsistencies (e.g. `Synthetic.pos` vs everyone else's `position` field).
  * Inline positions to save on keystrokes.
  * Split `Denotation.flags` into bits.

Looks like all these changes can be performed in a backward-compatible manner, so that existing programs written against the current protobuf schema continue working with the schema. Source compatibility of the lowlevel layer can be ensured via extension methods (or can be forgone completely since the lowlevel layer contains `internal` in its package name).

langmeta-sqlite/src/main/resources/semanticdb.ddl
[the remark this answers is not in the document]
ShaneDelmore: How about:
```
Create table documentLine(
  Document int,
  Line int,
  StartChar int,
  EndChar int,
)
```
Then you could just query for a single row using indices instead of a table scan.
```
Select line 
from documentLine dl 
where 
  my_document == dl.document 
  and my_start between dl.startChar and dl.endChar
```
Please ignore all of the casing.  I’m on iOS, it really likes to force capital letters in certain places.

langmeta-sqlite/src/main/resources/semanticdb.ddl
xeno-by: Moreover, speaking of the lowlevel layer of the protobuf schema (`org.langmeta.internal.semanticdb.schema`). What would it take to make it as attractive to users as `org.langmeta.semanticdb`? 

I think it's a shame that only Scala people can enjoy some of the conveniences that we provide in the highlevel API. Maybe we could push them to the lowlevel API, too?

langmeta-sqlite/src/main/resources/semanticdb.ddl
```diff
@@ -0,0 +1,49 @@
+create table document(
```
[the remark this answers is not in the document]
xeno-by: Great idea! How about we do the same to `semanticdb.proto` and then we make a pledge to version both `proto` and `ddl` schemas in lockstep?

langmeta-sqlite/src/main/resources/semanticdb.ddl
xeno-by: 👍

langmeta-sqlite/src/main/resources/semanticdb.ddl
ShaneDelmore: I think document needs a reference to a new table or a few new columns specifying the conditions under which it was created.
Compiler flags
Dependencies
Scala version

Basically all of the inputs to zinc.  Without this the results may be incorrect.  For example types inferred may change based on dependencies, or messages output may change based on -xlint flag.

This does not need to hold up this pr though.

langmeta-sqlite/src/main/resources/semanticdb.ddl
xeno-by: /cc @olafurpg

xeno-by: /cc @olafurpg

langmeta-sqlite/src/main/resources/semanticdb.ddl
```diff
@@ -0,0 +1,49 @@
...
+  signature int,
+  primary key (id),
+  foreign key (signature) references document (id)
+);
+
+create table synthetic(
+  id int,
+  start int,
+  end int,
+  text int,
```
ShaneDelmore: How about documentId instead of text?

langmeta-sqlite/src/main/resources/semanticdb.ddl
ShaneDelmore: I don’t understand the proposal here.  Is denotation just a symbol that refers to a document?

langmeta-sqlite/src/main/resources/semanticdb.ddl
ShaneDelmore: Why split denotation flags into bits?

ShaneDelmore: I think we could just combine the schemas easily enough and retain the performance of internal.semanticdb.schema.  Should just be adding a bunch of extension methods to the low level version right?

xeno-by: To make it easy to query for individual bits. `(s.flags & 135168) = 135168` looks awful to a newcomer. `s.is_class && s.is_final` is much better`. Same about SQL.

langmeta-sqlite/src/main/resources/semanticdb.ddl
xeno-by: Field names currently mirror whatever field names the protobuf schema provides, except positions which are always inlined.

langmeta-sqlite/src/main/resources/semanticdb.ddl
ShaneDelmore: Let’s test this.  It may be a large performance hit.  In general sql engines don’t consider Boolean flags for indexing, because they don’t offer enough discrimination to offset the lookup cost to the index, (there are two total values by definition) so besides the space usage of the fact that SQLite has no true Boolean type and will store these as an int under the covers, queries would get slower.  

It would be better I think to add a function as @olafurpg mentioned, to the client library if we are offering that for re-use or to the db itself if we expect people to connect directly to the db and ignore our library when analyzing it.

ShaneDelmore:
Ahh....we should change that I think but it doesn’t have to be now.  When I look at a relational db I expect columns that are FKs to somehow reference the other table, and I expect columns to not be named after reserved words.  This naming convention violates both of those expectations for me.

langmeta-sqlite/src/main/resources/semanticdb.ddl
xeno-by: @olafurpg I think we could give a database-centric approach a try. If an insert is sub-millisecond even on insanely big codebases + whatever single or double-digit overhead that JDBC brings to the table, then what's the big deal in requiring an extra insert?

xeno-by: What I think this pull request means is that we can try to write semantic tools in the same way how people write normal business apps - load up a database, send a few SQL queries, maybe do some updates and that's it. No compiler magic, no advanced data structures. It will be a huge breakthrough if it works, so how about we give it a try?

xeno-by:
@olafurpg To answer your non-inline comments:

1) Conflict handling. None of the columns (except for primary keys) are unique, so we can probably insert multiple copies of the same thing under different ids? But I agree that this is a complicated issues, which I conveniently sidestepped by requiring that the SQLite database that we're importing stuff into doesn't exist in advance.

2) Composability. Another thing I was thinking about, but didn't come to any satisfying conclusion. I wonder how costly it is to merge SQLite databases. If it's not, then maybe upgrading a library dependency could imply a remerge? However, I do agree that we should have multiple databases in flight - at least, one for semanticdbs on disks and one for the stuff that comes out of the PC.

ShaneDelmore:
SQLite has ability to connect to multiple dbs (32 or 64 if memory serves)

If there are no conflicts merging is as simple as insert into db2.table (select * from db1.table)

xeno-by:
I've just submitted a prototype implementation of an SQLite-based SymbolIndexer as a pull request to scalameta/language-server: https://github.com/scalameta/language-server/pull/94. As expected, the implementation provides instantaneous responses to "Go to Definition" and "Find All References" with negligible memory requirements even on huge codebases.

langmeta-sqlite/src/main/scala/org/langmeta/sqlite/Main.scala
```diff
@@ -0,0 +1,214 @@
...
+import scala.util.control.NonFatal
+import org.langmeta.internal.semanticdb.schema._
+
+object Main {
+  def main(args: Array[String]): Unit = {
+    args match {
+      case Array(sqliteFilename, semanticdbFilenames @ _*) =>
+        val appStart = System.nanoTime()
+        var genuineDocuments = 0
+        class Counter { var value = 0; def next() = { value += 1; value }; }
```
DavidDudson: This could be a scala Stream

xeno-by:
In December 2017, we've been further working on relational indexes for .semanticdb files for an internal project at Twitter. In a nutshell, we switched to MySQL for infrastructural reasons and made a couple cosmetic changes to the converter, but the concept itself hasn't changed much. Here's the branch with all the changes: https://github.com/xeno-by/scalameta/commits/topic/q4.

Based on our experience, I can say that the idea of relational indexes worked exceptionally well. These indexes are easy to build, easy to query and easy to deploy - even for huge codebases. I think that this functionality merits upstreaming.

In Q1 2018, we're planning to rebase topic/q4 on top of master, address the feedback from this pull request and submit a new pull request that makes relational indexing functionality available for everyone. In the meanwhile, I'll be closing this PR. Please stay tuned for updates.

PR #1178, opened 2017-12-08
Improve Position.Range.toString

MasseGuillaume:
I don't like the current Position.Range toString, it outputs the whole string of a file. I'm open to other ideas but the first one that came to my mind is to use ANSI colors to highlight the range and wrap it with some context (100 chars before and after). This is what it looks like:

a

To consider:

How to display Points such as `Position.Range(x, y) x = y` ?
What to do when `Position.Range(x, y)` y >>> x? Do we cut the text in the middle?
What to do when the range is the whole file?


olafurpg:
Superseded by #1186

PR #1179, opened 2017-12-09
Reorganize build

xeno-by:
Before jumping into performance optimizations of semanticdb-scalac, I wanted to split langmeta into langmeta-lowlevel and langmeta-highlevel, so that I could make semanticdb-scalac depend on langmeta-lowlevel and not worry about usages of the slow highlevel API unexpectedly leaking into the plugin.

That proved to be a tricky undertaking, as I learned that there's a bunch of tests scattered around our repository, and some of them are finicky about placement of various files. I fixed one thing, then another, and then realized that it probably would be a good idea to merge all these tests into one project, so that noone has to go through this again. While I was doing work in build.sbt, I took the liberty to clean up some stuff, too.

Four hours later, I merged all tests into one big project (`tests`, which to my surprise already existed) and was ready to proceed with my original goal. However, this pull request is getting too big, so I'll leave that story to a separate pull request.


build.sbt
```diff
@@ -337,6 +343,17 @@ lazy val tests = crossProject
 lazy val testsJVM = tests.jvm
 lazy val testsJS = tests.js
 
+lazy val tests210 = project
```
olafurpg: 👍 I had a comment about 2.10 in an earlier commit, glad this is taken care of

scalameta/semanticdb-scalac-core/src/main/scala/scala/meta/interactive/InteractiveSemanticdb.scala
```diff
@@ -9,7 +9,7 @@ import scala.tools.nsc.Settings
 import scala.tools.nsc.interactive.Global
 import scala.tools.nsc.interactive.Response
 import scala.tools.nsc.reporters.StoreReporter
-import org.langmeta.semanticdb.Document
```
olafurpg: FYI I reference org.langmeta in my code because IJ autoimports those modules when I type `Documen<COMPLETE>`. "Go to definition" on the scala.meta type aliases is also quite annoying.

PR #1180, opened 2017-12-09
Split langmeta into langmeta-lowlevel and langmeta-highlevel

xeno-by:
This pull request builds on https://github.com/scalameta/scalameta/pull/1179 and splits langmeta into lowlevel and highlevel parts. I'll need this split for my performance work, where I plan to downgrade the `semanticdbScalacCore -> scalametaJVM` dependency to `semanticDbScalacCore -> langmetaLowlevelJVM`.

Migration instructions:
  * `s/"org.scalameta" %% "langmeta"/"org.scalameta" %% "langmeta-highlevel"/g`
  * `s/org.langmeta.io/org.langmeta.highlevel.io/g`
  * `s/org.langmeta.inputs/org.langmeta.highlevel.inputs/g`
  * `s/org.langmeta.semanticdb/org.langmeta.highlevel.semanticdb/g`
  * `s/org.langmeta.internal.semanticdb.schema/org.langmeta.lowlevel.semanticdb/g`
  * `s/org.langmeta.internal.semanticdb.{schema => s}/org.langmeta.lowlevel.{schema => s}/g`


xeno-by:
@olafurpg How do we tell mima that this pull request intentionally breaks backward compatibility?

olafurpg:
I am not convinced this refactoring is worth its weight. This breaks close to every application I've written against semanticdb: scalafix, metadoc, doxy, language-server, corpus experiments. Even if we could automigrate it with scalafix then it take a lot of time to get all 3rd party scalafix rules to move to the `highlevel` package. As I mentioned in #1179, IJ is eager to auto-import `org.langmeta` modules so there are tons of references to those whether you like it or not 😉 

I propose we keep the package layouts unchanged as much as possible, including `internal.semanticdb.schema`. Let's prioritize performance improvements without moving APIs.

olafurpg:
I am btw taking compatibility extra seriously for scalafix because we compile rules from source, breaking changes in the API means that many scalafix rules that is shared via source may no longer compile.

xeno-by:
You convinced me with your point about scalafix, so I'll be giving up the `highlevel` part.

I think we're in agreement that the rest make sense, and I will do just that in a separate PR together with performance changes.

PR #1184, opened 2017-12-10
Disable semanticdb-scalac in presentation compiler

olafurpg:
It has taken me close to a full day to track down an infinite loop
in the pc to this change. Lesson learned, the presentation compiler
is a fragile beast.

I've manually verified that this patch fixes the infinite loop in language-server

screen shot 2017-12-10 at 19 49 04


xeno-by:
Lgtm

PR #1187, opened 2017-12-11
Fix #1183

### LINKED ISSUE #1183: langmeta tests210 is not 2.10 specific
those tests should run on all scala versions. Regressed #1173

xeno-by:


build.sbt
```diff
@@ -31,7 +31,7 @@ addCommandAlias("benchQuick", benchQuick.command)
 // it runs `test` sequentially in every defined module.
 commands += Command.command("ci-fast") { s =>
   if (ciScalaVersion.startsWith("2.10")) {
-    s"wow $ciScalaVersion" :: "tests210/test" :: s
+    s"wow $ciScalaVersion" :: "langmetaTestsJVM/test" :: s
```
olafurpg: `plz $ciScalaVersion test` should do this for you ;) but I don't mind as long as it works!

PR #1190, opened 2017-12-11
Cache g.Symbol.toSemantic

xeno-by:


scalameta/semanticdb-scalac-core/src/main/scala/scala/meta/internal/semanticdb/SymbolOps.scala
```diff
@@ -1,84 +1,96 @@
 package scala.meta.internal
 package semanticdb
 
+import java.util.HashMap
 import scala.{meta => m}
 import scala.meta.internal.inputs._
 
 trait SymbolOps { self: DatabaseOps =>
 
+  private lazy val symbolCache = new HashMap[g.Symbol, m.Symbol]
```
olafurpg: Can we maybe do this change without adding a 2 space indentation on the whole method? The diff makes it difficult to see the difference

scalameta/semanticdb-scalac-core/src/main/scala/scala/meta/internal/semanticdb/SymbolOps.scala
```diff
@@ -1,84 +1,96 @@
...
+          else m.Signature.Term(name(sym))
+        }
+        m.Symbol.Global(owner, signature)
+      }
+      val msym = symbolCache.get(sym)
+      if (msym != null) {
+        msym
+      } else {
+        val msym = uncached(sym)
+        symbolCache.put(sym, msym)
```
olafurpg: Is this cache cleared anywhere?

bench/README.md
```diff
@@ -38,6 +38,7 @@ Other Scalameta configurations progressively enable more and more features:
 Benchmark                      Mode    Cnt Score
 QuickScalacBaseline            sample  35  1595.274 ± 15.263  ms/op
 QuickScalacRangepos            sample  35  1703.097 ± 25.044  ms/op  (6% overhead)
+QuickScalametaOpt              sample  15  1934.902 ± 29.232  ms/op  (21% overhead)
```
olafurpg: Wow, is that a speedup from 42% to 21% overhead?

xeno-by: It's just 25% to 21% :)

scalameta/semanticdb-scalac-core/src/main/scala/scala/meta/internal/semanticdb/SymbolOps.scala
xeno-by: Good point.

scalameta/semanticdb-scalac-core/src/main/scala/scala/meta/internal/semanticdb/SymbolOps.scala
xeno-by: Would you like to have a separate commit that just changes indentation?

scalameta/semanticdb-scalac-core/src/main/scala/scala/meta/internal/semanticdb/SymbolOps.scala
```diff
@@ -1,84 +1,96 @@
...
 
+import java.util.HashMap
 import scala.{meta => m}
 import scala.meta.internal.inputs._
 
 trait SymbolOps { self: DatabaseOps =>
 
+  lazy val symbolCache = new HashMap[g.Symbol, m.Symbol]
   implicit class XtensionGSymbolMSymbol(sym: g.Symbol) {
     def toSemantic: m.Symbol = {
```
olafurpg: Can we keep the original function unchanged except named `toSemanticUncached` and then create new `toSemantic` that uses caching? I feel it's difficult to follow the logic when the nesting is too deep. The big diff also makes it harder to review this change.

olafurpg: Nevermind, I want to edit this method to fix #1194

Issue #1191, opened 2017-12-11
Add post order traversal macro


DavidDudson:
I trialed handwriting a custom traversal algorithm for scalagen. 

However we need to handle `Term.Block` which means we need to handle all `Terms`.
There's too many cases to handle via pattern matching.

In order to implement full transformations we need to have a post-order transform method in scalameta.

DavidDudson:
Actually, I have not actually looked. But is `transform` post-order already?

olafurpg:
Is this fixed with `leafFirstTranform` in https://github.com/scalameta/scalagen/pull/14/files#diff-430d2cf7cf04c40062d98b67c80416eeR8 ?

DavidDudson:
Yes

PR #1192, opened 2017-12-12
Implement leaf first traversal

DavidDudson:
Needed for scalagen

Will not affect existing code.

Note: Implicit param has been added to the traverse method, so this is a bin-compat issue.

### DIFF
--- a/scalameta/transversers/shared/src/main/scala/scala/meta/transversers/Api.scala
+++ b/scalameta/transversers/shared/src/main/scala/scala/meta/transversers/Api.scala
@@ -2,33 +2,58 @@ package scala.meta
 package transversers
 
 private[meta] trait Api {
+
+  sealed trait TraversalOrder
+  case object LeafFirst extends TraversalOrder
+  case object RootFirst extends TraversalOrder
+
+  implicit val order: TraversalOrder = RootFirst
+
   implicit class XtensionCollectionLikeUI(tree: Tree) {
-    def transform(fn: PartialFunction[Tree, Tree]): Tree = {
+    def transform(fn: PartialFunction[Tree, Tree])(implicit order: TraversalOrder): Tree = {
       object transformer extends Transformer {
         override def apply(tree: Tree): Tree = {
-          if (fn.isDefinedAt(tree)) super.apply(fn(tree))
-          else super.apply(tree)
+          order match {
+            case RootFirst =>
+              if (fn.isDefinedAt(tree)) super.apply(fn(tree))
+              else super.apply(tree)
+            case LeafFirst  =>
+              if (fn.isDefinedAt(tree)) fn(super.apply(tree))
+              else super.apply(tree)
+          }
         }
       }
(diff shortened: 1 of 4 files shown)


DavidDudson:
This fixes #1191

scalameta/transversers/shared/src/test/scala/meta/transversers/OrderTests.scala
```diff
@@ -0,0 +1,109 @@
...
+      "scala.meta.Template$TemplateImpl",
+      "scala.meta.Defn$Class$DefnClassImpl")
+
+  val input = q"class Foo { def bar = 1 }"
+
+  test("Traverse visits roots first") {
+    val visited: mutable.ArrayBuffer[String] = mutable.ArrayBuffer()
+
+    input.traverse {
+      case a => visited.append(a.getClass.getName)
```
olafurpg: FYI `.productPrefix` will give you a cleaner name per tree node

DavidDudson: Nice! Usually I use getClass.getSimpleName but of course that fails here.

DavidDudson:
Let's leave this PR outstanding for now. I'll try and whip something up for `scalagen` specfically for now.

If possible, I would like to convert the AST/Traversal macros to generators before scalameta 3.0. That way the generators would be a lot more approachable than the current macro based code for modifying.

scalameta/transversers/shared/src/main/scala/scala/meta/transversers/Api.scala
```diff
@@ -2,33 +2,58 @@ package scala.meta
 package transversers
 
 private[meta] trait Api {
+
+  sealed trait TraversalOrder
+  case object LeafFirst extends TraversalOrder
+  case object RootFirst extends TraversalOrder
```
xeno-by: How about factoring out these definitions into `TraversalOrder.scala` and then aliasing them here if you feel that these guys belong to the public API? This is the pattern that we're using in other modules of Scalameta.

scalameta/transversers/shared/src/test/scala/meta/transversers/OrderTests.scala
```diff
@@ -0,0 +1,109 @@
+package scala.meta.transversers
```
xeno-by: Recently we have moved the overwhelming majority of tests into a single big project at `scalameta/tests`. Let's continue doing this in future.

olafurpg:
@DavidDudson if we're gonna whip up something in scalagen I propose we close this PR. I'd love to expose more configurable traversal strategies in the main scalameta API but I think we should 
- consider how to do it without breaking the existing API
- get more hands-on experience with it in scalagen
In scalafix, I also have a lot of scalameta utilities that should probably go upstream at some point.

DavidDudson:
Fair enough

PR #1201, opened 2017-12-18
Use PartialFunction.lift to avoid duplicate evaluation. Fixes #1200.

### LINKED ISSUE #1200: .transform/.traverse should only trigger partial function once
I had to admit it but I was writing a partial function today that did have a side effect and I hit on this
```scala
@ var i = 0
i: Int = 0

@ q"A".transform {
    case q"A" if {
    i += 1
    i < 2
  } => q"b"
  }
scala.MatchError: A (of class scala.meta.Term$Name$TermNameImpl)
  scala.PartialFunction$$anon$1.apply(PartialFunction.scala:254)
  scala.PartialFunction$$anon$1.apply(PartialFunction.scala:252)
```
We should use `.applyOrElse` which is designed exactly for this use-case.

olafurpg:


DavidDudson:
I’ll do the same in leafFirstTraverse in scalagen. I still think that side effects happening when the partial function is not even applied is unusual, but oh well.

olafurpg:
> I still think that side effects happening when the partial function is not even applied is unusual

I agree, but this change is also a nice performance improvement since the guards are only run once instead of twice. For large patterns this can matter a lot!

PR #1205, opened 2017-12-31
Contrib more instances

DavidDudson:
Pushing some scalagen code upstream.

More changes in this area will probably come.


olafurpg:
s/ignore/exclude/

PR #1207, opened 2018-01-05
Fix bug in Tokens.slice causing incorrect length

marcelocenerine:
```scala
val tokens = "foo".tokenize.get

tokens.slice(1, 0).length
res0: Int = -1

tokens.slice(1, 0).isEmpty
res1: Boolean = false

tokens.slice(100, 200).length
res5: Int = -97
```

/cc @olafurpg

### DIFF
--- a/scalameta/tokens/shared/src/main/scala/scala/meta/tokens/Tokens.scala
+++ b/scalameta/tokens/shared/src/main/scala/scala/meta/tokens/Tokens.scala
@@ -29,7 +29,11 @@ import scala.meta.internal.prettyprinters._
 extends immutable.IndexedSeq[Token] with collection.IndexedSeqOptimized[Token, immutable.IndexedSeq[Token]] {
   def apply(idx: Int): Token = tokens(start + idx)
   def length: Int = end - start
-  override def slice(from: Int, until: Int): Tokens = Tokens(tokens, start + math.max(from, 0), start + math.min(math.max(until, 0), length))
+  override def slice(from: Int, until: Int): Tokens = {
+    val lo = start + from.max(0).min(length)
+    val hi = lo.max(start + until.min(length))
+    Tokens(tokens, lo, hi)
+  }
   override def toString = scala.meta.internal.prettyprinters.TokensToString(this)
 
   override def segmentLength(p: Token => Boolean, from: Int = 0): Int = super.segmentLength(p, from)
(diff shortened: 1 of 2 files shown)


scalameta/tokens/shared/src/main/scala/scala/meta/tokens/Tokens.scala
```diff
@@ -29,7 +29,11 @@ import scala.meta.internal.prettyprinters._
 extends immutable.IndexedSeq[Token] with collection.IndexedSeqOptimized[Token, immutable.IndexedSeq[Token]] {
   def apply(idx: Int): Token = tokens(start + idx)
   def length: Int = end - start
-  override def slice(from: Int, until: Int): Tokens = Tokens(tokens, start + math.max(from, 0), start + math.min(math.max(until, 0), length))
+  override def slice(from: Int, until: Int): Tokens = {
+    val lo = start + ((from max 0) min length)
```
olafurpg: Nitpick, can we use non-infix here?

```scala
    val lo = start + from.max(0).min(length)
    val hi = lo.max(start + until.min(length))
```

Infix should be reserved for symbolic operators.

tests/shared/src/test/scala/scala/meta/tests/tokens/TokensSuite.scala
```diff
@@ -51,14 +51,13 @@ class TokensApiSuite extends FunSuite {
...
     //Drop BOF and EOF to make tests more readable
-    var tokens = tokenize("((1 + 1) == 2)").drop(1).dropRight(1)
+    val tokens = tokenize("((1 + 1) == 2)").drop(1).dropRight(1)
 
     assert(tokens.length == 13)
     assert(tokens.segmentLength(_.is[LeftParen]) === 2)
     assert(tokens.segmentLengthRight(_.is[RightParen]) === 1)
     assert(tokens.take(2).syntax === "((")
     assert(tokens.slice(11, 13).syntax === "2)")
-    assert(tokens.slice(11, 13).syntax === "2)")
```
olafurpg: 👍

scalameta/tokens/shared/src/main/scala/scala/meta/tokens/Tokens.scala
marcelocenerine: done

Issue #1147, opened 2017-11-06
Improve semanticdb-scalac performance


olafurpg:
Currently, semanticdb-scalac adds quite a bit of overhead (I've heard 20-30%) on compilation. Memory usage is also very high.

This overhead is an issue for larger projects (where compile times are already slow enough) and more interactive usage of the plugin (editor integrations).

I've heard the scalameta parser is a big culprit, -Yrangepos is another one. I think we could actually support 98% of the current functionality by replacing the -Yrangepos requirement and scalameta parse step by running the scalac tokenizer on the source contents and collect all identifier `(start, end)` pairs. The scalameta parser is used correctly handle resolved names to for example constructors and `private[within]`, I'm sure we can come up with creative workarounds for those cases.

Another potential optimization, but likely not significant, is that we build `scala.meta.Database` that we convert to `org.langmeta.internal.semanticdb.schema.Database`, then to `Array[Byte]` and then write to file. We can probably remove `scala.meta.Database` and `Array[Byte]` without big loss of readability of maintainability.

olafurpg:
Our experience in https://github.com/scalameta/language-server is that removing the `scala.meta.Database => internal.schema.Database` conversion can lead up to a ~10x speedup improvement during symbol indexing, this might be a good first step to improve performance in the compiler plugin.

Another cheap win might also be to delegate I/O to a separate thread, although I haven't measured that overhead yet.

Before we start optimizations, we need a good benchmark suite to ensure we're making progress.

olafurpg:
I took a stab at this issue today and preliminary results are promising: overhead goes down from ~28% to ~17%. 

WIP diff is here https://github.com/scalameta/scalameta/compare/master...olafurpg:perf?expand=1 I still haven't ported some of the more advanced features like naming resolution of named arguments/within/ctors, but symbols/synthetics/messages work like before

I was hoping for a bigger speedup more but this is still a pretty good start. I used -P:semanticdb:profiling:console and to measure the overhead and used scalap as a corpus.

```bash
# 2.1.2 baseline
time for i in {1..5}; do time scalac -d target/out -Yrangepos -Xplugin:/Users/ollie/.coursier/cache/v1/https/repo1.maven.org/maven2/org/scalameta/semanticdb-scalac_2.12.4/2.1.2/semanticdb-scalac_2.12.4-2.1.2.jar -P:semanticdb:profiling:console $(find corpus/scalap/a8c43dc/scalax/ -type f); done
[...]

PR #1218, opened 2018-01-18
Improve semanticdb-scalac performance.

olafurpg:
First step towards #1147. This PR also removes the scalameta dependency in the semanticdb-scalac compiler plugin, which is another first step towards making it potentially easier to upstream into scalac.

This PR refactors the compiler plugin to avoid
- -Yrangepos requirement, run scalac tokenizer directly instead
- scalameta parse step, emit resolved names directly from scalac trees
- 4 layers of converters from plugin to bytes on disk
Previously, the plugin created high-level data structures that got
converted to scalapb data structures that got converted to Array[Byte]
that got written to file.
Now, the plugin writes directly scalapb messages that get persisted
directly to disk.

By removing the scalameta parse step, a few resolved names are now missing:
- constructor references
- private[within]
- named argument parameters

The named argument parameters are the most critical issue, in my
opinion. I would like to figure out to get those back asap. However, I
believe we should try and merge this before the PR gets too big.

I will post  benchmark numbers as they appear. Early numbers hint the overhead has dropped from ~32% to 19%.

### DIFF
--- a/langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
+++ b/langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
@@ -19,6 +19,23 @@ import org.langmeta.{semanticdb => d}
 package object semanticdb {
   implicit class XtensionSchemaDatabase(sdatabase: s.Database) {
 
+    def append(relpath: RelativePath, targetroot: AbsolutePath): Unit = {
+      write(relpath, targetroot, StandardOpenOption.CREATE, StandardOpenOption.APPEND)
+    }
+    def save(relpath: RelativePath, targetroot: AbsolutePath): Unit = {
+      write(relpath, targetroot, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING)
+    }
+    private def write(
+        relpath: RelativePath,
+        targetroot: AbsolutePath,
+        openOptions: OpenOption*): Unit = {
+      val semanticdbPath = targetroot.resolve(v.SemanticdbPaths.fromScala(relpath)).toNIO
+      Files.createDirectories(semanticdbPath.getParent)
+      val out = Files.newOutputStream(semanticdbPath, openOptions: _*)
+      try sdatabase.writeTo(out)
+      finally out.close()
+    }
+
     def mergeMessageOnlyDocuments: s.Database = {
       // returns true if this document contains only messages and nothing else.
       // deprecation messages are reported in refchecks and get persisted
@@ -38,7 +55,7 @@ package object semanticdb {
       } else {
         sdatabase.documents match {
           case Seq(doc, messages)
-            if doc.filename == messages.filename &&
+              if doc.filename == messages.filename &&
                 isOnlyMessages(messages) =>
             val x = doc.addMessages(messages.messages: _*)
             s.Database(x :: Nil)
@@ -59,15 +76,23 @@ package object semanticdb {
     }
 
     def toDb(sourcepath: Option[Sourcepath], sdoc: s.Document): d.Document = {
-      val s.Document(sunixfilename, scontents, slanguage, snames, smessages, ssymbols, ssynthetics) = sdoc
+      val s.Document(
+        sunixfilename,
+        scontents,
+        slanguage,
+        snames,
+        smessages,
+        ssymbols,
+        ssynthetics) = sdoc
       assert(sunixfilename.nonEmpty, "s.Document.filename must not be empty")
       val sfilename = PathIO.fromUnix(sunixfilename)
       val dinput = {
         if (scontents == "") {
           val uri =
-            sourcepath.getOrElse(sys.error("Sourcepath is required to load slim semanticdb."))
-                .find(RelativePath(sfilename))
-                .getOrElse(sys.error(s"can't find $sfilename in $sourcepath"))
+            sourcepath
+              .getOrElse(sys.error("Sourcepath is required to load slim semanticdb."))
+              .find(RelativePath(sfilename))
+              .getOrElse(sys.error(s"can't find $sfilename in $sourcepath"))
           dInput.File(AbsolutePath(uri.getPath))
         } else {
           dInput.VirtualFile(sfilename.toString, scontents)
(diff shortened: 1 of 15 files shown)


DavidDudson:
Do you have any performance numbers for his change? I would like to know how fast the semanticdb really is.

olafurpg:
I will post benchmark numbers as they appear @DavidDudson . Early numbers hint the overhead has dropped from ~40% to 17%.

xeno-by:
@olafurpg What about the overhead of the LSP mode, where: 1) documents are thin, 2) denotations are only saved for definitions, not for references, 3) synthetics are not generated?

olafurpg:
@xeno-by in LSP mode the overhead seems to be ~11% compared to scalac -Yrangepos. Note that -Yrangepos is not enabled in LSP mode because we no longer need it.

```
[info] QuickScalacRangepos.run              sample   41  1262.076 ± 21.051  ms/op
[info] QuickScalametaBaseline.run          semanticdb-s.jar  sample   37  1407.302 ± 40.350  ms/op
```

olafurpg:
I think a good goal to aim for is ~3-5% overhead similar to Index-While-Building in clang, so we still have more work to do 😄 

```
Benchmark               Mode    Cnt Score
QuickScalacRangepos     sample  45  1202.880 ± 15.870 ms/op
QuickScalametaBaseline  sample  40  1349.203 ± 21.157 ms/op (12.2% overhead)
```

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/TargetedSuite.scala
```diff
@@ -161,8 +162,6 @@ class TargetedSuite extends DatabaseSuite(SemanticdbMode.Slim) {
        |_root_.f.C1#T1# => abstract type T1
        |_root_.f.C1#T2# => type T2: Int
        |  [0..3): Int => _root_.scala.Int#
-       |_root_.f.C1#`<init>`()V. => secondaryctor <init>: (): C1
-       |  [4..6): C1 => _root_.f.C1#
```
xeno-by: Can you also mark this and other constructor-related issues as TODO?

tests/jvm/src/test/resources/semanticdb.expect
```diff
@@ -29,7 +28,6 @@ Names:
 
 Messages:
 [41..47): [warning] Unused import
-[72..107): [warning] class Stack in package mutable is deprecated (since 2.12.0): Stack is an inelegant and potentially poorly-performing wrapper around List. Use a List assigned to a var instead.
```
xeno-by: Why is this message gone?

scalameta/semanticdb-scalac-core/src/main/scala/scala/meta/internal/semanticdb/MessageOps.scala
```diff
@@ -1,34 +1,35 @@
 package scala.meta.internal.semanticdb
 
+import org.langmeta.internal.semanticdb.{schema => s}
 import org.scalameta.unreachable
 import scala.{meta => m}
 
 trait MessageOps { self: DatabaseOps =>
   implicit class XtensionCompilationUnitMessages(unit: g.CompilationUnit) {
-    def reportedMessages(mstarts: collection.Map[Int, m.Name]): List[m.Message] = {
+    def reportedMessages(mstarts: collection.Map[Int, s.Position]): List[s.Message] = {
```
xeno-by: 👍

scalameta/semanticdb-scalac-core/src/main/scala/scala/meta/internal/semanticdb/InputOps.scala
```diff
@@ -41,6 +42,10 @@ trait InputOps { self: DatabaseOps =>
   }
 
   implicit class XtensionGPositionMPosition(pos: GPosition) {
+    def toSchema: s.Position = {
+      val mpos = toMeta
+      s.Position(mpos.start, mpos.end)
+    }
     def toMeta: m.Position = {
```
xeno-by: Do we still need `toMeta` here and in other places?

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
```diff
@@ -19,6 +19,23 @@ import org.langmeta.{semanticdb => d}
 package object semanticdb {
   implicit class XtensionSchemaDatabase(sdatabase: s.Database) {
```
xeno-by: Now that we're working with `s.Database`, can we remove `scalametaJVM` from the dependencies of `semanticdb-scalac-core`?

olafurpg:
Here are the numbers comparing master and this PR, I ran the benchmarks both with -Yrangepos enabled and disabled for scalameta since the compiler plugin no longer requires -Yrangepos 

```// Master
Benchmark                      Mode    Cnt Score
QuickScalacBaseline            sample  48  1093.927 ± 20.610 ms/op
QuickScalacRangepos            sample  45  1133.068 ± 16.021 ms/op (3.7%)
QuickScalametaBaseline         sample  36  1448.724 ± 29.351 ms/op (32%)
QuickScalametaFullContents     sample  35  1495.988 ± 23.015 ms/op
QuickScalametaFullDenotations  sample  35  1573.523 ± 27.856 ms/op
QuickScalametaFullSynthetics   sample  35  1592.038 ± 30.538 ms/op (45%)


// This PR, with -Yrangepos enabled for scalameta
Benchmark                      Mode    Cnt Score
QuickScalacBaseline            sample  50  1043.711 ± 12.039 ms/op
QuickScalacRangepos            sample  48  1111.927 ± 15.659 ms/op (6%)
QuickScalametaBaseline         sample  40  1302.594 ± 18.033 ms/op (24%)
QuickScalametaFullContents     sample  40  1294.205 ± 17.843 ms/op
QuickScalametaFullDenotations  sample  39  1395.789 ± 17.234 ms/op
QuickScalametaFullSynthetics   sample  40  1382.443 ± 24.494 ms/op (32%)

// This PR, with -Yrangepos disabled for scalameta
Benchmark                      Mode    Cnt Score
QuickScalacBaseline            sample  49  1076.695 ± 47.137 ms/op
QuickScalacRangepos            sample  45  1155.135 ± 64.021 ms/op (7.3%)
QuickScalametaBaseline         sample  41  1281.564 ± 76.211 ms/op (19%)
QuickScalametaFullContents     sample  41  1280.848 ± 70.359 ms/op
QuickScalametaFullDenotations  sample  40  1369.912 ± 60.247 ms/op (27%)
QuickScalametaFullSynthetics   sample  40  1374.106 ± 55.381 ms/op
```

The difference was less dramatic than what I had hoped for. It would be interesting to see flamegraphs to learn more where the overhead is coming from. I suspect IO.

olafurpg:
I noticed that the numbers are least stable in the last benchmark, which makes sense since I was actively using the computer while running those. The other two benchmarks were run while I was asleep :)

xeno-by:
@olafurpg Here are some experiments that may be interesting:
  1) Instead of traversing ASTs in a separate phase, compute ResolvedNames directly in typedIdent. That's what we do in Rsc.
  2) Instead of creating intermediate ScalaPB data structures, directly write bytes to in-memory buffers.
  3) Instead of writing to multiple files, write to one big file.
  4) Instead of using huge symbol names, shorten the names to "1", "2", "3", etc. (This will be an irreversible transformation without a decoding table, but it'll give an idea about the performance potential of this technique).

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
olafurpg: We should be able to remove scalametaJVM from the dependencies.

scalameta/semanticdb-scalac-core/src/main/scala/scala/meta/internal/semanticdb/InputOps.scala
olafurpg: we shouldn't, these methods should be removed.

tests/jvm/src/test/resources/semanticdb.expect
olafurpg: I didn't manage to investigate it. I haven't been able to make use of deprecation messages yet so I was willing to give it up

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/TargetedSuite.scala
olafurpg: I would love to accomplish this refactoring with 0 breaking changes, and I believe it's possible without scalameta parse step but it requires more time to get right.

PR #1219, opened 2018-01-18
Add quickLSP bench command

olafurpg:
For simpler comparison between master and #1218


project/Build.scala
```diff
@@ -39,6 +39,11 @@ object Build extends AutoPlugin {
       }
     }
 
+    object benchLSP extends BenchSuite {
+      def scalacBenches = List("QuickScalacRangepos")
```
xeno-by: Why are we comparing with rangepos?

olafurpg: Good question, I changed to baseline.

PR #1220, opened 2018-01-19
SemanticDB Specification, Version 3.0.0

xeno-by:
## Motivation

It's been an exciting year for SemanticDB 🎉 

In the beginning of 2017, we created [semanticdb-scalac](http://scalameta.org/tutorial/#semanticdb-scalac) that was then immediately picked up by [Scalafix](https://scalacenter.github.io/scalafix/). June 2017 was the start of active development of [Metadoc](scalameta.org/metadoc). In November 2017, we welcomed [scalameta/language-server](https://github.com/scalameta/metals) (which is now called Metals, by the way 🤘). Finally, just a week ago we started working on [SemanticDB support for Rsc](https://github.com/xeno-by/reasonable-scala/commits/10).

Our experience with SemanticDB shows that it's not just useful for toy projects, but that it can scale to huge codebases that span millions of lines of code. Our experiments with indexing SemanticDB data into relational databases (https://github.com/scalameta/scalameta/pull/1174 and https://github.com/scalameta/metals/pull/94) demonstrate that resulting databases have reasonable size and single/double-digit millisecond response times.

So far, we've been more or less flying under the radar. Our protobuf schema lives in `package org.langmeta.internal.semanticdb.schema`, and our implementation is our specification. However, now that we've seen evidence that SemanticDB is a good idea, let's make things more official.

## Proposal

  * Let's write an official specification for SemanticDB and commit to having the specification drive the implementations.
  * Let's commit to proper versioning of the SemanticDB data model.
  * Let's extract the protobuf schema into a dedicated project.
  * Let's put the corresponding data structures into a public package.

## Implementation

In this pullrequest, I'd like to present one approach to implementing this proposal that @olafurpg and I have been working on this week. In a nutshell:
  * A specification document was added with the intention to document all the little implementation details that weren't properly communicated until now.
  * The data structures were refactored to accommodate your feedback and share as much vocabulary with Language Server Protocol as possible. After the schema becomes public, making breaking changes will be very complicated, so we decided to use this occasion to do some cleanup.
  * Thanks to the [recent breakthrough by the authors of ScalaPB](https://github.com/scalapb/ScalaPB/issues/387) and to the fact that our newly created project is so small, we've been able to crosscompile the new SemanticDB infrastructure to all three supported platforms: JVM, JS and Native.
  * The SemanticDB protobuf schema was moved to `org.scalameta %% semanticdb3` under the `scala.meta.semanticdb3` package.
  * All work is being done in a new sbt project that lives in a new package and publishes to a new artifact, which means that existing code that works now will continue working in the near future, and we'll be able to go through a civilized deprecation cycle. It may even be possible to support both the old and the new SemanticDB schemas in both the old and the new versions of Scalameta, but that remains to be seen.

There is still work to be done - some of it explicitly marked as TODO in the specification and some of it lies deep in the internals of `semanticdb-scalac` (with https://github.com/scalameta/scalameta/pull/1218 being the first step) - but it seems that we already have enough substance to have a productive discussion. Let us know what you think!

### DIFF
--- a/semanticdb3/semanticdb3.md
+++ b/semanticdb3/semanticdb3.md
@@ -1,0 +1,491 @@
+# SemanticDB Specification, Version 3.0.0
+
+  * [Motivation](#motivation)
+  * [Overview](#overview)
+  * [Model](#model)
+    * [TextDocument](#textdocument)
+    * [Range](#range)
+    * [Symbol](#symbol)
+    * [Identifier](#identifier)
+    * [SymbolInformation](#symbolinformation)
+    * [Diagnostic](#diagnostic)
+    * [Synthetic](#synthetic)
+  * [Schemas](#schemas)
+    * [Protobuf](#protobuf)
+    * [JSON](#json)
+    * [SQL](#sql)
+  * [Changelog](#changelog)
+    * [3.0.0](#300)
+
+## Motivation
+
+Nowadays, there is a clear trend towards standards for communication between developer tools.
+Language Server Protocol (LSP) [\[2\]][2], a protocol that connects programming
+language implementations and text editors, has gained strong industrial support
+and at the time of writing has implementations for over 25 programming languages.
+Build Server Protocol (BSP) [\[3\]][3] follows in LSP's tracks with an ambition to
+define a protocol for communication between language servers and build tools.
+
+While lots of work in the open-source community has been invested in unifying
+user experience (by codifying commonly used operations like go to definition or
+find all references), relatively less work went into unifying implementor experience.
+For example, at the moment, there exist five different LSP implementations for Scala
+[[4][4], [5][5], [6][6], [7][7], [8][8]]. They all implement the same protocol
+that works with code, but they all use different data structures to represent that code.
+
+Without a standard way to share information between tools, implementors have
+two unpleasant choices. First, they can use compiler internals, which are often
+underdocumented and lack compatibility guarantees. Otherwise, they reimplement
+compiler internals, which usually leads to duplication of effort and inconsistent UX.
+For example, Scala IDE [\[9\]][9] uses Scala compiler internals, which has known
+stability problems in interactive mode. To the contrast, IntelliJ [\[10\]][10] has
+its own Scala typechecker, which is more stable but is known for spurious red squiggles.
+
+This demonstrates the necessity for portable metaprogramming APIs -
+something that we have been working on within Scalameta [\[11\]][11].
+In the previous years, we shipped portable syntactic APIs for Scala,
+including abstract syntax trees, parsing and prettyprinting [\[12\]][12].
+SemanticDB is our take on portable semantic APIs.
+
+## Overview
+
+SemanticDB is a data model for semantic information about programs in Scala and
+other languages. SemanticDB decouples production and consumption of semantic information,
+establishing documented means for communication between tools.
+
+TODO: Once we upgrade semanticdb-scalac to support SemanticDB v3,
+demonstrate a hello world example. Show the original program, a prettyprinted
+version of its SemanticDB data, very briefly explain the sections.
+
+TODO: Once we ship metac and metap that support SemanticDB v3, explain that
+the SemanticDB data model is materialized as Protocol Buffers [\[13\]][13], JSON [\[14\]][14]
+and SQL [\[15\]][15]. Demonstrate how we can use metac to generate `.semanticdb` files
+from the hello world example above, show the binary payload, decode the files with metap.
+
+TODO: Explain how these protobuf files can be used in tools like scalafix, metadoc and metals.
+Mention that beyond protobuf files, there are many ways to use SemanticDB as an
+interchange format, including passing objects around in memory, communication via JSON, etc etc.
+
+## Model
+
+In this section, we describe the SemanticDB data model by going through the
+individual sections of the associated protobuf schema. However, in addition to protobuf,
+we also support other kinds of schemas, including JSON and SQL.
+See [Schemas](#schemas) for more information.
+
+### TextDocument
+
+```protobuf
+message TextDocuments {
+  repeated TextDocument textDocuments = 1;
+}
+
+message TextDocument {
+  string model = 255;
+  string uri = 1;
+  string text = 2;
+  string language = 3;
+  repeated Identifier identifiers = 4;
+  repeated SymbolInformation symbols = 5;
+  repeated Diagnostic diagnostics = 6;
+  repeated Synthetic synthetics = 7;
+}
+```
+
+`TextDocument` is the central data structure in the SemanticDB model.
+It provides semantic information about a snippet of code written in a programming
+language `language`.
+
+SemanticDB payloads must include the descriptor of the data model that was used
+to produce them in the `model` field. The following descriptors are supported:
+
+<table>
+  <tr>
+    <td><b>Descriptor</b></td>
+    <td><b>Explanation</b></td>
+    <td><b>Data model<b></td>
+  </tr>
+  <tr>
+    <td>&lt;missing&gt;</td>
+    <td>Legacy SemanticDB payloads</td>
+    <td><a href="https://github.com/scalameta/scalameta/blob/v2.1.5/langmeta/langmeta/shared/src/main/protobuf/semanticdb.proto">semanticdb.proto</a></td>
+  </tr>
+  <tr>
+    <td><code>semanticdb3</code></td>
+    <td>Current version of the model</td>
+    <td>Described in this document</td>
+  </tr>
+</table>
+
+Snippets of code referenced in `TextDocument` can be defined
+in one of three ways: 1) via a URI [\[16\]][16] provided in `uri`,
+2) via a string provided in `text`, 3) via a combination of both.
+
+Semantic information is stored in so called sections - repeated fields within
+the message definition - as described below. These sections are optional, which
+means that documents providing only part of semantic information for the corresponding snippet
+(or no semantic information at all) are completely legal.
+
+### Range
+
+```protobuf
+message Range {
+  int32 start_line = 1;
+  int32 start_character = 2;
+  int32 end_line = 3;
+  int32 end_character = 4;
+}
+```
+
+`Range` in SemanticDB directly corresponds to `Range` in LSP [\[2\]][2].
+It represents a range between start and end points in a document. Both points
+are represented by zero-based line and zero-based character offsets.
+The start point is inclusive, while the end point is exclusive.
+
+### Symbol
+
+Symbols are tokens that are used to correlate identifiers and definitions.
+In the SemanticDB model, symbols are represented as strings.
+At the moment, the symbol format is defined by the needs of the Scala implementations.
+In the future, we are planning to pay more attention to other languages,
+and we may decide to expand the format to accommodate namespacing rules not present in Scala.
+
+**Global symbols**. Correspond to a definition that can be referenced outside
+the compilation unit where the definition is defined.
+Global symbol format is a concatenation of signatures of the owner chain
+of the corresponding global definition, where:
+  * The owner chain of a definition is a list of its enclosing definitions
+    starting with the outermost one, with the outermost definition being
+    either `_empty_` (the special empty package [\[17\]][17]) or
+    `_root_` (the special root package [\[18\]][18]). For example,
+    for the standard `Int` class, the owner chain is `[_root_, scala, Int]`.
+  * The signature of a definition is:
+    * For a method, concatenation of its name, its JVM method descriptor [\[19\]][19]
+      and a dot (`.`). The JVM method descriptor is used to distinguish
+      overloaded methods as mandated by the Scala Language Specification [20].
+      For any other term definition (package, object, val or var),
+      concatenation of its name and a dot (`.`).
+    * For a type definition (class, type alias or type member),
+      concatentation of its name and a pound sign (`#`).
+
+For example, the standard `Int` class must be modelled by a global symbol `_root_.scala.Int#`.
+
+**Local symbols**. Correspond to a definition that isn't global (see above).
+Local symbol format is deliberately unspecified except for two restrictions:
+  * Local symbols must start with `local`, so that they can be easily
+    distinguished from global symbols.
+  * Local symbols must be unique within the underlying document.
+
+For example, `x` in `def identity[T](x: T): T` may be modelled by local symbols
+`local0`, `local_x`, `local_identity_x`, as long as these names are unique within
+the underlying document. The same logic applies to the type parameter `T`,
+which is also a local definition.
+
+**Placeholder symbols**. Are used to model original snippets of code in [Synthetics](#synthetic).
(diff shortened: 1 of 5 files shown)


semanticdb3/semanticdb3.md
```diff
@@ -0,0 +1,491 @@
...
+```
+
+`TextDocument` is the central data structure in the SemanticDB model.
+It provides semantic information about a snippet of code written in a programming
+language `language`.
+
+SemanticDB payloads must include the descriptor of the data model that was used
+to produce them in the `model` field. The following descriptors are supported:
+
+<table>
```
olafurpg: Why write manual table instead of markdown table?

semanticdb3/semanticdb3.md
```diff
@@ -0,0 +1,491 @@
...
+Semantic information is stored in so called sections - repeated fields within
+the message definition - as described below. These sections are optional, which
+means that documents providing only part of semantic information for the corresponding snippet
+(or no semantic information at all) are completely legal.
+
+### Range
+
+```protobuf
+message Range {
+  int32 start_line = 1;
```
olafurpg: I still have doubts about this change, I believe start/end offsets are a superior format and I have concerns about performance implications. For tooling you will always need to have the buffer available anyways, in which case converting between the two formats is easy.

semanticdb3/semanticdb3.md
```diff
@@ -0,0 +1,491 @@
...
+which is also a local definition.
+
+**Placeholder symbols**. Are used to model original snippets of code in [Synthetics](#synthetic).
+Must not be used outside `Synthetic.text` documents. Placeholder symbols are
+always equal to an asterisk (`*`).
+
+### Identifier
+
+```protobuf
+message Identifier {
```
olafurpg: One issue I have with `Identifier` is that it doesn't specify **what** it's identifying. In LSP there is `TextDocumentIdentifier`, BSP has `BuildTargetIdentifier` while here we have an unqualified `Identifier`. What about calling this `SymbolIdentifier` instead? That would make the relationship between `SymbolIdentifier` and `SymbolInformation` clearer IMO.

semanticdb3/semanticdb3.md
```diff
@@ -0,0 +1,491 @@
...
+  TextDocument signature = 6;
+  repeated string members = 7;
+}
+```
+
+"Symbols" is a section of a [TextDocument](#textdocument) that stores information
+about [Symbols](#symbol) that are defined in the underlying snippet of code.
+In a sense, this section is analogous to symbol tables [\[21\]][21] in compiler.
+
+`Definition` contains assorted metadata for a `symbol`, as explained below.
```
olafurpg: SymbolInformation

semanticdb3/semanticdb3.md
```diff
@@ -0,0 +1,491 @@
...
+message Identifier {
+  Range range = 1;
+  string symbol = 2;
+  int64 role = 3;
+}
+```
+
+"Identifiers" is a section of a [TextDocument](#textdocument) that represents
+the results of name resolution for identifiers in the underlying snippet of code.
+
```
MasseGuillaume: I would give an example of Identifier/role here.

semanticdb3/semanticdb3.md
```diff
@@ -0,0 +1,491 @@
...
+    * [SQL](#sql)
+  * [Changelog](#changelog)
+    * [3.0.0](#300)
+
+## Motivation
+
+Nowadays, there is a clear trend towards standards for communication between developer tools.
+Language Server Protocol (LSP) [\[2\]][2], a protocol that connects programming
+language implementations and text editors, has gained strong industrial support
+and at the time of writing has implementations for over 25 programming languages.
```
olafurpg: and most major editors. I don't think there is any editor with >2% marketshare that doesn't have some LSP client implementation, maybe nano but that still doesn't disqualify "most major editors"

semanticdb3/semanticdb3.md
```diff
@@ -0,0 +1,491 @@
...
+that the identifier performs in the snippet of code. Like many other enumerations
+in SemanticDB, this one is usecase-driven and will likely be updated in the future.
+
+<table>
+  <tr>
+    <td><b>Value</b></td>
+    <td><b>Explanation</b></td>
+  </tr>
+  <tr>
+    <td><code>1 << 1</code></td>
```
MasseGuillaume: I don't understand this part. This is a range? `[1, 1]`

semanticdb3/semanticdb3.md
```diff
@@ -0,0 +1,491 @@
...
+    <td>Reference</td>
```
MasseGuillaume: How is `1 << 1` Related to reference?

semanticdb3/semanticdb3.md
```diff
@@ -0,0 +1,491 @@
...
+TODO: Once we ship metac and metap that support SemanticDB v3, explain that
+the SemanticDB data model is materialized as Protocol Buffers [\[13\]][13], JSON [\[14\]][14]
+and SQL [\[15\]][15]. Demonstrate how we can use metac to generate `.semanticdb` files
+from the hello world example above, show the binary payload, decode the files with metap.
+
+TODO: Explain how these protobuf files can be used in tools like scalafix, metadoc and metals.
+Mention that beyond protobuf files, there are many ways to use SemanticDB as an
+interchange format, including passing objects around in memory, communication via JSON, etc etc.
+
+## Model
```
olafurpg: Data Model

semanticdb3/semanticdb3.md
MasseGuillaume: ahhh, It's a bit shift. I would just put 1, 2, 4, ...

semanticdb3/semanticdb3.md
olafurpg: After more thinking I don't like `SymbolIdentifier` since that intuitively means "the ID for this symbol". `SymbolOccurrence` is the most accurate and self-explanatory name I can think of, and it would match with clangs IndexWhileBuilding usage of "Occurrence".

semanticdb3/semanticdb3.md
olafurpg: The idea is that a symbol occurrence can have multiple roles. Currently, we only have reference/definition but other examples that come to mind include `extends`. See https://youtu.be/jGJhnIT-D2M?t=10m12s how they use this role information in xcode https://youtu.be/jGJhnIT-D2M?t=10m12s

semanticdb3/semanticdb3.md
xeno-by: Will do.

semanticdb3/semanticdb3.md
```diff
@@ -0,0 +1,491 @@
...
+At the moment, the supported metadata is usecase-driven and is not supposed to
+be comprehensive or language-agnostic. In the future, we may add support for
+more metadata, for example information about overriding, documentation strings
+or features from other languages.
+
+`kind`. Enumeration that defines the kind of the symbol:
+<table>
+  <tr>
+    <td width="100px"><b>Value</b></td>
+    <td><b>Explanation</b></td>
```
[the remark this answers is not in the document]
xeno-by: Good idea 👍

semanticdb3/semanticdb3.md
xeno-by: I'll think about diplomatic formulation that does LSP justice.

semanticdb3/semanticdb3.md
xeno-by: I thought about changing `Model` to `Data Model` and `Schemas` to `Data Schemas`, but didn't find that the added clarity is worth the added clutter. What's your take on this?

semanticdb3/semanticdb3.md
xeno-by: I find tables easier to maintain than the markdown ASCII art, which doesn't give control of column widths and breaks down on non-trivial formatting. If you'd like, I can try to migrate to the Markdown style.

semanticdb3/semanticdb3.md
xeno-by: Oops.

semanticdb3/semanticdb3.md
```diff
@@ -0,0 +1,491 @@
...
+
+### TextDocument
+
+```protobuf
+message TextDocuments {
+  repeated TextDocument textDocuments = 1;
+}
+
+message TextDocument {
+  string model = 255;
```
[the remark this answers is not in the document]
xeno-by: More flexibility, which I believe is essential. For example, semver uses strings.

xeno-by:
Thanks for the comments, folks - I left some replies above and will work on addressing your feedback in the coming days. As for performance, let's discuss it in https://github.com/scalameta/scalameta/pull/1218#issuecomment-359019052.

semanticdb3/semanticdb3.md
xeno-by: That's a valid argument. In PL, `Identifier` is unambiguous, but within a larger context this name may probably be not so clear. Let's think about alternatives.

What I like about `Identifier` is that it promotes a text-centric view of the program, where we aren't talking compiler internals, but are appealing to what is written in the source file. In that sense, I liked `Identifier` and `Definition`, because that's what's written in source code. 

`SymbolOccurrence` and `SymbolInformation` are pushing us away from that model. For example, with a text-centric view, you can have an identifier or a definition without any attached symbol, e.g. created by a parser. However, if you call them symbol occurrence or information, then it kinda doesn't make sense to have these data structures without semantic information. Now, this is not something that we have been doing with SemanticDB yet, but this illustrates the idea.

semanticdb3/semanticdb3.md
olafurpg: I don't think it adds clutter, jumping straight into the document without context it's not clear what "Model" means.

semanticdb3/semanticdb3.md
olafurpg: Interesting, I agree maintaining markdown tables is a drag. Sure, let's do this 👍

semanticdb3/semanticdb3.md
olafurpg: Fair enough, Index-While-Building seems to use line/column as well https://github.com/apple/swift-clang/blob/stable/test/Index/Store/Inputs/json.c.json

> Also, why does one always need to have the buffer available anyways?

Because you need the ability to account for minor changes in the file contents. Blindly returning reference positions may result in incorrect rename refactorings or bogus highlighted code regions. If you have something like sha256 it's better, but then a single character insert makes the entire snapshot unusable.

olafurpg: the readability of the raw data is improved with line/column, but readability will go down the drain anyways if we use symbol lookup tables to compact the data.

semanticdb3/semanticdb3.md
olafurpg: `SymbolOccurrence` seems to be a thing in swift-clang btw, https://github.com/apple/swift-clang/blob/stable/include/clang/Tooling/Refactor/SymbolOccurrenceFinder.h

My favorite remains `SymbolOccurrence`, and I think it's the most consistent name to pair up with `SymbolInformation`

semanticdb3/semanticdb3.md
olafurpg: Another potential usecase for role would be to label 

```scala
package scala {
  val Left = scala.util.Left
}
```
as an `alias` so that it's easy to "Find references" for `scala.util.Left` and include results for `scala.Left`

semanticdb3/semanticdb3.md
olafurpg: Metals btw already has support to recover from slightly stale semanticdb documents using token edit-distance, and that requires the full text contents. In this space, I think we are actually ahead of alternative technologies and the lesson learned is that you need the text contents anyways so you might as well use start/end.

olafurpg: Regardless, I'm fine with using line/character for the simplicity of manually constructing and validating raw data. I would personally prefer `column` over `character` since I think it's confusing to use `character` like is done in LSP. In my editor and in clang they use `col` 

screen shot 2018-01-19 at 19 13 53

semanticdb3/semanticdb3.md
xeno-by: Okay, I'll update.

semanticdb3/semanticdb3.md
xeno-by: I like the alias idea more and more every day.

semanticdb3/semanticdb3.md
xeno-by: Let me think about this over the weekend.

semanticdb3/semanticdb3.md
xeno-by: Speaking of always having the document text, what about SQL? When you're working with the SQL index, you don't have the luxury of positions on the fly. I suspect that many large-scale code analysis/rewrite jobs will be similarly restrictive.

I think a combination of a fingerprint and optional client smartness (like the one that is implemented in Metals) is the way to go, but let's defer this discussion to a follow-up update to the spec. There are multiple questions in play here - do we fingerprint just the contents of the source file, or its dependencies as well? do we allow fingerprinting to be buildtool-specific? - and they all seem orthogonal to the goal of this pullrequest.

xeno-by: Speaking of the choice between offset and line/column, let's put a hold on this until we have the benchmark data. Then we will be able to talk about readability, too.

semanticdb3/semanticdb3.md
ShaneDelmore: FWIW I like “data model” and schema, but not “data schema”

semanticdb3/semanticdb3.md
ShaneDelmore: I have noticed the usability of raw offsets isn’t great and will admit to pasting hunks of code in an online character counter to quickly figure out what offset I wanted 😊 
On the other hand, offsets are great for dbs that support interval types, especially for fast overlaps lookups.  We could use a datatype that allows encoding two ints in one value, say a decimal type.

xeno-by:
Very well, looks like we have no objections to this proposal 🎉 

I'll execute on the proposal as follows:
  * [X] Factor out SemanticDB v2 (https://github.com/scalameta/scalameta/pull/1228)
  * [ ] Migrate semanticdb-scalac to the low-level SemanticDB API
  * [ ] Introduce SemanticDB v3 addressing the comments here

In the meanwhile, this pull request will be closed. Thank you, everyone, for your extremely helpful feedback 🤝

xeno-by:
I would also like to ask for a moratorium on changes to the SemanticDB schema and the semanticdb-scalac compiler plugin until I finish the migration to v3. /cc @olafurpg @MasseGuillaume @gabro @laughedelic

semanticdb3/semanticdb3.md
xeno-by: Here's benchmark results: https://github.com/xeno-by/scalameta/commit/be54bca11542580f76b670870318ac238f997663. Looks like the effect of refactoring positions is within the fluctuation window.

PR #1225, opened 2018-01-19
Setup publish -SNAPSHOT on merge to master.

olafurpg:
To make it easier to consume latest fixes without triggering a stable
release. We should still be unafraid to cut stable releases, but this
is a nice way to give a pre-release a try before pushing the tag.

### DIFF
--- a/bin/ci-publish.sh
+++ b/bin/ci-publish.sh
@@ -3,10 +3,14 @@ set -eu
 PUBLISH=${CI_PUBLISH:-false}
 SECURE_VAR=${TRAVIS_SECURE_ENV_VARS:-false}
 
-if [[ "$TRAVIS_SECURE_ENV_VARS" == true && -n "$TRAVIS_TAG" && "$CI_PUBLISH" == true ]]; then
+if [[ "$TRAVIS_SECURE_ENV_VARS" == true && "$CI_PUBLISH" == true ]]; then
   git log | head -n 20
   echo "$PGP_SECRET" | base64 --decode | gpg --import
-  sbt ci-publish
+  if [ -n "$TRAVIS_TAG" ]; then
+    sbt ci-publish
+  else
+    sbt -Dscalameta.snapshot=true ci-publish
+  fi
 else
   echo "Skipping publish, branch=$TRAVIS_BRANCH publish=$PUBLISH test=$CI_TEST"
 fi
--- a/build.sbt
+++ b/build.sbt
@@ -19,6 +19,12 @@ lazy val LanguageVersion = LanguageVersions.head
 // ==========================================
 
 sharedSettings
+version.in(ThisBuild) ~= { old =>
+  val suffix =
+    if (sys.props.contains("scalameta.snapshot")) "-SNAPSHOT"
+    else ""
+  customVersion.getOrElse(old.replace('+', '-') + suffix)
+}
 name := {
   println(s"[info] Welcome to scalameta ${version.value}")
   "scalametaRoot"
@@ -40,8 +46,7 @@ commands += Command.command("ci-fast") { s =>
   }
 }
 commands += CiCommand("ci-publish")(
-  if (isTagPush) "publishSigned" :: Nil
-  else Nil
+  "publishSigned" :: Nil
 )
 commands += Command.command("mima") { s =>
   s"very mimaReportBinaryIssues" ::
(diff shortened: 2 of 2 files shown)


build.sbt
```diff
@@ -19,6 +19,12 @@ lazy val LanguageVersion = LanguageVersions.head
 // ==========================================
 
 sharedSettings
+version.in(ThisBuild) ~= { old =>
+  val suffix =
+    if (sys.props.contains("scalameta.snapshot")) "-SNAPSHOT"
+    else ""
+  customVersion.getOrElse(old.replace('+', '-') + suffix)
+}
```
xeno-by: Are these the usual SNAPSHOT releases, which make sbt do updates like crazy, or there's something special about them, since we seem to be publishing them to Sonatype releases?

bin/ci-publish.sh
```diff
@@ -3,10 +3,14 @@ set -eu
...
-if [[ "$TRAVIS_SECURE_ENV_VARS" == true && -n "$TRAVIS_TAG" && "$CI_PUBLISH" == true ]]; then
+if [[ "$TRAVIS_SECURE_ENV_VARS" == true && "$CI_PUBLISH" == true ]]; then
   git log | head -n 20
   echo "$PGP_SECRET" | base64 --decode | gpg --import
-  sbt ci-publish
+  if [ -n "$TRAVIS_TAG" ]; then
+    sbt ci-publish
+  else
+    sbt -Dscalameta.snapshot=true ci-publish
+  fi
```
xeno-by: Why not publish prereleases on every merge like we did previously?

olafurpg: bintray is unstable, has file size limits, and requires different credentials and sbt plugins to work. This change is a tiny change from our current setup

olafurpg: publishing non-SNAPSHOT on every merge to maven central creates noise

build.sbt
olafurpg: it's possible to configure sbt to treat -SNAPSHOT versions like normal versions, we never overwrite an old release because every merge has a unique commit that is included in the version number.

bin/ci-publish.sh
xeno-by: What do you mean by noise? Are we risking to run into quota issues? Is there an important webpage that would be spammed by frequent releases?

olafurpg:
PS. Publishing -SNAPSHOT does not require the `sonatypeRelease` step, `publishSigned` is enough. I currently run `sonatypeRelease` manually from the oss.sonatype.org UI for tag pushes because we have three parallel jobs doing the release where `sonatypeRelease` from one job can accidentally close a half-completed release from another job, this caused a lot of problems in the 2.0.0 release.

bin/ci-publish.sh
olafurpg: see https://github.com/scalameta/scalameta/pull/1225#issuecomment-359033250 publishing -SNAPSHOT to sonatype is logistically simpler than non-SNAPSHOT releases

xeno-by: Okay, that's a good reason. Can you talk a bit more about the workflow?

1) So, the idea is that we publish -SNAPSHOT releases to Sonatype releases but don't promote them to Maven Central?

2) Can there be a possibility to run `sonatypeRelease` after all the CI jobs have finished? If there was such a possibility, would we still need -SNAPSHOT, or we would be able to switch to regular prereleases?

olafurpg: 1. Correct, we publish -SNAPSHOT on every merge and forget about them, they never appear in search results in http://search.maven.org/, only https://oss.sonatype.org/content/repositories/snapshots/org/scalameta/

2. I don't want every merge to go to maven central even if I could. I don't want them to appear on http://search.maven.org/  Stable  releases should be accompanied with a tag, changelist in GitHub releases and attributes to contributors. Stable releases should be something worth sharing on gitter/twitter.

olafurpg: A stable release is still only a tag push away, I don't see any of my projects ending up depending on -SNAPSHOT releases for long periods. I intend to continue pushing tags liberally like I did this morning.

olafurpg: > Can there be a possibility to run sonatypeRelease after all the CI jobs have finished? 

I think it can be done with Travis build stages, which is an experimental feature. We're using it in metals with good success.

PR #1231, opened 2018-01-24
 Add Denotation.overrides for method symbols (fix #1209)

### LINKED ISSUE #1209: Add Denotation.overrides for method symbols
It would be nice if it was possible to get the symbol being overridden in `override def`. This would open opportunities for several interesting applications:

* add explicit `override` where missing
* SAM rewrite
* find all implementations
* go to override

An inspiration for how to approach this issue can be this PR here https://github.com/scalameta/scalameta/pull/1148. This issue can be broken down into roughly the following parts

* [x] add `Denotation.overrides: Option[Symbol]` to scala.meta.Denotation in binary backwards compatible way like we did with members https://github.com/scalameta/scalameta/blob/ecdaadfba7e883a93f4ddf2be7120631975a86ed/langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Denotation.scala
* [x] add same field to semanticdb.proto
* [x] come up with a representation of override symbols in `Denotation.toString`
* [x] Add `-P:semanticdb:overrides:def/none` options similar to MembersMode https://github.com/scalameta/scalameta/blob/ecdaadfba7e883a93f4ddf2be7120631975a86ed/scalameta/semanticdb-scalac-core/src/main/scala/scala/meta/internal/semanticdb/ConfigOps.scala#L85
* [ ] Update `saveDenotation()` in SymbolOps https://github.com/scalameta/scalameta/blob/ecdaadfba7e883a93f4ddf2be7120631975a86ed/scalameta/semanticdb-scalac-core/src/main/scala/scala/meta/internal/semanticdb/DocumentOps.scala#L173
* [ ] Add OverridesSuite similar to MemberSuite https://github.com/scalameta/scalameta/blob/ecdaadfba7e883a93f4ddf2be7120631975a86ed/tests/jvm/src/test/scala-2.12/scala/meta/tests/semanticdb/MemberSuite.scala

The diff in https://github.com/scalameta/scalameta/pull/1148 implements most of those steps.

MasseGuillaume:


scalameta/semanticdb-scalac-core/src/main/scala/scala/meta/internal/semanticdb/DenotationOps.scala
```diff
@@ -105,9 +105,17 @@ trait DenotationOps { self: DatabaseOps =>
...
+      else Nil
+
     def toDenotation: m.Denotation = {
       val (minfo, mnames) = info
-      m.Denotation(flags, name, minfo, mnames)
+      m.Denotation(flags, name, minfo, mnames, Nil, overrides)
     }
+
+    def overridesMembers: List[(m.Symbol, m.Denotation)] =
+      if (config.overrides.isAll) gsym.overrides.map(ov => (ov.toSemantic, ov.toDenotation))
```
xeno-by: This circumvents the `config.denotations` setting. Here's how I propose we take it into account:

```
if (config.overrides.isAll) {
  gsym.overrides.flatMap { ov => 
    if (!config.denotations.saveReferences && ov.symbol.pos.isEmpty) None
    else Some((ov.toSemantic, ov.toDenotation))
  }
}
```

PR #1235, opened 2018-01-29
Serialize Denotation.overrides

MasseGuillaume:


tests/jvm/src/test/resources/semanticdb.expect
```diff
@@ -88,6 +88,40 @@ Synthetics:
...
+Symbols:
+_root_.example. => package example.{+4 members}
+_root_.example.A# => trait A.{+1 members}
+_root_.example.A#foo()I. => abstract def foo: Int
+  [0..3): Int => _root_.scala.Int#
+_root_.example.B# => class B
+_root_.example.B#`<init>`()V. => primaryctor <init>: (): B
+  [4..5): B => _root_.example.B#
+_root_.example.B#foo()I. => def foo: Int
+  [0..3): Int => _root_.scala.Int#.{override _root_.example.A#foo()I.}
```
MasseGuillaume: ^

tests/jvm/src/test/resources/semanticdb.expect
```diff
@@ -107,7 +141,7 @@ Names:
 [95..106): stripPrefix => _root_.scala.collection.immutable.StringLike#stripPrefix(Ljava/lang/String;)Ljava/lang/String;.
 
 Symbols:
-_root_.example. => package example.{+2 members}
+_root_.example. => package example.{+4 members}
```
MasseGuillaume: `example.{A, B}` from Override

tests/jvm/src/test/resources/semanticdb.expect
```diff
@@ -123,9 +157,39 @@ _root_.scala.Int#`+`(I)I. => abstract def +: (x: Int): Int
   [10..13): Int => _root_.scala.Int#
 _root_.scala.Int#`<init>`()V. => primaryctor <init>: (): Int
   [4..7): Int => _root_.scala.Int#
+_root_.scala.collection.GenTraversableLike#headOption()Lscala/Option;. => abstract def headOption: Option[A]
```
MasseGuillaume: Maybe we could rewrite this test to avoid having overrides.

tests/jvm/src/test/resources/semanticdb.expect
olafurpg: Hmm, is `scala.Int` overriding `A#foo()` or am I misunderstanding the denotation syntax? I would expect the syntax to be something like this instead
```scala
_root_.example.B#foo()I. => def foo: Int
  [0..3): Int => _root_.scala.Int#.
  override _root_.example.A#foo()I.
```

wdyt?

tests/jvm/src/test/resources/semanticdb.expect
olafurpg: I think there is something wrong with overrides implementation if we are getting denotations for overrides of all referenced symbols. We should only persist the override denotation for symbols that are overridden in the document.

tests/jvm/src/test/resources/semanticdb.expect
```diff
@@ -123,9 +157,39 @@ _root_.scala.Int#`+`(I)I. => abstract def +: (x: Int): Int
...
+  [0..6): Option => _root_.scala.Option#
+  [7..8): A => _root_.scala.collection.GenTraversableLike#[A]
+_root_.scala.collection.GenTraversableLike#map(Lscala/Function1;Lscala/collection/generic/CanBuildFrom;)Ljava/lang/Object;. => abstract def map: [B, That] => (f: Function1[A, B])(implicit bf: CanBuildFrom[Repr, B, That]): That
```
olafurpg: These symbols are not overridden in the document, so they should not be included here even if `:overrides:all`

tests/jvm/src/test/resources/semanticdb.expect
MasseGuillaume:
```scala
trait A { def foo: Int }
class B() extends A { def foo: Int = 2 }
```
yep, I think it's not the correct symbol.

tests/jvm/src/test/resources/semanticdb.expect
MasseGuillaume: Yep, make sense. I'm adding this test to the `OverridesSuite`

```scala
  overrides(
    "object a { Array.empty[Int].headOption }",
    ""
  )
```

tests/jvm/src/test/resources/semanticdb.expect
```diff
@@ -88,6 +88,41 @@ Synthetics:
...
+Symbols:
+_root_.example. => package example.{+4 members}
+_root_.example.A# => trait A.{+1 members}
+_root_.example.A#foo()I. => abstract def foo: Int
+  [0..3): Int => _root_.scala.Int#
+_root_.example.B# => class B
+_root_.example.B#`<init>`()V. => primaryctor <init>: (): B
+  [4..5): B => _root_.example.B#
+_root_.example.B#foo()I. => def foo: Int
+  override _root_.example.A#foo()I.
```
MasseGuillaume: This is actually a List from the override chain, I think it makes sense to only include the top call.

tests/jvm/src/test/scala-2.12/scala/meta/tests/semanticdb/OverridesSuite.scala
```diff
@@ -5,6 +5,11 @@ import scala.meta.internal.semanticdb.OverrideMode
 import scala.meta.internal.semanticdb.SemanticdbMode
 
 class OverridesSuite extends DatabaseSuite(SemanticdbMode.Slim, overrides = OverrideMode.All) {
+  // Dont save overrides for reference, only for definition
+  overrides(
+    "object a { Array.empty[Int].headOption }",
```
olafurpg: 👍

tests/jvm/src/test/resources/semanticdb.expect
olafurpg: Looks good!

scalameta/semanticdb-scalac-core/src/main/scala/scala/meta/internal/semanticdb/DocumentOps.scala
```diff
@@ -183,10 +183,10 @@ trait DocumentOps { self: DatabaseOps =>
               names(mtree.pos) = symbol
               if (mtree.isDefinition) binders += mtree.pos
 
-              def saveDenotation(): Unit = {
+              def saveDenotation(isDefinition: Boolean): Unit = {
```
olafurpg: `val isDefinition = gsym/mtree.isDefinition`?

tests/jvm/src/test/resources/semanticdb.expect
xeno-by: Thanks for explaining! 👍

build.sbt
```diff
@@ -339,6 +339,8 @@ lazy val semanticdbIntegration = project
         s"-P:semanticdb:failures:error", // fail fast during development.
         s"-P:semanticdb:members:all",
         s"-P:semanticdb:exclude:Exclude.scala",
+         "-P:semanticdb:overrides:all",
+         "-P:semanticdb:denotations:all",
```
xeno-by: Wdyt about making this s"..." for consistency?

scalameta/semanticdb-scalac-core/src/main/scala/scala/meta/internal/semanticdb/DenotationOps.scala
```diff
@@ -109,14 +109,15 @@ trait DenotationOps { self: DatabaseOps =>
       if (config.overrides.isAll) gsym.overrides.map(_.toSemantic)
       else Nil
 
-    def toDenotation: m.Denotation = {
+    def toDenotation(isDefinition: Boolean): m.Denotation = {
```
xeno-by: Let's rename this flag to `saveOverrides`.

scalameta/semanticdb-scalac-core/src/main/scala/scala/meta/internal/semanticdb/DocumentOps.scala
xeno-by: I second this request.

tests/jvm/src/test/scala-2.12/scala/meta/tests/semanticdb/OverridesSuite.scala
```diff
@@ -37,7 +42,8 @@ class OverridesSuite extends DatabaseSuite(SemanticdbMode.Slim, overrides = Over
        |_empty_.B#`<init>`()V. => primaryctor <init>: (): B
        |  [4..5): B => _empty_.B#
        |_empty_.B#foo()I. => def foo: Int
-       |  [0..3): Int => _root_.scala.Int#.{override _empty_.A#foo()I.}
+       |  override _empty_.A#foo()I.
```
xeno-by: I like this prettyprint format ❤️

MasseGuillaume:
Follow up: 2c096d03e549b62379cf34acfe77955ba45bad32

xeno-by:
Looks like a legit failure.

xeno-by:
To avoid rebase problems for my work on SemanticDB v3, I'm going to supersede this pull request with a fix to the typo that failed the build and then merge the new pull request using the approvals from this one. See https://github.com/scalameta/scalameta/pull/1236 for details.

MasseGuillaume:
@xeno-by ops, Thank for fixing the PR!

PR #1241, opened 2018-02-01
rewrite Symbol parser with fastparse

mberndt123:
Hey,

I found the Symbol parser ugly, and since the syntax is currently undocumented, I wanted to make it easier for other people who, like me, reverse engineer the syntax from the parser 😆 (also there was a comment about rewriting it with fastparse, so why not). I like to think I succeeded, but let me know what you think.

I've done some rudimentary manual testing and afaics it behaves the same as the old parser, except it gives better error messages. It also fixes a crash: ```Symbol.unapply("_.``#")``` currently crashes (rather than returning None).

### DIFF
--- a/bench/suite/src/main/scala/org/scalameta/bench/SymbolParser.scala
+++ b/bench/suite/src/main/scala/org/scalameta/bench/SymbolParser.scala
@@ -1,0 +1,123 @@
+package org.scalameta.bench
+
+import java.util.concurrent.TimeUnit
+
+import org.openjdk.jmh.annotations.{Benchmark, BenchmarkMode, Measurement, Mode, Fork}
+import org.langmeta.semanticdb.Symbol
+import org.langmeta.semanticdb.Signature
+
+object SymbolParser {
+  val symbols: Array[String] =
+    Array(
+      "_root_.scala.",
+      "_root_.scala.Int#",
+      "_root_.scala.Predef.`???`().",
+      "_root_.scala.`<repeated>`#",
+      "_root_.scala.annotation.",
+      "_root_.scala.annotation.StaticAnnotation#",
+      "_root_.scala.language.",
+      "_root_.scala.language.existentials.",
+      "_root_.scala.language.higherKinds.",
+      "_root_.scala.package.List#",
+      "_root_.types.",
+      "_root_.types.B#",
+      "_root_.types.B#`<init>`().",
+      "_root_.types.C#",
+      "_root_.types.P#",
+      "_root_.types.P#C#",
+      "_root_.types.P#C#`<init>`().",
+      "_root_.types.P#X#",
+      "_root_.types.P#X#`<init>`().",
+      "_root_.types.P#`<init>`().",
+      "_root_.types.P#x.",
+      "_root_.types.T#",
+      "_root_.types.T#C#",
+      "_root_.types.T#C#`<init>`().",
+      "_root_.types.T#X#",
+      "_root_.types.T#X#`<init>`().",
+      "_root_.types.T#`<init>`().",
+      "_root_.types.T#x.",
+      "_root_.types.Test.",
+      "_root_.types.Test.C#",
+      "_root_.types.Test.C#$anon#L#",
+      "_root_.types.Test.C#$anon#L#[T]",
+      "_root_.types.Test.C#$anon#k().",
+      "_root_.types.Test.C#ByNameType.",
+      "_root_.types.Test.C#ByNameType.m1(Lscala/Function0;).",
+      "_root_.types.Test.C#ByNameType.m1(Lscala/Function0;).(x)",
+      "_root_.types.Test.C#ClassInfoType1.",
+      "_root_.types.Test.C#ClassInfoType2#",
+      "_root_.types.Test.C#ClassInfoType2#`<init>`().",
+      "_root_.types.Test.C#ClassInfoType2#x().",
+      "_root_.types.Test.C#ClassInfoType3#",
+      "_root_.types.Test.C#ClassInfoType3#[T]",
+      "_root_.types.Test.C#MethodType.",
+      "_root_.types.Test.C#MethodType.m3().",
+      "_root_.types.Test.C#MethodType.m4().",
+      "_root_.types.Test.C#MethodType.m5(I).",
+      "_root_.types.Test.C#MethodType.m5(I).(x)",
+      "_root_.types.Test.C#MethodType.m6(Ljava/lang/Object;).",
+      "_root_.types.Test.C#MethodType.m6(Ljava/lang/Object;).(x)",
+      "_root_.types.Test.C#MethodType.m6(Ljava/lang/Object;).T#",
+      "_root_.types.Test.C#MethodType.x1().",
+      "_root_.types.Test.C#MethodType.x2().",
+      "_root_.types.Test.C#RepeatedType.",
+      "_root_.types.Test.C#RepeatedType.m1(Lscala/collection/Seq;).",
+      "_root_.types.Test.C#RepeatedType.m1(Lscala/collection/Seq;).(x)",
+      "_root_.types.Test.C#TypeType.",
+      "_root_.types.Test.C#TypeType.T1#",
+      "_root_.types.Test.C#TypeType.T4#",
+      "_root_.types.Test.C#TypeType.T5#",
+      "_root_.types.Test.C#TypeType.T5#[U]",
+      "_root_.types.Test.C#TypeType.m2().",
+      "_root_.types.Test.C#TypeType.m2().T2#",
+      "_root_.types.Test.C#TypeType.m3().",
+      "_root_.types.Test.C#TypeType.m3().M3#",
+      "_root_.types.Test.C#`<init>`().",
+      "_root_.types.Test.C#annType1.",
+      "_root_.types.Test.C#annType2.",
+      "_root_.types.Test.C#compoundType1.",
+      "_root_.types.Test.C#compoundType2.",
+      "_root_.types.Test.C#compoundType3.",
+      "_root_.types.Test.C#existentialType1.",
+      "_root_.types.Test.C#existentialType1.T#",
+      "_root_.types.Test.C#p.",
+      "_root_.types.Test.C#singleType1.",
+      "_root_.types.Test.C#singleType2.",
+      "_root_.types.Test.C#superType1.",
+      "_root_.types.Test.C#superType2.",
+      "_root_.types.Test.C#superType3.",
+      "_root_.types.Test.C#thisType1.",
+      "_root_.types.Test.C#thisType2.",
+      "_root_.types.Test.C#typeLambda1().",
+      "_root_.types.Test.C#typeLambda1().M#",
+      "_root_.types.Test.C#typeRef1.",
+      "_root_.types.Test.C#typeRef2.",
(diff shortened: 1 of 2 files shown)


DavidDudson:
I'll review once the tests pass :)

xeno-by:
@mberndt123 Thank you for your work! Just this morning, we merged an initial version of the SemanticDB spec, and I hope it'll be useful in the future: https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md#symbol.

mberndt123:
@DavidDudson, the tests do pass, it's just that I haven't signed the CLA yet.

mberndt123:
Well, I've done my best to optimize this new parser, i. e. many cuts and no flatMap (so no construction of new parsers while parsing). That said, the fastparse parser is much higher level and does more (e. g. error handling), so I'd be surprised if it weren't slower than the old parser. Especially given that the old parser is written in a C-like style which should be easy for the JIT to optimize. Of course, all this is pure conjecture for now, I haven't measured any of this.

The question is, what kind of slowdown would be acceptable? I guess the primary question is how much time something like metadoc spends parsing symbols. If it's only 1% of the time anyway, parsing two or three times slower is probably acceptable. If it's 20%, then probably not. I don't suppose you have any numbers about that?

olafurpg:
Good question @mberndt123 ! I agree if symbol parsing is only a fraction of the metadoc runtime then it won't matter if parsing symbols is ~2x slower. When you put it that way, I suspect the symbol parsing performance may turn out to not be a bottle-neck.

I will try to setup a representative benchmark fixture sometime next week to get a better feel.

olafurpg:
I haven't forgotten this PR @mberndt123 Our PR queue has been pretty busy this week so I will try to do better next week to look into this. Sorry for the slowness!

mberndt123:
I've rebased the PR on current master, it now also sorts the symbols for the `Multi` case.

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Symbol.scala
```diff
@@ -33,128 +33,49 @@ object Symbol {
...
+  private val name =
+    P(quotedName | unquotedName)
+
+  private val signature = P {
+    ("[" ~/ name ~ "]").map(Signature.TypeParameter) |
+    ("(" ~/ name ~ ")").map(Signature.TermParameter) |
+    (name ~/ (
+      "#" ~/ PassWith(Signature.Type) |
+      "." ~/ PassWith(Signature.Term) |
+      (("(" ~/ charsExcept('.')).! ~ ".").map(sig => Signature.Method(_: String, sig)) |
```
allanrenucci: I think you can cut after `charsExcept('.')).!`

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Symbol.scala
```diff
@@ -33,128 +33,49 @@ object Symbol {
...
   }
-  def unapply(sym: String): Option[Symbol] = scala.util.Try(apply(sym)).toOption
+
+  private val local: Parser[Symbol] = P {
+    name.filter(_.startsWith("local")).map(Symbol.Local)
+  }
+
+  private val multi = P {
+    ((&("_") ~/ global) | local).rep(min = 0, sep = ";").map {
+      case Nil => Symbol.None
```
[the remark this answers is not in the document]
mberndt123: Hm, Nil works too for e. g. an empty `Vector()`. But yeah, I guess `Seq()` expresses intent a bit more clearly

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Symbol.scala
```diff
@@ -33,128 +33,49 @@ object Symbol {
...
-      def entryPoint(): Symbol = {
-        readChar()
-        parseMulti(Nil)
-      }
+  import fastparse.all._
+  private def charsExcept(c: Char) =
+    (!c.toString ~ AnyChar).rep(min = 0).!
+
+  private val quotedName =
+    P("`" ~ charsExcept('`').! ~ "`")
```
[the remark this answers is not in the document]
mberndt123: right…

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Symbol.scala
```diff
@@ -33,128 +33,49 @@ object Symbol {
...
-        }
-      }
-
-      def entryPoint(): Symbol = {
-        readChar()
-        parseMulti(Nil)
-      }
+  import fastparse.all._
+  private def charsExcept(c: Char) =
+    (!c.toString ~ AnyChar).rep(min = 0).!
```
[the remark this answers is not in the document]
mberndt123: With my naive Benchmark this didn't perform better, and the performance varied more between iterations.

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Symbol.scala
```diff
@@ -33,128 +33,49 @@ object Symbol {
...
+  private val quotedName =
+    P("`" ~ charsExcept('`').! ~ "`")
+  private val unquotedName =
+    P(CharPred(Character.isJavaIdentifierStart) ~ CharsWhile(Character.isJavaIdentifierPart, min = 0)).!
+  private val name =
+    P(quotedName | unquotedName)
+
+  private val signature = P {
+    ("[" ~/ name ~ "]").map(Signature.TypeParameter) |
+    ("(" ~/ name ~ ")").map(Signature.TermParameter) |
```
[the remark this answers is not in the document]
mberndt123: Ok

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Symbol.scala
mberndt123: fine

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Symbol.scala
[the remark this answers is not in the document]
allanrenucci: The documentation says:
> This is a very fast parser, ideal for quickly consuming large numbers of characters. Note that this builds a high-performance bit-packed lookup table, the size of the range of valid characters, up to 65k. That means that creating a parser like this has a one-time cost in terms of memory (up to 8k bytes) and time.

So I believe it is expected that the first iterations are slower, but then it should be much faster. Maybe you can try increasing the number of warmup and measurement  iterations, and see if you get more stable results.

bench/suite/src/main/scala/org/scalameta/bench/SymbolParser.scala
```diff
@@ -0,0 +1,29 @@
...
+      Signature.Term("Predef")),
+    Signature.Type("Int")).syntax
+}
+class SymbolParser {
+
+  @Benchmark
+  @BenchmarkMode(Array(Mode.Throughput))
+  @Fork(value = 1)
+  @Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+  def run: Unit = {
```
allanrenucci: The return type should be `Symbol`. [Otherwise it is dead code and might be optimised away](https://github.com/ktoso/sbt-jmh/blob/a6f5b8e8e077f77c52aab3f5c7c4a18581bb8ece/plugin/src/sbt-test/sbt-jmh/run/src/main/scala/org/openjdk/jmh/samples/JMHSample_08_DeadCode.scala#L52-L54).

mberndt123:
Alright, I've changed the `charsExcept` parser and the benchmark.

bench/suite/src/main/scala/org/scalameta/bench/SymbolParser.scala
```diff
@@ -0,0 +1,29 @@
+package org.scalameta.bench
+
+import java.util.concurrent.TimeUnit
+
+import org.openjdk.jmh.annotations.{Benchmark, BenchmarkMode, Measurement, Mode, Fork}
+import org.langmeta.semanticdb.Symbol
+import org.langmeta.semanticdb.Signature
+
+object SymbolParser {
+  val symbol: String = Symbol.Global(
```
olafurpg: You can use the parser to create a symbol from a string `Symbol("_root_.scala.Predef.Int#")`

bench/suite/src/main/scala/org/scalameta/bench/SymbolParser.scala
```diff
@@ -0,0 +1,29 @@
...
+    Signature.Type("Int")).syntax
+}
+class SymbolParser {
+
+  @Benchmark
+  @BenchmarkMode(Array(Mode.Throughput))
+  @Fork(value = 1)
+  @Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+  def run: Symbol = {
+    Symbol(SymbolParser.symbol)
```
olafurpg: Here is a random sample of symbols that we can benchmark https://gist.github.com/olafurpg/88b041fc027b45d451114284ba40afa5

mberndt123: Oops, I didn't really mean to push the commit with the Benchmark just yet 😆. Oh well, thanks for the representative sample of symbols, I'll make a benchmark based on that.

olafurpg:
After more thinking I am afraid I think we should reject this PR for now. The timing is quite unfortunate right now since there is a lot of ongoing to refine the symbol spec. For example #1323 makes conflicting changes to the parser.

The quality of this PR is great and I apologize for the comment in the code that invited a contribution to rewrite the parser to fastparse.

mberndt123:
@olafurpg, if you're making a lot of modifications to the symbol parser, doesn't that mean you should merge it? After all the whole point is to make such changes easier…

mberndt123:
I've rebased the change on top of master, it now parses the changed symbol format that was merged from #1323.
I have also changed the benchmark to run with the sample symbols that @olafurpg posted earlier. On my machine, the old parser is around 6x faster than the fastparse based one ☹️. So I guess you need to make a decision on whether the improved maintainability, robustness and error reporting is worth the degraded performance…

mberndt123:
Ok, so is there any chance this can move forward? I don't see any PRs right now that make conflicting changes to the parser, so to me it seems like now would be a good time… Or do you need more performance data or something?

olafurpg:
> On my machine, the old parser is around 6x faster than the fastparse based one 

I'm afraid that is an unacceptable slowdown. It's not only metadoc that uses the symbol parser on a hot path, scalafix also parses all symbols before running semantic rules. For large projects, this will be a noticable regression where scalafix is already slow enough (and I'm looking for opportunities to speed things up).

With the recent changes in semanticdb the symbol parser may no longer appear on the critical path for applications like metadoc and scalafix. Symbols are becoming more and more opaque string IDs so their parsed structure matters less. However, that will require refactorings in both projects that I can't give a good estimate on when will happen (it'll be sometime this year, maybe next couple months). Until then, I think it's best to close this PR. I'm sorry.

mberndt123:
Hey, I understand the rationale for rejecting this. No need to apologize :smile:

PR #1243, opened 2018-02-02
[WIP] First stab at crosscompilation to Scala Native

xeno-by:
Long story short, I have managed to successfully crosscompile Scalameta to Scala Native.

```
~$ time metap **/*.semanticdb
Library.scala
-------------
...
real	0m0.433s
user	0m0.435s
sys	0m0.055s

~$ time semanticdb/metap/.native/target/scala-2.11/metap-out **/*.semanticdb
Library.scala
-------------
...
real	0m0.015s
user	0m0.007s
sys	0m0.006s
```

However, there are some caveats. First, there's a blocker bug in Scala Native that breaks our public API (https://github.com/scala-native/scala-native/issues/1155) - I haven't found a way to work around it. Second, there's a bunch of minor issues - all workaroundable. See the discussion below for details.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,6 +3,7 @@ import scala.util.Try
 import scala.xml.{Node => XmlNode, NodeSeq => XmlNodeSeq, _}
 import scala.xml.transform.{RewriteRule, RuleTransformer}
 import org.scalajs.sbtplugin.ScalaJSCrossVersion
+import sbtcrossproject.{crossProject, CrossType}
 import org.scalameta.build._
 import org.scalameta.build.Versions._
 import org.scalameta.os
@@ -73,7 +74,7 @@ console := console.in(scalametaJVM, Compile).value
 
 /** ======================== SEMANTICDB ======================== **/
 
-lazy val semanticdb3 = crossProject
+lazy val semanticdb3 = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .crossType(CrossType.Pure)
   .in(file("semanticdb/semanticdb3"))
   .settings(
@@ -89,8 +90,10 @@ lazy val semanticdb3 = crossProject
   .jsSettings(
     crossScalaVersions := List(LatestScala211, LatestScala212)
   )
+  .nativeSettings(nativeSettings)
 lazy val semanticdb3JVM = semanticdb3.jvm
 lazy val semanticdb3JS = semanticdb3.js
+lazy val semanticdb3Native = semanticdb3.native
 
 lazy val semanticdbScalacCore = project
   .in(file("semanticdb/scalac/library"))
@@ -144,7 +147,7 @@ lazy val metac = project
   .disablePlugins(BackgroundRunPlugin)
   .dependsOn(semanticdbScalacPlugin)
 
-lazy val metap = crossProject
+lazy val metap = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .crossType(CrossType.Pure)
   .in(file("semanticdb/metap"))
   .settings(
@@ -153,15 +156,17 @@ lazy val metap = crossProject
     description := "SemanticDB decompiler",
     mainClass := Some("scala.meta.cli.Metap")
   )
+  .nativeSettings(nativeSettings)
   // NOTE: workaround for https://github.com/sbt/sbt-core-next/issues/8
   .disablePlugins(BackgroundRunPlugin)
   .dependsOn(semanticdb3)
 lazy val metapJVM = metap.jvm
 lazy val metapJS = metap.js
+lazy val metapNative = metap.native
 
 /** ======================== LANGMETA ======================== **/
 
-lazy val langmeta = crossProject
+lazy val langmeta = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .in(file("langmeta/langmeta"))
   .settings(
     publishableSettings,
@@ -173,13 +178,15 @@ lazy val langmeta = crossProject
   .jsSettings(
     crossScalaVersions := List(LatestScala211, LatestScala212)
   )
+  .nativeSettings(nativeSettings)
   .dependsOn(semanticdb3)
 lazy val langmetaJVM = langmeta.jvm
 lazy val langmetaJS = langmeta.js
+lazy val langmetaNative = langmeta.native
 
 /** ======================== SCALAMETA ======================== **/
 
-lazy val common = crossProject
+lazy val common = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .in(file("scalameta/common"))
   .settings(
     publishableSettings,
@@ -187,41 +194,49 @@ lazy val common = crossProject
     description := "Bag of private and public helpers used in scalameta APIs and implementations",
     enableMacros
   )
+  .nativeSettings(nativeSettings)
 lazy val commonJVM = common.jvm
 lazy val commonJS = common.js
+lazy val commonNative = common.native
 
-lazy val io = crossProject
+lazy val io = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .in(file("scalameta/io"))
   .settings(
     publishableSettings,
     description := "Scalameta APIs for input/output"
   )
+  .nativeSettings(nativeSettings)
   .dependsOn(langmeta, common)
 
 lazy val ioJVM = io.jvm
 lazy val ioJS = io.js
+lazy val ioNative = io.native
 
-lazy val dialects = crossProject
+lazy val dialects = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .in(file("scalameta/dialects"))
   .settings(
     publishableSettings,
     description := "Scalameta dialects",
     enableMacros
   )
+  .nativeSettings(nativeSettings)
   .dependsOn(common)
 lazy val dialectsJVM = dialects.jvm
 lazy val dialectsJS = dialects.js
+lazy val dialectsNative = dialects.native
 
-lazy val inputs = crossProject
+lazy val inputs = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .in(file("scalameta/inputs"))
   .settings(
     publishableSettings,
     description := "Scalameta APIs for source code",
     enableMacros
   )
+  .nativeSettings(nativeSettings)
   .dependsOn(langmeta, common, io)
 lazy val inputsJVM = inputs.jvm
 lazy val inputsJS = inputs.js
+lazy val inputsNative = inputs.native
 
 lazy val interactive = project
   .in(file("scalameta/interactive"))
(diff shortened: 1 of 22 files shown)


build.sbt
```diff
@@ -406,13 +439,29 @@ lazy val tests = crossProject
...
   .jvmConfigure(_.dependsOn(testkit, interactive, metac))
+  .jvmSettings(
+    // TODO: Workaround for what seems to be a bug in ScalaTest 3.2.0-SNAP10.
+    // Without adding scalacheck to library dependencies, we get the following error:
+    // > testsJVM/test
+    // [info] Compiling 79 Scala sources to /Users/eburmako/Projects/scalameta/tests/jvm/target/scala-2.12/test-classes...
+    // [trace] Stack trace suppressed: run last testsJVM/test:executeTests for the full output.
+    // [error] (testsJVM/test:executeTests) java.lang.NoClassDefFoundError: org/scalacheck/Test$TestCallback
+    // [error] Total time: 19 s, completed Feb 1, 2018 3:12:34 PM
+    libraryDependencies += "org.scalacheck" %% "scalacheck" % "1.13.5" % "test"
```
xeno-by: @bvenners This only happens in ScalaTest 3.2.0-SNAP10, but not in 3.2.0-SNAP6. Looks like a bug?

build.sbt
```diff
@@ -406,13 +439,29 @@ lazy val tests = crossProject
...
+  )
+  .nativeSettings(
+    nativeSettings,
+    // TODO: Is it alright to test Scala Native libraries under debug?
+    nativeMode := "debug"
```
xeno-by: @densh Is it alright to test Scala Native libraries under debug? If unit tests succeed under debug, does this guarantee that they will also succeed under release?

build.sbt
```diff
@@ -638,6 +687,20 @@ lazy val hasLargeIntegrationTests = Seq(
...
+  crossScalaVersions := List(LatestScala211),
+  nativeGC := "immix",
+  nativeMode := "release",
+  // TODO: I'd love to switch nativeLinkStubs to false, but I can't.
+  // No idea where the following nativeLink error is coming from.
+  // [error] cannot link: @java.lang.Thread::getStackTrace_scala.scalanative.runtime.ObjectArray
+  // [error] unable to link
+  // [error] (testsNative/nativetest:nativeLinkNIR) unable to link
+  // [error] Total time: 44 s, completed Feb 1, 2018 3:59:15 PM
+  nativeLinkStubs := false
```
xeno-by: @densh Are there best practices for debugging linkage errors? I can't seem to find out who uses Thread.getStackTrace. I'm reasonably sure that the direct caller is not in Scala Native core libs or in Scalameta, but I've no idea how to pinpoint the problematic 3rd-party library without bruteforcing through everything.

semanticdb/semanticdb3/semanticdb3.proto
```diff
@@ -44,7 +44,10 @@ message SymbolInformation {
     OBJECT = 10;
     PACKAGE = 11;
     PACKAGE_OBJECT = 12;
-    CLASS = 13;
+    // TODO: Can't say CLASS here, because of a Scala Native bug.
+    // Unlike other todos in this pull request, this one doesn't have a workaround.
+    // https://github.com/scala-native/scala-native/issues/1155.
+    CLAZZ = 13;
```
xeno-by: @densh This is a hard blocker, because this name is part of our API. Do you think you could help us out with this issue? If you don't have time for that, can you give some pointers so that I can fix the issue myself?

tests/jvm/src/test/scala/scala/meta/tests/io/FragmentSuite.scala
```diff
@@ -8,10 +8,14 @@ import org.scalatest.FunSuite
...
-    val expected = new File(".").getAbsoluteFile.toURI.normalize().toString
-    assert(obtained == expected)
+    // TODO: This test is broken cause of the recent changes to Fragment.
+    // Not sure whether this is a problem for us or not.
+    // [info] - Fragment.normalize *** FAILED ***
+    // [info]   "file:/[//]Users/eburmako/Proje..." did not equal "file:/[]Users/eburmako/Proje..." (FragmentSuite.scala:14)
+    // val obtained =
+    //   Fragment(AbsolutePath(new File(".", "a")), RelativePath(new File(".."))).uri.toString
+    // val expected = new File(".").getAbsoluteFile.toURI.normalize().toString
+    // assert(obtained == expected)
```
xeno-by: @olafurpg Do you think we should get to the bottom of this, or it's fine as is?

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
```diff
@@ -317,8 +317,11 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
   test("Lit.Double") {
     assert(templStat("1.4d").structure == """Lit.Double(1.4d)""")
     assert(templStat("1.40d").structure == """Lit.Double(1.40d)""")
-    assert(Lit.Double(1.40d).structure == "Lit.Double(1.4d)") // trailing 0 is lost
-    assert(Lit.Double(1.4d).structure == "Lit.Double(1.4d)")
+    // TODO: This fails under Scala Native:
+    // [info] - Lit.Double *** FAILED ***
+    // [info]   "Lit.Double(1.4[00000]d)" did not equal "Lit.Double(1.4[]d)" (SyntacticSuite.scala:321)
+    // assert(Lit.Double(1.40d).structure == "Lit.Double(1.4d)") // trailing 0 is lost
+    // assert(Lit.Double(1.4d).structure == "Lit.Double(1.4d)")
```
xeno-by: @olafurpg How much of a problem would it be for your tools if we allowed some slack in Lit.structure?

tests/shared/src/test/scala/scala/meta/tests/quasiquotes/ErrorSuite.scala
```diff
@@ -1,13 +1,24 @@
...
 import org.scalatest._
 import org.scalameta.tests._
 import typecheckError.Options.WithPositions
 import compat.Platform.EOL
 
+// TODO: Workaround for what seems to be a bug in ScalaTest 3.2.0-SNAP10.
+// I had to remove $ characters from all test names in this file.
+// This is because ScalaTest seems to erroneously consider dollars to be name terminators,
+// so it would spuriously crash with "duplicated test" exceptions for e.g.:
+// test("...$ in Pat.Extract") { ... } and test("...$ in Pat.ExtractInfix")  { .. }.
```
xeno-by: @bvenners Does this ring a bell? If it doesn't, I can try to minimize and submit a ticket.

build.sbt
shadaj: This is probably ScalaTest. When porting over to Scala Native I was seeing these linking errors and enabled `nativeLinkStubs` to get around them, but never got to actually finding where this method was used. I'll try to do some more investigation to find the source.

xeno-by: @shadaj Thank you for your work! I really appreciate that we now have a native-compatible build of ScalaTest 💯

shadaj: @xeno-by just pushed https://github.com/scalatest/scalatest/pull/1322, which should make it possible to run tests without linking stubs!

build.sbt
olafurpg: are we missing scalacheck on our classpath?

langmeta/langmeta/native/src/main/scala/org/langmeta/internal/io/PlatformFileIO.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  def isFile(path: AbsolutePath): Boolean =
+    Files.isRegularFile(path.toNIO)
+
+  def isDirectory(path: AbsolutePath): Boolean =
+    Files.isDirectory(path.toNIO)
+
+  def listAllFilesRecursively(root: AbsolutePath): ListFiles = {
+    // TODO: Some Java stream APIs aren't yet available in Scala Native,
+    // so I had to steal the Scala.js implementation from js/.
```
olafurpg: We should have shared source directories for js+native and jvm+native

scalameta/tokenizers/native/src/main/scala/scala/meta/internal/tokenizers/PlatformTokenizerCache.scala
```diff
@@ -0,0 +1,20 @@
...
+package internal
+package tokenizers
+
+import scala.collection.mutable
+import scala.meta.inputs._
+import scala.meta.tokens._
+
+import java.{util => ju}
+
+object PlatformTokenizerCache {
```
olafurpg: Ditto, we should not copy-paste between js/native since that will be hard to maintain

tests/jvm/src/test/scala/scala/meta/tests/io/FragmentSuite.scala
olafurpg: We rely on file URIs being printed as `file:///` in metals https://github.com/scalameta/metals/blob/325610ac10e1cfc37346fae61708b5f2d6003e02/metaserver/src/main/scala/scala/meta/languageserver/Uri.scala#L26 and we use Fragment to read zip files. We can of course work around this, but I think we should avoid a breaking change like.

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
olafurpg: I use structure a lot for testing purposes. I think it would be fantastic if we can achieve uniform syntax across all platforms. Also `1.400000d` output is quite ugly

build.sbt
densh: Compiler uses different pipeline for debug and release. Modulo compiler bugs, the semantics should be the same as long as you don't trigger undefined behavior.

build.sbt
xeno-by: We don't have a dependency on scalacheck, so yeah it's not on the classpath.

tests/jvm/src/test/scala/scala/meta/tests/io/FragmentSuite.scala
xeno-by: Do you think we should include this expectation in the SemanticDB spec?

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
xeno-by: I'll see what I can do.

langmeta/langmeta/native/src/main/scala/org/langmeta/internal/io/PlatformFileIO.scala
xeno-by: Do you know whether there is such functionality in sbt-crossproject?

semanticdb/semanticdb3/semanticdb3.proto
densh: https://github.com/scala-native/scala-native/pull/1156

langmeta/langmeta/native/src/main/scala/org/langmeta/internal/io/PlatformFileIO.scala
MasseGuillaume: nope.

MasseGuillaume: You can use: `CrossProject { def configurePlatform(platforms: Platform*)(f: Project => Project): CrossProject }`

xeno-by:
Very well! Thank you for the feedback, everyone. Let's wait for the Scala Native fix to land in a release, and then we'll reconvene 👋

xeno-by:
Superseded by https://github.com/scalameta/scalameta/pull/1263.

PR #1252, opened 2018-02-02
 Give local symbols reference equality.

olafurpg:
While upgrading to semanticdb v3.0 in scalafix we encountered that our
SemanticdbIndex was giving bogus results for local symbols. The bug was
not caught in our test suite, Guillaume discovered it while working on
SAM rewrite. This commit makes Symbol.Local use reference equality and
changes Database.load to use the same Symbol.Local instance for the same
id in the same document.

Applications that use alternative indexing techniques like SQL will need
to figure out another solution to distinguish between conflicting local
symbols between different files. One trick is to compute
definition/references on the fly from a Semanticdb for local symbols
because they cannot be referenced elsewhere.
1ad7790

### DIFF
--- a/langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Symbol.scala
+++ b/langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Symbol.scala
@@ -19,6 +19,10 @@ object Symbol {
     override def toString = syntax
     override def syntax = id
     override def structure = s"""Symbol.Local("$id")"""
+    // Local symbols have reference equality because two local symbols from
+    // different source files can have the same id (for example, local0).
+    override def equals(obj: scala.Any): Boolean = this.eq(obj.asInstanceOf[AnyRef])
+    override def hashCode(): Int = System.identityHashCode(this)
   }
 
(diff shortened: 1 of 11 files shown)


olafurpg:
Just noticed I'm not testing this fix properly, need to change the targeted test a bit.

olafurpg:
I've fixed the implementation and tests to properly validate that this commit fixes the problem

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Symbol.scala
```diff
@@ -19,6 +19,10 @@ object Symbol {
     override def toString = syntax
     override def syntax = id
     override def structure = s"""Symbol.Local("$id")"""
+    // Local symbols have reference equality because two local symbols from
+    // different source files can have the same id (for example, local0).
+    override def equals(obj: scala.Any): Boolean = this.eq(obj.asInstanceOf[AnyRef])
+    override def hashCode(): Int = System.identityHashCode(this)
```
xeno-by: How about we just revert to the previous local symbol format? It is regrettable that with the current format we have to push identity management to schema users.

olafurpg:
Reference equality is a bad idea. Let's not do this.

PR #1253, opened 2018-02-03
Explain uniqueness guarantees for symbols

xeno-by:
Supersedes https://github.com/scalameta/scalameta/pull/1252, fixes https://github.com/scalacenter/scalafix/issues/596.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -78,6 +78,7 @@ lazy val semanticdb3 = crossProject
   .in(file("semanticdb/semanticdb3"))
   .settings(
     publishableSettings,
+    ignoreMimaSettings,
     protobufSettings,
     PB.protoSources.in(Compile) := Seq(file("semanticdb/semanticdb3")),
     version := "3.1.0"
--- a/langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
+++ b/langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
@@ -62,6 +62,11 @@ package object semanticdb {
     def toDb(sourcepath: Option[Sourcepath], sdoc: s.TextDocument): d.Document = {
       val s.TextDocument(sschema, suri, stext, slanguage, ssymbols, soccurrences, sdiagnostics, ssynthetics) = sdoc
       assert(sschema == s.Schema.SEMANTICDB3, "s.TextDocument.schema must be ${s.Schema.SEMANTICDB3}")
+      def dsymbol(ssymbol: String) = {
+        def suffix = suri.replace("/", "_").replace(".", "_")
+        if (ssymbol.startsWith("local")) d.Symbol(ssymbol + "_" + suffix)
+        else d.Symbol(ssymbol)
+      }
       val dinput = {
         val sfilename = {
           assert(suri.nonEmpty, "s.TextDocument.uri must not be empty")
@@ -104,7 +109,8 @@ package object semanticdb {
       }
       object sSymbolInformation {
         def unapply(ssymbolInformation: s.SymbolInformation): Option[d.ResolvedSymbol] = ssymbolInformation match {
-          case s.SymbolInformation(d.Symbol(dsym), _, skind, sproperties, sname, _, ssignature, smembers, soverrides) =>
+          case s.SymbolInformation(ssym, _, skind, sproperties, sname, _, ssignature, smembers, soverrides) =>
+            val dsym = dsymbol(ssym)
             val dflags = {
               var dflags = 0L
               def dflip(dbit: Long) = dflags ^= dbit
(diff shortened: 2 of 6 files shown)


build.sbt
```diff
@@ -78,9 +78,10 @@ lazy val semanticdb3 = crossProject
   .in(file("semanticdb/semanticdb3"))
   .settings(
     publishableSettings,
+    ignoreMimaSettings,
```
xeno-by: @thesamet Are there best practices for achieving source/binary compatibility of ScalaPB-generated code during compatible Protobuf schema evolution?

For example, this pull request removes the `range` field and adds the `location` field to the `SymbolInformation` message. As far as I understand, this change is both backward and forward compatible in the sense that old Protobuf payloads are valid under the new schema and new Protobuf payloads are valid under the old schema.

However, the changes to the `SymbolInformation` Scala class generated by ScalaPB are not compatible (so I had to disable [MiMa](https://github.com/lightbend/migration-manager) for this project), which is regrettable. Can you provide recommendations for dealing with this?

thesamet: AFAIK there is no way around the binary compatibility issue as long as ScalaPB generates case classes. It has been suggested that we switch to normal classes, but there wasn't much of a demand for that.

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -181,12 +221,13 @@ always equal to an asterisk (`*`).
...
 ```protobuf
 message SymbolInformation {
+  reserved 6;
   string symbol = 1;
   string language = 2;
   Kind kind = 3;
   int32 properties = 4;
   string name = 5;
-  Range range = 6;
+  Location location = 10;
```
xeno-by: Instead of deprecating `range`, I removed it. This is an exceptional case, but that field never worked properly, and it was introduced just a couple days ago, so I think that it's alright to outright remove it.

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
```diff
@@ -62,6 +62,11 @@ package object semanticdb {
     def toDb(sourcepath: Option[Sourcepath], sdoc: s.TextDocument): d.Document = {
       val s.TextDocument(sschema, suri, stext, slanguage, ssymbols, soccurrences, sdiagnostics, ssynthetics) = sdoc
       assert(sschema == s.Schema.SEMANTICDB3, "s.TextDocument.schema must be ${s.Schema.SEMANTICDB3}")
+      def dsymbol(ssymbol: String) = {
+        def suffix = suri.replace("/", "_").replace(".", "_")
```
MasseGuillaume: suri is always a file?

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -165,6 +195,16 @@ For example, `x` in `def identity[T](x: T): T` may be modelled by local symbols
 the underlying document. The same logic applies to the type parameter `T`,
 which is also a local definition.
 
+Local symbols must be unique within the underlying document, but they don't have
+to be unique across multiple documents. For example, at the time of writing
+`semanticdb-scalac` produces local symbols named `local0`, `local1`, etc,
+with the counter resetting to zero for every new document.
```
MasseGuillaume: Above you suffix with the file.

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
MasseGuillaume: What if you have a cross-project?

build.sbt
xeno-by: Thank you! I think we'll just bite the bullet for now, since this is an internal API anyway.

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/internal/package.scala
xeno-by: `suri` is a relative path (original source's path minus the original sourceroot).

In case of multiple projects sharing the same files, you end up with SemanticDB payloads that have the same uris. This means that local symbols aren't guaranteed to be unique within universes that mix such projects.

However, this problem was also present in SemanticDB v2. This pull requests doesn't aim to fix this problem though - just to achieve feature parity between v2 and v3.

semanticdb/semanticdb3/semanticdb3.md
xeno-by: The specification describes the contents of protobuf payloads (the so-called "low-level semantic API"). High-level APIs are free to apply custom transformations to these payloads.

semanticdb/semanticdb3/semanticdb3.md
ShaneDelmore: 👍

PR #1254, opened 2018-02-03
Specify the order of underlying symbols in multi symbols

xeno-by:


xeno-by:
I think that in the long run it would make sense to split multi symbols into multiple occurrences. However, implementing the new scheme and migrating existing tools to support that scheme will probably take a while.

Since this improvement is not on our immediate critical path, I decided to simply codify the state of the art, and then make it a bit more tractable.

PR #1255, opened 2018-02-03
Synchronize the spec and the implementation of global symbol format

xeno-by:
Inspired by https://github.com/scalameta/scalameta/pull/1241. /cc @mberndt123


xeno-by:
The versioning issue suggests that we may have a separate repository trying to get out. 

I think that everything under `semanticdb/` can go into https://github.com/scalameta/semanticdb to be versioned and published separately from https://github.com/scalameta/scalameta.

However, infrastructure setup will take lots of time, so since this is not on our critical path, let's just keep an eye on the situation and do nothing for the time being.

xeno-by:
This pull request only updates the spec, so I'm going to merge it without waiting for Travis to finish.

PR #1260, opened 2018-02-05
Make local symbol suffix more robust.

olafurpg:
- replace all non-alphanumeric characters with underscore handle
  filenames like `foo-bar.scala`.
- strip the suffix when converting from high-level API to low-level
  schema to make toDb/toSchema bijective.


olafurpg:
Good catch @MasseGuillaume I added a new expect test that tests both fixes in this PR.

olafurpg:
We never tested before for the property that

- save -> load == save -> load -> save -> load

PR #1267, opened 2018-02-06
Update travis script to only run native install on native entry

olafurpg:
To speed up non-native entries, native installation is kind of slow.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,35 +3,44 @@ git:
   depth: 9999
 install:
   - git fetch --tags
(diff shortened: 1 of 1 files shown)


olafurpg:
I copied the layout from the  fastparse build, this setup allows very fine grained control of all steps for each entry. Haven't looked at logs yet to validate if everything is still working as expected https://travis-ci.org/scalameta/scalameta/builds/338113625?utm_source=github_status&utm_medium=notification

PR #1268, opened 2018-02-07
Codify compatibility and versioning policies

xeno-by:


COMPATIBILITY.md
```diff
@@ -0,0 +1,39 @@
+## Public API
+
+"Public API" means all public definitions in the artifacts under the
+`org.scalameta` Maven groupId whose fully-qualified name does not include
+`internal` and `contrib`, e.g. `scala.meta.Tree` or `org.langmeta.Position`:
+  * `x.y.z` must be backward compatible with `x.0.0`:
```
olafurpg: I think we should make this claim stronger and guarantee compatibility against x.y1.z1 for all `y1 < y` and `z1 < z`. This will add a tax to our mima checks, maybe we can make it more digestible somehow. wdyt?

COMPATIBILITY.md
```diff
@@ -0,0 +1,39 @@
+## Public API
+
+"Public API" means all public definitions in the artifacts under the
+`org.scalameta` Maven groupId whose fully-qualified name does not include
+`internal` and `contrib`, e.g. `scala.meta.Tree` or `org.langmeta.Position`:
```
olafurpg: missing `org.scalameta` package, which is mentioned as private api below.

xeno-by: Not sure what you mean by "below".

COMPATIBILITY.md
xeno-by: See updates.

xeno-by:
The AppVeyor failure is spurious.

build.sbt
```diff
@@ -559,12 +553,39 @@ lazy val publishableSettings = Def.settings(
...
+      // TODO: Figure out whether there is a more satisfying solution.
+      // Here's what I'd like to do, but I can't because of deprecations:
+      //   val isJVM = crossPlatform.value == JVMPlatform
+      // Here's my second best guess, but it doesn't work due to some reason:
+      //   val isJVM = platformDepsCrossVersion.value == CrossVersion.binary
+      val isJVM = {
+        val isJS = platformDepsCrossVersion.value == ScalaJSCrossVersion.binary
+        val isNative = platformDepsCrossVersion.value == ScalaNativeCrossVersion.binary
+        !isJS && !isNative
+      }
```
xeno-by: @sjrd Is there a better way to do this?

sjrd: > Here's my second best guess, but it doesn't work due to some reason

That's because `CrossVersion.binary` is a `def` which returns a new instance of `CrossVersion` every time, and in sbt 0.13.x, `==` is useless on `CrossVersion`s (it's equivalent to `eq`).

There is no elegant way to test the platform of a project in the most general case. The closest you can get is testing for the presence (or absence) of `scalajs-library` or `nativelib` in the `libraryDependencies` of the project.

Alternatively, in your own build, you could define your own very simple `lazy val platform = settingKey[String]` and configure it to your liking.

You could also test whether `thisProject.value.id` (or whatever it's called) ends with `"JVM"`, `"JS"` or `Native`.

And I'm sure you can invent other creative ways to do it.

xeno-by: Thanks for the swift response!

Issue #819, opened 2017-04-29
Bug in interpolator with escape quotes


olafurpg:
```scala
@ """
  s"a\"${escape(v)}\""
  """.parse[Term].get.tokens.map(x => x.structure -> x.getClass)
res8: collection.immutable.IndexedSeq[(String, Class[?0]) forSome { type ?0 <: Token }] = Vector(
  (BOF [0..0),class scala.meta.tokens.Token$BOF),
  (\n [0..1),class scala.meta.tokens.Token$LF),
  (s [1..2),class scala.meta.tokens.Token$Interpolation$Id),
  (" [2..3),class scala.meta.tokens.Token$Interpolation$Start),
  (a\ [3..5),class scala.meta.tokens.Token$Interpolation$Part),
  (" [5..6),class scala.meta.tokens.Token$Interpolation$End),
  ($ [6..7),class scala.meta.tokens.Token$Ident),
  ({ [7..8),class scala.meta.tokens.Token$LeftBrace),
  (escape [8..14),class scala.meta.tokens.Token$Ident),
  (( [14..15),class scala.meta.tokens.Token$LeftParen),
  (v [15..16),class scala.meta.tokens.Token$Ident),
  () [16..17),class scala.meta.tokens.Token$RightParen),
  (} [17..18),class scala.meta.tokens.Token$RightBrace),
  (\ [18..19),class scala.meta.tokens.Token$Ident),
  ("" [19..21),class scala.meta.tokens.Token$Constant$String), // expected SpliceEnd
  (\n [21..22),class scala.meta.tokens.Token$LF),
  (EOF [22..22),class scala.meta.tokens.Token$EOF)
)

 // expected something similar to 
@ """
  s"aa${escape(v)}"
  """.parse[Term].get.tokens.map(x => x.structure -> x.getClass)
res7: collection.immutable.IndexedSeq[(String, Class[?0]) forSome { type ?0 <: Token }] = Vector(
  (BOF [0..0),class scala.meta.tokens.Token$BOF),
  (\n [0..1),class scala.meta.tokens.Token$LF),
  (s [1..2),class scala.meta.tokens.Token$Interpolation$Id),
  (" [2..3),class scala.meta.tokens.Token$Interpolation$Start),
  (aa [3..5),class scala.meta.tokens.Token$Interpolation$Part),
  ($ [5..6),class scala.meta.tokens.Token$Interpolation$SpliceStart),
[...]

PR #1270, opened 2018-02-07
Fix vanish unicode escape in string interpolation

avdv:
This fixes a problem where unicode escape sequences are removed from interpolated string literals.

See #819. 

Where should I add a test case?


avdv:
> Thank you for this contribution @avdv All the tests still pass, which is a good sign 😄 Can you please sign the Scala CLA? See link on travis.

Sure, I already did in the meantime.
> I took a stab at testing this fix in scalameta/scalameta/compare/master...olafurpg:fix-unicode-escape-vanish?expand=1 but couldn't get it to work. I think we need to write the input in an external file in order to preserve the unicode escapes.

Oh, yes. That's a problem.
> Can you try to reproduce the bug?

I'll look into it.

avdv:
Hi @olafurpg,

your test case seems to work just fine for me: 🎉 
```
[info] - s"${x}\uef17" *** FAILED ***
[info]   ===========
[info]   => Obtained
[info]   ===========
[info]   s"${x}7"
[info]   
[info]   =======
[info]   => Diff
[info]   =======
[info]   -s"${x}7"
[info]   +s"${x}\uef17" (BaseTokenizerSuite.scala:22)
[info] - s"]2;SBT${titleName}${titleBranch}\u0007" *** FAILED ***
[info]   ===========
[info]   => Obtained
[info]   ===========
[info]   s"]2;SBT${titleName}${titleBranch}7"
[info]   
[info]   =======
[info]   => Diff
[info]   =======
[info]   -s"]2;SBT${titleName}${titleBranch}7"
[info]   +s"]2;SBT${titleName}${titleBranch}\u0007" (BaseTokenizerSuite.scala:22)
```
(after commenting my additional lines)

olafurpg:
Whoop whoop, CI is happy 🎉

avdv:
Thanks, @olafurpg for actually doing the heavy lifting here... !! 👍

olafurpg:
Thank you for digging into LegacyScanner! It's a place I avoid as I can, testing is easy 😄

PR #1277, opened 2018-02-08
Handle accessor method for repeated type.

### LINKED ISSUE #1273: MatchError with semanticdb-scalac
While compiling ammonite with 3.2
```
[warn] failed to generate semanticdb for /Users/ollie/dev/Ammonite/ops/src/main/scala/ammonite/ops/Extensions.scala:
[warn] scala.MatchError: => String* (of class scala.reflect.internal.Types$NullaryMethodType)
[warn] 	at scala.meta.internal.semanticdb.scalac.TypeOps$RepeatedType$.unapply(TypeOps.scala:171)
```

olafurpg:
Fixes #1273.


semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TypeOps.scala
```diff
@@ -168,7 +168,11 @@ trait TypeOps { self: DatabaseOps =>
...
     def unapply(gtpe: g.Type): Option[g.Type] =
       if (g.definitions.isRepeatedParamType(gtpe)) {
-        gtpe match { case g.TypeRef(_, _, gtpe :: Nil) => Some(gtpe) }
+        gtpe match {
+          case g.TypeRef(_, _, gtpe :: Nil) => Some(gtpe)
+          // accessor for repeated type
+          case g.NullaryMethodType(ref) => unapply(ref)
+        }
       } else None
   }
```
xeno-by: This collapses the type structure of repeated accessors, making things inconsistent with other accessors. Instead, I propose that we give up on `isRepeatedParamType` altogether, since it seems to have been written for a different use case, and implement `RepeatedType` as follows:

```
  object RepeatedType {
    def unapply(gtpe: g.Type): Option[g.Type] =
      tpe match {
        case g.TypeRef(_, g.definitions.RepeatedParamClass, List(garg)) => Some(garg)
        case g.TypeRef(_, g.definitions.JavaRepeatedParamClass, List(garg)) => Some(garg)
        case _ => None
      }
  }
```

Same for `ByNameType`.

Whether accessors should or should not be typed as `NullaryMethodType` is a good question, but it's separate from this pr / associated ticket.

olafurpg: Done.

PR #1278, opened 2018-02-08
Add ENUM literal type.

### LINKED ISSUE #1274: Unsupported const RuntimeException with semanticdb-scalac
While compiling ammonite with 3.2

```
[warn] failed to generate semanticdb for /Users/ollie/dev/Ammonite/ops/src/main/scala/ammonite/ops/Model.scala:
[warn] java.lang.RuntimeException: unsupported const Constant(value NOFOLLOW_LINKS): Constant(TermName("NOFOLLOW_LINKS"))
[warn] 	at scala.sys.package$.error(package.scala:27)
[warn] 	at scala.meta.internal.semanticdb.scalac.TypeOps$XtensionGTypeSType.loop$1(TypeOps.scala:62)
```

olafurpg:
Fixes #1274. Java enum values have types with `g.TermSymbol` constant
values. This commit adds a new kind of literal types for java enums that
are represented as the field symbol.  This commit also adds expect test
for literal types.


tests/jvm/src/test/resources/lowlevel.expect
```diff
@@ -588,6 +594,27 @@ _root_.types.Test.C#x. => val x: : C.this.p.X
   C => _root_.types.Test.C#
   p => _root_.types.Test.C#p.
   X => _root_.types.P#X#
+_root_.types.Test.Literal. => final object Literal
+_root_.types.Test.Literal.bool. => final val bool: : true
```
olafurpg: @xeno-by do you know why there are two colons `: :`? I expected there to be only one colon, the type is
```scala
tag: METHOD_TYPE
methodType {
  result {
    tag: LITERAL_TYPE
    literalType {
      tag: ENUM
      primitive: 0
      string: "_root_.java.nio.file.LinkOption.NOFOLLOW_LINKS."
    }
  }
}
```

olafurpg:
`TermSymbol.isJavaEnum` doesn't seem to exist on 2.11, what do we do?

tests/jvm/src/test/resources/lowlevel.expect
xeno-by: `: T` corresponds to `NullaryMethodType`, which is the signature of the underlying getter.

xeno-by:
@olafurpg `hasJavaEnum` is just a tiny helper. Use `hasFlag(JAVA_ENUM)` directly, which is available on both platforms.

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -354,6 +354,7 @@ message LiteralType {
     DOUBLE = 9;
     STRING = 10;
     NULL = 11;
+    ENUM = 12;
```
xeno-by: Let's reflect on whether we should copy/paste what Scalac does here. 

To me, [enum constants](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.9.1), e.g. `NOFOLLOW_LINKS` look pretty similar to objects in that they are singletons and therefore should be modelled with `SingleType`.

Now, why didn't Scalac go the same route? I wasn't around when this functionality was introduced. However, I have an educated guess. 

I think it's an unfortunate side-effect of the split between `StaticAnnotation` and `ClassfileAnnotation`. The latter only supports `javaArgs` (/cc @MasseGuillaume) that only accept `Constant`, nested annotations with `javaArgs` and arrays thereof. Now, because the JLS allows enum constants in annotations and because Scala must interoperate with Java, enum constants were stuffed into `Constant` for the simplest possible implementation. This decision to prioritize implementation simplicity looks consistent with a number of other hacks in other compiler modules that were introduced to support annotations.

Clearly we don't have the same constraints as Scalac developers had, which is why I propose we do the following, which seems to be the cleanest conceptually:
  * Add `SymbolInformation.ENUM` and `SymbolInformation.ENUM_CONSTANT`.
  * Model `LiteralType(Constant(sym: Symbol))` with `SingleType(sym.owner.thisPrefix, sym)`.
  * Update the specification for `LiteralType` and `SingleType`.

The test case from below will become:
```
_root_.java.nio.file.LinkOption. => final enum LinkOption
_root_.java.nio.file.LinkOption.NOFOLLOW_LINKS. => final enum constant NOFOLLOW_LINKS: NOFOLLOW_LINKS.type
  NOFOLLOW_LINKS => _root_.java.nio.file.LinkOption.NOFOLLOW_LINKS.
```

olafurpg:
Thanks for the review @xeno-by I did not manage to get the test working as you expected, but I removed the ENUM constant type and model them as SingleType instead.

I propose we merge this as is and fix the ENUM/ENUM_CONSTANT properties in a followup PR. Why?
- adding new properties is tricky because of Denotation => s.SymbolInformation conversion, we should bite the bullet and refactor semanticdb-scalac to go straight to s.TextDocument.
- this PR is already fixing a critical bug blocking scalafix from migrating to the new low-level schema

If you agree I can open a ticket documenting the status.

olafurpg:
I had imagined that we would refactor semanticdb-scalac to go directly to s.TextDocument and leave out new features from the old org.langmeta._ api. However, with the current situation we are forced to add new features to org.langmeta._ and scala.meta.semanticdb3._, which adds a lot of complexity.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TypeOps.scala
```diff
@@ -59,7 +62,8 @@ trait TypeOps { self: DatabaseOps =>
               case g.Constant(x: Double) => s.LiteralType(l.DOUBLE, doubleBits(x), "")
               case g.Constant(x: String) => s.LiteralType(l.STRING, 0, x)
               case g.Constant(null) => s.LiteralType(l.NULL, 0, "")
-              case gother => sys.error(s"unsupported const ${gother}: ${g.showRaw(gother)}")
+              case gother =>
+                sys.error(s"unsupported const ${gother}: ${g.showRaw(gother)}")
```
xeno-by: Why this formatting change? Our `.scalafmt.conf` says `maxColumn = 100`.

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Flags.scala
```diff
@@ -28,6 +28,8 @@ private[semanticdb] trait Flags {
   final val CONTRAVARIANT: Long = 1 << 23
   final val INLINE: Long = 1 << 24
   final val JAVADEFINED: Long = 1 << 25
+  final val ENUM: Long = 1 << 26
+  final val ENUM_CONSTANT: Long = 1 << 26
```
xeno-by: Is it intended that these constants map to the same value?

xeno-by: If `ENUM_CONSTANT` is unused for now, let's just remove it.

semanticdb/semanticdb3/semanticdb3.proto
```diff
@@ -175,6 +175,8 @@ message SymbolInformation {
     PACKAGE_OBJECT = 12;
     CLASS = 13;
     TRAIT = 14;
+    ENUM = 15;
+    ENUM_CONSTANT = 16;
```
xeno-by: Can you update the specification, too?

xeno-by:
@olafurpg Okay, let's defer the `ENUM`/`ENUM_CONSTANT` split to a dedicated ticket. I left some follow-up comments.

olafurpg:
I've rebased on top of master and addressed your feedback. I hit on https://github.com/scalameta/scalameta/issues/1283 and "fixed" it with a try/catch.

PR #1281, opened 2018-02-08
Add Denotation.annotations

MasseGuillaume:
This only adds the AnnotationInfo.symbol, parameters are not saved.
For example, given

```scala
class A[@specialized(Int, Long) T](v: T)
//                              ^
```

The T under the carret will have the symbol `_root_.scala.specialized` in it's annotations. The parameters `Int` and `Long` are lost. Annotations parameter are available, see https://github.com/scala/scala/blob/2.12.x/src/reflect/scala/reflect/internal/AnnotationInfos.scala. The difficulty is to encode them, since they are typed as Tree.

One idea would be to special case some of the annotation with parameters under the `scala.annotation` package and the `scala` package. Here is a list of annotations with non-empty parameter list:

```scala
class SerialVersionUID(value: Long)
class elidable(level: Int)
class implicitAmbiguous(msg: String)
class implicitNotFound(msg: String)
class languageFeature(feature: String, enableRequired: Boolean)
class showAsInfix(enabled: Boolean)
class deprecated(message: String, since: String)
class deprecatedInheritance(message: String, since: String)
class deprecatedName(name: Symbol, since: String)
class deprecatedOverriding(message: String, since: String)
class throws(clazz: Class[T])
class throws[T <: Throwable](cause: String)
class specialized(types: Specializable*)
class specialized(group: SpecializedGroup)
```

simple types like `Long`, `Int`, `String` and `Boolean` can be encoded easily.

more complicated types encoding such as `Class[T]`, `[T <: Throwable]`, `Specializable` and `SpecializedGroup` can be hardcoded.

There are also arguments for java annotations!

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/DenotationOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/DenotationOps.scala
@@ -122,24 +122,29 @@ trait DenotationOps { self: DatabaseOps =>
         if (saveOverrides && config.denotations.saveReferences) gsym.overrides
         else Nil
       }
-      config.signatures match {
-        case SignatureMode.None =>
-          val denot = m.Denotation(flags, name, "", Nil, Nil, over, None)
-          DenotationResult(denot, todoOverrides, Nil)
-        case SignatureMode.Old =>
-          val (signature, names) = oldInfo
-          val denot = m.Denotation(flags, name, signature, names, Nil, over, None)
-          DenotationResult(denot, todoOverrides, Nil)
-        case SignatureMode.New =>
-          val (tpe, todoTpe) = newInfo
-          val denot = m.Denotation(flags, name, "", Nil, Nil, over, tpe)
-          DenotationResult(denot, todoOverrides, todoTpe)
-        case SignatureMode.All =>
-          val (signature, names) = oldInfo
-          val (tpe, todoTpe) = newInfo
-          val denot = m.Denotation(flags, name, signature, names, Nil, over, tpe)
-          DenotationResult(denot, todoOverrides, todoTpe)
-      }
+      val todoAnnotations = gsym.annotations.map(_.symbol)
+      val annot = gsym.annotations.map(_.symbol.toSemantic)
+      val (denot, todoTpe) =
+        config.signatures match {
+          case SignatureMode.None =>
+            val denot = m.Denotation(flags, name, "", Nil, Nil, over, None, annot)
+            (denot, Nil)
+          case SignatureMode.Old =>
+            val (signature, names) = oldInfo
+            val denot = m.Denotation(flags, name, signature, names, Nil, over, None, annot)
+            (denot, Nil)
+          case SignatureMode.New =>
+            val (tpe, todoTpe) = newInfo
+            val denot = m.Denotation(flags, name, "", Nil, Nil, over, tpe, annot)
+            (denot, todoTpe)
+          case SignatureMode.All =>
+            val (signature, names) = oldInfo
+            val (tpe, todoTpe) = newInfo
+            val denot = m.Denotation(flags, name, signature, names, Nil, over, tpe, annot)
+            (denot, todoTpe)
+        }
+
+      DenotationResult(denot, todoOverrides, todoTpe, todoAnnotations)
     }
   }
 
@@ -147,5 +152,6 @@ trait DenotationOps { self: DatabaseOps =>
   case class DenotationResult(
       denot: m.Denotation,
       todoOverrides: List[g.Symbol],
-      todoTpe: List[g.Symbol])
+      todoTpe: List[g.Symbol],
+      todoAnnotations: List[g.Symbol])
 }
--- a/semanticdb/semanticdb3/semanticdb3.proto
+++ b/semanticdb/semanticdb3/semanticdb3.proto
@@ -202,6 +202,7 @@ message SymbolInformation {
   repeated string members = 8;
   repeated string overrides = 9;
   Type tpe = 11;
+  repeated string annotations = 12;
 }
 
 message SymbolOccurrence {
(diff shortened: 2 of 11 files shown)


semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/DenotationOps.scala
```diff
@@ -122,30 +122,36 @@ trait DenotationOps { self: DatabaseOps =>
...
-        case SignatureMode.All =>
-          val (signature, names) = oldInfo
-          val (tpe, todoTpe) = newInfo
-          val denot = m.Denotation(flags, name, signature, names, Nil, over, tpe)
-          DenotationResult(denot, todoOverrides, todoTpe)
-      }
+      val todoAnnotations = gsym.annotations.map(_.symbol)
+      val annot = gsym.annotations.map(_.symbol.toSemantic)
+      val (denot, todoTpe) =
+        config.signatures match {
```
olafurpg: Please try to minimize the diffs by moving `config.signatures` right after `=`. Scalafmt preserves the line break if you insert it

semanticdb/semanticdb3/semanticdb3.proto
```diff
@@ -202,6 +202,7 @@ message SymbolInformation {
   repeated string members = 8;
   repeated string overrides = 9;
   Type tpe = 11;
+  repeated string annotations = 12;
```
olafurpg: I wonder if we should prepare for handling annotation parameters in the future.

```proto
message Annotation {
  message ParameterList { string parameter  = 1; }
  string symbol = 1;
  repeated ParameterList args = 2;
}
```
We already store synthetics, which are trees, as syntax.

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/AnnotationsSuite.scala
```diff
@@ -0,0 +1,76 @@
...
+      |  // Variable annotation
+      |  @transient @volatile var m: Int = 0
+      |
+      |  // Expression annotation
+      |  def f(x: Option[String]) = (x: @unchecked) match { case Some(y) => y }
+      |}
+    """.stripMargin,
+    """
+      |_root_.scala.volatile#{
+      |  _root_.scala.annotation.meta.field#
```
olafurpg: I feel like a `@` would help readability, wdyt?

MasseGuillaume: this is for testing. There is a `@` in `Denotation.toString`.

semanticdb/semanticdb3/semanticdb3.proto
MasseGuillaume: There is also `javaArgs`.

MasseGuillaume:
>  I agree it would be really nice to have annotation arguments. Maybe we should go all the way and handle it right now

@xeno-by What do you think?

semanticdb/semanticdb3/semanticdb3.proto
xeno-by: Here's what I would propose towards that end:

```
message Annotation {
  Type tpe = 1;
}

message AnnotatedType {
  reserved 2;
  Type tpe = 1;
  repeated Annotation annotations = 3;
}
```

Having a dedicated structure will allow us to freely add arguments when/if we are ready to support that. I should've thought about that when adding `AnnotatedType`.

The distinction between `javaArgs` and `scalaArgs` is an extremely ugly implementation detail of Scalac. I don't think there's a need to bring it to SemanticDB.

xeno-by:
@MasseGuillaume See my inline comment above.

semanticdb/semanticdb3/semanticdb3.proto
MasseGuillaume: what about the parameter name?

xeno-by:
@olafurpg I don't think that annotation arguments are feasible at this point. That would require storing AST payloads, which would explode the schema (I'd say that'd be 300-500 extra lines) and invalidate the specification (and I'm not sure whether it will recover from that). I'm really afraid that the benefit of modelling annotation arguments will not be worth the cost.

olafurpg:
@xeno-by why not do the same trick as with Synthetic? Store the arguments as unspecified syntax. Alternatively, model arguments as literal types as long as they're literals (most common case).

xeno-by:
@olafurpg I suggest that we don't add more implementation-specific features to the spec. This year, we made a huge push towards making SemanticDB portable (in fact, only synthetics are unspecified at this point), and I think it would be great to keep it this way.

If there's a very strong use case, we can consider doing the same as Synthetic, but even then I would probably try to come up with something limited but portable (like literals).

olafurpg:
Fair enough. Something supporting only literals would go a long way however. `@deprecated` requires literal strings for example.

MasseGuillaume:
I think the first PR was better in term of semantic. Annotations are only a vehicle to add information to the Denotation. We should encode deprecation & etc the same way we encode private & al. Those annotations are wired with the compiler.

xeno-by:
@MasseGuillaume Let's stick to the distinction between modifiers and annotations as established in the language specification and manifested in the language syntax and the underlying platform. 

I am also unsatisfied with this distinction. For example, I would prefer if `private` and friends were annotations in `scala.annotation`. However, I think that for Scala that ship has long sailed. In the early days of Scalameta, we tried to force unification, but that didn't end well.

@olafurpg Do we have an immediate use case for deprecation messages or other annotation arguments? If yes, let's discuss the argument issue further. Otherwise, I think we may be overdesigning.

semanticdb/semanticdb3/semanticdb3.proto
xeno-by: Parameter names can be obtained by analyzing `tpe` and its constructor signature(s).

MasseGuillaume:
@xeno-by 

`class deprecatedName(name: Symbol, since: String)` would be useful for scalafix

see https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/websocket/Websocket.scala#L7 for example.

MasseGuillaume:
@xeno-by I will hand off this to you. My focus was mainly for the SAM rewrite and it ended up being a bit more complex than what I thought. https://github.com/scalameta/scalameta/issues/1282

xeno-by:
Superseded by https://github.com/scalameta/scalameta/issues/1308.

PR #1301, opened 2018-02-10
Fully automate enforcement of the compatibility policy

xeno-by:
We should strive for the compatibility policy to be enforced
completely automatically - without easy-to-reach blanket overrides
like ignoreMimaSettings.

### DIFF
--- a/project/Mima.scala
+++ b/project/Mima.scala
@@ -6,13 +6,14 @@ import com.typesafe.tools.mima.core._
 // More details about Mima:
 // https://github.com/typesafehub/migration-manager/wiki/sbt-plugin#basic-usage
 object Mima {
-  val ignoredABIProblems: Seq[ProblemFilter] = {
-    Seq(
-      ProblemFilters.exclude[Problem]("org.langmeta.internal.*"),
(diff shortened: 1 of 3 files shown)


project/Mima.scala
```diff
@@ -6,13 +6,14 @@ import com.typesafe.tools.mima.core._
 // More details about Mima:
 // https://github.com/typesafehub/migration-manager/wiki/sbt-plugin#basic-usage
 object Mima {
-  val ignoredABIProblems: Seq[ProblemFilter] = {
-    Seq(
```
olafurpg: How do we manually exclude/ignore individual problems? I think we should still keep this `Seq` so we can copy-paste the mima errors when we intentionally break the API. Example when we need to manually exclude errors, package private members are breaking changes on the bytecode level but according to the language contract it's not a breaking change.

xeno-by: Let's do this on by-need basis. Now that we have a compatibility policy, adding these ignores won't be trivial anyway, since we'll likely need to update the policy, too.

olafurpg: changes to package private members are within our compatibility policy and they will require custom ProblemFilters, but sure we can deal with it then

Issue #1309, opened 2018-02-14
Add support for private/protected to this and within


xeno-by:

xeno-by:
Fixed in https://github.com/scalameta/scalameta/pull/1316.

Issue #1308, opened 2018-02-14
Add SymbolInformation.annotations


xeno-by:

xeno-by:
Fixed in https://github.com/scalameta/scalameta/pull/1316.

Issue #1295, opened 2018-02-09
Rethink method symbols


xeno-by:
I've run into a hard blocker implementing metacp with scalap. 

From within scalap, there is no way to compute an erasure of a type, which means that we can't generate specification-compliant method symbols (https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md#symbol).

This is another situation that highlights that current method symbols are problematic. During the SemanticDB v3 design period, we had a discussion about  using a different symbol format, but we didn't have time to follow through. Looks like now we have to do that.

olafurpg:
I am both excited about improving the format and scared of changing it.

Just brainstorming, would it be possible to use the dealiased symbol for all arguments?

```
def foo(a: Int) = a
// foo(scala.Int#)
def foo(a: String) = foo(a.length)
// foo(java.lang.String#)
def foo(a: Either[String, Int]) = foo(a.length)
// foo(scala.util.Either#)
```

I had problems reproducing the symbol method format in mtags for metals because there we only have syntax trees. The information have syntactically from the source is

* position of definition
* parameter names
* syntax for parameter type

With that information we could do a best-effort for compute the actual symbols, it would at least likely be easier than matching erased signature.

xeno-by:
Let's summarize our current experience with using unique identifiers for modeling method symbols (well, not unique unique, but unique-within-universe unique - see [the spec](https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md#symbol) for details):
* Symbol generation in producers
    * Current approach has proven to be really bad in this aspect, since it requires computing erasures, which requires lots of semantic APIs (dealiasing, parents + members - for value classes, maybe more). These semantic APIs are hard to come by, as our experience with mtags and metacp shows.
* Symbol lookup in consumers
    * We don't really provide the API to look up symbols within other symbols, so this is currently a moot point.
* Manually writing down symbols
    * Current approach is really bad at this, too. A human would have to perform type erasure in their head to be able to easily write down a symbol. The state of the art solution is to write example code and see what symbols are generated for it by semanticdb-scalac.
* Eyeballing
    * Current approach produces walls of text, which is pretty bad.
* Incremental compilation
[...]

Issue #1294, opened 2018-02-09
Rethink how class parameters are modelled


xeno-by:
Here is an excerpt from lowlevel.expect for `class AA(x: Int, val y: Int, var z: Int)`:

```
_root_.flags.p.package.AA# => class AA.{+7 decls}
  extends AnyRef
_root_.flags.p.package.AA#(x) => param x: Int
  Int => _root_.scala.Int#
_root_.flags.p.package.AA#(y) => val param y: : Int
  Int => _root_.scala.Int#
_root_.flags.p.package.AA#(z) => var param z_=: (x$1: <?>): Unit
  x$1 => _root_.flags.p.package.AA#`z_=`(I)V.(x$1)
  Unit => _root_.scala.Unit#
_root_.flags.p.package.AA#`<init>`(III)V. => primaryctor <init>: (x: <?>, y: <?>, z: <?>): AA
  x => _root_.flags.p.package.AA#`<init>`(III)V.(x)
  y => _root_.flags.p.package.AA#`<init>`(III)V.(y)
  z => _root_.flags.p.package.AA#`<init>`(III)V.(z)
  AA => _root_.flags.p.package.AA#
```

The original idea was to pretend that a parameter in a primary constructor is just one symbol. This decision was caused by the scala.reflect madness, in which class parameters may give rise to up to three symbols: 1) a constructor parameter, 2) an accessor, 3) an underlying field.

However, while writing metacp I realized that this approach is unfortunately flawed. Now that we have SymbolInformation.tpe, we can no longer pretend that there is just one symbol to rule them all (which causes the <?> thingies in the printout above).

I think we should bite the bullet and admit that every parameter in a primary constructor creates two symbols: 1) a class parameter, 2) a constructor parameter.

xeno-by:
Here's something even more interesting. I think we'll have to have separate symbols for getters and setters. See https://github.com/scalameta/scalameta/issues/1293#issuecomment-364530629 for discussion.

xeno-by:
Implemented in https://github.com/scalameta/scalameta/pull/1326.

PR #1310, opened 2018-02-14
[WIP] Major updates to the SemanticDB spec to unblock our progress

xeno-by:
This pull request only contains updates to the SemanticDB specification and the associated schema. It does not contain any implementation. Still, I think it's worth reviewing, since the changes are huge already. The build is going to fail, and that's expected.

Summary:
  * Added `Language`.
  * Separated language-agnostic and language-dependent parts of the spec.
  * Changed method symbols (https://github.com/scalameta/scalameta/issues/1295).
  * Refactored `Type` to be language-agnostic by merging all singleton types into `SINGLETON_TYPE`, renaming `COMPOUND_TYPE` to `STRUCTURAL_TYPE` and renaming `TYPE_LAMBDA` to `UNIVERSAL_TYPE`. This uses the terminology that is shared across languages.
  * Added `Kind.GETTER` and `Kind.SETTER` (https://github.com/scalameta/scalameta/issues/1293 and https://github.com/scalameta/scalameta/issues/1294).
  * Removed `Property.VALPARAM` and `Property.VARPARAM` (https://github.com/scalameta/scalameta/issues/1294).
  * Added `SymbolInformation.annotations` (https://github.com/scalameta/scalameta/issues/1308).
  * Removed `Property.PRIVATE` and `Property.PROTECTED` (https://github.com/scalameta/scalameta/issues/1309).
  * Added `SymbolInformation.accessibility` (https://github.com/scalameta/scalameta/issues/1309).
  * Exhaustively documented how Scala language features map onto SemanticDB.

### DIFF
--- a/semanticdb/semanticdb3/semanticdb3.md
+++ b/semanticdb/semanticdb3/semanticdb3.md
@@ -92,7 +92,7 @@ to produce them in the `schema` field. The following versions are supported:
     <td><b>Data model<b></td>
   </tr>
   <tr>
-    <td>LEGACY</td>
+    <td><code>LEGACY<code></td>
     <td>Legacy SemanticDB payloads</td>
     <td><a href="https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb2/semanticdb2.proto">semanticdb2.proto</a></td>
   </tr>
@@ -668,12 +668,12 @@ documentation strings or features from other languages.
   <tr>
     <td><code>0x400</code></td>
     <td><code>VALPARAM</code></td>
-    <td>Is a `val` parameter of a primary constructor?</td>
+    <td>Is a <code>val</code> parameter of a primary constructor?</td>
   </tr>
   <tr>
     <td><code>0x800</code></td>
     <td><code>VARPARAM</code></td>
-    <td>Is a `var` parameter of a primary constructor?</td>
+    <td>Is a <code>var</code> parameter of a primary constructor?</td>
   </tr>
 </table>
 
--- a/semanticdb/semanticdb3/semanticdb3.md
+++ b/semanticdb/semanticdb3/semanticdb3.md
@@ -7,9 +7,10 @@
   * [Range](#range)
   * [Location](#location)
   * [Symbol](#symbol)
-  * [Annotation](#annotation)
   * [Type](#type)
   * [SymbolInformation](#symbolinformation)
+  * [Annotation](#annotation)
+  * [Accessibility](#accessibility)
   * [SymbolOccurrence](#symboloccurrence)
   * [Diagnostic](#diagnostic)
   * [Synthetic](#synthetic)
@@ -19,9 +20,10 @@
   * [Scala](#scala)
     * [Language](#scala-language)
     * [Symbol](#scala-symbol)
-    * [Annotation](#scala-annotation)
     * [Type](#scala-type)
     * [SymbolInformation](#scala-symbolinformation)
+    * [Annotation](#scala-annotation)
+    * [Accessibility](#scala-accessibility)
     * [Synthetic](#scala-synthetic)
 
 ## Motivation
@@ -187,7 +189,7 @@ Global symbol format accommodates scoping rules of the underlying language,
 and is therefore language-dependent. For example, the `Int` class in
 the Scala standard library must be modelled as `_root_.scala.Int#`.
 The `_root_` prefix stands for the root package defined in the Scala Language
-Specification [\[18\]][18], and the pound sign (`#`) at the end of the symbol
+Specification [\[17\]][17], and the pound sign (`#`) at the end of the symbol
 means that the symbol corresponds to a class as opposed to objects that end
 with a dot (`.`). See [Languages](#languages) for more information.
 
@@ -241,18 +243,6 @@ is not a valid symbol.
 [Synthetics](#synthetic). Must not be used outside `Synthetic.text` documents.
 Placeholder symbols are always equal to an asterisk (`*`).
 
-### Annotation
-
-```protobuf
-message Annotation {
-  Type tpe = 1;
-}
-```
-
-`Annotation` represents annotations. See [Languages](#languages) for
-information on how annotations in supported languages map onto this
-data structure.
-
 ### Type
 
 ```protobuf
@@ -292,7 +282,7 @@ message Type {
 
 The SemanticDB type system is a language-agnostic superset of the type systems
 of supported languages - currently modelled after the Scala type system
-[\[19\]][19]. This section describes the language-agnostic model, while
+[\[18\]][18]. This section describes the language-agnostic model, while
 [Languages](#languages) elaborates on how language types map onto this model.
 
 ```protobuf
@@ -456,13 +446,15 @@ message SymbolInformation {
   repeated string members = 8;
   repeated string overrides = 9;
   Type tpe = 11;
+  repeated Annotation annotations = 13;
+  Accessibility accessibility = 14;
 }
 ```
 
 "Symbols" is a section of a [TextDocument](#textdocument) that stores
 information about [Symbols](#symbol) that are defined in the underlying
 document. In a sense, this section is analogous to a symbol table
-[\[20\]][20] in a compiler.
+[\[19\]][19] in a compiler.
 
 `SymbolInformation` contains assorted metadata for a `symbol`,
 as explained below. At the moment, the supported metadata is usecase-driven
@@ -472,7 +464,7 @@ we may add support for more metadata, e.g. documentation strings.
 `language`. [Language](#language) that defines this symbol.
 
 `kind`. Enumeration that defines the kind of the symbol.
-See [Languages](#language) to learn how definitions in supported
+See [Languages](#language) for information on how definitions in supported
 languages map onto these kinds.
 
 <table>
@@ -499,12 +491,12 @@ languages map onto these kinds.
   <tr>
     <td><code>15</code></td>
     <td><code>GETTER</code></td>
-    <td>Getter method, e.g. <code>def x: Int</code> generated for <code>val x = 42</code>.</td>
+    <td>Getter, e.g. <code>def x: Int</code> generated for <code>val x = 42</code>.</td>
   </tr>
   <tr>
     <td><code>16</code></td>
     <td><code>SETTER</code></td>
-    <td>Setter method, e.g. <code>def x_=(x$1: Int): Unit</code> generated for <code>var x = 42</code>.</td>
+    <td>Setter, e.g. <code>def x_=(x$1: Int): Unit</code> generated for <code>var x = 42</code>.</td>
   </tr>
   <tr>
     <td><code>4</code></td>
@@ -564,6 +556,8 @@ languages map onto these kinds.
 </table>
 
 `properties`. Bitmask of miscellaneous bits of metadata:
+See [Languages](#language) for information on how definitions in supported
+languages map onto these properties.
 
 <table>
   <tr>
@@ -573,13 +567,13 @@ languages map onto these kinds.
   </tr>
   <tr>
     <td><code>0x1</code></td>
-    <td><code>PRIVATE</code></td>
-    <td>Has a <code>private</code> modifier?</td>
+    <td><code>RESERVED</code></td>
+    <td>Reserved for backward compatibility.</td>
   </tr>
   <tr>
     <td><code>0x2</code></td>
-    <td><code>PROTECTED</code></td>
-    <td>Has a <code>protected</code> modifier?</td>
+    <td><code>RESERVED</code></td>
+    <td>Reserved for backward compatibility.</td>
   </tr>
   <tr>
     <td><code>0x4</code></td>
@@ -623,6 +617,16 @@ languages map onto these kinds.
     <td><code>CONTRAVARIANT</code></td>
     <td>Has a contravariant (<code>-</code>) modifier?</td>
   </tr>
+  <tr>
+    <td><code>0x400</code></td>
+    <td><code>RESERVED</code></td>
+    <td>Reserved for backward compatibility.</td>
+  </tr>
+  <tr>
+    <td><code>0x800</code></td>
+    <td><code>RESERVED</code></td>
+    <td>Reserved for backward compatibility.</td>
+  </tr>
 </table>
 
 `name`. String that represents the name of the corresponding definition.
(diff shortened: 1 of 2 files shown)


semanticdb/semanticdb3/semanticdb3.proto
```diff
@@ -36,13 +41,11 @@ message Location {
 
 message Type {
   enum Tag {
+    // reserved 2, 3, 4, 5;
```
xeno-by: @thesamet Is it expected that this line fails? https://github.com/google/protobuf/issues/3458 suggests that Protocol Buffers have support for `reserved` in enums.

thesamet: @xeno-by that depends on the version of protoc you are using. As far as I can tell protoc 3.5.x supports `reserved`. It looks like your build is using a custom protoc (https://github.com/scalameta/scalameta/blob/7a4f642c6173598f7edd8ceb4398af906f927d12/build.sbt#L543) I'm not using which version it is.

xeno-by: This part of build configuration only uses custom protoc on `NixOS`, otherwise it falls back to the standard `(PB.runProtoc in Compile).value`. How do I find what version of `protoc` is bundled with ScalaPB?

thesamet: The default behavior is to call into protoc-jar and passing a version to it. It can be controlled/examined using [`PB.protocVersion`](https://github.com/thesamet/sbt-protoc/blob/c29d4d6c6074b25edda1efdabd89eddea27fe770/src/main/scala/sbtprotoc/ProtocPlugin.scala#L66). The command invoked is also printed so you might be able to see the version there as well. Can you link/paste the full output of whatever is failing?

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -64,23 +87,24 @@ message TextDocuments {
...
 message TextDocument {
+  reserved 4;
   Schema schema = 1;
   string uri = 2;
   string text = 3;
-  string language = 4;
-  repeated SymbolOccurrence occurrences = 5;
-  repeated SymbolInformation symbols = 6;
+  Language language = 9;
+  repeated SymbolInformation symbols = 5;
```
olafurpg: Should remain unchanged

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -64,23 +87,24 @@ message TextDocuments {
...
+  repeated SymbolOccurrence occurrences = 6;
```
olafurpg: ditto

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -141,129 +187,100 @@ It represents a location inside a document, such as a line inside a text file.
...
-  * Local symbols must start with `local`, so that they can be easily
-    distinguished from global symbols.
-  * Local symbols must be unique within the underlying document.
 
-For example, `x` in `def identity[T](x: T): T` may be modelled by local symbols
-`local0`, `local_x`, `local_identity_x`, as long as these names are unique within
-the underlying document. The same logic applies to the type parameter `T`,
-which is also a local definition.
+Local symbol format is language-agnostic and is a concatenation of `local`
+and a decimal number. For example, `x` in a Scala method
```
olafurpg: we append non-decimal numbers to local symbols in the converter to the legacy api, I think "starts with local" is enough

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -141,129 +187,100 @@ It represents a location inside a document, such as a line inside a text file.
...
 tool authors are advised to accompany local symbols with `TextDocument.uri`.
 
-**Multi symbols**. Are used to model references to a set of multiple definitions at once.
-This is occasionally useful to support corner cases of Scala, e.g. identifiers
-in imports that can refer to both a class and an object with the same name,
-or references to unresolved overloaded methods.
+**Multi symbols**. Represent references to multiple definitions at once.
+For example, this is occasionally useful to support corner cases of Scala,
+e.g. identifiers in imports that can refer to both a class and an object with
+the same name, or references to unresolved overloaded methods.
```
olafurpg: overloaded methods is a better motivating example than "corner case of scala"

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -272,141 +289,102 @@ message Type {
...
-In this section, we describe various alternatives of `Type`, providing example
-SemanticDB data that corresponds to different Scala types,
-inspired by the scala.reflect documentation [\[27\]][27].
-
-In these examples, we will be using a simple notation to describe SemanticDB data.
-In this notation, `M(v1, v2, ...)` corresponds a message `M` with fields set to values
-`v1`, `v2`, etc. Literals correspond to scalar values, and `List(x1, x2, ...)`
-corresponds to repeated values. Moreover, `<X>` corresponds to a message that
-represents `X`.
+The SemanticDB type system is a language-agnostic superset of the type systems
```
olafurpg: I'm weary of claiming "language-agnostic" when there is only support for one language. I'm sure there are tons of missing categories of types that not covered here but are used in other mainstream languages.

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -647,41 +627,73 @@ or features from other languages.
...
+```protobuf
+message Accessibility {
+  enum Tag {
+    UNKNOWN_TAG = 0;
+    PRIVATE = 1;
+    PRIVATE_THIS = 2;
+    PRIVATE_WITHIN = 3;
+    PROTECTED = 4;
+    PROTECTED_THIS = 5;
+    PROTECTED_WITHIN = 6;
```
olafurpg: No public?

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -647,41 +627,73 @@ or features from other languages.
...
 
-`overrides`. Symbols that are overridden by this symbol either directly or transitively.
+`accessibility`. [Accessibility](#accessibility) of the corresponding definition.
 
-`tpe`. [Type](#type) that represents the type signature of the definition.
+### Annotation
+
+```protobuf
+message Annotation {
+  Type tpe = 1;
```
olafurpg: I'm curious if we could use `Type` to encode literal arguments for annotations like `@deprecated("Message", "version")`, basically support term arguments only if they have a translation to `Type`

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -773,38 +792,1079 @@ message Synthetic {
...
+* The encoded name of a definition is:
+  * For a Java identifier [\[22\]][22], the name itself.
+  * Otherwise, concatenation of a backtick, the name itself and another backtick.
+* The SemanticDB descriptor of a method is a concatenation of simplifications of
+  its parameter types interspersed by a comma (`,`), where a simplification
+  of a type is:
+  * For a `TYPE_REF`, the `SymbolInformation.name` of the underlying symbol.
+  * For a `SINGLETON_TYPE`, its Scala syntax.
+  * For a `STRUCTURAL_TYPE`, `{}`.
+  * For an `ANNOTATED_TYPE`, the simplification of `tpe`.
```
olafurpg: Missing example below

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -773,38 +792,1079 @@ message Synthetic {
...
+    encoded name and a right bracket (`]`).
+* The encoded name of a definition is:
+  * For a Java identifier [\[22\]][22], the name itself.
+  * Otherwise, concatenation of a backtick, the name itself and another backtick.
+* The SemanticDB descriptor of a method is a concatenation of simplifications of
+  its parameter types interspersed by a comma (`,`), where a simplification
+  of a type is:
+  * For a `TYPE_REF`, the `SymbolInformation.name` of the underlying symbol.
+  * For a `SINGLETON_TYPE`, its Scala syntax.
+  * For a `STRUCTURAL_TYPE`, `{}`.
```
olafurpg: Missing example below

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -773,38 +792,1079 @@ message Synthetic {
...
+  * For an `ANNOTATED_TYPE`, the simplification of `tpe`.
+  * For an `EXISTENTIAL_TYPE`, the simplification of `tpe`.
```
olafurpg: Missing example below

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -773,38 +792,1079 @@ message Synthetic {
...
+    and therefore are not supported.
+
+For example, this is how some of the definitions from the Scala standard library
+must be modelled:
+
+* The `scala` package: `_root_.scala.`
+* The `Int` class: `_root_.scala.Int#`
+* The `def println(x: Any)` method: `_root_.scala.Predef.println(Any).`
+* The `def +:[B >: A, That](elem: B)(implicit bf: CanBuildFrom[List[A], B, That]): That` method: ``_root_.scala.collection.immutable.List#`+:`(B,CanBuildFrom).``
+* The integer addition method: ``_root_.scala.Int#`+`(Int).``
```
olafurpg: Missing example for overloaded methods that have conflicting descriptors 

```scala
trait A {
  object b { type Foo = String }
  object c { type Foo = Int }
  def foo(arg: a.Foo): Int // A#foo(Foo)
  def foo(arg: b.Foo): Int // A#foo(Foo)
}
```

olafurpg: Another example worth covering might be context bounds
```scala
def foo[T:Ordering] // foo(Ordering)
```

semanticdb/semanticdb3/semanticdb3.md
xeno-by: This is just synchronizing the spec with the protobuf file.

semanticdb/semanticdb3/semanticdb3.md
xeno-by: High-level API isn't governed by the spec though. Let's make this as prescriptive as possible.

semanticdb/semanticdb3/semanticdb3.md
xeno-by: This formulation was around like that for a long time. Unresolved overloaded methods is a corner case though, I think. You don't see them in typechecked Scala programs.

semanticdb/semanticdb3/semanticdb3.md
xeno-by: We can do `PUBLIC`. That's a great idea actually.

semanticdb/semanticdb3/semanticdb3.md
xeno-by: If we end up having an immediate use case for that, I'd rather we extract `Literal` from `SingletonType` and use it for arguments.

semanticdb/semanticdb3/semanticdb3.md
xeno-by: I'll try to provide examples.

xeno-by:
Thanks for the feedback! I'll be closing this PR now and will be working on implementation today.

semanticdb/semanticdb3/semanticdb3.proto
xeno-by:
```
> semanticdb3JVM/compile
[info] Compiling 1 protobuf files to /Users/eburmako/Projects/scalameta/semanticdb/semanticdb3/.jvm/target/scala-2.12/src_managed/main
[info] Compiling schema /Users/eburmako/Projects/scalameta/semanticdb/semanticdb3/semanticdb3.proto
protoc-jar: protoc version: 340, detected platform: mac os x/x86_64
protoc-jar: executing: [/var/folders/pg/v06175sd2_qb9jbg7c28xr040000gn/T/protocjar5934144415890866447/bin/protoc.exe, --plugin=protoc-gen-scala=/var/folders/pg/v06175sd2_qb9jbg7c28xr040000gn/T/protocbridge4997102442195402814, --scala_out=flat_package,grpc:/Users/eburmako/Projects/scalameta/semanticdb/semanticdb3/.jvm/target/scala-2.12/src_managed/main, -I/Users/eburmako/Projects/scalameta/semanticdb/semanticdb3, -I/Users/eburmako/Projects/scalameta/semanticdb/semanticdb3/.jvm/target/protobuf_external, /Users/eburmako/Projects/scalameta/semanticdb/semanticdb3/semanticdb3.proto]
semanticdb3.proto:44:14: Missing numeric value for enum constant.
[trace] Stack trace suppressed: run last semanticdb3JVM/compile:protocGenerate for the full output.
[error] (semanticdb3JVM/compile:protocGenerate) protoc returned exit code: 1
[error] Total time: 1 s, completed Feb 14, 2018 8:59:13 AM
```

https://github.com/xeno-by/scalameta/commit/142aa521aa6ed295836400a5448eede59e949ab3

semanticdb/semanticdb3/semanticdb3.md
xeno-by: The next line says "of supported languages", which adds up to "The SemanticDB type system is a language-agnostic superset of the type systems of supported languages", so this is not a claim that the SemanticDB type system should be enough for everyone.

What I meant here is that we make a commitment for the type system to be abstracted away from concrete languages, and I think this diff is a nice improvement in this direction. For example, we have renamed compound types and type lambdas (Scala-isms) to the language-agnostic notions of structural types and universal types.

xeno-by: No idea how whether this will work actually, but I think we ought to try.

olafurpg: I would try at least to dampen down the claims, it's easy to misinterpret "language agnostic" as naive wishful thinking if you are type system researcher. And non-experts might believe this is the holy grail, which it's not either.

xeno-by: Sounds good.

semanticdb/semanticdb3/semanticdb3.md
olafurpg: I agree that would be a nicer implementation but it would not be able to encode singleton types (java enum, objects). Maybe we should support only the same terms as java annotations:

* enum
* literal
* array

probably missed something else.

xeno-by: Also nested annotations. All in all, it's a rabbit hole.

Issue #1293, opened 2018-02-09
Rethink how vals/vars are modelled


xeno-by:
```
Symbols:
_root_.flags.p.package.AA# => class AA
_root_.flags.p.package.AA#(x) => param x: Int
  [0..3): Int => _root_.scala.Int#
_root_.flags.p.package.AA#(y) => val param y: Int
  [0..3): Int => _root_.scala.Int#
_root_.flags.p.package.AA#(z) => var param z_=: (x$1: Int): Unit
  [6..9): Int => _root_.scala.Int#
  [12..16): Unit => _root_.scala.Unit#
```

xeno-by:
I think we'll have to have separate symbols for getters and setters. If not, how do we sanely model something like:

```scala
class B {
  var x = 42
}

class C extends B {
  override def x_=(value: Int) = sys.error("nuh-uh")
}
```

xeno-by:
Here's a proposal for the new scheme of things (includes parts of the proposed design for https://github.com/scalameta/scalameta/issues/1294 and ignores the problems with methods described in https://github.com/scalameta/scalameta/issues/1295)

```scala
class C(xp: T, val yp: T, var zp: T) {
  // NOTE: ym/zm aren't different from yp/zp in any interesting way.
  // val ym: T = ???
  // var zm: T = ???
  def m = {
    val yl: T = ???
    var zl: T = ???
  }
}
```

```
_empty_.C#yp. => getter def yp: : Int
_empty_.C#zp(0). => getter def zp: : Int
_empty_.C#zp_=(1). => setter def zp_=: (x$1: Int): Unit
_empty_.C#`<init>`(3). => primaryctor ...
_empty_.C#`<init>`(3).(xp) => param xp: Int
_empty_.C#`<init>`(3).(yp) => param yp: Int
_empty_.C#`<init>`(3).(zp) => param zp: Int
local0 => val yl: Int
local1 => var yl: Int
```

Proposed changes:
  * Classes no longer have parameters (but still have type parameters, I guess).
  * `val` and `var` parameters/members become `DEF` methods with appropriate signatures.
  * `VAL` and `VAR` aren't going anywhere, because they are still needed for locals.
  * We're not shying away from synthesized setter signatures, because we'll need them if someone decides to override those setters.
  * New `SymbolInformation.Property` properties: `GETTER` and `SETTER` to disambiguate regular methods from getters and setters.
[...]

PR #1316, opened 2018-02-15
Further updates to the SemanticDB spec

xeno-by:
Supersedes #1310, proposes major updates to the SemanticDB specification, implements many of these updates.

Summary:
  * Added `Language`.
  * Separated language-agnostic and language-dependent parts of the spec.
  * Changed method symbols (https://github.com/scalameta/scalameta/issues/1295).
  * Refactored `Type` to be language-agnostic by merging all singleton types into `SINGLETON_TYPE`, renaming `COMPOUND_TYPE` to `STRUCTURAL_TYPE` and renaming `TYPE_LAMBDA` to `UNIVERSAL_TYPE`. This uses the terminology that is shared across languages.
  * Added `Kind.GETTER` and `Kind.SETTER` (https://github.com/scalameta/scalameta/issues/1293 and https://github.com/scalameta/scalameta/issues/1294).
  * Removed `Property.VALPARAM` and `Property.VARPARAM` (https://github.com/scalameta/scalameta/issues/1294).
  * Added `SymbolInformation.annotations` (https://github.com/scalameta/scalameta/issues/1308).
  * Removed `Property.PRIVATE` and `Property.PROTECTED` (https://github.com/scalameta/scalameta/issues/1309).
  * Added `SymbolInformation.accessibility` (https://github.com/scalameta/scalameta/issues/1309).
  * Exhaustively documented how Scala language features map onto SemanticDB.

Todo: 
  * [ ] Implement the new method symbol format.
  * [ ] Implement the new encoding for vals and vars.
  * [ ] Decide whether `hasNontrivialPrefix` should stay or go.

I think that even in the current state this pull request may be mergeable. The diff is already big enough to present potential rebase problems. Action items on todo list can be addressed after the merge.


langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Denotation.scala
```diff
@@ -108,6 +134,18 @@ object Denotation extends AbstractFunction4[Long, String, String, List[ResolvedN
             tpe: Option[s.Type]): Denotation =
     new Denotation(flags, name, signature, names, members, overrides, tpe)
 
+  def apply(flags: Long,
```
olafurpg: Why do we still bother with apply when we can use the primary ctor?

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -210,11 +215,16 @@ object Main {
...
 
   private def sname(sym: Symbol): String = {
-    if (sym.name == "<no symbol>") ""
-    else if (sym.name == "<root>") "_root_"
-    else if (sym.name == "<empty>") "_empty_"
-    else if (sym.name == "<init>") "<init>"
-    else NameTransformer.decode(sym.name)
+    def loop(name: String): String = {
+      val i = name.lastIndexOf("$$")
+      if (i > 0) return loop(name.substring(i + 2))
```
olafurpg: can we please reserve `return` for early returns in large method bodies and use if/else in small methods like here?

olafurpg: return is inherently harder to reason about

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -359,6 +374,33 @@ object Main {
     }
   }
 
+  def sanns(sym: SymbolInfoSymbol): List[s.Annotation] = {
+    // TODO: Not supported by scalap.
```
olafurpg: 😭 seems we'll need to roll our own solution sooner or later

tests/jvm/src/test/resources/lowlevel.expect
```diff
@@ -39,7 +114,7 @@ _root_.scala.Unit# => abstract final class Unit.{+2 decls}
   extends AnyVal
 _root_.scala.collection. => package collection
 _root_.scala.collection.mutable. => package mutable
-_root_.scala.collection.mutable.Stack# => class Stack.{+20 decls}
+_root_.scala.collection.mutable.Stack# => @deprecated @deprecated class Stack.{+20 decls}
```
olafurpg: Why `@deprecated` twice?

tests/jvm/src/test/resources/metac.owners
```diff
@@ -0,0 +1,206 @@
+_empty_.Empty# => _empty_.
```
olafurpg: I'm still sceptical about the value of owners since I believe we can compute this information directly from a prefix tree with children/owner links. But I can see the value in not having to rely on sophisticated data structures

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/DenotationOps.scala
```diff
@@ -113,6 +114,38 @@ trait DenotationOps { self: DatabaseOps =>
...
+        else Some(s.Accessibility(a.PUBLIC))
+      } else {
+        val ssym = gsym.privateWithin.toSemantic.syntax
+        if (gsym.isProtected) Some(s.Accessibility(a.PROTECTED_WITHIN, ssym))
+        else Some(s.Accessibility(a.PRIVATE_WITHIN, ssym))
+      }
+    }
+
+    private def owner: m.Symbol = {
+      if (gsym.isSemanticdbGlobal) gsym.owner.toSemantic
```
olafurpg: We should have a flag `-P:semanticdb:owners:all/none` to skip `owners`, I have no use for this in scalafix at the moment and I can imagine it has non-trivial impact on the semanticdb filesize.

olafurpg:
Mima is complaining about additions to `Flags` which is a trait nobody should inherit

```
[error]  * abstract method GETTER()Long in interface org.langmeta.semanticdb.Flags is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("org.langmeta.semanticdb.Flags.GETTER")
[error]  * abstract method SETTER()Long in interface org.langmeta.semanticdb.Flags is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("org.langmeta.semanticdb.Flags.SETTER")
[error]  * abstract synthetic method org$langmeta$semanticdb$Flags$_setter_$GETTER_=(Long)Unit in interface org.langmeta.semanticdb.Flags is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("org.langmeta.semanticdb.Flags.org$langmeta$semanticdb$Flags$_setter_$GETTER_=")
[error]  * abstract synthetic method org$langmeta$semanticdb$Flags$_setter_$SETTER_=(Long)Unit in interface org.langmeta.semanticdb.Flags is present only in current version
```


We can probably add a blank filter for any complaints about ReversedMissingMethodProblem in org.langmeta.semanticdb.Flags

langmeta/langmeta/js/src/main/scala/org/langmeta/internal/io/PlatformFileIO.scala
```diff
@@ -33,7 +34,12 @@ object PlatformFileIO {
     result
   }
 
-  def write(path: AbsolutePath, proto: GeneratedMessage): Unit = {
+  def readAllDocuments(path: AbsolutePath): List[TextDocument] = JSIO.inNode {
+    val bytes = readAllBytes(path)
+    TextDocuments.parseFrom(bytes).documents
```
olafurpg: Missing `.toList`, although I question why the signature is `List[T]` instead of `Seq[T]` which is what we have in the scalapb generated classes

olafurpg: We only care about iterating over the list, so `Seq[T]` is perfectly fine for that IMO

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Denotation.scala
xeno-by: I don't know. For me it's inertia. I'm fine with either.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
xeno-by: Will do.

tests/jvm/src/test/resources/lowlevel.expect
xeno-by: You wouldn't believe: https://github.com/scala/scala/blob/c2a5883891a68180b143eb462c8b0cebc8d3b021/src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala#L828-L831.

tests/jvm/src/test/resources/metac.owners
xeno-by: The idea that I want to try is to treat symbols as completely opaque tokens, with everything that we need to know living in `SymbolInformation`. I know that this is not how we've been rolling so far, but I think it's an interesting goal to have. It also simplifies the API.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/DenotationOps.scala
xeno-by: Will do.

langmeta/langmeta/js/src/main/scala/org/langmeta/internal/io/PlatformFileIO.scala
xeno-by: Will do.

langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Denotation.scala
olafurpg: I think we we should not bother with apply anymore.

olafurpg: as in, we should not keep on adding new apply constructors. What we should have done in v3.0.0 was to remove all the boilerplate and add `.withAnnotations()` builder methods.

build.sbt
```diff
@@ -68,6 +68,12 @@ commands += Command.command("ci-metac") { s =>
   }
   "testsJVM/test:runMain scala.meta.tests.semanticdb.MetacScalaLibrary" :: s
 }
+commands += Command.command("save-expect") { s =>
```
olafurpg: 👍

olafurpg:
Seems to be a sorting bug causing test failure on appveyor

```
  [439..450): specialized => _root_.scala.specialized#
[info]   -[451..451):  => _root_.scala.specialized#`<init>`()V.
[info]    [451..452): T <= _root_.flags.p.package.S#[T]
[info]   +[451..451):  => _root_.scala.specialized#`<init>`()V.
[info]    [454..454):  <= _root_.flags.p.package.S#`<init>`()V. (ExpectSuite.scala:44)
[info] - metac.owners
[info] MirrorConstructionSuite:

```

xeno-by:
@olafurpg Yep, fixed in the latest commit.

PR #1318, opened 2018-02-15
Use nio file visitors in metacp.

### LINKED ISSUE #1305: Metacp tries to read non-class files
While running metacp on spark-sql I got the following error

```
 java.net.URISyntaxException: Illegal character in opaque part at index 183: jar:file:///Users/ollie/.coursier/cache/v1/https/repo1.maven.org/maven2/org/apache/hadoop/hadoop-yarn-common/2.6.5/hadoop-yarn-common-2.6.5.jar!/webapps/static/dt-1.9.4/images/Sorting icons.psd
```
I expect metacp to ignore non-class files.

olafurpg:
Previously, metacp did a custom traversal of the classpath using a mix
of java.io and custom solutions via Fragment.  This commit adds
`Classpath.visit` which simplifies the implementations by using built-in
`java.nio.Files.walkFileTree`. This automatically fixes #1305.

### DIFF
--- a/langmeta/langmeta/shared/src/main/scala/org/langmeta/io/Multipath.scala
+++ b/langmeta/langmeta/shared/src/main/scala/org/langmeta/io/Multipath.scala
@@ -4,9 +4,13 @@ package io
 import java.net._
 import java.io._
 import java.io.File.pathSeparator
+import java.nio.file.FileVisitor
+import java.nio.file.Files
+import java.nio.file.Path
 import java.util.zip._
 import scala.collection.mutable
 import org.langmeta.internal.io.FileIO
+import org.langmeta.internal.io.PathIO
 
 sealed trait Multipath {
   def shallow: List[AbsolutePath]
@@ -63,6 +67,26 @@ sealed trait Multipath {
   def find(name: RelativePath): Option[URI] = {
     deep.find(_.name == name).map(_.uri)
   }
+
+  def visit(getVisitor: AbsolutePath => FileVisitor[Path]): Unit = {
+    shallow.foreach { base =>
+      val path = base.toNIO
+      if (Files.isDirectory(path)) {
+        val visitor = getVisitor(base)
+        Files.walkFileTree(path, visitor)
+      } else {
+        PathIO.extension(path) match {
+          case "jar" =>
+            val root = FileIO.jarRootPath(base)
+            val visitor = getVisitor(root)
+            Files.walkFileTree(root.toNIO, visitor)
+          case _ =>
+            sys.error(s"Expected jar file, obtained $base")
+        }
+      }
+    }
+  }
+
 }
 
 final case class Classpath(shallow: List[AbsolutePath]) extends Multipath {
--- a/semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
+++ b/semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
@@ -30,9 +30,7 @@ object Main {
     var failed = false
     def fail(file: Path, ex: Throwable): FileVisitResult = {
       println(s"error: can't convert $file")
-      if (ex != null) {
-        ex.printStackTrace()
-      }
+      ex.printStackTrace()
       failed = true
       FileVisitResult.CONTINUE
     }
@@ -85,7 +83,8 @@ object Main {
         }
 
         override def postVisitDirectory(dir: Path, e: IOException): FileVisitResult = {
-          fail(dir, e)
+          if (e != null) fail(dir, e)
+          else FileVisitResult.CONTINUE
         }
 
         override def visitFileFailed(file: Path, e: IOException): FileVisitResult = {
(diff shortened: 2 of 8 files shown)


langmeta/langmeta/shared/src/main/scala/org/langmeta/io/Multipath.scala
```diff
@@ -63,6 +67,26 @@ sealed trait Multipath {
   def find(name: RelativePath): Option[URI] = {
     deep.find(_.name == name).map(_.uri)
   }
+
+  def visit(getVisitor: AbsolutePath => FileVisitor[Path]): Unit = {
```
xeno-by: How about we make this an extension method? This will avoid the need to silence mima.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -25,44 +26,64 @@ object Main {
...
-          println(s"error: can't convert $fragment")
-          ex.printStackTrace()
-          failed = true
+
+        override def postVisitDirectory(dir: Path, e: IOException): FileVisitResult = {
+          if (e != null) throw e
+          FileVisitResult.CONTINUE
+        }
+
+        override def visitFileFailed(file: Path, e: IOException): FileVisitResult = throw e
```
xeno-by: What about exceptions that we get here? Shall we also catch and println them without crashing?

langmeta/langmeta/shared/src/main/scala/org/langmeta/io/Multipath.scala
olafurpg: This method deserves inclusion in Multipath IMO, and the mima warning is standard procedure when dealing with scala-specific features that are not first-class on the jvm like `sealed` and `private[pkg]`

xeno-by: What exactly is the error saying? I think we need to make sure that our `Every program linkable against x.y.0 must be linkable against x.{y+1}.z` rule is not violated.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -28,6 +28,14 @@ object Main {
     val semanticdbRoot = AbsolutePath(settings.d).resolve("META-INF").resolve("semanticdb")
     Files.createDirectories(semanticdbRoot.toNIO)
     var failed = false
+    def fail(file: Path, ex: Throwable): FileVisitResult = {
+      println(s"error: can't convert $file")
+      if (ex != null) {
+        ex.printStackTrace()
+      }
```
xeno-by: How can ex be null here?

olafurpg: In postVisitDirectory, 

```
     * @param   exc
     *          {@code null} if the iteration of the directory completes without
     *          an error; otherwise the I/O exception that caused the iteration
     *          of the directory to complete prematurely
```

langmeta/langmeta/shared/src/main/scala/org/langmeta/io/Multipath.scala
olafurpg: The error says that if a client from external repo implements that interface then it will fail at runtime. Because the trait is `sealed` the language contract promises us that nobody extends `Multipath`

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
xeno-by: If `ex` being `null` indicates success, then you can't call `fail` from `postVisitDirectory`.

langmeta/langmeta/shared/src/main/scala/org/langmeta/io/Multipath.scala
olafurpg: In Scala.js they ignore mima reports all the time because of package private and sealed, one example https://github.com/scala-js/scala-js/blob/ec3c3bd685ff32f9fbbe4e19db085cb651babf8c/project/BinaryIncompatibilities.scala#L9

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
olafurpg: guess failed should not become false in that case, let me fix this asap

langmeta/langmeta/shared/src/main/scala/org/langmeta/io/Multipath.scala
xeno-by: Oh right, `sealed`. Makes sense then.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
olafurpg: Comforting to know CI caught this bug https://ci.appveyor.com/project/olafurpg/scalameta/build/744#L1797

xeno-by:
Given that Travis is acting up, I'll be closing this pull request and incorporating it into mine (https://github.com/scalameta/scalameta/pull/1316). @olafurpg will need to sleep soon, and in my timezone it's still morning, so I'll have more time to fight with Travis.

PR #1320, opened 2018-02-15
 Reformat scalameta/semanticdb

xeno-by:


.scalafmt.conf
```diff
@@ -3,18 +3,25 @@ align.openParenCallSite=false
 align.openParenDefnSite=false
 align.tokens = []
 assumeStandardLibraryStripMargin = true
+binPack.parentConstructors = false
```
olafurpg: I believe this is default

.scalafmt.conf
```diff
@@ -3,18 +3,25 @@ align.openParenCallSite=false
...
+danglingParentheses = false
```
olafurpg: ditto

.scalafmt.conf
```diff
@@ -3,18 +3,25 @@ align.openParenCallSite=false
...
+docstrings = ScalaDoc
```
olafurpg: ditto

.scalafmt.conf
```diff
@@ -3,18 +3,25 @@ align.openParenCallSite=false
...
 maxColumn = 100
 optIn.annotationNewlines = true
+optIn.breakChainOnFirstMethodDot = true
```
olafurpg: ditto

.scalafmt.conf
```diff
@@ -3,18 +3,25 @@ align.openParenCallSite=false
...
+optIn.selfAnnotationNewline = true
```
olafurpg: ditto

xeno-by:
@olafurpg I just want to explicitly say that I'm interested in these particular values, because I have tried other values and didn't like the result. 

By settings these values explicitly, we are making sure that: 
  1) Potential changes in defaults in future versions of scalafmt will not affect us.
  2) Hypothetical future contributors that would like to change .scalafmt.conf will know that we're not just accidentally using the defaults, but are conscious in that choice.

Issue #253, opened 2015-10-02
Rethink Dialect


xeno-by:
## Background

In 0.0.4, I'm going to obviate the need to write `import scala.meta.dialects.Scala211` everywhere by introducing an implicit materializer that will expand to the dialect corresponding to the current version of Scala.

I was (and still kinda am) against this, because having default language version preselected doesn't feel like the right thing to do in a platform- and version-independent library.

However, as @densh has rightfully pointed out, this doubles the amount of boilerplate for simple tasks (requiring 2 imports instead of 1 imports for semantic APIs), plus it requires explaining an extra concept to the newcomers.

Thinking about this in more depth, I realized that our behavior wrt dialects is inconsistent with our behavior wrt contexts, because concrete implementations in scalahost are called Mirror, Proxy, Toolbox instead of Mirror211, Proxy211 and Toolbox211.

Since it's all so messed up, I'm going to err on the side of simplicity and lift the requirement to import a dialect explicitly. In the meanwhile, I'll keep this issue open so that we don't forget about it.
## Actionable items
- [ ] Figure out how to / whether to switch context implementations on the fly. 
  - Currently it's flat out impossible, because if we try scalahost211 and scalahost212 as separate library linked together in an umbrella scalahost library, it's not going to work, because mixing scala-compiler for 2.11 and scala-compiler for 2.12 (transitive dependencies of scalahosts) is going to lead to jar hell.
  - Maybe Jigsaw will make things better, but I have no idea. Maybe @darkdimius follows the developments there and could provide some info.
  - At some point we might want to rename `scala.meta.internal.hosts.scalac` into `scala.meta.internal.hosts.scalac211`.
- [ ] Find out a way to ensure that `implicitly[Dialect] == implicit[Context].dialect`.
  - We could have `Context.apply` take an implicit dialect and then make sure that it's equal to the dialect of the context that it's going to create. However, it's only going to work if you write `import some.dialect; { implicit val c = Context(...) }`, but it won't catch discrepancies in `implicit val c = Context(...); { import some.dialect }`.

Atry:
What is Context?

Atry:
Is there a macro to create Dialect according to the current compiler setting?

xeno-by:
This issue is pretty dated. `Context` is something that is called `Mirror` now.

There used to be a macro called `currentDialect`, but now it's just a normal method implemented differently by different platforms that we support:
[...]

PR #1322, opened 2018-02-15
Add a cross-compat dialect

DavidDudson:
Name up for discussion.

Introduce a restricted subset of all dialects to retain cross-compatibility. 
This is the exact opposite of what scalafmt needs (To support every feature)

This is important for code generation, as we want to generate code to a shared folder, and have it compiled by multiple targets, including Dotty.

This allows longevity of code generators, as they will not need to be updated when scala's version changes.

Note: We lose Sbt 0.13.5 here, but I would expect very few people will target sbt with code gen, and even fewer will be using 0.13.5 rather than 0.13.6.

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
@@ -220,6 +220,18 @@ package object dialects {
     allowXmlLiterals = false // Dotty parser doesn't have the corresponding code, so it can't really support xml literals
   )
 
+  /** Dialect that tries to be compatible with all language versions. */
+  implicit val CrossCompat = Scala210.copy(
+    allowAtForExtractorVarargs = false, // Removed in Dotty
+    allowCaseClassWithoutParameterList = false, // Removed in 2.11
+    allowInlineIdents = false, // Only in Paradise
+    allowMultilinePrograms = true, // Remains true, as this only affect quasiquotes
+    allowViewBounds = false,  // Removed in Dotty
+    allowWithTypes = false,   // Only in Dotty
+    allowXmlLiterals = false, // Removed in Dotty
+    toplevelSeparator = "" // Technically will break with Sbt 0.13.6, but it cannot be helped.
+  )
+
   // TODO: https://github.com/scalameta/scalameta/issues/380
   private[meta] def QuasiquoteTerm(underlying: Dialect, multiline: Boolean) = {
     require(!underlying.allowUnquotes)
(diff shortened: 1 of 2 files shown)


DavidDudson:
sigh, is @xeno-by is there a way to get ci to recheck? Other then resubmitting the PR, it has not detected my force pushed changes

olafurpg:
@DavidDudson I restarted the CI, it was behaving weirdly last night

olafurpg:
There's a legitimate test failure in SurfaceSuite

```
[info] - prettyprinters for statics (core) *** FAILED ***
[info]   isTested was false scala.meta.dialects.CrossCompat prettyprinting is not tested (SurfaceSuite.scala:133)
```

olafurpg:
I'm still not sure if this dialect needs to be in the scalameta repo, would it be OK to keep this in scalagen instead? I maintain custom dialects in both the scalafix and scalafmt repos, for better or worse. In https://github.com/scalameta/scalameta/issues/253#issuecomment-312634924 we discuss that dialects should probably be removed from pretty printing.

DavidDudson:
@olafurpg Dialects could definitely be removed from pretty printing, this is more for quasiquotes and parsing then emitting the right syntax. I'm happy to keep this in scalagen, unless we could move the scalafmt and scalafix ones to contrib? I thought this seemed a bit out of place given that the rest of the Dialects are backend specific.

We can either do...

1. Move this to contrib.
2. Leave it in scalagen.

I can definitely see the want for these "tooling" dialects being somewhere, rather than each project making up their own version though.

I'll close this.

olafurpg:
I agree it's not optimal to have every tool customize it's own dialect, but as you say these custom dialects appear out of place in scala.meta.dialects when all the other ones are backed by an specific compiler/tool version. I think contrib is a better place for dialects like CrossCompat, I'd be happy to merge it here. What do you think? While we're at it, should we maybe add the scalafix default dialect too as the opposite of `CrossCompat`? We could name it `Lenient` or `BestEffort`, any suggestions?

https://github.com/scalacenter/scalafix/blob/3d4de59322d8f63eff043f33e19e335ef59aa15a/scalafix-core/shared/src/main/scala/scalafix/internal/config/ScalafixConfig.scala#L64

DavidDudson:
I like the name `Lenient`.

We could use `Restricted` in place of CrossCompat for an alternative.

olafurpg:
I like `Restricted` as the dual of `Lenient`. The docstrings can further explain the use-cases for each dialect. For restricted the doc can say something like "least common denominator for all dialects, a program that parses with restricted dialect should be parseable with as many dialects as possible."

Issue #1333, opened 2018-02-17
Make a decision on typerefs with trivial prefixes


xeno-by:
Should `Int` be represented as `TypeRef(None, <Int>, List())` or `TypeRef(<scala.type>, <Int>, List())`? `metac` and `metacp` think the former, the spec thinks the latter. Both decisions have their pros and cons, though.

olafurpg:
Can you elaborate on the pros/cons? I like `TypeRef(None, <Int>, List())` better since I struggle to comphrehend what `scala.this.Int` means

xeno-by:
This comes from [SLS 3.2.3](https://www.scala-lang.org/files/archive/spec/2.12/03-types.html#type-designators):

> Specifically, the unqualified type name `t` where `t` is bound in some class, object, or package `C` is taken as a shorthand for `C.this.type#t`. If t is not bound in a class, object, or package, then `t` is taken as a shorthand for `ε.type#t`.

I'm undecided whether we should follow this rule or not in the SemanticDB specification (as you can see from me oscillating between `TypeRef(None, <Int>, List())` ("R1", for representation number 1) and `TypeRef(<scala.this.type>, <Int>, List())` ("R2").

I guess, in Scalac, they chose R2 because of consistency reasons. If you don't prohibit R1, then probably you need a bit more normalization and probably you spend more memory on hashconsing. However, R2 is clearly less convenient to users, because they often have to conjure useless prefixes

The design dilemma here boils down to whether we can compute an omitted prefix from a symbol without using a symbol table. If yes, then R1 and R2 are equally powerful, which means that according to the Scalameta spirit, R1 wins, because it's more convenient to users. If no, we'll have a much harder choice.

The good news is that I think the answer to the question is "yes". For a global symbol, we can always compute its owner without a symbol table and use the owner to auto-conjure the trivial prefix if necessary. For a local symbol, we have the entire compilation unit anyway, so it doesn't matter.

I'll make the changes to the spec and the implementation (https://github.com/scalameta/scalameta/pull/1339#pullrequestreview-97398196) later today.

xeno-by:
Implemented in https://github.com/scalameta/scalameta/pull/1339.

xeno-by:
Followed up by https://github.com/scalameta/scalameta/issues/1343.

PR #1339, opened 2018-02-18
Assorted cleanups

xeno-by:


tests/jvm/src/test/resources/metacp.expect
```diff
@@ -38,26 +67,36 @@ Schema => SemanticDB v3
...
+_root_.example. => package example
 _root_.example.Acc# => class Acc.{+8 decls}
-  extends AnyRef
-_root_.example.Acc#`<init>`(). => primaryctor <init>: (): Acc
+  extends scala.this.AnyRef
+_root_.example.Acc#`<init>`(). => primaryctor <init>: (): example.this.Acc
+  example => _root_.example.
   Acc => _root_.example.Acc#
-_root_.example.Acc#m1(). => private def m1: : Nothing
+_root_.example.Acc#m1(). => private def m1: : scala.this.Nothing
```
olafurpg: Is this change necessary? I prefer the old type

olafurpg: `scala.this.Int` is not even valid unless `scala` is an enclosing class

xeno-by: See https://github.com/scalameta/scalameta/issues/1333 for discussion.

PR #1342, opened 2018-02-18
Automate sbt sonatypeReleaseAll [skip ci]

olafurpg:


olafurpg:
I'm not 100% sure if this is correct, but it's worth a shot.

.travis.yml
```diff
@@ -9,6 +9,11 @@ script:
 jdk:
   - oraclejdk8
 
+stages:
+  - name: test
+  - name: release
+    if: (branch = master) AND (type = push)
```
xeno-by: This seems to be inconsistent with `if [[ "$TRAVIS_TAG" ]];` below?

olafurpg: Good catch! This is copied from the metals repo. I removed the `branch = master` guard but kept `type = push`

.travis.yml
```diff
@@ -45,6 +50,8 @@ matrix:
       - CI_SCALA_JS: true
     - env:
       - CI_TEST: mima
+    - stage: release
+      script: if [[ "$TRAVIS_TAG" ]]; then sbt "sonatypeReleaseAll"; else echo "skipping release"; fi
```
xeno-by: How about `"only releasing on tag pushes"`?

olafurpg: Done.

PR #1349, opened 2018-02-19
Preserve parent chain's origin when doing Tree.tranform. (fix #1345)

### LINKED ISSUE #1345: Tree.transform does not preserve parent chain origins
```scala
import scala.meta._

val code = """
  object a {
    def bar = 4
    // comment
    def foo = 2
  }
""".parse[Source].get

code.collect { case t => t.productPrefix -> (t.pos.start, t.pos.end) }

List(
  ("Source", (0, 65)),
  ("Defn.Object", (3, 64)),
  ("Term.Name", (10, 11)),
  ("Template", (12, 64)),
  ("Self", (18, 18)),
  ("Name.Anonymous", (18, 18)),
  ("Defn.Def", (18, 29)),
  ("Term.Name", (22, 25)),
  ("Lit.Int", (28, 29)),
  ("Defn.Def", (49, 60)),
  ("Term.Name", (53, 56)),
  ("Lit.Int", (59, 60))
)

code.transform { case q"2" => q"3" }.collect { case t => t.productPrefix -> (t.pos.start, t.pos.end) }

// obtained

List(
  ("Source", (-1, -1)),
  ("Defn.Object", (-1, -1)),
  ("Term.Name", (10, 11)),
  ("Template", (-1, -1)),
  ("Self", (18, 18)),
  ("Name.Anonymous", (18, 18)),
  ("Defn.Def", (18, 29)),
  ("Term.Name", (22, 25)),
  ("Lit.Int", (28, 29)),
  ("Defn.Def", (-1, -1)),
  ("Term.Name", (53, 56)),
  ("Lit.Int", (-1, -1))
)

// expected

List(
  ("Source", (0, 65)),
  ("Defn.Object", (3, 64)),
  ("Term.Name", (10, 11)),
  ("Template", (12, 64)),
  ("Self", (18, 18)),
  ("Name.Anonymous", (18, 18)),
  ("Defn.Def", (18, 29)),
  ("Term.Name", (22, 25)),
  ("Lit.Int", (28, 29)),
  ("Defn.Def", (49, 60)),
  ("Term.Name", (53, 56)),
  ("Lit.Int", (-1, -1))
)
```

MasseGuillaume:


scalameta/common/shared/src/main/scala/scala/meta/internal/transversers/transformer.scala
```diff
@@ -88,7 +88,7 @@ class TransformerMacros(val c: Context) extends TransverserMacros {
       var same = true
       ..$transformedFields
       if (same) tree
-      else $constructor(..${transformedFields.map(_.name)})
+      else $constructor(..${transformedFields.map(_.name)}).withOrigin(tree.origin)
```
olafurpg: Wow, I can't believe it was this simple!

tests/jvm/src/test/scala/scala/meta/tests/trees/ParentSuite.scala
```diff
@@ -0,0 +1,26 @@
+package scala.meta
+package tests
+package trees
+
+import org.scalatest._
+
+class ParentSuite extends FunSuite {
+
+  test("Tree.transform does not preserve parent chain origins") {
```
olafurpg: Can we add more detailed tests?

- validate only `Lit.Int(3)` is missing pos
- validate all other nodes have same position as their corresponding node have before

PR #1351, opened 2018-02-20
Keep origin on Tree.copy (fix #1350)

### LINKED ISSUE #1350: Tree.copy does not preserve the origin
```scala
import scala.meta._
val code ="object A { }".parse[Source].get
code.copy(stats = Nil).pos
```

MasseGuillaume:


tests/shared/src/test/scala/scala/meta/tests/parsers/PositionSuite.scala
```diff
@@ -60,6 +64,103 @@ class PositionSuite extends ParseSuite {
...
+              |Name.Anonymous{11..11}(), 
+              |None
+            |), 
+            |Nil))))""".trimPos
+
+    assert(code.show[Positions] === codePositions)
+
+    val refactored = code.copy(stats = Nil)
+    val refactoredPositions = """Source{0..12}(Nil)"""
+    assert(refactored.show[Positions] === refactoredPositions)
```
olafurpg: `assertNoDiff` with `extends DiffAssertsions` produces nicer error messages

olafurpg:
I think we should go ahead with Origin.Transformed for both transforms and copy as explained in https://github.com/scalameta/scalameta/issues/1350#issuecomment-367038022 this PR should be mostly unchanged with Origin.Transformed, but let's close this PR in favor of a dedicated PR for Origin.Transformed

PR #1363, opened 2018-02-26
Assorted fixes

xeno-by:


semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Javacp.scala
```diff
@@ -166,7 +166,10 @@ object Javacp {
 
     val classParents = classSignature match {
       case Some(c: ClassSignature) =>
-        c.parents.map(_.toType)
+        try c.parents.map(_.toType)
+        catch {
+          case _: NullPointerException => Nil
```
xeno-by: @olafurpg This is pretty outrageous, but I didn't have time to think of a proper solution. Gotta crash now.

olafurpg: Every field on asm nodes is nullable, so i'm not surprised one bit. Can take a look.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -122,6 +144,76 @@ object Main {
         }
       }
     }
+    locally {
+      def synthesizeBuiltin(name: String): s.SymbolInformation = {
```
olafurpg: What do you think about synthesizing these on the fly at runtime instead of re-generating scala/Any.class and friends for every entry in each classpath?

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -434,8 +526,26 @@ object Main {
       }
     }
 
-    try loop(sym.infoType)
-    catch {
+    try {
+      if (sym.isAlias) {
```
olafurpg: Looks like the logic inside this try body can be extracted into a method. I like it best when the try body is as small as possible `try process(foo) catch ...`

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -122,6 +144,76 @@ object Main {
...
+          language = Some(s.Language("Scala")),
+          kind = k.CLASS,
+          name = name,
+          tpe = Some(s.Type(tag = t.CLASS_INFO_TYPE, classInfoType = Some(tpe))),
+          owner = "_root_.scala."
+        )
+      }
+      def synthesizeAny(): List[s.SymbolInformation] = {
+        val decls = {
+          // TODO: Implement me.
```
olafurpg: Opened https://github.com/scalameta/scalameta/issues/1364 to track this.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
xeno-by: Could you elaborate on the reasons? Is this a purely stylistic preference, or this also influences IDE support and/or runtime performance?

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Javacp.scala
olafurpg: Fixed in #1367

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
olafurpg: Stylistic reasons, I get uneasy with massive try/catch blocks

Issue #1362, opened 2018-02-25
Address PR feedback from #1359


xeno-by:
* [x] Get rid of global mutable state
  * [x] Run on JDK
  * [x] Get rid of extraneous symbols
    * [x] Anonymous classes
    * [x] this$ constructor parameters
    * [x] this$ fields
  * [x] Data model (Major)
    * [x] p.STATIC
  * [x] Cleanup
    * [x] Main/Metacp/Javacp interface
    * [x] Visitors vs lists
    * [x] PackageSpecifier (JVMS deviation)
    * [x] V (JVMS deviation)
    * [x] * (JVMS deviation)
    * [x] [Array#[T: <?>]
    * [x] Uniform API for process(out)
    * [x] Other action items from https://github.com/scalameta/scalameta/pull/1359
  * [ ] Specification
    * [ ] Symbol (symbol format)
    * [ ] Type (which language types map to which SemanticDB types)
    * [ ] SymbolInformation (which language definitions map to which SemanticDB infos)
  * [ ] Data model (Minor)
    * [ ] k.FIELD, k.LOCAL, p.VAL, p.VAR
    * [ ] k.METHOD
    * [ ] k.INTERFACE
    * [ ] t.WITH_TYPE, t.AND_TYPE
    * [ ] k.ENUM etc

xeno-by:
Superseded by https://github.com/scalameta/scalameta/issues/1378.

PR #1359, opened 2018-02-23
WIP - Add Java support for metacp

olafurpg:
This PR extends metacp to support Java class files. Opening PR now to get feedback on the diff (which should be mostly polished) but still need to clean up the commit history.

PR #1367, opened 2018-02-26
Follow up from #1359

olafurpg:
Partially addressed the checkboxes in https://github.com/scalameta/scalameta/issues/1362.

* remove global mutable scope
* skip anonymous classes
* merge inner classes with outer classes
* add p.STATIC
* run javacp against JDK in tests


semanticdb/integration/src/main/java/com/javacp/Javacp.java
```diff
@@ -59,4 +59,7 @@ void packagePrivateMethod() { }
 
     public Serializable anonymous = new Serializable() { };
 
+    static void staticMethod() {}
+    static class StaticClass {}
```
xeno-by: Let's test a static field too, if you don't mind.

tests/jvm/src/test/resources/metacp.expect
```diff
@@ -36,73 +36,6 @@ _empty_.B#`<init>`(). => primaryctor <init>: (): B
 _empty_.B#a(). => def a: : A
   A => _empty_.A#
 
-com/javacp/ClassSuffix$Inner$Bar$Fuz.class
-------------------------------------------
```
xeno-by: Nested classes are now stored in their owner's semanticdb files, right?

tests/jvm/src/test/resources/metacp.expect
```diff
@@ -189,13 +94,13 @@ Schema => SemanticDB v3
...
 Language => Java
-Symbols => 56 entries
+Symbols => 66 entries
 
 Symbols:
 _root_. => package _root_
 _root_.com. => package com
 _root_.com.javacp. => package javacp
-_root_.com.javacp.Javacp# => class Javacp[A <: CharSequence with Serializable, B <: Object].{+30 decls}
+_root_.com.javacp.Javacp# => class Javacp[A <: CharSequence with Serializable, B <: Object].{+31 decls}
```
xeno-by: Why is it just +1 decl when you added both a static class and a static method?

tests/jvm/src/test/resources/metacp.expect
```diff
@@ -121,65 +54,37 @@ _root_.com.javacp.ClassSuffix# => class ClassSuffix.{+2 decls}
...
-Symbols => 4 entries
-
-Symbols:
-_root_.com.javacp.Javacp#1# => private[javacp] class 1.{+2 decls}
+_root_.com.javacp.ClassSuffix#Inner# => private[javacp] class Inner[A <: Object].{+1 decls}
   extends Object
-  extends Serializable
-_root_.com.javacp.Javacp#1#<init>(Javacp). => private[javacp] def <init>: (arg0: Javacp): Unit
-  arg0 => _root_.com.javacp.Javacp#1#<init>(Javacp).(arg0)
+_root_.com.javacp.ClassSuffix#Inner#<init>(ClassSuffix). => private[javacp] def <init>: (arg0: ClassSuffix): Unit
```
xeno-by: I like that `this$N` outer fields no longer show up in SemanticDB. Can we also filter out the corresponding outer parameters from inner class constructors? The heuristic would be "if the class contains a field such that `field.name.startsWith("this$")`, then drop the first parameters of the constructor".

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -128,7 +126,10 @@ object Main {
           files
             .iterator()
             .asScala
-            .filter(f => Files.isRegularFile(f))
+            .filter { f =>
+              !isVisited(f) &&
+              Files.isRegularFile(f)
+            }
```
xeno-by: Do we still need sorting? If not, let's go back to having all logic in `visitFile`.

tests/jvm/src/test/resources/metacp.expect
olafurpg: Correct, these entries are included below see https://github.com/scalameta/scalameta/pull/1367/files#diff-026fd03a8437acdb4213e4e80bf077d1R71

tests/jvm/src/test/resources/metacp.expect
olafurpg: Done.

tests/jvm/src/test/resources/metacp.expect
olafurpg: Good catch, I've updated the logic to include direct inner classes as ClassInfo.declarations.

Issue #1325, opened 2018-02-16
Incorrect accessibility for local definitions, parameters and type parameters


xeno-by:
Currently, if a definition is not private or protected, it's assigned with `Some(s.Accessibility(PUBLIC))`. This is incorrect, because things like local vals or pattern variables in expressions cannot be public. I would say that non-global symbols must have their `accessibility` set to `None`.

There's also a question of whether it is appropriate to say that some global symbols (namely, `PARAMETER` and `TYPE_PARAMETER` symbols) have accessibility. The spec is silent on that matter too - it neither lists valid accessibilities for parameters and type parameters, nor says that no accessibilities are supported.

PR #1373, opened 2018-02-27
Further follow up from #1359

xeno-by:
Builds upon #1367 to address some of the remaining action items in #1362.


scalameta/interactive/src/main/scala/scala/meta/interactive/InteractiveSemanticdb.scala
```diff
@@ -46,11 +52,12 @@ object InteractiveSemanticdb {
     * @param filename the name of the source file.
     * @param timeout max number of milliseconds to allow the presentation compiler
     *                to typecheck this file.
+    * @param options -P:semanticdb:* options to influence how the document is built.
```
olafurpg: Mention explicitly if the `-P:semanticdb:` prefix should be included

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
```diff
@@ -0,0 +1,532 @@
...
+          case _ => None
+        }
+      }
+      val stoplevels = toplevels.flatMap(sinfo)
+      val sothers = toplevels.flatMap(spackages).distinct ++ others.flatMap(sinfo)
+      ToplevelInfos(classfile, stoplevels, sothers)
+    }
+  }
+
+  private def spackages(sym: SymbolInfoSymbol): List[s.SymbolInformation] = {
```
olafurpg: Nitpick. I find it a bit weird to make as much as possible private in the internal package. These methods are not protected by any bincompat guarantees.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/package.scala
```diff
@@ -0,0 +1,17 @@
+package scala.meta.internal
+
+import java.nio.file._
+import scala.tools.asm._
+import scala.tools.asm.ClassReader._
+import scala.tools.asm.tree._
+
+package object metacp {
+  implicit class XtensionAsmPathOps(path: Path) {
+    def toClassNode: ClassNode = {
```
olafurpg: Nitpick. I personally think `toX` extension methods should be pure data transformations and not do I/O. I think java.nio and scala.concurrent get this right by making unsafe/blocking operations static methods on objects like `Await.result` and `Files.readAllBytes`.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
```diff
@@ -0,0 +1,532 @@
...
+      val (owner, name) = {
+        transitivePackagePath.split("\\.").toList match {
+          case List(name) if name.nonEmpty => ("", name)
+          case ownerSteps :+ name if name.nonEmpty => (ownerSteps.mkString(".") + ".", name)
+          case _ => sys.error(s"unsupported top-level symbol: $sym")
+        }
+      }
+      s.SymbolInformation(
+        symbol = transitivePackagePath + ".",
+        language = Some(s.Language("Scala")),
```
olafurpg: I think we should consider making `language`  an enum and add a `version: String` field. I find it awkward to hardcode `"Scala"` and `"Java"` all over the place and it opens up a can of worms related to equality (case-sensitivity etc.)

semanticdb/metap/src/main/scala/scala/meta/internal/metap/Settings.scala
```diff
@@ -0,0 +1,31 @@
+package scala.meta.internal.metap
+
+final case class Settings(
+    paths: List[String] = Nil,
+    pretty: Boolean = true,
```
olafurpg: > every boolean option is an enum screaming to get out
> -- Ólafur Páll Geirsson

```scala
sealed trait OutputFormat
object OutputFormat {
  case object Pretty extends OutputFormat
  case object Proto extends OutputFormat
}
```

olafurpg: Also helps follow the "Make invalid states unrepresentable" principle

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
```diff
@@ -1,86 +1,64 @@
...
+      access: Int,
+      scope: Scope): Seq[s.SymbolInformation] = {
 
     val buf = ArrayBuffer.empty[s.SymbolInformation]
     val decls = ListBuffer.empty[String]
 
     def addInfo(
         symbol: String,
         kind: s.SymbolInformation.Kind,
-        access: Int,
```
olafurpg: The ordering here was already following the ordering in semanticdb3.proto, `access` is used to set properties which come after kind 😉 The name "access" comes from asm, not semanticdb "accessibility" (which should maybe be called "visibility"?).

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
```diff
@@ -1,86 +1,64 @@
...
 
-  def sinfosClassNode(
+  private def sinfos(
+      toplevel: ToplevelClassfile,
       node: ClassNode,
-      outerClassAccess: Int,
-      scope: Scope,
-      root: Path,
-      isVisited: mutable.Set[Path]): Seq[s.SymbolInformation] = {
+      access: Int,
```
olafurpg: I intentionally renamed this from access to outerClassAccess in hope to make it clearer. No objections from me if you prefer it this way.

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
```diff
@@ -92,9 +70,9 @@ object Javacp {
           addInfo(
             pkgSymbol,
             k.PACKAGE,
-            o.ACC_PUBLIC,
             pkgName,
             None,
+            o.ACC_PUBLIC,
```
olafurpg: NB. I'm not sure if "Some(PUBLIC)" is the right accessibility for packages. Maybe it should be None instead. Doesn't matter for me.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
xeno-by: Yep, I was thinking about that yesterday. I've opened a ticket: https://github.com/scalameta/scalameta/issues/1374.

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
xeno-by: We already have a related ticket: https://github.com/scalameta/scalameta/issues/1325. Let's follow up there.

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
xeno-by: I think that `outerClassAccess` is somewhat misleading. It's not `access` of an outer class, but `access` of an `InnerClass` entry in outer class. I don't have better ideas though, so let's use `access` for the time being if you don't mind.

olafurpg: I agree.

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
xeno-by: Uh oh. My bad.

semanticdb/metap/src/main/scala/scala/meta/internal/metap/Settings.scala
xeno-by: 👍

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
xeno-by: This emphasizes that the methods aren't part of the interface between metacp, javacp and scalacp. However if you'd like, I can make them public again.

olafurpg: I hit on myself the other day that I wanted to access a private method in Metap from external repo. In this particular case it's probably better to come up with a better public API for metap than un-private methods. Nothing to act upon now.

PR #1380, opened 2018-02-28
Add documentation for Trees.scala

MasseGuillaume:
This is a WIP, but I would like to have some early feedback.

At a first glance, it looks like a duplication of quasiquote.md. However, it's really useful to see how you can construct a given `scala.meta.Tree` from a scala syntax.

### DIFF
--- a/notes/trees.md
+++ b/notes/trees.md
@@ -1,0 +1,174 @@
+
+## Expressions (`meta.Term`)
+
+```scala
+"1 + 1".parse[Term].get.structure
+```
+
+Ref             | Examples
+----------------|-----------
+Term.This       | `this`, `foo.this`
+Term.Super      | `super`, `foo.super`, `super[bar]"`, `foo.super[bar]`
+Term.Name       | `foo`
+Term.Select     | `bar.foo`
+Term.ApplyUnary | `!foo`, `~foo`, `-foo`, `foo`
+
+Term                  | Examples
+----------------------|------------------------------
+Term.Apply            | `buzz(42)`
+Term.ApplyType        | `implicitly[Ordering[Int]]`
+Term.ApplyInfix       | `a + a`
+Term.Assign           | `a = 1`
+Term.Return           | `return bar`
+Term.Throw            | `throw exception`
+Term.Ascribe          | `a: Int`
+Term.Annotate         | `(x: @unchecked)`
+Term.Tuple            | `(1, 1)`
+Term.Block            | `{ bar; buzz }`
+Term.If               | `if (p) t else f`
+Term.Match            | `x match { case _ => }`
+Term.Try              | `try foo catch { case _ => }`
+Term.TryWithHandler   | `try foo catch (h)`
+Term.Function         | `(a, b) => a + b`
+Term.PartialFunction  | `{ case _ =>  }`
+Term.While            | `while(p){ d }`
+Term.Do               | `do { d } while(p)`
+Term.For              | `for{x <- xs} f(x)`
+Term.ForYield         | `for{x <- xs} yield f(x)`
+Term.New              | `new A`
+Term.NewAnonymous     | `new A { }`
+Term.Placeholder      | `_`
+Term.Eta              | `foo _`
+Term.Repeated         | `foo(x: _*)`
+Term.Param            | `def foo<<(x: Int)>>`
+Term.Interpolate      | `s"|foo ~ $bar ~ ${42}|"`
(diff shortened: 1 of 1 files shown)


notes/trees.md
```diff
@@ -0,0 +1,174 @@
...
+Type.Refine           | `A { def f: Int }`
+Type.Existential      | `A forSome { type T }`
+Type.Annotate         | `T @A`
+Type.Lambda           | `[X] => (X, X)`
+Type.[Method]         | ???
+Type.Placeholder      | `T[_ >: Nothing <: Any]`
+Type.Bounds           | `def F[T >: Nothing <: Any] = 1`
+Type.ByName           | `def f(v: => T) = v`
+Type.Repeated         | `def print(v: Any*): Unit`
+Type.Var              | ???
```
olafurpg:
```scala
ollie-scala@ p"case _: List[t] =>"
res4: Case = Case(Pat.Typed(_, Type.Apply(Type.Name("List"), List(Type.Var(Type.Name("t"))))), None, Term.Block(List()))
```

notes/trees.md
```diff
@@ -0,0 +1,174 @@
...
+Type.ImplicitFunction | `implicit A => B`
+Type.Tuple            | `(A, B)`
+Type.With             | `A with B`
+Type.And              | `A & B`
+Type.Or               | `A \| B`
+Type.Refine           | `A { def f: Int }`
+Type.Existential      | `A forSome { type T }`
+Type.Annotate         | `T @A`
+Type.Lambda           | `[X] => (X, X)`
+Type.[Method]         | ???
```
olafurpg: n/a without a custom dialect. Can't be written in source
```scala
ollie-scala@ implicit val d = dialects.Scala212.copy(allowMethodTypes = true)
ollie-scala@ "(e: T): T".parse[Type]
res6: Parsed[Type] = Success(Type.Method(List(List(Term.Param(List(), Term.Name("e"), Some(Type.Name("T")), None))), Type.Name("T")))
```

notes/trees.md
```diff
@@ -0,0 +1,174 @@
...
+Term.Eta              | `foo _`
+Term.Repeated         | `foo(x: _*)`
+Term.Param            | `def foo<<(x: Int)>>`
+Term.Interpolate      | `s"|foo ~ $bar ~ ${42}|"`
+Term.Xml              | `<h1>|Hello ~ {bar} ~ {42}|</h1>`
+
+## Types (`meta.Type`)
+
+
+Ref            | Examples
```
olafurpg: Why is type refs above parse example? In terms it's reversed

notes/trees.md
```diff
@@ -0,0 +1,174 @@
...
+Type.Var              | ???
+Type.Param            | `trait A[X]`
+
+[Method]: https://www.scala-lang.org/files/archive/spec/2.12/03-types.html#method-types
+
+
+## Patterns (meta.Pat) and Cases (meta.Case)
+
+```scala
+"""(foo, bar)""".parse[Pat].get.structure
```
olafurpg: I would use a more obvious pattern like `_: Foo | _: Bar`

ShaneDelmore:
This is pretty wonderful, especially now that you can’t jump to the code in scalameta, only to the protobuf classes.  I don’t know that it needs to hold up this pr but if it was updated with the actual Params to the ADT members (so users can tell how to construct/deconstruct) it would become the canonical reference for users I think.

DavidDudson:
I frequently search in Tree's.scala for the apply/unapply methods for traits and case class constructors to see what parameters are.

Can we add what @ShaneDelmore suggested (parameter lists) and also potentially add the Apply/Unapply for some traits (Although this may deserve going into the tutorial and not this document).

The 2 trait apply/unapply methods are:
- Name
- Lit

MasseGuillaume:
This also might be helpful.

https://github.com/scalacenter/scala-syntax/blob/master/tests/unit/src/test/scala/org/scalafmt/tests/CoverageSuite.scala#L15

olafurpg:
> Can we add what @ShaneDelmore suggested (parameter lists) and also potentially add the Apply/Unapply for some traits (Although this may deserve going into the tutorial and not this document).

We can probably auto-generate that from the sources. I think it's a great idea but I see that complementing the document in this PR. 

@MasseGuillaume saw two rendering issues

screen shot 2018-03-07 at 12 56 56
screen shot 2018-03-07 at 12 57 15

olafurpg:
Closing due to inactivity. Please reopen once the document is more complete @MasseGuillaume ! I'm looking forward to provide this reference to users, I think it's a great idea 😄

PR #1393, opened 2018-03-06
Metacp cache

olafurpg:
While working with metacp and scalalib past week I found it awkward to convert classpaths into metacp classpaths. In a nutshell, roughly the following should be possible now

```
metap $(metacp $(coursier fetch scala-library))
```

Metacp can be considered as a `Classpath => Classpath` function. 

In more detail,  this PR cleans up the biggest pain points I experienced with metacp:

* provide high-level API methods to run metacp on the boot classpath (jdk) and scala-library-synthetics (Any/AnyVal/AnyRef/Nothing). 
* add Metacp.process(List[Path]): List[Path]. The idea is that tools like scalafix can accept a --classpath from end users and invoke Metacp.process on the provided arguments.
* remove scalalib and scalalib-generator projects
* compress output in jar files reducing output by ~6x.
* cache output from processed jars in global directory, avoiding reprocessing for example the jdk on every metacp invocation. Metacp is an application now, with global state.
* metap supports printing classpaths, so it's fine to pass in jars and directories now.


scalameta/testkit/src/main/scala/scala/meta/testkit/DiffAssertions.scala
```diff
@@ -47,7 +47,7 @@ trait DiffAssertions extends FunSuiteLike {
   private def compareContents(original: Seq[String], revised: Seq[String]): String = {
     import collection.JavaConverters._
     // NOTE: If we don't trim lines, AppVeyor gets really upset.a
-    def trim(lines: Seq[String]) = lines.map(_.trim).asJava
+    def trim(lines: Seq[String]) = lines.asJava
```
xeno-by: Maybe you have looked into why AppVeyor is unhappy about this?

langmeta/langmeta/jvm/src/main/scala/org/langmeta/internal/io/PlatformFileIO.scala
```diff
@@ -2,11 +2,14 @@ package org.langmeta.internal.io
...
 import java.net.URI
 import java.nio.charset.Charset
+import java.nio.file.FileSystem
 import java.nio.file.FileSystemAlreadyExistsException
 import java.nio.file.FileSystems
 import java.nio.file.Files
 import java.nio.file.Path
+import java.util
 import java.util.stream.Collectors
+
```
xeno-by: Let's avoid newlines between imports. Same for other files.

semanticdb/integration/src/main/java/com/javacp/Javacp.java
```diff
@@ -34,8 +34,8 @@ void primitiveParams(int a, long b, float c, double d, short e, byte f, boolean
     void typeParams(ArrayList<HashMap<A, String[]>> a, Hashtable<String, B> b) { }
     <C extends Integer> void methodTypeParams(C c) { }
 
-    public void overload(com.javacp.Javacp a) { }
-    public void overload(example.Javacp a) { }
+    public void overload(java.io.File a) { }
+    public void overload(sourcecode.File a) { }
```
xeno-by: This does not compile for me. Same for CI.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Settings.scala
```diff
@@ -1,31 +1,56 @@
...
+import org.langmeta.io.Classpath
+
+import scala.meta.cli.Metacp
+
 final case class Settings(
-    cps: List[String] = Nil,
-    d: String = "."
-)
+    out: PrintStream = System.out,
+    err: PrintStream = System.err,
```
xeno-by: If we need an abstraction for `out` and `err`, I suggest we define a `Reporter` that lives outside `Settings`. 

Things like `Settings.parse(args, out, err)` look weird enough to suggest that reporting is a concern that is unrelated to settings.

xeno-by: Whatever we decide here, I think we should do exactly the same for `metap` and `metac` for consistency reasons.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Settings.scala
```diff
@@ -1,31 +1,56 @@
...
+    classpath: Classpath = Classpath(Nil),
+    d: AbsolutePath = Metacp.defaultCacheDir,
+    directories: DirectoryOutput = DirectoryOutput.TempDirectory
+) {
+  def withClasspath(e: String) =
```
xeno-by: `withX` usually means "take existing value of X and replace it with the argument". Let's call this method `appendX` instead?

semanticdb/metap/src/main/scala/scala/meta/internal/metap/Main.scala
```diff
@@ -1,43 +1,100 @@
...
+            buf += entry
           }
-        } catch {
-          case NonFatal(ex) =>
-            out.println(s"error: can't decompile $arg")
-            ex.printStackTrace(out)
-            failed = true
         }
+        buf.sortBy(_.getName).foreach { entry =>
+          processPath(Paths.get(entry.getName), jarfile.getInputStream(entry))
```
xeno-by: Does `Paths.get(entry.getName)` include the underlying `path`?

semanticdb/metacp/src/main/scala/scala/meta/cli/Metacp.scala
```diff
@@ -1,25 +1,94 @@
...
       case None =>
         1
     }
   }
+
+  def process(settings: Settings): List[AbsolutePath] = {
+    settings.classpath.shallow.toParArray.map(p => processPath(p, settings)).toList
+  }
+
+  def processPath(in: AbsolutePath, settings: Settings): AbsolutePath = {
```
xeno-by: I think we have way too many public APIs here. Instead of `main/process/process/process/processPath`, I suggest that we have just two methods:
  * `def main(args: Array[String]): Unit` (this one we can't avoid)
  * `def process(settings: Settings, reporter: Reporter): Result` (where `Result` carries both an indication of success and the list of absolute paths).

Sure, this will be less convenient to clients, because they will have to manually construct and deconstruct the API structures to use `process`, but this way we keep the public API minimal.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Settings.scala
```diff
@@ -1,31 +1,56 @@
 package scala.meta.internal.metacp
 
+import java.io.PrintStream
+
+import org.langmeta.io.AbsolutePath
+import org.langmeta.io.Classpath
+
+import scala.meta.cli.Metacp
+
 final case class Settings(
```
xeno-by: This can no longer be an internal class if it's used in the public API.

semanticdb/metacp/src/main/scala/scala/meta/cli/Metacp.scala
```diff
@@ -1,25 +1,94 @@
...
+        PlatformFileIO.withJarFileSystem(out) { jar =>
+          val exit = Main.process(settings.copy(classpath = Classpath(in :: Nil), d = jar))
+          assert(exit == 0, s"Failed to process $in")
+        }
+      }
+      out
+    }
+  }
+
+  def scalaLibrarySynthetics(scalaVersion: String, settings: Settings): AbsolutePath = {
```
xeno-by: Do metacp clients need to invoke this method manually? Can we instead invoke it automatically whenever someone passes us scala-library.jar? This will further simplify the public API.

olafurpg: One problem with relying on finding a classpath entry containing the string "scala-library" is that it won't work for

* fat jar classpaths
* scala-library compiled from source

By exposing this method we delegate to the client to deal with these cases instead of trying to provide an automatic solution that fails in some cases.

semanticdb/metacp/src/main/scala/scala/meta/cli/Metacp.scala
```diff
@@ -1,25 +1,94 @@
...
+    val out = cacheFile(settings, "scala-library-synthetics")
+    if (!out.isDirectory) {
+      PlatformFileIO.withJarFileSystem(out) { jar =>
+        ScalaLibrarySynthetics.process(scalaVersion, jar)
+      }
+    }
+    out
+  }
+
+  def bootClasspath(settings: Settings): List[AbsolutePath] = {
```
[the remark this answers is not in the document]
olafurpg: Sure, I don't mind doing this from the client side in scalafix. I suspect every client will end up having to copy-paste this again and again however 😜

semanticdb/metacp/src/main/scala/scala/meta/cli/Metacp.scala
```diff
@@ -1,25 +1,94 @@
...
+            .iterator
+            .map(p => AbsolutePath(p))
+            .filter(_.isFile)
+            .toList
+      }
+      .getOrElse(sys.error("failed to detect JDK classpath"))
+    process(settings.copy(classpath = Classpath(classpath)))
+  }
+
+  def defaultCacheDir: AbsolutePath = AbsolutePath(
```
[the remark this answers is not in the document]
olafurpg: I was thinking that clients should be able to explicitly set the output directory to `Array("-d", defaultCacheDir.toString)`. The alternative is implicitly depending on the default value of `Settings.d` which is an internal API. I don't mind if this is moved under scala.meta.internal with no bincompat guarantees, but I'd like it to be at least publicly available.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -10,28 +9,28 @@ import scala.util.control.NonFatal
 import org.langmeta.internal.io._
 import org.langmeta.io._
 
-class Main(settings: Settings, out: PrintStream, err: PrintStream) {
```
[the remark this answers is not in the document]
olafurpg: Done.

semanticdb/metap/src/main/scala/scala/meta/internal/metap/Settings.scala
```diff
@@ -1,7 +1,11 @@
 package scala.meta.internal.metap
 
+import java.io.File
+import java.nio.file.Path
+import java.nio.file.Paths
+
 final case class Settings(
-    paths: List[String] = Nil,
+    paths: Vector[Path] = Vector.empty,
```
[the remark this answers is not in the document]
olafurpg: I'm still not convinced "always use List unless you have JMH benchmarks to prove it's a perf problem" is a good strategy. The collection should communicate what operations are expected to be done with it, `List` indicates prepend and iteration are fine. If you repeatedly do concats then Vector is the best default immutable collection type IMO.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Settings.scala
olafurpg: Good catch! Will fix.

semanticdb/metacp/src/main/scala/scala/meta/cli/Metacp.scala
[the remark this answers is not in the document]
xeno-by: Well, that one is on JDK, I think. If this becomes a widespread problem, we can revisit.

semanticdb/metacp/src/main/scala/scala/meta/cli/Metacp.scala
xeno-by: I've been thinking about looking into `META-INF/MANIFEST.MF`, which must say:

```
15:00 ~/Downloads$ cat META-INF/MANIFEST.MF
Manifest-Version: 1.0
Bnd-LastModified: 1507705065521
Bundle-ManifestVersion: 2
Bundle-Name: Scala Standard Library
Bundle-RequiredExecutionEnvironment: JavaSE-1.8
Bundle-SymbolicName: org.scala-lang.scala-library
Bundle-Version: 2.12.4.v20171010-220511-VFINAL-c2a5883
Created-By: 1.8.0_121 (Oracle Corporation)
...
```

This probably won't work well with far jars though.

xeno-by: If we don't do this automatically, how would generation of scala library synthetics work from the command line? Will `metap $(metacp $(coursier fetch scala-library))` include the synthetics?

olafurpg: Let's add --include-scala-library-synthetics  to metacp to optionally include scala library synthetics in the output or not.

semanticdb/metacp/src/main/scala/scala/meta/cli/Metacp.scala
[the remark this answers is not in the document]
xeno-by: `Settings.defaultCacheDir`?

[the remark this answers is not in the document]
olafurpg: Sounds good! `--cache-dir`

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Settings.scala
olafurpg: I agree that `withX` is not appropriate here, appendX would be better.

semanticdb/metap/src/main/scala/scala/meta/internal/metap/Main.scala
olafurpg: I haven't checked, this value is only used for reporting errors on failure.

semanticdb/metap/src/main/scala/scala/meta/internal/metap/Settings.scala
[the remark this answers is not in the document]
xeno-by: My motivation here is that Scala already provides lots of ways of doing the same thing in lots of areas (aka being unopinionated).

Personally, when writing Scala, I spend a lot of mental effort to consistently pick the same ways of doing the same things for those areas. I'm constantly on the lookout for techniques and tools that help to reduce this burden. Scalafmt helps a lot, which is a big plus - thank you for that.

If we look at collections, Scala gives lots of choice as usual. Even if we take only linear immutable collections, there's `Traversable`, `Iterable`, `Seq`, `IndexedSeq`, `LinearSeq`, `List`, `Vector`, `Array`, etc. This is another decision that I don't want to make, so I simply default to `List` everywhere.

I realize that in performance-critical areas, the choice between `List` or `Vector` can make a difference between success and failure, but luckily for my sanity these areas are very rare in our tools.

xeno-by:
Superseded by https://github.com/scalameta/scalameta/pull/1394.

semanticdb/metacp/src/main/scala/scala/meta/cli/Metacp.scala
xeno-by: Implemented in https://github.com/scalameta/scalameta/pull/1394.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Settings.scala
xeno-by: https://github.com/scalameta/scalameta/pull/1394

semanticdb/metacp/src/main/scala/scala/meta/cli/Metacp.scala
xeno-by: Implemented in https://github.com/scalameta/scalameta/pull/1394.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Settings.scala
xeno-by: No longer relevant.

semanticdb/metacp/src/main/scala/scala/meta/cli/Metacp.scala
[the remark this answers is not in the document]
xeno-by: Implemented in https://github.com/scalameta/scalameta/pull/1394.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Settings.scala
xeno-by: Implemented in https://github.com/scalameta/scalameta/pull/1394.

tests/jvm/src/test/scala/scala/meta/tests/metacp/MetacpCrashSuite.scala
```diff
@@ -1,8 +1,8 @@
 package scala.meta.tests.metacp
 
-// TODO(olafur): move to slow/integration tests. Each run takes ~30s on my laptop
-// and produces ~350mb of artifacts in a temporary directory.
-class MetacpSuite extends BaseMetacpSuite {
+// TODO(olafur): move to slow/integration tests. Each run takes ~20s on my laptop
+// and produces ~60mb of artifacts in a temporary directory.
```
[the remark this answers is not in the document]
xeno-by: Implemented in https://github.com/scalameta/scalameta/pull/1394.

PR #1394, opened 2018-03-07
Metacp cache, take two

xeno-by:
Build upon https://github.com/scalameta/scalameta/pull/1393 to address review feedback and add additional functionality.


olafurpg:
Fixed the Windows failure by sorting directory entries by lowercase Path.toString.

PR #1395, opened 2018-03-07
 First stab at semanticidx support for metac

xeno-by:


semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Index.scala
```diff
@@ -1,40 +0,0 @@
-package scala.meta.internal.metacp
-
-import org.langmeta.internal.io._
-import org.langmeta.io._
-import scala.collection.mutable
-import scala.meta.internal.{semanticdb3 => s}
-
-class Index {
```
olafurpg: What is supposed to replace this class? I was trying to rebase this PR on top of master but it fails compilation because this class is missing.

xeno-by: Oops. Looks like the file got accidentally gitignored on my machine. Fixed & pushed.

semanticdb/metacp/src/main/scala/scala/meta/internal/index/Index.scala
```diff
@@ -29,7 +28,7 @@ class Index {
   }
 
   def save(out: AbsolutePath): Unit = {
-    val indexAbspath = out.resolve("META-INF").resolve("semanticdb.semanticidx")
+    val indexAbspath = out.resolve("META-INF/semanticdb.semanticidx")
```
olafurpg: I've always wondered if this works correctly on windows, Appveyor doesn't seem to complain but it feels wrong to hardcode `/`

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -85,7 +86,8 @@ class Main(settings: Settings, reporter: Reporter) {
                 }
               }
               result.foreach { infos =>
-                index.append(infos)
+                val uri = infos.classfile.relative + ".semanticdb"
```
olafurpg: This is duplicated with the code above, why remove the method? Seems like a regression rather than improvement.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SemanticdbPipeline.scala
```diff
@@ -62,9 +62,38 @@ trait SemanticdbPipeline extends DatabaseOps { self: SemanticdbPlugin =>
...
       }
 
+      private def synchronizeSourcesAndSemanticdbFiles(): Unit = {
+        val vdb = v.Database.load(Classpath(scalametaTargetroot))
+        val orphanedVentries = vdb.entries.filter(ventry => {
+          val scalaName = v.SemanticdbPaths.toScala(ventry.fragment.name)
+          !config.sourceroot.resolve(scalaName).isFile
+        })
+        orphanedVentries.map(ve => {
+          def cleanupUpwards(file: File): Unit = {
```
olafurpg: Unrelated to this PR, but this logic has always bugged me. I think we should try to use idiomatic java.nio wherever possible. Here's how it's possible to delete a directory upwards

```scala
  Files.walkFileTree(
    Paths.get(""),
    new SimpleFileVisitor[Path] {
      override def visitFile(file: Path, attrs: BasicFileAttributes): FileVisitResult = {
        if (someLogic() { Files.delete(file) }
        FileVisitResult.CONTINUE
      }
      override def postVisitDirectory(dir: Path, exc: IOException): FileVisitResult = {
        val ds = Files.newDirectoryStream(dir)
        val isEmpty = !ds.iterator().hasNext
        ds.close()
        if (isEmpty) {
          Files.delete(dir)
        }
        FileVisitResult.CONTINUE
      }
    }
  )
```
Sure it's verbose, but by using conventional APIs for IO the code is more approachable for anyone dropping into the codebase.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/DocumentOps.scala
```diff
@@ -189,7 +193,44 @@ trait DocumentOps { self: DatabaseOps =>
               todo -= mtree
 
               names(mtree.pos) = symbol
-              if (mtree.isDefinition) binders += mtree.pos
+              if (mtree.isDefinition) {
+                val isToplevel = {
```
olafurpg: This logic does not make sense to me, would `isToplevel=true` for `x` in something like
```scala
object a {
  locally {
    class A {
      val x = 2 // toplevel=true ?
    }
  }
}
```

Would it be sufficient to check `gsym.owner.hasPackageFlag`? I don't recall the package index containing method symbols.

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/ExpectSuite.scala
```diff
@@ -156,6 +164,29 @@ trait ExpectHelpers extends FunSuiteLike {
...
+      buf.append("=========" + EOL)
+      index.packages.sortBy(_.symbol).foreach { entry =>
+        buf.append(entry.symbol + EOL)
+        entry.members.sorted.foreach(member => buf.append("  " + member + EOL))
+      }
+      buf.append(EOL)
+      buf.append("Toplevels:" + EOL)
+      buf.append("==========" + EOL)
+      index.toplevels.sortBy(_.symbol).foreach { entry =>
+        buf.append(s"${entry.symbol} => ${entry.uri}" + EOL)
```
olafurpg: Seems we need `PathIO.toUnix(entry.uri)` on the right hand side to please Appveyor.

semanticdb/metacp/src/main/scala/scala/meta/internal/index/Index.scala
xeno-by: Done.

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/ExpectSuite.scala
```diff
@@ -125,6 +117,23 @@ trait ExpectHelpers extends FunSuiteLike {
     sorted
   }
 
+  protected def normalizedSymbols(path: Path): Map[String, s.SymbolInformation] = {
```
olafurpg: FYI, I had this method in ExpectHelpers because I was in fact reusing it between two test suites at one point, but then I figured out how to make the other suite redundant.

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/ExpectSuite.scala
```diff
@@ -125,6 +117,23 @@ trait ExpectHelpers extends FunSuiteLike {
...
+    for {
+      file <- FileIO.listAllFilesRecursively(AbsolutePath(path)).iterator
+      if PathIO.extension(file.toNIO) == "semanticdb"
+      doc <- s.TextDocuments.parseFrom(file.readAllBytes).documents
+      sym <- doc.symbols
+    } yield {
+      val normalizedSym = sym.copy(
```
olafurpg: 👍

tests/jvm/src/test/resources/metac-metacp.index.diff
```diff
@@ -0,0 +1,89 @@
+--- metac
```
olafurpg: I'm not sure if my suggestion to diff metac.index and metacp.index was such a good idea after all. It's clearly expected that the output be the two will never be the same, neither top-levels or recorded packages. We can keep it for now, but we might want to consider removing it at some point.

ShaneDelmore: Is there any level of similarity that can be expected?  Is it worth doing a partial diff?

olafurpg: I don't think so, the differences are inherent and expected

* metac does not emit symbols for *.java sources
* metac supports excluding files
* the relative URIs point to X.scala.semanticdb files in metac but X.class.semanticdb in metacp.

Issue #1378, opened 2018-02-28
Synchronize javacp, metac and the spec


xeno-by:
* [ ] Update semanticdb-scalac to ensure that its references to Java definitions are compliant to the new specification
* [x] Add tests to ensure metacp/metac output is identical.  (pending fix in #1399 )
* [ ] Update the specification to account for recently developed javacp
  * [ ] Symbol (symbol format)
  * [ ] Type (which language types map to which SemanticDB types)
  * [ ] SymbolInformation (which language definitions map to which SemanticDB infos)
* [ ] Update the data model to harmonize the differences between Java and Scala
  * [ ] k.FIELD, k.LOCAL, p.VAL, p.VAR
  * [ ] k.METHOD
  * [ ] k.INTERFACE
  * [ ] t.WITH_TYPE, t.AND_TYPE
  * [ ] k.ENUM etc

xeno-by:
Superseded by:
  * https://github.com/scalameta/scalameta/issues/1411
  * https://github.com/scalameta/scalameta/issues/1404
  * https://github.com/scalameta/scalameta/issues/1405
  * https://github.com/scalameta/scalameta/issues/1406
  * https://github.com/scalameta/scalameta/issues/1407
  * https://github.com/scalameta/scalameta/issues/1408
  * https://github.com/scalameta/scalameta/issues/1412
  * https://github.com/scalameta/scalameta/issues/1162

PR #1399, opened 2018-03-07
Add new expect test to check diff between metacp and metac.

### LINKED ISSUE #1286: Test the diff between metac and metacp
"@olafurpg: I feel like we should test the diff between metacp and metac for symbols that are covered by metacp. In theory they should produce identical types making the diff empty but I suspect that may not be the case, if there are inconsistencies I want them documented in our tests."

olafurpg:
Fixes #1286. The goal is to make metac-metacp.diff empty. Currently, it seems the two
disagree on several fronts.

Towards #1378. The first step in syncronizing metacp and metac is to know where they differ.


olafurpg:
I have a pending change that improves this test quite a bit. Will push soon!

PR #1430, opened 2018-03-11
Centralize symbol construction and deconstruction as much as possible

xeno-by:


semanticdb/semanticdb3/src/main/scala/scala/meta/internal/semanticdb3/Scala.scala
```diff
@@ -0,0 +1,253 @@
...
+
+  object Names {
+    val None: String = ""
+    val RootPackage: String = "_root_"
+    val EmptyPackage: String = "_empty_"
+    val Constructor: String = "<init>"
+    val Anonymous: String = "_"
+  }
+
+  sealed trait TypeDescriptor {
```
olafurpg: Is this level of detail necessary? I personally prefer `Method(name: String, disambiguator: String)` in langmeta.Symbol. By making the descriptor a plain string instead of `TypeDescriptor` it seems we can avoid the `?` simplification (keeping the more readable `Int*`/`=>Int`/`{}`/...).

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/TargetedSuite.scala
```diff
@@ -611,7 +611,7 @@ class TargetedSuite extends DatabaseSuite(SemanticdbMode.Slim) {
        |  [38..39): * => _star_.
        |[17..17): *.apply[Int]
        |  [0..1): * => _star_.
-       |  [2..7): apply => scala.collection.immutable.List.apply(A*).
+       |  [2..7): apply => scala.collection.immutable.List.apply(?).
```
olafurpg: This is a good example where `?` hurts readability

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
```diff
@@ -368,22 +367,32 @@ object Javacp {
...
     }
-    throw new IllegalArgumentException(s"Missing $$ or / from symbol '$asmName'")
+    asmName
   }
 
-  private def ssym(asmName: String): String =
-    asmName.replace('$', '#').replace('/', '.') + "#"
+  private def ssym(asmName: String): String = {
+    var fqName = asmName.replace('$', '#').replace('/', '.')
+    locally {
```
olafurpg: why locally?

olafurpg: Can we at least guard against `if (fqName.endsWith("package-info") || fqName.endsWith("#."))`?

Sure, it's premature optimization but

* this method is VERY hot
* `replace(String, String)` is not cheap

I already felt bad enough doing to `replace(Char, Char)` twice instead of rolling a while loop 😅

semanticdb/semanticdb3/src/main/scala/scala/meta/internal/semanticdb3/Scala.scala
```diff
@@ -0,0 +1,253 @@
...
+    def Local(symbol: String): String = symbol
+  }
+
+  implicit class ScalaSymbolOps(symbol: String) {
+    def isNone: Boolean = symbol == Symbols.None
+    def isRootPackage: Boolean = symbol == Symbols.RootPackage
+    def isEmptyPackage: Boolean = symbol == Symbols.EmptyPackage
+    def isGlobal: Boolean = !isNone && Descriptor.descriptorLasts.contains(symbol.last)
+    def isLocal: Boolean = !isNone && !isGlobal
+    def isValid: Boolean = ??? // TODO: If we find ourselves needing this, we'll need a parser.
```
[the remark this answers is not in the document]
xeno-by: Sure. Just wanted to draw your attention to the bit in comments.

semanticdb/semanticdb3/src/main/scala/scala/meta/internal/semanticdb3/Scala.scala
```diff
@@ -0,0 +1,253 @@
...
+    def ownerChain: List[String] = if (!owner.isNone) owner.ownerChain :+ symbol else List(symbol)
```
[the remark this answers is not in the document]
xeno-by: Done.

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
xeno-by: I'll just reimplement this method as a while loop.

semanticdb/semanticdb3/src/main/scala/scala/meta/internal/semanticdb3/Scala.scala
xeno-by: The old scheme also exposes quite a lot of string manipulation and magic strings to clients of the API. But I agree that the new scheme doesn't look so pretty. I'll revert and we'll think about this later.

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/TargetedSuite.scala
xeno-by: See above.

PR #1434, opened 2018-03-12
Add a dependency on pprint.

olafurpg:
pprint is indispensable for my productivity, it's a small module that
deserves inclusion in the standard library. This commit adds an explicit
dependency on pprint because I'm tired of adding this dependency in my
private local setup causing compilation errors when other people try my
WIP branches.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -206,7 +206,8 @@ lazy val langmeta = crossProject(JSPlatform, JVMPlatform)
   .in(file("langmeta/langmeta"))
   .settings(
     publishableSettings,
-    description := "Langmeta umbrella module that includes all public APIs"
+    description := "Langmeta umbrella module that includes all public APIs",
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -224,7 +224,10 @@ lazy val common = crossProject(JSPlatform, JVMPlatform)
   .in(file("scalameta/common"))
   .settings(
     publishableSettings,
-    libraryDependencies += "com.lihaoyi" %% "sourcecode" % "0.1.3",
+    libraryDependencies ++= List(
+      "com.lihaoyi" %% "sourcecode" % "0.1.3",
+      "com.lihaoyi" %% "pprint" % "0.5.3"
+    ),
```
xeno-by: We can make these dependencies even more ubiquitous by adding them to `langmeta`. I think that `semanticdb3` would be an overkill though.

olafurpg: Done.

PR #1450, opened 2018-03-22
Add tests to validate symbol parser behavior for _root_. prefix.

### LINKED ISSUE #1424: Temporarily support _root_. as a valid prefix for Scala symbols
Context: https://github.com/scalameta/scalameta/pull/1422#discussion_r173620804

olafurpg:
While working on fixing #1424 I discovered that the symbol parser
already works as I expected. A `_root_.` prefix is discarded.  This
commit introduces no functional changes to the symbol parser, just adds
test to validate the current existing behavior. Fixes #1424


olafurpg:
Turns out that the parser already works as expected!

Issue #1262, opened 2018-02-05
Improve Type


xeno-by:
https://github.com/scalameta/scalameta/pull/1259#discussion_r166001942

xeno-by:
I'll be working on this ticket this week. The overhead of dealing with the current implementation of `Type` is becoming unbearable.

olafurpg:
I had thought about this and have some ideas

* all suptypes of `Type` use `.extends`

```
  option (scalapb.message).extends = "scala.meta.semanticdb3.Type"
```


* `Type` is not defined in protobuf, it's hand written in scala and has a custom `TypeMapper` as documented here https://scalapb.github.io/customizations.html where

olafurpg:
I am working on this. It seems that the best option is to extend ScalaPB to generate more idiomatic Scala code for special cases of `oneof` that have the following shape
```proto
message Lit {
  int32 value = 1
}
message Add {
  Expr lhs = 1;
  Expr rhs = 2;
}
message Expr {
  oneof value {
    Lit lit = 1;
    Add add = 2;
  }
}
```

The message `Expr` will be detected as the supertype of a sealed ADT and generate roughly the following code
```scala
case class Lit(value: Int) extends Expr { ... }
case class Add(lhs: Expr, rhs: Expr) extends Expr { ... }
sealed trait Expr {
  def asExprMessage: Expr.Message
}
object Expr {
  case object Empty extends Expr { ... }
  class Message(
      underlying: scalapb.Message[Expr] with scalapb.GeneratedMessage,
      tag: Int
  ) extends scalapb.Message[Expr] with scalapb.GeneratedMessage
}
```

I will open a discussion in ScalaPB to see if this encoding makes sense and could be contributed upstream. It's possible to write custom scalapb code generators, but I prefer to use the standard ScalaPB encoding.

olafurpg:
Proposal to add "sealed oneof" option to ScalaPB https://github.com/scalapb/ScalaPB/issues/455

olafurpg:
Current status: https://github.com/scalameta/scalameta/blob/74e7c070ae90363857b7abe576ed9f6369b2f6de/semanticdb/semanticdb3/src/main/generated/scala/meta/internal/semanticdb3/TypeMessage.scala

with this protobuf encoding for `Type`
```proto
message Type {
 //   reserved 1, 3, 4, 5, 6;
  oneof sealed_value {
    TypeRef typeRef = 2;
    SingletonType singletonType = 16;
[...]

PR #1456, opened 2018-03-23
Fix #1442

### LINKED ISSUE #1442: Javacp does not produce RepeatedType for Java vargargs
```diff
- obtained
+ expected
class Foo {
  void main(String... args) {}
}
- _empty_.Foo#main(Array). => private[_empty_] method main: (param0: Array[String]): Unit
+ _empty_.Foo#main(String*). => private[_empty_] method main: (param0: String*): Unit
  param0 => _empty_.Foo#main(Array).(param0)
-  Array => scala.Array#
  String => java.lang.String#
  Unit => scala.Unit#
```

This bug causes method symbols to be inconsistent between metac and metacp. It's possible to detect varargs with ASM through the o.ACC_VARARGS `access` flag

gabro:
This PR fixes #1442


semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
```diff
@@ -194,18 +195,29 @@ object Javacp {
             method.signature.params
           }
 
-        val parameterSymbols = params.zipWithIndex.map {
+        val parameterSymbols: List[String] = params.zipWithIndex.map {
```
gabro: for some reason, removing this annotation causes this error


```
[error] /Users/gabro/buildo/scalameta/semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala:231: recursive value parameterSymbols needs type
[error]               parameters = s.MethodType.ParameterList(parameterSymbols) :: Nil,

File: semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala [hunk 2]
```diff
@@ -194,18 +195,29 @@ object Javacp {
...
               else method.node.parameters.get(i).name
             }
             val paramSymbol = Symbols.Global(methodSymbol, d.Parameter(paramName))
+            val isRepeatedType = method.node.access.hasFlag(o.ACC_VARARGS) && i == params.length - 1
+            val tpe =
+              if (isRepeatedType)
+                s.Type(
+                  tag = t.REPEATED_TYPE,
+                  repeatedType = Some(s.RepeatedType(
+                    param.toType(methodScope).typeRef.flatMap(_.typeArguments.headOption)
```
gabro: this is probably not the most elegant way to achieve this.
I'm essentially converting the type to an `Array[T]`, and then accessing `T`.

Can this be improved?

tests/jvm/src/test/resources/metacp.expect
```diff
@@ -503,6 +503,16 @@ com.javacp.Test#typeParams(ArrayList,Hashtable).(b) => param b: Hashtable[String
...
+com.javacp.Test#vararg(Int,Array). => private[javacp] method vararg: (a: Int, args: String*): Unit
+a => com.javacp.Test#vararg(Int,Array).(a)
+Int => scala.Int#
+args => com.javacp.Test#vararg(Int,Array).(args)
+String => java.lang.String#
+Unit => scala.Unit#
+com.javacp.Test#vararg(Int,Array).(a) => param a: Int
+Int => scala.Int#
+com.javacp.Test#vararg(Int,Array).(args) => param args: String*
+String => java.lang.String#
```
xeno-by: If one compares these `foo => bar` lines with other lines nearby, they look misaligned. @olafurpg and I typically use `sbt save-expect` to update expect files automatically to avoid this kind of mishaps.

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
```diff
@@ -194,18 +195,29 @@ object Javacp {
...
 
-        val parameterSymbols = params.zipWithIndex.map {
+        val parameterSymbols: List[String] = params.zipWithIndex.map {
           case (param: JavaTypeSignature, i) =>
             val paramName = {
               if (method.node.parameters == null) "param" + i
               else method.node.parameters.get(i).name
             }
             val paramSymbol = Symbols.Global(methodSymbol, d.Parameter(paramName))
+            val isRepeatedType = method.node.access.hasFlag(o.ACC_VARARGS) && i == params.length - 1
```
gabro: This was hard to find, but the `ACC_VARARGS` flag applies to the whole method.

A parameter is then a vararg if it's the last of a method with `ACC_VARARGS`.

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
xeno-by: s.Type in general is quite ugly at this point. We have https://github.com/scalameta/scalameta/issues/1262 to track that, but I don't feel we have enough experience with s.Type to design a reasonable API for it. Hence, we're just biting the bullet and writing verbose code for the time being.

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
xeno-by: Nice find! SemanticDB exists so that tool authors don't have to bother themselves with stuff like that.

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
gabro: 👍 I'll leave it as it is then

tests/jvm/src/test/resources/metacp.expect
gabro: nice trick, the latest commit should fix it

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
```diff
@@ -194,18 +195,31 @@ object Javacp {
...
+            val paramTpe =
+              if (isRepeatedType) {
+                val tpe = param.toType(methodScope)
+                require(
+                  tpe.typeRef.isDefined && tpe.typeRef.get.symbol == "scala.Array#",
+                  s"expected $paramName to be a scala.Array#, found $tpe"
+                )
```
gabro: I've introduced this specific `require` since this should always be an `Array`.

tests/jvm/src/test/resources/metacp.expect
```diff
@@ -503,6 +503,16 @@ com.javacp.Test#typeParams(ArrayList,Hashtable).(b) => param b: Hashtable[String
   Hashtable => java.util.Hashtable#
   String => java.lang.String#
   B => com.javacp.Test#[B]
+com.javacp.Test#vararg(Int,Array). => private[javacp] method vararg: (a: Int, args: String*): Unit
```
xeno-by: Oops, wait! Why does the symbol say `(Int,Array)` instead of `(Int,String*)`?

gabro: Ah, good question, the original issue does the same so I followed that example, but I'm still getting acquainted to the concepts so I wasn't sure

xeno-by: I've updated the ticket with the expected output. According to the spec (https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md#scala-symbol, see the "Type descriptor is" part), repeated types in parameters should be reflected in the corresponding symbol, too.

olafurpg: Yup the expected type descriptor in my original ticket was wrong 😅

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
olafurpg: Making vararg a property of the method makes sense to me semantically. One awkward artifact of s.ByNameType is that it's possible to encode invalid states
```scala
def foo(a: *A, b: *B)
```
On the other hand, by-name types makes sense for syntax trees where Defn.Def has no attributes/modifiers to encode varargs 🤔

tests/jvm/src/test/resources/metacp.expect
gabro: Makes sense, I'll try to fix this too!

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
```diff
@@ -145,12 +145,16 @@ object Javacp {
         new MethodSignatureVisitor
       )
       val typeDescriptor = {
-        val paramTypeDescriptors = signature.params.map {
+        val hasVarArg = method.access.hasFlag(o.ACC_VARARGS)
+        def toTypeDescriptor(t: JavaTypeSignature, i: Int): String = t match {
```
gabro: had to make this recursive in order to print the `javaTypeSignature` of the Array type parameter.

gabro:
@xeno-by it should be ok now!

xeno-by:
@gabro Sorry for lagging behind. This weekend I was busy with Rsc, but I'll do my best to review on Monday or Tuesday (PDT).

olafurpg:
Woot, nice fix @gabro !

PR #1458, opened 2018-03-26
Fix #1392

### LINKED ISSUE #1392: Wrong type for scala.package.Either
```scala
$  metacp -cp $(coursier fetch org.scala-lang:scala-library:2.12.4)
$ metap -proto META-INF/semanticdb/scala/package.class.semanticdb | grep "symbol: .*Either()." -A 40
    symbol: "_root_.scala.package.Either()."
    kind: GETTER
    properties: 0
    name: "Either"
    tpe {
      tag: METHOD_TYPE
      methodType {
        return_type {
          tag: SINGLETON_TYPE
          singletonType {
            tag: SYMBOL
            prefix {
              tag: SINGLETON_TYPE
              singletonType {
                tag: SYMBOL
                prefix {
                  tag: SINGLETON_TYPE
                  singletonType {
                    tag: SYMBOL
                    symbol: "_root_.scala#"
                    primitive: 0
                    string: ""
                  }
                }
                symbol: "_root_.scala.util#"
                primitive: 0
                string: ""
              }
            }
            symbol: "_root_.scala.util.Either#"
            primitive: 0
            string: ""
          }
        }
      }
    }
```
Expected
```scala

$ cat foo.scala
package object a {
  val b = scala.util.Either
}
$ scalac foo.scala
$ metap -proto META-INF/semanticdb/a.class.semanticdb
...
symbol: "_empty_.a.b()."
    kind: GETTER
    properties: 0
    name: "b"
    tpe {
      tag: METHOD_TYPE
      methodType {
        return_type {
          tag: SINGLETON_TYPE
          singletonType {
            tag: SYMBOL
            symbol: "_root_.scala.util.Either."
            primitive: 0
            string: ""
          }
        }
      }
    }
```

xeno-by:


semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
```diff
@@ -194,7 +194,15 @@ object Scalacp {
...
             val spre = if (tpe.hasNontrivialPrefix) loop(pre) else None
-            val ssym = ssymbol(sym)
+            val ssym = {
+              // NOTE: Due to some unclear reason, Scalac sometimes saves
+              // (or Scala sometimes loads) single types that point to
+              // companion classes, not module classes. We assume that it's
+              // a mistake and work around accordingly.
+              val raw = ssymbol(sym)
+              if (raw.endsWith("#")) raw.stripSuffix("#") + "."
+              else raw
```
gabro: glorious fix

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
```diff
@@ -194,7 +194,15 @@ object Scalacp {
           val stpe = {
             val stag = st.SYMBOL
             val spre = if (tpe.hasNontrivialPrefix) loop(pre) else None
-            val ssym = ssymbol(sym)
+            val ssym = {
+              // NOTE: Due to some unclear reason, Scalac sometimes saves
+              // (or Scala sometimes loads) single types that point to
```
gabro: you mean "Scalap sometimes"?

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
```diff
@@ -194,7 +194,15 @@ object Scalacp {
           val stpe = {
             val stag = st.SYMBOL
             val spre = if (tpe.hasNontrivialPrefix) loop(pre) else None
```
gabro: should we also take care of `hasNonTrivialPrefix`? For the same corner case it currently returns `true` instead of `false`, because the `isObject` test fails.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
```diff
@@ -194,7 +194,15 @@ object Scalacp {
...
-            val ssym = ssymbol(sym)
+            val ssym = {
+              // NOTE: Due to some unclear reason, Scalac sometimes saves
```
gabro: I would link to #1392 in the comment for extra context

tests/jvm/src/test/resources/metac-metacp.expect.diff
```diff
@@ -1076,7 +1076,7 @@ types.Test.C#singleType2.
 +++ metacp
          string: ""
 -    symbol: "types.P#x()."
-+    symbol: "types.P#x#"
++    symbol: "types.P#x."
```
olafurpg: 👍 

This passed our scrutiny, `x#` makes no sense

semanticdb/integration/src/main/scala/example/Types.scala
```diff
@@ -43,6 +43,7 @@ object Test {
 
     val singleType1: x.type = ???
     val singleType2: p.x.type = ???
+    val Either = scala.util.Either
```
olafurpg: For future reference, we should put test cases like this in a separate `Bugs.scala` file (name up for debate) where we only append new code. Adding a new line to Types.scala shifts all positions in the following statements which causes unnecessary diffs making it harder to review the changes.

gabro: I agree: it's also probably interesting to keep track of this bugs (and respective workaround) in case they get fixed upstream

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
xeno-by: Do you have a test case in mind? Either is now looking good:

```
  symbols {
    symbol: "_empty_.C#Either()."
    kind: METHOD
    properties: 0
    name: "Either"
    tpe {
      tag: METHOD_TYPE
      methodType {
        return_type {
          tag: SINGLETON_TYPE
          singletonType {
            tag: SYMBOL
            symbol: "scala.util.Either."
            primitive: 0
            string: ""
          }
        }
      }
    }
```

gabro: ah, strange. Given the change, I would expect the prefix to be still there. 👍 then

PR #1469, opened 2018-04-03
Fail when parsing multi symbols

### LINKED ISSUE #1468: Symbol parser discards anything after semicolon
```
@ import $ivy.`org.scalameta:testkit_2.12:3.7.0`, scala.meta._, testkit._
import $ivy.$                                 , scala.meta._, testkit._

@ Symbol("_root_.Buz.;_root_.Foo.a#")
res1: Symbol = Global(Global(, Term("_root_")), Term("Buz"))

```

I expected the parser either to fail because it did not consume the full symbol or parse into `Symbol.Multi`.

olafurpg:
Fixes #1468

### DIFF
--- a/langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Symbol.scala
+++ b/langmeta/langmeta/shared/src/main/scala/org/langmeta/semanticdb/Symbol.scala
@@ -42,10 +42,9 @@ object Symbol {
   def apply(s: String): Symbol = {
     object naiveParser {
       var i = 0
-      def fail() = {
-        val message = "invalid symbol format"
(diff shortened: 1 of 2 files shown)


olafurpg:
This opens the question whether we should deprecate Symbol.Multi given that it's not possible to parse or print them.

xeno-by:
@olafurpg I don't think we can deprecate multi symbols, because we're using them internally to declare aliases. Deprecation warnings from those usages translate to compilation errors because of `-Xfatal-warning`.

olafurpg:
In that case, why do we throw errors on `Symbol.Multi.syntax`? Either multi symbols should be deprecated of we should support printing/parsing them.

PR #1471, opened 2018-04-04
Fix an oversight in newJarFileSystem

xeno-by:
Files.createDirectories doesn't throw if a parent exists and is a dir,
but it does throw if a parent exists and is a symlink.


olafurpg:
Build failure seems to be flake, restarted.

olafurpg:
The build failure continues to cause problems despite restart. I cleaned the build cache and now the error seems to be gone. The error was about permission problems reading downloaded/cached jars so keep an eye out for similar problems in the future.

PR #1474, opened 2018-04-05
Use `.par.foreach` and introduce metacp --par option.

### LINKED ISSUE #1398: Investigate making metacp parallel
In 1cd839e462a7cf9ac1414d9dc553d2e220d44e86 we removed `.par` from metacp to prevent a deadlock. We should consider making metacp parallel again because it's processing steps are trivially parallelizable and in my brief experiments processing spark-sql (110mb of jars) metacp was ~2-3x faster with parallel collections on my machine.

olafurpg:
This commit replaces the usage of `.par.flatMap` with `.par.foreach` as
an attempt to prevent a difficult-to-reproduce deadlock problem in
parallel collections.  The parallel behavior is disabled by default and
can be enabled with the --par option.

Using --par seems to yield ~40-50% faster performance in
MetacpCrashSuite on my Macbook. It takes ~36s to run with --par and ~57s
without --par.

The theory for why `.par.foreach` is better than `.par.flatMap` is based
on the experience in scalafmt/scalafix where I've used that same pattern
for over two years and have not seen reports of deadlock behavior.
I have never used `.par.flatMap` in my applications.

Fixes #1398

### DIFF
--- a/semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
+++ b/semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
@@ -2,6 +2,7 @@ package scala.meta.internal.metacp
 
 import java.nio.file._
 import java.nio.file.attribute.BasicFileAttributes
+import java.util.concurrent.ConcurrentLinkedQueue
 import scala.collection.JavaConverters._
 import scala.meta.internal.index._
 import scala.meta.internal.javacp._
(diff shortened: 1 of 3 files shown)


semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -11,55 +12,77 @@ import scala.util.control.NonFatal
...
       }
     }
-    val synthetics = {
+
+    val synthetics: Option[AbsolutePath] = {
       if (settings.scalaLibrarySynthetics) {
         val cacheEntry = settings.cacheDir.resolve("scala-library-synthetics.jar")
         if (cacheEntry.toFile.exists) {
-          List(cacheEntry)
+          Some(cacheEntry)
```
xeno-by: `buffer.add(cacheEntry)`? This will allow to remove the need to declare `synthetics` as well as the pattern matching bit on lines 74-78.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -11,55 +12,77 @@ import scala.util.control.NonFatal
...
+      else settings.classpath.shallow
+
+    val buffer = new ConcurrentLinkedQueue[AbsolutePath]()
+
+    classpath.foreach { entry =>
+      val processedEntry: Option[AbsolutePath] = if (entry.isDirectory) {
+        val out = AbsolutePath(Files.createTempDirectory("semanticdb"))
+        val res = convertClasspathEntry(entry, out)
+        success.compareAndSet(true, res)
+        Some(out)
```
xeno-by: `buffer.add(out)`? This will allow to remove the need to declare `processedEntry` as well as the pattern matching bit on lines 51-54.

tests/jvm/src/test/scala/scala/meta/tests/metacp/BaseMetacpSuite.scala
```diff
@@ -17,7 +14,7 @@ abstract class BaseMetacpSuite extends BaseCliSuite {
 
   def checkMetacp(name: String, classpath: () => Classpath): Unit = {
     test(name) {
-      val settings = Settings().withClasspath(classpath()).withCacheDir(tmp)
+      val settings = Settings().withClasspath(classpath()).withCacheDir(tmp).withPar(true)
```
xeno-by: I don't mind this bit. If our tests deadlock, that's fine.

PR #1475, opened 2018-04-05
Don't close nio.FileSystem.

olafurpg:
While testing metacp in scalafix I encountered frequent flaky test
failures due to "writing to closed file system" errors caused by
`FileSystems.getFileSystem` reusing open file systems.  This leaks
resources, but I see no alternative that does not involve refactoring
everything to java.io or global mutable state for reference counting
open file systems per zip file.


olafurpg:
> Am I right in thinking that an alternative is to get rid of the unified handling of paths provided by nio and use JarFile/ZipFile instead.

Correct, after more adventures with nio FileSystem (tests failing in 2.11 but not 2.12) I give up and think the best long-term solution is to stick to JarFile and friends.

PR #1476, opened 2018-04-05
DO NOT MERGE Test metacp with default settings on windows

olafurpg:


olafurpg:
Hmm, the CI passes on windows. I was expecting it to fail with URI error.

PR #1483, opened 2018-04-09
Make metacp write to temporary files during processing.

olafurpg:
Once processing completes, the jar files are atomically moved to the
destination cache jar. This prevents concurrent clients from reading
half-processed cached jars.

### DIFF
--- a/semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
+++ b/semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
@@ -24,6 +24,36 @@ class Main(settings: Settings, reporter: Reporter) {
 
     val buffer = new ConcurrentLinkedQueue[AbsolutePath]()
 
+    if (!Files.exists(Files.createDirectories(settings.cacheDir.toNIO))) {
+      Files.createDirectories(settings.cacheDir.toNIO)
+    }
+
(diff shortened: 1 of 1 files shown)


semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -24,6 +24,36 @@ class Main(settings: Settings, reporter: Reporter) {
...
+      }
+      try {
+        Files.move(
+          tmp,
+          cacheEntry.toNIO,
+          StandardCopyOption.ATOMIC_MOVE,
+          StandardCopyOption.REPLACE_EXISTING
+        )
+      } catch {
+        case _: AtomicMoveNotSupportedException =>
```
xeno-by: Shall we crash instead?

olafurpg: We have no guarantee on which systems ATOMIC_MOVE is not supported, but I would prefer to run successfully on those systems instead of making it impossible to use metacp. Errors from reading half-complete symbols will still be reported.

olafurpg: btw, using `move` even without atomic significantly reduces the chances or reproducing the error compared to the previous behavior where we write directly to the target jar while processing.

PR #1489, opened 2018-04-11
Exclude parens tokens in Term.ApplyInfix.args and Term.Select.qual (fix #1083)

### LINKED ISSUE #1083: Wrong position for parens wrapped identifier
```scala
@ """val x = (a).value""".parse[Stat].get.collect {
    case n: Name => n.syntax
  }
res1: List[String] = List("x", "(a)", "value") // obtained
res1: List[String] = List("x", "a", "value") // expected
```

This causes problems when looking up symbols of names by their positions.

MasseGuillaume:


MasseGuillaume:
Weird this test passes locally

`ErrorSuite `
`test("...$ in Term.ApplyInfix") {`

https://travis-ci.org/scalameta/scalameta/jobs/365068847#L1140

tests/shared/src/test/scala/scala/meta/tests/parsers/PositionSuite.scala
```diff
@@ -29,7 +29,7 @@ class PositionSuite extends ParseSuite {
     val tree = term("(1 + 2).foo")
     assert(tree.show[Positions] === """
       |Term.Select{0..11}(
-        |Term.ApplyInfix{0..7}(
+        |Term.ApplyInfix{1..6}(
```
olafurpg: I'm not sure if this is good expected output since the parentheses are required here. The motivation to open #1083 was that redundant parentheses are included in the position for example in `(a).b`. In `(1 + 2).foo` the parens are required to construct that Term.ApplyInfix. I had imagined that the fix for this ticket would involve adapting `def termName` constructor for Term.Name nodes.

Any thoughts @xeno-by ?

MasseGuillaume: @olafurpg the parenthesis would not be necessary for the `Term.ApplyInfix` if we move it outside of the `Term.Select`.

olafurpg: This PR handles infix applications, but what about other nodes?

```scala
@ val Term.Select(qual, _) = "(x match { case _ => 2 }).b".parse[Term].get
qqual: Term = Term.Match(Term.Name("x"), List(Case(_, None, Lit.Int(2))))

@ qual.tokens.syntax
res3: String = "(x match { case _ => 2 })"
```

xeno-by:
What's the expected behavior that this PR intends to achieve? I.e. could you guys exhaustively list the cases that involve parentheses right now and highlight the desired changes to positions?

MasseGuillaume:
@xeno-by 

Good idea, I started the work here:
https://github.com/MasseGuillaume/scalameta/blob/f481dfb032d16f34fcee41f12f79c13c93bcde4c/tests/shared/src/test/scala/scala/meta/tests/tokenizers/TokenizerCoverageSuite.scala

sub

I found out that the parser is a bit inconsistent with regards to parenthesis.

olafurpg:
#1498 was just merged, which should unblock changes in this PR

olafurpg:
Please reopen when you have time again to look into this @MasseGuillaume !

PR #1498, opened 2018-04-13
Formalize how trees are parsed with respect of their children

MasseGuillaume:
This is my reply to the comment about exhaustivity of parenthesis parsing: https://github.com/scalameta/scalameta/pull/1489#issuecomment-380499309.


tests/shared/src/test/scala/scala/meta/tests/tokenizers/TokenizerCoverageSuite.scala
```diff
@@ -0,0 +1,60 @@
...
+  check[Term.Apply]("→(f)←(→{ case a => a }←)")
+  check[Term.Apply]("→(f)←(→{ x }←)")
+  check[Term.ApplyInfix]("→(a)← →op← →(b)←")
+  check[Term.ApplyInfix]("→(a, b)← →op← (→c←, →d←)")
+  check[Term.ApplyInfix]("→(a, b)← →op← →((c, d))←")
+  check[Term.ApplyInfix]("→1← →+← →1←")
+  check[Term.ApplyInfix]("→a← →f← →(b)←")
+  check[Term.ApplyType]("→(f)← [→A←,→B←]")
+  check[Term.ApplyType]("→(f)← [→A←]")
+  check[Term.ApplyUnary]("→-← →(a)←")
```
olafurpg: Is this expected that the parentheses belong to `Term.Name`?

olafurpg: There seems to be several related places where I would expect redundant parentheses to belong to the parent node.

MasseGuillaume: @olafurpg Right now, my focus is to describe the current state. I will do a followup with the positions we expect.

This is the pattern I saw in the parser for lists:

* if there is a single element, it will capture the parenthesis (ex: `a f (b)`).
* If there is no element, it will not capture the parenthesis (ex `a f ()`)
* if there is more than one element, it will not capture the parenthesis (ex `a f (b, c)`)

For structures (like if, while, etc) it will never capture parenthesis when required:

* `if (p) t` gives `p`
* `return (a)` gives `(a)`

one exception is classes:

* `class A(b: B)` gives `(b: B)`

tests/shared/src/test/scala/scala/meta/tests/tokenizers/TokenizerCoverageSuite.scala
```diff
@@ -0,0 +1,114 @@
...
+  check[Term.Ascribe]("→(a)←: →(A)←")
+  check[Term.Assign]("→(a)← = →(b)←")
+  check[Term.Block]("{ →(a)←; →(b)← }")
```
MasseGuillaume: this does not capture the semicolon

tests/shared/src/test/scala/scala/meta/tests/tokenizers/TokenizerCoverageSuite.scala
```diff
@@ -0,0 +1,114 @@
...
+  check[Term.If]("if (→p←) →(t)← else →(f)←")
+  check[Term.If]("if (→p←) →(t)←")
+  check[Term.If]("if (→p←) →if (p2) t←")
+  check[Term.If]("if (→p←) →{}←")
+  check[Term.Interpolate](""" →s←"→start ←${→(a)←}→ end←" """)
+  check[Term.Match]("→(a)← match { →case x => x← }")
+  check[Term.New]("new →(A)←")
+  check[Term.NewAnonymous]("new →(A){}←")
+  check[Term.Param, Decl.Def]("def f(→a←: →A← = →(da)←): A")
+  check[Term.PartialFunction]("{ →case x => x;← →case y => y← }")
```
MasseGuillaume: this does capture the semicolon (why ?)

tests/shared/src/test/scala/scala/meta/tests/tokenizers/TokenizerCoverageSuite.scala
```diff
@@ -0,0 +1,114 @@
...
+  check[Term.Repeated, Term.Apply]("f(→(x)←: _*)")
+  check[Term.Return]("return →(a)←")
+  check[Term.Select]("→(a)←.→b←")
+  check[Term.Super]("→a.super[B]←.→c←")
+  check[Term.This]("→a←.this")
+  check[Term.Throw]("throw →(e)←")
+  check[Term.Try]("try (→f←) catch { →case x => x;← →case y => y← } finally →{ }←")
```
MasseGuillaume: semicolon ?

tests/shared/src/test/scala/scala/meta/tests/tokenizers/TokenizerCoverageSuite.scala
xeno-by: Semicolon is considered to be a statement separator, just like newline. I think it makes sense for it to not be part of any statements, just like newline.

tests/shared/src/test/scala/scala/meta/tests/tokenizers/TokenizerCoverageSuite.scala
```diff
@@ -0,0 +1,114 @@
...
+  checkSelf[Mod.Override, Defn.Def]("→override← def f = 1")
+  checkSelf[Mod.Case, Defn.Object]("→case← object B")
+  checkSelf[Mod.Abstract, Defn.Class]("→abstract← class A")
+  checkSelf[Mod.Covariant, Defn.Class]("class A[→+← T]")
+  checkSelf[Mod.Contravariant, Defn.Class]("class A[→-← T]")
+  checkSelf[Mod.Lazy, Defn.Val]("→lazy← val a = 1")
+  checkSelf[Mod.ValParam, Defn.Class]("class A(→val← b: B)")
+  checkSelf[Mod.VarParam, Defn.Class]("class A(→var← b: B)")
+  // check[Mod.Inline, Defn.Def]("→inline← def f = 1", dotty)
+}
```
xeno-by: Very very good. Super helpful visualization 😍

tests/shared/src/test/scala/scala/meta/tests/tokenizers/TokenizerCoverageSuite.scala
olafurpg: That's a good strategy!

tests/shared/src/test/scala/scala/meta/tests/tokenizers/BaseTokenizerCoverageSuite.scala
```diff
@@ -0,0 +1,331 @@
...
+   * We can write an arbitary function to extract a child node.
+   * ex: checkSome[Term.Param, Decl.Def](_.tokenEqual)("def f(x: A →=← 1): B")
+   *   R: Term.Param, T: Decl.Def, apply: extract first parameter
+   */
+  trait Projection[R, T] {
+    def apply(root: R): T
+  }
+
+  implicit val applyRepeated: Projection[Term.Apply, Term.Repeated] =
+    new Projection[Term.Apply, Term.Repeated] {
```
olafurpg: Can we have a `projection[Term.Apply, Term.Repeated] { case Term.Apply(_, (r: Term.Repeated) :: Nil) => r }` helper to reduce `def apply(...)` boilerplate for these instances?

MasseGuillaume:
@densh any clues what is wrong here: https://travis-ci.org/scalameta/scalameta/jobs/367107521#L2544 ?

MasseGuillaume:
@olafurpg CI is green, comments addressed.

PR #1500, opened 2018-04-13
Fix metacp AccessDeniedException on Windows

### LINKED ISSUE #1499: Metacp move should gracefully handle AccessDeniedException on windows
I hit a test failure on Windows in the scalafix CI using metacp https://ci.appveyor.com/project/scalacenter/scalafix/build/215#L1007

```
[info] scalafix.tests.core.BasePrettyTypeSuite *** ABORTED *** (15 milliseconds)
[info]   java.nio.file.AccessDeniedException: C:\Users\appveyor\AppData\Local\Temp\1\metacp6981278481363537711\jsse-E46B1924A1D00ABDBA3438467208B53E.jar -> C:\Users\appveyor\AppData\Local\semanticdb\cache\3.7.3\jsse-E46B1924A1D00ABDBA3438467208B53E.jar
```

The cause is likely that the file was being used elsewhere, judging by answers here 

- http://grokbase.com/t/openjdk/nio-dev/098x6jeh0b/wrong-exception-thrown
- https://stackoverflow.com/questions/22988210/java-nio-file-accessdeniedexception-while-using-files-move/22988462
- http://www.javaquizplayer.com/blogposts/blogpost3.html

I propose to 

1. catch this exception on windows only,
2. fallback to a non-atomic move
3. catch exception if non-atomic move fails, with no fallback.

under the theory that the file already exists. Any Windows people who can recommend a less hacky solution?

olafurpg:
Opening PR first to see if we can reproduce the bug on Window.

Fixes #1499


olafurpg:
Hmm, was unable to reproduce the problem 🤔 Let's see if reading the resulting jar right away causes the exception on windows

olafurpg:
Still no luck in reproducing the issue, I removed the unnecessary test suite but included the hypothetical "fix". I propose we go ahead and merge and see if the problem pops up again.

PR #1501, opened 2018-04-14
Add scala.Singleton trait to synthesized definitions

### LINKED ISSUE #1496: scala.Singleton is missing from scala-library-synthetics
I can't find reference to `scala.Singleton` in scala-library so I suspect it needs to be synthesized like Any/AnyRef and friends.
https://github.com/scalameta/scalameta/blob/0b41ec6a0cac00eef35fc043a74e6cffda12afdc/semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalalib.scala

gabro:
Closes #1496

### DIFF
--- a/semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalalib.scala
+++ b/semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalalib.scala
@@ -48,11 +48,34 @@ object Scalalib {
     builtinClass(List(p.ABSTRACT, p.FINAL), "Null", List("scala.AnyRef#"), Nil)
   }
 
+  def singletonTrait: ToplevelInfos = {
+    builtinTrait(List(p.FINAL), "Singleton", List("scala.Any#"), Nil)
+  }
+
   private def builtinClass(
       props: List[s.SymbolInformation.Property],
       name: String,
       bases: List[String],
-      symbols: List[s.SymbolInformation]): ToplevelInfos = {
+      symbols: List[s.SymbolInformation]
+  ): ToplevelInfos = {
+    builtinClassOrTrait(props, name, bases, symbols, isTrait = false)
+  }
+
+  private def builtinTrait(
+      props: List[s.SymbolInformation.Property],
+      name: String,
+      bases: List[String],
+      symbols: List[s.SymbolInformation]
+  ): ToplevelInfos = {
+    builtinClassOrTrait(props, name, bases, symbols, isTrait = true)
+  }
+
+  private def builtinClassOrTrait(
+      props: List[s.SymbolInformation.Property],
+      name: String,
+      bases: List[String],
+      symbols: List[s.SymbolInformation],
+      isTrait: Boolean): ToplevelInfos = {
     val parents = bases.map { base =>
       s.Type(tag = t.TYPE_REF, typeRef = Some(s.TypeRef(None, base, Nil)))
     }
@@ -77,7 +100,7 @@ object Scalalib {
     val builtin = s.SymbolInformation(
       symbol = symbol,
       language = l.SCALA,
-      kind = k.CLASS,
+      kind = if (isTrait) k.TRAIT else k.CLASS,
       properties = props.foldLeft(0)((acc, prop) => acc | prop.value),
       name = name,
       tpe = Some(builtinSig),
--- a/semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalalib.scala
+++ b/semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalalib.scala
@@ -94,7 +94,9 @@ object Scalalib {
     )
     val builtinSig = {
       val decls = symbols.filter(_.kind == k.METHOD)
-      val tpe = s.ClassInfoType(Nil, parents, ctor.symbol +: decls.map(_.symbol))
+      val declSymbols = decls.map(_.symbol)
+      val declarations = if (isTrait) declSymbols else ctor.symbol +: declSymbols
+      val tpe = s.ClassInfoType(Nil, parents, declarations)
       s.Type(tag = t.CLASS_INFO_TYPE, classInfoType = Some(tpe))
     }
     val builtin = s.SymbolInformation(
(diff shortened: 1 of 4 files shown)


tests/jvm/src/test/resources/scalalib.expect
```diff
@@ -126,3 +126,19 @@ scala.Null# => abstract final class Null.{+1 decls}
...
+Uri => scala/Singleton.class
+Text => empty
+Language => Scala
+Symbols => 2 entries
+
+Symbols:
+scala.Singleton# => final trait Singleton.{+1 decls}
+  extends Any
+scala.Singleton#`<init>`(). => primary ctor <init>: (): Singleton
+  Singleton => scala.Singleton#
```
gabro: not sure about this, should traits have a constructor? I suspect no, right?

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalalib.scala
```diff
@@ -48,11 +48,34 @@ object Scalalib {
...
+  ): ToplevelInfos = {
+    builtinClassOrTrait(props, name, bases, symbols, isTrait = true)
+  }
+
+  private def builtinClassOrTrait(
+      props: List[s.SymbolInformation.Property],
+      name: String,
+      bases: List[String],
+      symbols: List[s.SymbolInformation],
+      isTrait: Boolean): ToplevelInfos = {
```
olafurpg: can we replace boolean with `Kind`? That should allow us to have a single `builtin` method

tests/jvm/src/test/resources/scalalib.expect
olafurpg: The singleton trait should not have a constructor. I think traits have synthetic constructors, but you can't reference them in code so there is no need to include it.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalalib.scala
gabro: 👍 sounds right. This is where I really miss union types:

```scala
private def builtinClassOrTrait(
    kind: k.CLASS | k.TRAIT,
    props: List[s.SymbolInformation.Property],
    name: String,
    bases: List[String],
    symbols: List[s.SymbolInformation]): ToplevelInfos

File: tests/jvm/src/test/resources/scalalib.expect (hunk 1 above)
Comment (gabro) [reply tests/jvm/src/test/resources/scalalib.expect]: I've removed it when it's a trait. It still has a constructor from `Any` but it's not a declared member of that trait.

File: tests/jvm/src/test/resources/scalalib.expect [hunk 3]
```diff
@@ -126,3 +126,19 @@ scala.Null# => abstract final class Null.{+1 decls}
...
+Schema => SemanticDB v3
+Uri => scala/Singleton.class
+Text => empty
+Language => Scala
+Symbols => 2 entries
+
+Symbols:
+scala.Singleton# => final trait Singleton
+  extends Any
+scala.Singleton#`<init>`(). => primary ctor <init>: (): Singleton
```
olafurpg: Why is the constructor here? I thought the guard against `k.CLASS` prevented traits from having constructors.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalalib.scala
```diff
@@ -37,18 +37,23 @@ object Scalalib {
...
+    builtin(k.CLASS, List(p.ABSTRACT, p.FINAL), "Null", List("scala.AnyRef#"), Nil)
   }
 
-  private def builtinClass(
+  def singletonTrait: ToplevelInfos = {
+    builtin(k.TRAIT, List(p.FINAL), "Singleton", List("scala.Any#"), Nil)
+  }
+
+  private def builtin(
+      kind: s.SymbolInformation.Kind,
```
olafurpg: I agree union type would be nice but I can live with limiting it to SymbolInformation.Kind, which is a medium sized enum

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalalib.scala
```diff
@@ -71,13 +76,15 @@ object Scalalib {
     )
     val builtinSig = {
       val decls = symbols.filter(_.kind == k.METHOD)
-      val tpe = s.ClassInfoType(Nil, parents, ctor.symbol +: decls.map(_.symbol))
+      val declSymbols = decls.map(_.symbol)
+      val declarations = if (kind == k.CLASS) ctor.symbol +: declSymbols else declSymbols
```
olafurpg: ps. `kind.isClass` works, I prefer those helpers over using `==`

tests/jvm/src/test/resources/scalalib.expect
gabro: Not sure about this either: the interesting thing is that the `Singleton` trait doesn't have own declarations anymore (`{+1 decls}` is gone). Could that constructor be inherited from the `Any` class?

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalalib.scala
gabro: I'll change also `== k.METHOD` above then.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalalib.scala
```diff
@@ -37,18 +37,23 @@ object Scalalib {
...
   }
 
   def nullClass: ToplevelInfos = {
-    builtinClass(List(p.ABSTRACT, p.FINAL), "Null", List("scala.AnyRef#"), Nil)
+    builtin(k.CLASS, List(p.ABSTRACT, p.FINAL), "Null", List("scala.AnyRef#"), Nil)
   }
 
-  private def builtinClass(
+  def singletonTrait: ToplevelInfos = {
+    builtin(k.TRAIT, List(p.FINAL), "Singleton", List("scala.Any#"), Nil)
```
xeno-by: The spec says that traits can only be `SEALED`. Is `FINAL` here essential?

tests/jvm/src/test/resources/scalalib.expect
xeno-by: I suggest we remove the constructor here. The spec says "Unlike some other metaprogramming systems for Scala, we do not create synthetic constructor symbols for traits and objects".

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalalib.scala
gabro: Right, not essential. It was in this error message so I've applied:

```scala
scala> new scala.Singleton {}
<console>:18: error: illegal inheritance from final trait Singleton
```

tests/jvm/src/test/resources/scalalib.expect
gabro: Sure, I had tried already, but I've missed an `if` on the top-level info symbols.

gabro:
Thanks for the review @xeno-by, I've addressed your comments.

PR #1512, opened 2018-04-24
Merge langmeta back into scalameta

xeno-by:
Separating langmeta from scalameta was a mistake, since our use cases
ended up not requiring this level of generality, but the abstraction
overhead (aliases, education, extra artifacts, etc) was real.

Now that we're going for a major release because of the recent removal
of the high-level semantic API, we might as well fix this mistake.


scalameta/semanticdb/shared/src/main/scala/scala/meta/semanticdb/Symbol.scala
```diff
@@ -1,5 +1,4 @@
-package org.langmeta
-package semanticdb
+package scala.meta.semanticdb
```
xeno-by: @olafurpg Do we still need `scala.meta.Symbol`?

olafurpg: Yes, `Symbol` will need to live with us another cycle, although we might want to deprecate it 🤔

PR #1513, opened 2018-04-26
Remove scala.meta.io.{Multipath,Sourcepath,Fragment}

olafurpg:
These classes are no longer used.


scalameta/io/shared/src/main/scala/scala/meta/io/Classpath.scala
```diff
@@ -0,0 +1,23 @@
+package scala.meta.io
+
+import java.io.File.pathSeparator
+
+final case class Classpath(entries: List[AbsolutePath]) {
+
+  @deprecated("Use .entries instead", "v4.0.0")
```
xeno-by: 4.0.0

PR #1518, opened 2018-05-02
Add SYNTHETIC property to SemanticDB

### LINKED ISSUE #1492: Missing SYNTHETIC or CASE property for generated case class methods
Synthetic members for case classes are now included by metac
```scala
$ cat foo.scala
case class Foo()
```

However, the synthetic methods don't have properties to indicate they are synthetic

```diff
- obtained
+ expected
$ metac foo.scala && metap -proto META-INF/semanticdb/foo.scala.semanticdb | grep "Foo#equals" -A 10
  symbols {
    symbol: "_empty_.Foo#equals(Any).(x$1)"
    kind: PARAMETER
-    properties: 0
+    properties: SYNTHETIC | CASE
    name: "x$1"
    tpe {
      tag: TYPE_REF
      typeRef {
        symbol: "scala.Any#"
      }
    }
```

This change would need to be accompanied with a spec. The same SYNTHETIC property could be added to java enum `values().` and `valueOf(String).`

jeffreyolchovy:
Fixes #1492

### DIFF
--- a/semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
+++ b/semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
@@ -130,6 +130,7 @@ object Scalacp {
     if (sym.isImplicit) sflip(p.IMPLICIT)
     if (sym.isLazy) sflip(p.LAZY)
     if (sym.isCase) sflip(p.CASE)
+    if (sym.isSynthetic) sflip(p.SYNTHETIC)
     if (sym.isType && sym.isCovariant) sflip(p.COVARIANT)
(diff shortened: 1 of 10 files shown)


olafurpg:
It might be worth documenting what SYNETHTIC means in Java. The access flag ACC_SYNTHETIC is represented with the ARTIFACT flag in scalac/dotc and has a different meaning than SYNTHETIC.

olafurpg:
I will take care of finishing the last remaining pieces in this PR. I propose that we restrict SYNTHETIC to only generated members for case classes. That should be easy to both spec and implement as well as unblock the main use-case I had in mind for SYNTHETIC. wdyt @xeno-by ?

xeno-by:
@olafurpg Sounds good. We can always expand the spec later.

PR #1524, opened 2018-05-07
Add docstring to protobuf schema

mads-hartmann:
**Goal**: Get docstrings into the semanticdb schema. I intend to use this information in metals to show documentation for symbols on hover.

**WIP**: This is, as you can see, very much a work in progress. I just wanted to float the idea by you early to make sure I'm not going down a wrong path. My next step would be to find all the places we create an instance of `SymbolInformation` and include the docstring once I've figured out how to get ahold of it :wink:

### DIFF
--- a/semanticdb/semanticdb3/semanticdb3.md
+++ b/semanticdb/semanticdb3/semanticdb3.md
@@ -1201,6 +1201,7 @@ message SymbolInformation {
   Kind kind = 3;
   int32 properties = 4;
   string name = 5;
+  string docstring = 17;
   Type tpe = 11;
   repeated Annotation annotations = 13;
   Accessibility accessibility = 14;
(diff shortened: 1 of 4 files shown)


olafurpg:
two more things I forgot to mention, these two might come in handy

- https://github.com/scalameta/scalameta/blob/afbd949b70144d86973ac9c573474d4b777fa007/scalameta/contrib/shared/src/main/scala/scala/meta/contrib/AssociatedComments.scala
- https://github.com/scalameta/scalameta/blob/afbd949b70144d86973ac9c573474d4b777fa007/scalameta/contrib/shared/src/main/scala/scala/meta/contrib/ScaladocParser.scala

They're in contrib but we could make them accessible in the semanticdb-scalac compiler plugin if necessary. The scaladoc parser is quite rough and likely super slow, which might become a problem when indexing ~2 million lines of code (all dependencies on classpath) in the IDE during build import. But we can start by focusing on functionality and optimize later

mads-hartmann:
@olafurpg Thanks a lot for the pointers, that definitely helps ☺️ It varies quite a lot from week to week how much time I can find to hack around on this but I'll keep you posted 👍

olafurpg:
Please reopen to continue with this! I'm cleaning up the PR queue :)

PR #1533, opened 2018-05-16
Add materialized types to symbol occurrences

romanowski:
New mode for symbol config is added that when enabled fills materializedType filed in each symbol occurence that is reference (there is no point in filling it on definition side since type will be the same as one stored for symbol)


romanowski:
@olafurpg PR as promised yesterday. I've realized that I forgot to rebase and I need to add one more test. So it is WIP for now :)

semanticdb/metap/src/main/scala/scala/meta/internal/metap/Main.scala
```diff
@@ -170,26 +170,23 @@ class Main(settings: Settings, reporter: Reporter) {
     lineIndices(line)
   }
 
+  def pprint(range: Range, doc: Option[TextDocument]): String = {
```
xeno-by: Other `pprint` methods simply print to `out` instead of returning a string. How about changing this method to do the same for consistency reasons?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/ConfigOps.scala
```diff
@@ -160,12 +160,14 @@ sealed abstract class SymbolMode {
...
   def saveDefinitions: Boolean = this == All || this == Definitions
   def saveReferences: Boolean = this == All
+  def saveMaterializedTypes: Boolean = this == AllWithMaterializedTypes
 }
 object SymbolMode {
   def name: String = toString.toLowerCase
   def unapply(arg: String): Option[SymbolMode] =
     all.find(_.toString.equalsIgnoreCase(arg))
   def all = List(All, Definitions, None)
+  case object AllWithMaterializedTypes extends SymbolMode
```
xeno-by: Let's introduce a dedicated `OccurrenceMode`. `SymbolMode` traditionally only affects the `SymbolInformation` section in SemanticDB.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/Inferred.scala
```diff
@@ -4,15 +4,16 @@ package internal.semanticdb.scalac
 import scala.meta.internal.inputs._
 import scala.meta.internal.{semanticdb3 => s}
 
-case class SyntheticRange(start: Int, end: Int, symbol: Symbol) {
-  def addOffset(offset: Int) = SyntheticRange(start + offset, end + offset, symbol)
+case class SyntheticRange(start: Int, end: Int, symbol: Symbol, tpe: Option[s.Type]) {
```
xeno-by: Uh oh. Synthetics definitely complicate things.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/PrinterOps.scala
```diff
@@ -6,25 +6,27 @@ import java.io.StringWriter
 import java.io.Writer
 import scala.collection.mutable
 import scala.reflect.internal.ModifierFlags._
+import `scala`.meta.internal.{semanticdb3 => s}
```
xeno-by: This is not the first time I see `scala` backquoted in Scala code. Is this some corner case in the Intellij plugin? In some other tool? Really curious.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -159,6 +159,29 @@ trait TextDocumentOps { self: SemanticdbOps =>
 
       locally {
         object traverser extends g.Traverser {
+
+          private var currentTress = List[g.Tree]()
```
xeno-by: Looks like there's a typo in the name of the local variable.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -159,6 +159,29 @@ trait TextDocumentOps { self: SemanticdbOps =>
...
+                None
+              case _ if currentStack.isEmpty =>
+                None
+              case tpe =>
+                tpe.widen.toSemantic._1
+            }
+
+            currentStack.tail match { // Current tree is head...
+              case (tpeApply @ g.TypeApply(_, _)) :: _ =>
+                computeTpe(tpeApply)
```
xeno-by: Can you give an example of where this check is used? Looks like a lot of infrastructure for the sake of these two lines of code, so I'd like to better understand the motivation.

semanticdb/semanticdb3/semanticdb3.proto
```diff
@@ -250,6 +250,7 @@ message SymbolOccurrence {
   Range range = 1;
   string symbol = 2;
   Role role = 3;
+  Type materializedTpe = 4;
```
xeno-by: I suggest we call this simply `tpe`. Along the same lines, I think it would make sense to rename associated fields, methods and local variables to `tpe` (or `gtpe` / `mtpe` in DocumentOps).

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/PreciseTypeSuite.scala
```diff
@@ -0,0 +1,78 @@
...
+
+class PreciseTypeSuite extends SemanticdbSuite(symbols = SymbolMode.AllWithMaterializedTypes) {
+
+  private def testActualTypes(code: String)(occurencesTypes: String) = test(code){
+    val db = computeDatabaseFromSnippet(code)
+    val allOccurences = db.occurrences ++ db.synthetics.flatMap(_.text).flatMap(_.occurrences)
+
+    def printOccurence(occ: s.SymbolOccurrence) = occ.materializedTpe.map( tpe =>
+          s"${Print.range(occ.range.get)} ${Print.tpe(db, tpe)}"
+      )
```
xeno-by: What do you guys think about changing `metap` to always print types associated with occurrences. This will probably be useful for `metap` users and will simplify testing. For example:

Before:
```
Occurrences:
[0:7..0:8): M <= _empty_.M.
```

After:
```
Occurrences:
[0:7..0:8): M <= _empty_.M. (M.type)
```

xeno-by: This may easily end up being too verbose, so let's see how this will look in practice.

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/PreciseTypeSuite.scala
```diff
@@ -0,0 +1,78 @@
...
+    val typeLines = db.occurrences.map(printOccurence) ++ occurencesStrings
+    val typeString = typeLines.flatten.sorted.mkString("\n")
+    assert(typeString == occurencesTypes)
+  }
+
+  testActualTypes(
+    "object L2 { val head = Array(1, 2, 3)(0) }"
+  )(
+    """[0:23..0:28) scala.Array.
+      |[0:28..0:28) ~> [0:2..0:7) scala.Array.apply(Int,Int*).(x)scala.Array.apply(Int,Int*).(xs)scala.Array#scala.Int#
```
xeno-by: I'm not sure what `scala.Array.apply(Int,Int*).(x)scala.Array.apply(Int,Int*).(xs)scala.Array#scala.Int#` means. This looks like multiple concatenated types, but I don't quite get why there are several of them. Could you elaborate?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/PrinterOps.scala
romanowski: Probably, somehow I missed this change. I will try to repro and report to Intellij guys

olafurpg: It's a ScalaPB bug, `scala` is treated as a keyword https://github.com/scalapb/ScalaPB/blob/f1beff9dee6dfe38494d281a730e544ac520e810/compiler-plugin/src/main/scala/scalapb/compiler/DescriptorPimps.scala#L24

olafurpg: I opened a PR updating scalapb https://github.com/scalapb/ScalaPB/pull/447

ShaneDelmore:
Very excited about this.  I have had a couple of people asking me how to get type of expression to use for rewrites.  One specific case I recall was someone wanting to rewrite futures to something that is lazy, but only when the futures were called solely for side-effect (indicated in their use case by an expression return type of Unit)

olafurpg:
Thanks for working on this @romanowski It's a great idea, I am excited about this feature and I look forward to using in scalafix! Since this is critical functionality, it may take several iterations to make sure it's done well. I am interested in helping out, it might be possible to refactor the current implementation to simplify this PR.

I am afraid I can't personally prioritize this in at least the next month. My current focus is making semanticdb-scalac as stable and robust as possible, we have identified 31 tickets while stress-testing SemanticDB on larger projects (see https://github.com/scalameta/scalameta/milestone/26). I'd like to fix those for the next release before taking on feature additions.

Can you give me and estimate for your timeline on this PR? Is it urgent or can you afford to wait a couple months?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
romanowski: In case like `foo[Int](1)` scalac use generic type for `foo` and precise type is stored in `foo[Int]`.
Probably we could keep only last tree but full stack is pretty much the same cost to work with.

I am open for any suggestion how to rewrite it.

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/PreciseTypeSuite.scala
romanowski: Changes are there and output is quite nice

romanowski:
I should be fine - TBH I can release it somewhere myself and use custom version. By default my feature is turn off and can be marked as experimental or something so (in theory) it should not cause problems. 

@ShaneDelmore I've added simple test case for Future - let me know if it contains all information needed

olafurpg:
Cool @romanowski would it help to publishLocal if the modules were aggregated? Currently it's a hassle to publishLocal IME, I'm happy to refactor build.sbt to make it easier.

xeno-by:
@romanowski @olafurpg Unfortunately, I don't think I will have time to help out with shepherding this pull request in the near future either - not in May/June anyway. Let's close this pull request then and keep it in mind for the potential 4.x roadmap.

Issue #897, opened 2017-05-27
AssociatedComments skips over first Scaladoc in a file


Atry:
Create a file A.scala:
``` scala
/** Scaladoc for class A
  */
class A
/** Scaladoc for object A
  */
object A
```

Then parse it:
``` scala
Welcome to the Ammonite Repl 0.9.3
(Scala 2.12.2-bin-typelevel-4 Java 1.8.0_131)
If you like Ammonite, please support our development at www.patreon.com/lihaoyi
@ import scala.meta.contrib._ 
import scala.meta.contrib._
@ import scala.meta.contrib.DocToken._ 
import scala.meta.contrib.DocToken._
@ import scala.meta.tokens.Token.Comment 
import scala.meta.tokens.Token.Comment
@ import scala.meta.Defn 
import scala.meta.Defn
@ import scala.meta._ 
import scala.meta._
@ import java.io.File 
import java.io.File
@ val source@Source(stats) = new File("A.scala").parse[Source].get 
source: Source = Source(
  List(
    Defn.Class(
      List(),
      Type.Name("A"),
      List(),
      Ctor.Primary(List(), Ctor.Ref.Name("this"), List()),
      Template(List(), List(), Term.Param(List(), , None, None), None)
    ),
    Defn.Object(List(), Term.Name("A"), Template(List(), List(), Term.Param(List(), , None, None), None))
  )
)
stats: collection.immutable.Seq[Stat] = List(
  Defn.Class(
    List(),
    Type.Name("A"),
    List(),
    Ctor.Primary(List(), Ctor.Ref.Name("this"), List()),
    Template(List(), List(), Term.Param(List(), , None, None), None)
  ),
  Defn.Object(List(), Term.Name("A"), Template(List(), List(), Term.Param(List(), , None, None), None))
)
@ val comments = AssociatedComments(source.tokens) 
comments: AssociatedComments = scala.meta.contrib.AssociatedComments$$anon$1@d1b0709
@ stats.map(comments leading _ map ScaladocParser.parseScaladoc) 
res8: collection.immutable.Seq[Set[Option[Seq[DocToken]]]] = List(
  Set(),
  Set(Some(List(DocToken(Description, None, Some("Scaladoc for object A")))))
)
```

There should be a `Scaladoc for class A`. Unfortunately, there is only a `Set()`.

olafurpg:
Thank you for reporting! I am able to reproduce. I suspect the bug is the AssociatedComments class.

[...]

PR #1534, opened 2018-05-17
Fix: handle comments before EOF

marcelocenerine:
Fixes https://github.com/scalacenter/scalafix/issues/627

This issue is similar to #897, but affecting comments at the bottom of a file (== ignored).

/cc. @olafurpg

### DIFF
--- a/scalameta/contrib/shared/src/main/scala/scala/meta/contrib/AssociatedComments.scala
+++ b/scalameta/contrib/shared/src/main/scala/scala/meta/contrib/AssociatedComments.scala
@@ -59,16 +59,22 @@ object AssociatedComments {
         if (isLeading) leading += c
         else trailing += c
       case Token.LF() => isLeading = true
+      case Token.EOF() =>
+        val l = leading.result()
+        val t = trailing.result()
+        if (l.nonEmpty || t.nonEmpty) {
+          trailingBuilder += lastToken -> (l ::: t)
+        }
       case Trivia() =>
       case currentToken =>
         val t = trailing.result()
         if (t.nonEmpty) {
-          trailingBuilder += lastToken -> trailing.result()
+          trailingBuilder += lastToken -> t
           trailing.clear()
         }
         val l = leading.result()
         if (l.nonEmpty) {
-          leadingBuilder += currentToken -> leading.result()
+          leadingBuilder += currentToken -> l
           leading.clear()
         }
         if (!currentToken.is[Comma]) {
--- a/tests/shared/src/test/scala/scala/meta/tests/contrib/AssociatedCommentsTest.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/contrib/AssociatedCommentsTest.scala
@@ -3,7 +3,6 @@ package contrib
 
 import scala.meta._
 import scala.meta.contrib._
-import scala.meta.tokens.Token
 import org.scalatest.FunSuite
 
 class AssociatedCommentsTest extends FunSuite {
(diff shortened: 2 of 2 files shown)


scalameta/contrib/shared/src/main/scala/scala/meta/contrib/AssociatedComments.scala
```diff
@@ -59,16 +59,22 @@ object AssociatedComments {
         if (isLeading) leading += c
         else trailing += c
       case Token.LF() => isLeading = true
+      case Token.EOF() =>
+        val l = leading.result()
+        val t = trailing.result()
+        if (l.nonEmpty || t.nonEmpty) {
+          trailingBuilder += lastToken -> (l ::: t)
+        }
```
marcelocenerine: this could also be implemented outside of the `foreach`. Please let me know if you prefer that instead

tests/shared/src/test/scala/scala/meta/tests/contrib/AssociatedCommentsTest.scala
```diff
@@ -39,19 +38,234 @@ class AssociatedCommentsTest extends FunSuite {
...
+    )
+  }
+
+  test("multiple trailing comments in different lines at the end of a file") {
+    val input =
+      """|trait A {
+         |} /** trailing 1 */
+         |/* trailing 2 */
+         |// trailing 3
+         |""".stripMargin.parse[Source].get
```
marcelocenerine: the fix makes this scenario consistent with the one where multiple leading comments appear after BOF. This means that all comments (regardless of how many lines) are associated to the closest non-trivia token. Please let me know if you think it is not correct

olafurpg: This looks great!

tests/shared/src/test/scala/scala/meta/tests/contrib/AssociatedCommentsTest.scala
```diff
@@ -39,19 +38,234 @@ class AssociatedCommentsTest extends FunSuite {
...
+        defnVal -> Set("// trailing"),
+        litInt -> Set("// trailing")
+      )
+    )
+  }
+
+  test("multiple comments interleaved with trees in a single line") {
+    val input =
+      """|object A {
+         | /* comment 1 */ val /* comment 2 */ foo = /* comment 3 */ 0 /* comment 4 */
```
olafurpg: What happens to comments 2 and 3? I would expect comment 2 to be leading for `foo` and comment 3 leading for `0`. Looks like a bug unrelated to this PR.

marcelocenerine: @olafurpg that's something that got me confused too. It turns out that the parsing logic associates comments to **tokens**. If you inspect the internal state you will see that comments 2 and 3 are there:

```
AssociatedComments(
  Leading =
    val [28..31) => List(/*∙comment∙1∙*/)

  Trailing =
    val [28..31) => List(/*∙comment∙2∙*/)
    = [52..53) => List(/*∙comment∙3∙*/)
    0 [70..71) => List(/*∙comment∙4∙*/)
)
```

However, the `AssociatedCommentsTest` public api let you query comments associated to **trees**. The `leading` and `trailing` methods only search for comments associated to the first and last tokens (respectively) of the given tree node. Comments 2 and 3 are neither associated to one or the other of any tree. Not sure if this is a bug though. Wdyt?

olafurpg: I think this is a bug, good catch. I opened https://github.com/scalameta/scalameta/issues/1539 to track it.

PR #1536, opened 2018-05-21
Include diagnostics with offset positions

### LINKED ISSUE #1530: Unused private val warning are not reported in messages
Hi there,

We are not seeing the private val warning in the messages.

Related to: https://github.com/scalacenter/scalafix/issues/626
See: https://github.com/scalacenter/scalafix/pull/723#issuecomment-387017310

Cheers

lgirault:
Relax RelevantMessage condition to forward unused local warnings.

Fix #1530

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/ReporterOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/ReporterOps.scala
@@ -14,7 +14,6 @@ trait ReporterOps { self: SemanticdbOps =>
         case r: StoreReporter =>
           object RelevantMessage {
             def unapply(info: r.Info): Option[(gPosition, Int, String)] = {
-              if (!info.pos.isRange) return None
               if (info.pos.source != unit.source) return None
               Some((info.pos, info.severity.id, info.msg))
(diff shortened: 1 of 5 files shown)


lgirault:
My first reflex was to add a `Position.Offset` to `Position.Range` and `Position.None` (see https://github.com/scalameta/scalameta/compare/master...lgirault:offset ) .
Is there a reason why it does not exists ?

olafurpg:
> My first reflex was to add a Position.Offset to Position.Range and Position.None (see master...lgirault:offset ) .
> Is there a reason why it does not exists ?

A large number of bugs and pains working with positions in scala-compiler revolve around different kinds of positions. In Scalameta we try to keep the number of concepts as few as possible. Since offset positions can be treated as zero-width range positions there is no need for `Position.Offset`

lgirault:
You're welcome ! I'm trying to get more involved in the oss ^_^

lgirault:
and thank you for the answer!

lgirault:
ok that's weird. I have a case of "work on my machine" :s 
I'll try to check tonight the problem

lgirault:
thank you @joan38 
So I found the problem, the warnings are different in scala 2.11 and 2.12 but I'm not sure if the solution I used is right

PR #1558, opened 2018-05-24
Replace isSnapshot.value with version.value.endsWith("SNAPSHOT")

olafurpg:
The former does not work.

```
[info] dialectsJVM/*:isSnapshot
[info]     true
[info] dialectsJVM/*:version
[info]     4.0.0-M1
[info] dialectsJVM/*:publishTo
[info]     Some(sonatype-staging: https://oss.sonatype.org/service/local/staging/deploy/maven2)
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -640,7 +640,8 @@ lazy val publishableSettings = Def.settings(
     platformDepsCrossVersion.value == ScalaNativeCrossVersion.binary,
   publishTo := Some {
     if (isCustomRepository) "adhoc" at adhocRepoUri
-    else if (isSnapshot.value) Opts.resolver.sonatypeSnapshots
+    // NOTE: isSnapshot.value does not work with sbt-dynver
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -640,7 +640,8 @@ lazy val publishableSettings = Def.settings(
     platformDepsCrossVersion.value == ScalaNativeCrossVersion.binary,
   publishTo := Some {
     if (isCustomRepository) "adhoc" at adhocRepoUri
-    else if (isSnapshot.value) Opts.resolver.sonatypeSnapshots
+    // NOTE: isSnapshot.value does not work with sbt-dynver
+    else if (version.value.endsWith("SNAPSHOT")) Opts.resolver.sonatypeSnapshots
```
xeno-by: Looks like the same change will have to be made to the Scalafix build.

olafurpg: This is only necessary when publishing to maven from local machine. Scalafix is only released via ci.

PR #1575, opened 2018-05-31
Fix missing setter symbol for private[this] vars

xeno-by:


semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
```diff
@@ -496,10 +496,21 @@ object Scalacp {
...
+      val decls = sym.children.filter(decl => decl.isUseful && !decl.isTypeParam)
+      SemanticdbDecls(decls.toList)
+    }
   }
 
-  private implicit class ScopeOps(decls: Seq[Symbol]) {
-    def useful: Seq[Symbol] = decls.filter(decl => decl.isUseful && !decl.isTypeParam)
+  case class SemanticdbDecls(syms: List[Symbol]) {
+    lazy val ssyms: List[String] = {
+      val sbuf = List.newBuilder[String]
```
olafurpg: why not `syms.map(ssymbol)`?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolOps.scala
```diff
@@ -102,6 +94,46 @@ trait SymbolOps { self: SemanticdbOps =>
...
+          }
+          loop(sym.info)
+        }
+      }
+    }
+  }
+
+  case class SemanticdbDecls(gsyms: List[g.Symbol]) {
+    lazy val ssyms: List[String] = {
+      val sbuf = List.newBuilder[String]
```
olafurpg: ditto

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
xeno-by: This implementation gets an upgrade in a subsequent commit. It looks dumb when introduced, but that allows for an easier to read diff later on.

Issue #1426, opened 2018-03-11
Tighten the spec around refinement members, existential skolems and type lambda parameters


xeno-by:
When used in signatures of global symbols, refinement members are global symbols, too. However, symbol generation for refinement members is not very well understood at this moment. The specification says: 

> Owner is:
>   ...
>   * For other global definition, the innermost enclosing definition, i.e. the definition whose Location in source code most tightly encloses the Location of the original definition.

But unfortunately: 1) that's not how metac and scalacp work, 2) this strategy leads to name collisions, as illustrated by https://github.com/scalameta/scalameta/issues/1297.

xeno-by:
Example code:
```scala
val compoundType1: { def k: String } = ???
val compoundType3: M with N { def k: Int } = ???
```

Existing behavior:
```
_root_.types.Test.C#$anon#k()I. => abstract def k: : Int
```

Proposed behavior:
```
_root_.types.Test.C#1#k()I. => abstract def k: : String
_root_.types.Test.C#2#k()I. => abstract def k: : Int
```

In the snippet above, refinement member symbols consist of three parts:
  * Top-level enclosing class symbol, i.e. `_root_.types.Test.C#` (necessary to achieve the required degree of uniqueness across documents).
  * Fake owner, i.e. `NNN`, where `NNN` is a fresh decimal numeric id (digits are not valid Java identifier starts, so this is unambiguous with regular owners).
  * Definition descriptor, i.e. `k()l.` (the usual stuff).

xeno-by:
Same concerns apply to existential skolems and type lambda parameters - the other symbols that are defined outside the traditional owner chain.

PR #1580, opened 2018-06-02
Über PR to continue the 4.0.0-RC1 development cycle

xeno-by:
A bunch more tickets down.


xeno-by:
What has started as a small pull request, has grown to incorporate more and more stuff. The branch name no longer corresponds to reality, but I don't really want to close and reopen the PR.

xeno-by:
Looks like this has become a yet another mega PR. I've started working on refinement and existential types, and it turns out I needed lots of foundational work that can be upstreamed right away. #1426 is still pending, but I hope to be done today.

xeno-by:
@olafurpg This was a productive weekend. I think I solved the problem with refinements, existentials and type lambdas, but that required a significant refactoring of our entire toolchain:
  * After thinking real hard, I realized that refinement declarations, existential declarations and lambda type parameters cannot be global symbols. [Summary of my current thoughts](https://github.com/scalameta/scalameta/pull/1580/commits/946136b476df89d2bd319c513be4f8a12b067a4b).
  * In order to accommodate the consequences, `StructuralType`, `ExistentialType`, `UniversalType`, `ClassInfoType`, `MethodType` and `PolyType`  now have a slightly different way to model children: [spec](https://github.com/xeno-by/scalameta/blob/wip/mini/semanticdb/semanticdb3/semanticdb3.md#scope). In addition to carrying symlinks to their children, these types can now hardlink to their children metadata by embedding it directly inside their payloads. As a result, if necessary, these types can become mini symbol tables in themselves. This is the solution to the problem that I have never managed to solve in `reify`.
  * To adopt the new way to model children, I had to modify Metac, Metacp and Metap. Of these tools, only Metac was able to accommodate the changes relatively easily. Metacp and Metap were pretty deep in technical debt, so I had to modularize them before attempting important changes like that. As a side effect, these refactorings made it possible to fix lots of other small bugs.

In addition to implementing the functionality that was the direct goal of this pull request, I've also made some quality of life improvements:
  * Modularized `testsJVM/test` into `testsJVM/fast:test` (default) and `testsJVM/slow:test`, and as a result typical test invocations on my machine are down from 44.7s to 8.3s. Now I can finally run testsJVM/test for basically all changes without losing focus.
  * Killed `-P:semanticdb:symbols:all` and refactored Type.toSemantic and friends to avoid carrying todos around. That ended up being a significant sweeping simplification across almost entire Metac.
  * Streamlined compiler plugin options ([link](https://github.com/xeno-by/scalameta/blob/wip/mini/semanticdb/semanticdb3/guide.md#scalac-compiler-plugin)). Names and deprecations are up for discussion - I just thought that it's high time to clean things up in that area, given that we've been accumulating options for close to 1.5 years now.
  * Implemented `metap -compact` - the format that doesn't print fully-qualified names of signature components alongside `TextDocument.symbols` entries. This unexpectedly significantly improved readability.

I could probably attempt to split these changes into multiple pull requests, but those pull requests would end up pretty intertwined between each other anyway. With this scope of changes (which involve fundamental cleanups basically everywhere), it ended up being really hard to modularize the diffs. I did my best to offset this by being even more fine grained with individual commits to at least achieve isolation.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/SymbolOps.scala
```diff
@@ -0,0 +1,244 @@
...
+      sym.isSyntheticValueClassCompanion ||
+      sym.isUselessField ||
+      sym.isSyntheticCaseAccessor ||
+      sym.isRefinementClass
+    }
+    def isUseful: Boolean = !sym.isUseless
+  }
+
+  private var nextId = 0
+  private def freshSymbol(): String = {
```
olafurpg: I'm a concerned about producing local symbols from metacp since those can get mixed with local symbols in a text document. For example, scalafix symtab is per-source which means after this PR, it's not clear how to implement `info("local1")` since `local1` may come from a hardlinked symbol from a dependency or from a local variable in the current text document.

I've wondered if we should adopt a similar scheme as JVM signatures where the first character indicates the type

* prefix global symbols with `G`
* prefix local symbols with `L`
* prefix hardlink symbols with `H`

Hardlinked symbols could additionally be suffixed by the enclosing owner symbol so that 

```diff
- this PR
+ proposal

advanced.Structural#s2(). => method s2: AnyRef { val def x: Int }
  AnyRef => scala.AnyRef#
-  x => local1
+  x => Hadvanced.Structural#s2().1
  Int => scala.Int#
```

This would mean hardlinked symbols could remain globally unique.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -223,7 +223,8 @@ trait TextDocumentOps { self: SemanticdbOps =>
                   }
                 }
               } else {
-                occurrences(mtree.pos) = symbol
+                val selectionFromStructuralType = gsym.owner.isRefinementDummy
+                if (!selectionFromStructuralType) occurrences(mtree.pos) = symbol
```
olafurpg: It took me a while to understand the motivation for this change but I think I get it now. Structural types can be produced on the fly so the structure `{ def foo: Int }` does not have a global unique symbols. I agree with the strategy to edge on the conservative side and not emit occurrences here, poking around IntelliJ it seems to struggle with structural types as well



In the long term, we need to come up with a solution for this case since it's highly relevant for tools to know which occurrences are selections on structural types. One crazy idea might be to introduce `SymbolOccurrence.Role.STRUCTURAL` and make `SymbolOccurrence.symbol` be empty. The structural type could be embedded in the `SymbolOccurrence.tpe` of the prefix. Anyways, we can leave this discussion to an issue to unblock this PR. This commit LGTM 👍

olafurpg: Can we open a ticket with the description of this commit + add a link in the code to the issue? It's not clear from the source code why we skip on `selectionFromStructuralType`

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolOps.scala
```diff
@@ -66,6 +66,7 @@ trait SymbolOps { self: SemanticdbOps =>
           sym.isSelfParameter ||
           sym.isRefinementDummy ||
           sym.isAnonymousClass ||
+          sym.isAnonymousFunction ||
```
olafurpg: \o/ Thank you 🙏

build.sbt
```diff
@@ -470,7 +471,12 @@ lazy val tests = crossProject(JSPlatform, JVMPlatform, NativePlatform)
     ),
     buildInfoPackage := "scala.meta.tests",
     libraryDependencies += "com.lihaoyi" %%% "fansi" % "0.2.5" % "test",
-    libraryDependencies += "org.scalatest" %%% "scalatest" % "3.2.0-SNAP10" % "test"
+    libraryDependencies += "org.scalatest" %%% "scalatest" % "3.2.0-SNAP10" % "test",
+    testOptions.in(Test) += Tests.Argument("-l", "org.scalatest.tags.Slow"),
+    inConfig(Fast)(Defaults.testTasks),
```
olafurpg: Is `Fast` necessary? Seems like `testsJVM/test` runs the same

build.sbt
```diff
@@ -470,7 +471,12 @@ lazy val tests = crossProject(JSPlatform, JVMPlatform, NativePlatform)
     ),
     buildInfoPackage := "scala.meta.tests",
     libraryDependencies += "com.lihaoyi" %%% "fansi" % "0.2.5" % "test",
-    libraryDependencies += "org.scalatest" %%% "scalatest" % "3.2.0-SNAP10" % "test"
+    libraryDependencies += "org.scalatest" %%% "scalatest" % "3.2.0-SNAP10" % "test",
+    testOptions.in(Test) += Tests.Argument("-l", "org.scalatest.tags.Slow"),
```
olafurpg: I love this change ❤️ Been meaning to come up with a solution for this problem for a long time. Thank you @laughedelic for the SO answer!

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalalib.scala
```diff
@@ -106,17 +106,7 @@ object Scalalib {
...
-        // FIXME: https://github.com/scalameta/scalameta/issues/1550
-        val _ :+ last = symbol.split("[\\.|#]").toList
-        val last1 = last.stripPrefix("(").stripPrefix("[")
-        val last2 = last1.stripSuffix(")").stripSuffix("]").stripSuffix("#")
-        last2.stripPrefix("`").stripSuffix("`")
-      }
-      paramTypeDescriptors.mkString(",")
-    }
-    val methodSymbol = classSymbol + encodedMethodName + "(" + disambiguator + ")."
+    val methodSymbol = classSymbol + encodedMethodName + "()."
```
olafurpg: Note to self. We should harden our expect tests to catch bugs like this in the future. One idea, assert that all symbol occurrences have a SymbolInformation with a naive symbol table implementation.

tests/jvm/src/test/resources/metac.expect
```diff
@@ -1977,13 +1977,13 @@ types.Test.C#thisType1(). => val method thisType1: : C.this.type
   C => types.Test.C#
 types.Test.C#thisType2(). => val method thisType2: : C.this.type
   C => types.Test.C#
-types.Test.C#typeLambda1(). => method typeLambda1: [M [M][_: <?>] => >: Nothing <: Any] => : Nothing
+types.Test.C#typeLambda1(). => method typeLambda1: [M [_: <?>] => >: Nothing <: Any] => : Nothing
   M => types.Test.C#typeLambda1().[M]
-  M][_ => types.Test.C#typeLambda1().[M][_]
```
olafurpg: 😂 glad to see symbol parsing logic unified

tests/jvm/src/test/resources/metac-metacp.expect.diff
```diff
@@ -1,186 +1,218 @@
...
-
-
-===============================
-com.javacp.MetacJava#Overload1#
-===============================
+                 type_arguments {
+                   tag: TYPE_REF
+                   typeRef {
+-                    symbol: "local8"
++                    symbol: "advanced.Test.T#"
```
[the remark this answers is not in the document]
xeno-by: Yeah, and an unfixable one at that, as long as we're using Scalap: https://github.com/scalameta/scalameta/issues/1586.

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -248,6 +249,48 @@ but we haven't yet found a way to do that without sacrificing performance
 and payload size. In the meanwhile, when global uniqueness is required,
 tool authors are advised to accompany local symbols with `TextDocument.uri`.
 
+### Scope
+
+```protobuf
+message Scope {
+  repeated string symlinks = 1;
```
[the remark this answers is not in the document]
xeno-by: Aren't softlinks and symlinks the same thing?

tests/jvm/src/test/resources/metac-metacp.expect.diff
```diff
@@ -1,186 +1,218 @@
...
--              symbol: "scala.annotation.unchecked.uncheckedVariance#"
-
-
-====
-com.
-====
+             type_arguments {
+               tag: TYPE_REF
+               typeRef {
+-                symbol: "local7"
```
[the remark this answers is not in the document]
xeno-by: I thought I submitted a ticket for that, but it turns out that I didn't. Maybe, like the design of hardlinks, this was part of my dream the other night 🤣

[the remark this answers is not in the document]
xeno-by: https://github.com/scalameta/scalameta/issues/1587

semanticdb/metap/src/main/scala/scala/meta/internal/metap/RangePrinter.scala
```diff
@@ -0,0 +1,50 @@
...
+    out.print(range.endCharacter)
+    out.print(")")
+  }
+
+  private val lineToOffsetCache = new mutable.HashMap[TextDocument, Array[Int]]
+  implicit class DocumentOps(doc: TextDocument) {
+    def substring(range: Option[Range]): Option[String] = {
+      range.flatMap { range =>
+        if (doc.text.nonEmpty) {
+          val lineToOffset = lineToOffsetCache.getOrElseUpdate(doc, {
```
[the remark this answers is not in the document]
xeno-by: https://github.com/scalameta/scalameta/issues/1588

tests/jvm/src/test/resources/metac-metacp.expect.diff
```diff
@@ -1,186 +1,218 @@
...
+-            tag: TYPE_REF
+-            typeRef {
+-              symbol: "scala.annotation.unchecked.uncheckedVariance#"
 
 
-===============================
-com.javacp.MetacJava#Overload3#
-===============================
+============================
+classes.C4#copy$default$1().
```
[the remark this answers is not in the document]
xeno-by: Synthetics are a gray area for Scala, but here's my personal rule of thumb. 

There are synthetics that are necessary for typer to operate (e.g. implicit evidences, default getters, etc), and there are synthetics that are not necessary for typer to operate (e.g. extension methods from value classes, MODULE$, etc). In my opinion, all synthetics from the first category should be included and specified in SemanticDB.

Also, interestingly, default getters are included in SLS: https://www.scala-lang.org/files/archive/spec/2.12/04-basic-declarations-and-definitions.html#default-arguments.

tests/jvm/src/test/resources/metac-metacp.expect.diff
```diff
@@ -1,186 +1,218 @@
...
+             declarations {
+               hardlinks {
+-                symbol: "local14"
++                symbol: "advanced.D#_#"
+                 kind: TYPE
+                 properties: 4
+                 name: "_"
+       symlinks: "advanced.D#[CC]"
+     declarations {
++      symlinks: "advanced.D#_#"
```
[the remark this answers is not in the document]
xeno-by: Yeah, this is also an instance of https://github.com/scalameta/scalameta/issues/1586.

build.sbt
xeno-by: I included it for consistency. We could remove it if you'd like.

semanticdb/semanticdb3/semanticdb3.md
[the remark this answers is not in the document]
olafurpg: Talking with colleagues and reading `man ln` seems to indicate that symlinks are indeed opposites, fair enough 👍

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
xeno-by: https://github.com/scalameta/scalameta/issues/1591

build.sbt
olafurpg: Let's remove it to avoid "two ways to do the same thing". Let's also update the "This is a big project with a lot of tests :)" message in build.sbt to reflect the current state of things

> - testsJVM/test      # run fast unit tests
> - testsJVM/slow:test # run slow unit tests

olafurpg: We can remove `testQuick` from that too.

xeno-by:
I'm willing to take you up on your suggestion and merge as is and then figure out hardlinked symbol format by rc1. I opened a ticket: https://github.com/scalameta/scalameta/issues/1592.

tests/jvm/src/test/resources/metac-metacp.expect.diff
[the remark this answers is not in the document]
olafurpg: Interesting, I wasn't aware the spec mentioned the encoding of default params. I agree with your assessment that some synthetics are unavoidable, and it seems default method fall in that camp 😢

olafurpg:
Let's remove `Fast` and update the test message in build.sbt, then this PR is good to merge 🔥

xeno-by:
Thank you so much for a very thorough review!! This is a huge PR with 94 commits, and I am very grateful that we can casually deal with things like that when there is a need. It is a genuine pleasure to work side by side with you and a great open-source success story ❤️ 

I'm willing to take you up on your suggestion to merge as is and follow up later. I believe that: 1) changes in this PR are a strict improvement over the current stable version of Scalameta, 2) the aforementioned follow-ups don't mean "throw the code away and rewrite from scratch while breaking clients", but rather "incrementally improve on the already sound idea". Therefore, I think it would be justified to merge even though the code isn't perfect yet. 

I have submitted detailed tickets that you requested, so that we won't forget whether/how to follow up. Of those tickets, I think that https://github.com/scalameta/scalameta/issues/1592 merits fixing before 4.0.0-RC1, and the rest can be delayed.

P.S. Of course, I'm not going to make a habit of casually submitting über PRs every other week. This time was an exception because the changes were so fundamental across our entire infrastructure. That also triggered some long delayed garbage collection, which contributed to the complexity of the patch, too.

PR #1602, opened 2018-06-12
Fix #1601

### LINKED ISSUE #1601: The option '-P:semanticdb:exclude:<string>.*' doesn't work with scalameta (semanticdb-scalac plugin) revision 4.0.0-M2
In trying to upgrade to scalameta rev 4.0.0-M2 from 3.7.4,  discovered that this :exclude option string is no longer taking effect. This issue is a blocker for the upgrade.

xeno-by:


semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/RemoveOrphanSemanticdbFiles.scala
```diff
@@ -12,30 +12,34 @@ import scala.meta.io.AbsolutePath
...
-        override def visitFile(file: Path, attrs: BasicFileAttributes): FileVisitResult = {
-          if (PathIO.extension(file.getFileName) == SemanticdbPaths.semanticdbExtension) {
-            val scalafile = SemanticdbPaths.toScala(AbsolutePath(file), sourceroot, targetroot)
-            if (!scalafile.isFile) {
-              Files.delete(file)
+  def process(config: SemanticdbConfig): Unit = {
+    val sourceroot = config.sourceroot
+    val targetroot = config.targetroot
+    val semanticdbroot = targetroot.resolve(SemanticdbPaths.semanticdbPrefix).toNIO
+    if (Files.exists(semanticdbroot)) {
```
olafurpg: nitpick, this is a moment where `if (!Files.exists(semanticdbroot)) return` might be nice to avoid the added indent 😉

Issue #1595, opened 2018-06-07
Remove numbers from artifact and package names related to SemanticDB


xeno-by:
When we introduced a dedicated artifact for the schema bindings, we weren't sure whether we would want to version Scalameta and schema in lockstep. Therefore, we called the artifact `semanticdb3` and the package `scala.meta.internal.semanticdb3`.

However, over time it's become clear that it's pretty inconvenient to version Scalameta and `semanticdb3` differently. Currently, versions are computed automatically from git history, which means that to do something different for `semanticdb3`, we had to assign versions by hand. That quickly got old, so we reverted to lockstep versioning.

Now when we're about to release Scalameta 4, it's a good time to codify the de-facto convention. I suggest that we call the artifact `semanticdb` and the package `scala.meta.internal.semanticdb`.

olafurpg:
I'm in favor of this change. We should still keep `TextDocument.schema` in order to provide a way to programmatically assert against the version of SemanticDB payloads.

xeno-by:
Totally. Perhaps, we could change `Schema` from an enum to a proper data structure that carries the version of the generator, e.g. 

```protobuf
message Version {
  string major = 1;
  string minor = 2;
  string patch = 3;
  string sha = 4;
  string timestamp = 5;
}
```

PR #1609, opened 2018-06-19
Remove scala.meta.{Symbol,Signature}

### LINKED ISSUE #1525: Remove scala.meta.{Symbol,Signature}
Now that SemanticDB symbols are primarily strings, we should no longer expose `scala.meta.{Symbol,Signature}` in the public API.

olafurpg:
Fixes #1525. This PR replaces usage of `scala.meta.{Symbol,Signature}` with `String`. The refactoring was straightforward in most cases except for `Symbol.Multi`. Multi symbols are not in the spec since they are not persisted and this PR does not change that.


build.sbt
```diff
@@ -364,18 +364,6 @@ lazy val treesJVM = trees.jvm
...
-  .in(file("scalameta/semanticdb"))
-  .settings(
-    publishableSettings,
-    description := "Scalameta semantic database APIs"
-  )
-  .nativeSettings(nativeSettings)
-  .dependsOn(common)
-lazy val semanticdbJVM = semanticdb.jvm
-lazy val semanticdbJS = semanticdb.js
-lazy val semanticdbNative = semanticdb.native
```
xeno-by: 😍 I will proceed with https://github.com/scalameta/scalameta/issues/1595 when I return from ScalaDays.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolOps.scala
```diff
@@ -8,34 +8,35 @@ import scala.meta.internal.{semanticdb3 => s}
 import scala.meta.internal.semanticdb3.Scala._
 import scala.meta.internal.semanticdb3.Scala.{Descriptor => d}
 import scala.util.control.NonFatal
+import scala.meta.internal.semanticdb3.Scala._
```
xeno-by: There is already the same import on line 8.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolOps.scala
```diff
@@ -8,34 +8,35 @@ import scala.meta.internal.{semanticdb3 => s}
...
+        if (sym.isOverloaded) return Symbols.Multi(sym.alternatives.map(_.toSemantic))
         if (sym.isModuleClass) return sym.asClass.module.toSemantic
         if (sym.isTypeSkolem) return sym.deSkolemize.toSemantic
         if (sym.isSemanticdbLocal) return freshSymbol(sym.pos.toMeta.input)
 
         val owner = sym.owner.toSemantic
         val signature = {
           if (sym.isMethod || sym.isUsefulField) {
-            m.Signature.Method(sym.name.toSemantic, sym.disambiguator)
+            Descriptor.Method(sym.name.toSemantic, sym.disambiguator)
```
xeno-by: The convention that I've been used in other parts of Scalameta is to say `d.Method`, `d.TypeParameter`, etc. There is already `import scala.meta.internal.semanticdb3.Scala.{Descriptor => d}` in scope in this file.

scalameta/semanticdb/shared/src/main/scala/scala/meta/semanticdb/Symbol.scala
```diff
@@ -1,203 +0,0 @@
...
-
-  final case class Local(id: String) extends Symbol {
-    override def toString = syntax
-    override def syntax = Symbols.Local(id)
-    override def structure = s"""Symbol.Local("$id")"""
-  }
-
-  final case class Global(owner: Symbol, signature: Signature) extends Symbol {
-    override def toString = syntax
-    override def syntax = Symbols.Global(owner.syntax, signature.syntax)
```
xeno-by: Now that this line is gone, I suggest we delete `def Global(owner: String, desc: String): String = ???` from the public API.

scalameta/semanticdb/shared/src/main/scala/scala/meta/semanticdb/Symbol.scala
```diff
@@ -1,203 +0,0 @@
...
-object Symbol {
-  case object None extends Symbol {
-    override def toString = syntax
-    override def syntax = Symbols.None
-    override def structure = s"""Symbol.None"""
-  }
-
-  final case class Local(id: String) extends Symbol {
-    override def toString = syntax
-    override def syntax = Symbols.Local(id)
```
xeno-by: Now that this line is gone, I suggest we delete `def Local(id: String): String = ???` from the public API.

semanticdb/semanticdb3/src/main/scala/scala/meta/internal/semanticdb3/Scala.scala
```diff
@@ -14,6 +13,8 @@ object Scala {
     def Global(owner: String, desc: String) = if (owner != RootPackage) owner + desc else desc
     def Local(id: Int): String = Local("local" + id.toString)
     def Local(symbol: String): String = symbol
+    def Multi(symbols: List[String]): String = symbols.mkString(";", ";", "")
+    def parse(symbol: String): (String, Descriptor) = DescriptorParser(symbol).swap
```
xeno-by: Looks like this method is only used in tests. If that's the intent, let's move the method into tests.

semanticdb/semanticdb3/src/main/scala/scala/meta/internal/semanticdb3/Scala.scala
```diff
@@ -22,6 +23,31 @@ object Scala {
     def isEmptyPackage: Boolean = symbol == Symbols.EmptyPackage
     def isGlobal: Boolean = !isNone && Descriptor.descriptorLasts.contains(symbol.last)
     def isLocal: Boolean = !isNone && !isGlobal
+    def isMulti: Boolean = symbol.startsWith(";")
+    def flattenMulti: List[String] = {
```
xeno-by: How about we rename this to `symbols` or `parts`?

semanticdb/semanticdb3/src/main/scala/scala/meta/internal/semanticdb3/Scala.scala
```diff
@@ -22,6 +23,31 @@ object Scala {
     def isEmptyPackage: Boolean = symbol == Symbols.EmptyPackage
     def isGlobal: Boolean = !isNone && Descriptor.descriptorLasts.contains(symbol.last)
```
xeno-by: This should be updated to take multi symbols into account.

semanticdb/semanticdb3/src/main/scala/scala/meta/internal/semanticdb3/Scala.scala
```diff
@@ -22,6 +23,31 @@ object Scala {
...
     def isLocal: Boolean = !isNone && !isGlobal
```
xeno-by: Ditto.

semanticdb/semanticdb3/src/main/scala/scala/meta/internal/semanticdb3/Scala.scala
olafurpg: `DescriptorParser.apply` is private, can I make it `private[meta]` instead?

xeno-by: Sure.

semanticdb/semanticdb3/src/main/scala/scala/meta/internal/semanticdb3/Scala.scala
olafurpg: I agree this name is not ideal, it replaced a `def flatten(sym: Symbol): Symbol` in metac but `String#flatten` is already used by the scala-library. I would like to have "multi" somewhere in there, how about `asMulti: List[String]`?

semanticdb/semanticdb3/src/main/scala/scala/meta/internal/semanticdb3/Scala.scala
olafurpg: You mean by adding `!isMulti && ...`? I don't think we should do `asMulti.exists(_.isGlobal)`

semanticdb/semanticdb3/src/main/scala/scala/meta/internal/semanticdb3/Scala.scala
xeno-by: Sounds good.

semanticdb/semanticdb3/src/main/scala/scala/meta/internal/semanticdb3/Scala.scala
olafurpg: Done, made it
```scala
    def isGlobal: Boolean =
      !isNone &&
        Descriptor.descriptorLasts.contains(symbol.last) &&
        (!isMulti || asMulti.exists(_.isGlobal))
    def isLocal: Boolean =
      !isNone &&
        !isGlobal &&
        (!isMulti || asMulti.exists(_.isLocal))
```

scalameta/semanticdb/shared/src/main/scala/scala/meta/semanticdb/Symbol.scala
olafurpg: Done.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolOps.scala
olafurpg: Removed, good catch!

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolOps.scala
olafurpg: Fixed.

semanticdb/semanticdb3/src/main/scala/scala/meta/internal/semanticdb3/Scala.scala
olafurpg: Done.

PR #1610, opened 2018-06-19
Upgrade to directories-jvm 10.

### LINKED ISSUE #1599: Update soc/directories-jvm
We are currently using version 5 of https://github.com/soc/directories-jvm and the latest version is 10. There were breaking changes between the two so we should upgrade to avoid binary conflicts with other libraries.

olafurpg:
To follow Windows and macOS conventions, we are expected to qualify the
project name with a domain name.  Given that "SemanticDB" is also used
in Emacs, there's a valid risk of conflict.

Fixes #1599


olafurpg:
cc/ @usrinivasan this should fix the class error you encountered in Scalafix.

PR #1611, opened 2018-06-19
Add TextDocument.md5

### LINKED ISSUE #1593: Add TextDocument.md5
This is a refined feature request based on https://github.com/scalameta/scalameta/issues/886

Currently, metac supports `-P:semanticdb:text:on` to include the full source file contents in `TextDocument.text`. Our benchmarks have revealed that embedding the full source contents is expensive and adds up to 15% overhead to **total** compilation compared to a baseline with `-P:semanticdb:text:off` accounting for 1/3 of the total overhead of enabling semanticdb-scalac. It might be worth investigating if we can lower the overhead by including an md5 fingerprint of the file contents instead and delegate to tools like scalafix to read the source file contents from disk.

olafurpg:
This allows users to use `-P:semanticdb:text:off` while ensuring that
SemanticDB files are in sync with file contents on disk.

Fixes #1593.


olafurpg:
This opens the question whether we should disable `TextDocument.text` by default. The MD5 fingerprint is sufficient for Scalafix to replace `.text`.

semanticdb/semanticdb3/guide.md
```diff
@@ -266,6 +266,19 @@ be passed through Scalac in the form of `-P:semanticdb:<option>:<value>`
...
+    <td>
+        <code>on</code>,<br/>
+        <code>off</code>
+    </td>
+    <td>
+      Specifies whether to save a hexadecimal formatted MD5 fingerprint of the source
+      file contents in <code>TextDocument.md5</code> (<code>on</code> for yes,
+      <code>off</code> for no).
+    </td>
+    <td><code>on</code></td>
```
xeno-by: If this is default, let's disable text:on by default.

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -130,9 +131,12 @@ SemanticDB payloads must include the version of the SemanticDB model in the
   </tr>
 </table>
 
-Code snippets referenced in `TextDocument` can be defined
-in one of three ways: 1) via a [URI](#uri) provided in `uri`,
-2) via a string provided in `text`, 3) via a combination of both.
+Code snippets referenced in `TextDocument` can be defined as any combination of
+the following three fields:
```
xeno-by: Isn't `uri` always mandatory?

semanticdb/semanticdb3/src/main/generated/scala/meta/internal/semanticdb3/Semanticdb3Proto.scala
```diff
@@ -40,93 +40,94 @@ object Semanticdb3Proto extends _root_.scalapb.GeneratedFileObject {
...
+  W5nZRIWCgZzeW1ib2wYAiABKAlSBnN5bWJvbBJKCgRyb2xlGAMgASgOMjYuc2NhbGEubWV0YS5pbnRlcm5hbC5zZW1hbnRpY2RiM
+  y5TeW1ib2xPY2N1cnJlbmNlLlJvbGVSBHJvbGUiNwoEUm9sZRIQCgxVTktOT1dOX1JPTEUQABINCglSRUZFUkVOQ0UQARIOCgpER
+  UZJTklUSU9OEAIiiwIKCkRpYWdub3N0aWMSPAoFcmFuZ2UYASABKAsyJi5zY2FsYS5tZXRhLmludGVybmFsLnNlbWFudGljZGIzL
+  lJhbmdlUgVyYW5nZRJQCghzZXZlcml0eRgCIAEoDjI0LnNjYWxhLm1ldGEuaW50ZXJuYWwuc2VtYW50aWNkYjMuRGlhZ25vc3RpY
+  y5TZXZlcml0eVIIc2V2ZXJpdHkSGAoHbWVzc2FnZRgDIAEoCVIHbWVzc2FnZSJTCghTZXZlcml0eRIUChBVTktOT1dOX1NFVkVSS
+  VRZEAASCQoFRVJST1IQARILCgdXQVJOSU5HEAISDwoLSU5GT1JNQVRJT04QAxIICgRISU5UEAQijAEKCVN5bnRoZXRpYxI8CgVyY
+  W5nZRgBIAEoCzImLnNjYWxhLm1ldGEuaW50ZXJuYWwuc2VtYW50aWNkYjMuUmFuZ2VSBXJhbmdlEkEKBHRleHQYAiABKAsyLS5zY
+  2FsYS5tZXRhLmludGVybmFsLnNlbWFudGljZGIzLlRleHREb2N1bWVudFIEdGV4dColCgZTY2hlbWESCgoGTEVHQUNZEAASDwoLU
+  0VNQU5USUNEQjMQAyo1CghMYW5ndWFnZRIUChBVTktOT1dOX0xBTkdVQUdFEAASCQoFU0NBTEEQARIICgRKQVZBEAJiBnByb3RvM
+  w=="""
```
xeno-by: Oh boy. Can't wait until https://github.com/scalapb/ScalaPB/pull/458 is ready.

semanticdb/semanticdb3/guide.md
olafurpg: What will that mean for metap? Should we support --sourceroot

semanticdb/semanticdb3/semanticdb3.md
olafurpg: If I read it correctly, the original text implied option 2) was valid which implied only `.text` with no `.uri` is OK. I can't see it mentioned that `uri` is required. What should we do?

semanticdb/semanticdb3/guide.md
xeno-by: If there's no text, then we don't prettyprint identifiers. I suggest we don't introduce additional options just yet.

semanticdb/semanticdb3/semanticdb3.md
xeno-by: Huh indeed you're right - too little sleep recently. How about not changing this paragraph at all? It's not like `md5` can be used to define code snippets - it's only verifying snippets that are defined by other means.

olafurpg:
I updated the PR to let metac use `text:off` by default and updated the spec to document `text`, `uri` and `md5` separately.

olafurpg:
The guide.md is not incorrect since the examples for `metac Test.scala` assume `text:on`. Should we?

- update the metap output for the case when `text:off`, or
- make `text:on` by default for the metac cli, or
- use `-P:semanticdb:text:on` in guide.md

xeno-by:
Let's go for the first option, i.e. update the metap output for the case when text:off.

xeno-by:
I propose we handle the guide update in a separate ticket to get some executable code out of the door asap.

PR #1612, opened 2018-06-20
Make metac index incremental.

### LINKED ISSUE #1528: semanticdb.semanticidx does not play nicely with incremental compilation
Steps to reproduce https://github.com/olafurpg/semanticdb-zinc/commit/60ac13efe7604c2177a3d4fc6523210fa3007615

We need to figure out how to make semantic.semanticidx handle incremental compilation.

olafurpg:
This commit fixes a bug in metac so that it plays nicely with
incremental compilation under zinc. We use bloop in the tests to
simplify the integration with zinc and reproduce the bug. This
setup provides a modestly rapid edit/test/debug workflow.

Notes:

- To make semanticdb.semanticidx incremental, we merge an existing index
  file if it exists with the index metadata that we are about to
  persist. Along the way, we prune outdated data that references
  non-existing URIs.
- Zinc depends on an outdated version of ScalaPB (v0.6), which we
  exclude in build.sbt in favor of the latest version of ScalaPB (v0.7).
  Fortunately, Zinc uses only ScalaPB APIs that are available in v0.7
- Zinc is only compiled against 2.12 so the incremental tests run only
  for 2.12.
- The incremental tests take ~10-15s to run so they're labeled as slow.
- Thanks to the new testing infrastructure, we're able to integration test
  that orphan SemanticDB files are indeed removed by metac.

Fixes #1528

### DIFF
--- a/scalameta/testkit/src/main/scala/scala/meta/testkit/StringFS.scala
+++ b/scalameta/testkit/src/main/scala/scala/meta/testkit/StringFS.scala
@@ -1,0 +1,94 @@
+package scala.meta.testkit
+
+import scala.meta.io.AbsolutePath
+import java.io.File
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.StandardOpenOption
+import scala.meta.internal.io.FileIO
+import scala.meta.io.RelativePath
+
+object StringFS {
+
+  /**
+    * Creates a temporary directory with a layout matching the markup in the string.
+    *
+    * Example syntax of the expected markup in the string:
+    * {{{
+    *   fromString("""
+    *   /build.sbt
+    *   lazy val core = project
+    *   /src/main/scala/core/Foo.scala
+    *   package core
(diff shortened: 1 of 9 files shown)


scalameta/testkit/src/main/scala/scala/meta/testkit/StringFS.scala
```diff
@@ -0,0 +1,94 @@
...
+    * Creates a temporary directory with a layout matching the markup in the string.
+    *
+    * Example syntax of the expected markup in the string:
+    * {{{
+    *   fromString("""
+    *   /build.sbt
+    *   lazy val core = project
+    *   /src/main/scala/core/Foo.scala
+    *   package core
+    *   object Foo.scala
```
xeno-by: Fancy!

olafurpg: I already use `StringFS` in scalafmt and scalafix for a long time, it's very helpful!

PR #1614, opened 2018-06-20
Simplify isX logic for symbols

olafurpg:


semanticdb/semanticdb3/src/main/scala/scala/meta/internal/semanticdb3/Scala.scala
```diff
@@ -14,21 +14,30 @@ object Scala {
       else desc.toString
     def Local(id: Int): String =
       "local" + id.toString
-    def Multi(symbols: List[String]): String =
-      symbols.mkString(";", ";", "")
+    def Multi(symbols: List[String]): String = symbols match {
+      case Nil => None
+      case head :: _ =>
+        if (head.isMulti) symbols.mkString(";")
+        else symbols.mkString(";", ";", "")
```
xeno-by: What if one of the symbols in the middle is a multi symbol?

olafurpg: Good question, fixed.

PR #1621, opened 2018-06-21
Improve metacp global cache handling

olafurpg:
I've encountered several difficult-to-reproduce bugs related to the global metacp cache while testing scalafix on larger project. This PR should hopefully improve the situation, but unfortunately I don't know how to best test it so there are no accompanying tests.

### DIFF
--- a/semanticdb/semanticdb/guide.md
+++ b/semanticdb/semanticdb/guide.md
@@ -438,7 +438,7 @@ following scheme:
     The default cache location depends on the OS and is computed using
     [directories-jvm](https://github.com/soc/directories-jvm).
     For example, on macOS the location for scala-library would be
-    `$HOME/Library/Caches/semanticdb/3.7.4/scala-library-FINGERPRINT.jar`.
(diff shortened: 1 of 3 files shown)


semanticdb/semanticdb/guide.md
```diff
@@ -438,7 +438,7 @@ following scheme:
     The default cache location depends on the OS and is computed using
     [directories-jvm](https://github.com/soc/directories-jvm).
     For example, on macOS the location for scala-library would be
-    `$HOME/Library/Caches/semanticdb/3.7.4/scala-library-FINGERPRINT.jar`.
+    `$HOME/Library/Caches/org.scalameta.SemanticDB/3.7.4/scala-library-FINGERPRINT.jar`.
```
xeno-by: Let's open a ticket about keeping Scalameta version in sync with guide.md.

olafurpg: This diff is unrelated to the scalameta version, I think it's exceptional enough that the cache location changed to not have to keep a not around here.

Issue #1540, opened 2018-05-22
Consider removing SymbolInformation.owner from the schema


xeno-by:
Now that we have a symbol parser in `semanticdb3`, `SymbolInformation.owner` seems redundant. We should consider removing it from the schema to avoid confusion and potential discrepancies with updates to the implementation of the symbol parser.

xeno-by:
Unfortunately, this doesn't work. `ScalaSymbolOps.owner` only returns correct results for global symbols. Local members, e.g. members of local classes, must obtain their owner information elsewhere.

xeno-by:
I have experimented with this in https://github.com/xeno-by/scalameta/commit/f435f58b347f17c9c3c9c49218427644c63c1125 and must say that this is not a clear-cut win. Scalac's notion of owners is pretty crazy, so we'd have to clean it up and spec before readding SymbolInformation.owner. Since this doesn't look critical for our work in 4.0.0, I'm removing the milestone assignment.

olafurpg:
One situation where we may need `SymbolInformation.owner` is to handle static inner java classes. The proposed solution for #1491 is to give the symbol syntax `java.util.Map.Entry#` for the inner class `Entry` while it's owner should be `java.util.Map#`.

xeno-by:
The argument that @olafurpg has mentioned above is no longer applicable. In https://github.com/scalameta/scalameta/pull/1638, we went the other way and retained syntactic owners for global symbols.

xeno-by:
I'm working on pickling deeply nested structural/existential types, and I can attest that the lack of `SymbolInformation.owner` is pretty annoying. I'm left with no choice but to maintain an owner chain while I'm traversing SemanticDB payloads. Perhaps, we could reintroduce `owner`, but only fill it in for local symbols?

olafurpg:
I have experienced the same problem, I am in favor of adding `.owner` but only fill it for local symbols 👍

olafurpg:
Encountering the lack of `owner` when implementing rename in Metals for anonymous classes. A workaround is to filter out class symbols that contain the local method symbol as part of `ClassSignature.declarations`.

jozanek:
Metals recovers a local symbol's owner by scanning all `SymbolInformation`s for a `ClassSignature.declarations` containing it, and `RenameProvider` refuses local classes outright.

Concrete proposal (per the consensus above): reintroduce owner **only for local symbols**:

```protobuf
string owner = 21;
```

[...]

PR #1629, opened 2018-06-24
Change symbol syntax for static Java inner classes and interfaces.

### LINKED ISSUE #1491: Scalacp emits wrong symbols for Java references
For the signature `  val f: java.util.Map.Entry[Int, Int]` metacp produces a wrong java symbol
```diff
- metacp
+ metac
_empty_.a.f(). => val method f: : Entry[Int, Int]
-  Entry => java.util.Map.Entry#
+  Entry => java.util.Map#Entry#
```

olafurpg:
Fixes #1491


semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
```diff
@@ -350,7 +360,7 @@ object Javacp {
     asmName
   }
 
-  private def ssym(asmName: String): String = {
+  private def ssym(asmName: String, scope: Scope): String = {
```
xeno-by: How about turning Javacp from an object to a class like Scalacp. This way we won't have to pass `scope` to `ssym`, and the `Scope` itself won't have to be stretched to also accommodate inner classes (which turns it from a `Scope` to more like a `SymtabLite`).

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolOps.scala
```diff
@@ -21,7 +22,11 @@ trait SymbolOps { self: SemanticdbOps =>
         if (sym.isTypeSkolem) return sym.deSkolemize.toSemantic
         if (sym.isSemanticdbLocal) return freshSymbol(sym.pos.toMeta.input)
 
-        val owner = sym.owner.toSemantic
+        val owner = {
+          val result = sym.owner.toSemantic
+          if (sym.isJavaStaticInnerClass) result.stripSuffix("#") + "."
+          else result
+        }
```
xeno-by: Ooooohhh... The fact that owners become non-syntactic is super unfortunate. If we follow this route, we pretty much must have https://github.com/scalameta/scalameta/issues/1540 on our roadmap for RC1.

semanticdb/semanticdb/specification.md
```diff
@@ -2257,6 +2257,10 @@ In this section, we describe the Java symbol format.
         <li>
           For top-level package, its descriptor.
         </li>
+        <li>
+          For static inner classes and interfaces, concatenation of
+          of owner's simple name, dot (`.`) and definition descriptor.
```
xeno-by: A simple name for `foo.Bar#` is just `Bar`, which is probably not what you wanted. Probably instead you meant "concatenation of the owner's owner symbol, owner's simple name, dot (`.`) and definition descriptor".

tests/jvm/src/test/resources/metac-metacp.expect.diff
```diff
@@ -210,27 +210,6 @@ example.Acc#
...
-example.MetacJava#staticInner().
-================================
---- metac
-+++ metacp
-       typeRef {
-         prefix {
--        symbol: "com.javacp.MetacJava#StaticInner#"
-+        symbol: "com.javacp.MetacJava.StaticInner#"
-
-
```
xeno-by: This is a small diff, but it took a lot of effort! 👍

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolOps.scala
xeno-by: Olaf and I had a chat, and the non-syntactic owner thing looks like a deal-breaker, since it: 1) will require us to give up `String.owner` and `Symbols.Global(owner, desc)`, which will be a pretty big deal internally, 2) will require a symbol table to construct a symbol from a string.

olafurpg:
The changes to `.owner` is a deal breaker. Back to the whiteboard brainstorming solutions 😊

Issue #1460, opened 2018-03-27
Store Java-defined members in their declaration order


xeno-by:
Currently, the SemanticDB spec imposes a weird requirement on the order of members in Java-defined ClassInfoTypes. All non-static members must come first, and only after them come the static members.

This is an artifact of how Scalac models Java classes (they are split into a class symbol with all non-static members and an object symbol with all static members, with no way to restore the original order). And we care about this here, because we need member order to compute stable disambiguators.

When reviewing the Java portion of the SemanticDB spec, I realized that if the only reason for reordering members is disambiguators, then maybe we should order members in a natural way in the general case and then use the weird order virtually, only when computing disambiguators. Let's try this out.

xeno-by:
@olafurpg Do we want this in 4.0.0?

PR #1631, opened 2018-06-24
Preserve declaration order of methods in javacp.

olafurpg:
Fixes https://github.com/scalameta/scalameta/issues/1460.
Note that we can't preserve interleaved order of fields/methods since
ASM separates them in different lists. If we write a custom ASM
ClassNodeVisitor we could probably retrieve it but that can be left for
another PR.

### DIFF
--- a/semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
+++ b/semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
@@ -128,11 +128,6 @@ object Javacp {
       }
     }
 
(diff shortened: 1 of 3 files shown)


olafurpg:
> Can you add a test where fields are also involved? The thing is that fields should not qualify as overloads, since they have a different namespace in Java and different symbol format in SemanticDB.

What behavior should the test assert?  That they appear in declaration order? All fields currently appear before all methods

olafurpg:
Fixed a minor typo in the markdown update and accidentally force pushed. The change was tiny.

PR #1632, opened 2018-06-24
Move spec and guide to old location

### LINKED ISSUE #1630: Move SemanticDB spec and guide to the old location
One of the recent PRs moved the spec from `semanticdb3/semanticdb3.md` to `semanticdb/specification.md` and the guide from `semanticdb3/guide.md` to `semanticdb/guide.md`. This made sense to me at the time, since `semanticdb3/` is now dead (code moved to `semanticdb/`).

Since then, I've learned about the upcoming work on centralizing and improving documentation about Scalameta. The approach that @olafurpg wants to pursue is to have everything in `docs/` and then use one of the website generators to produce a website from that.

Given that the upcoming work is planned shortly after the release of v4, I don't see a point of moving the docs into a new place and then have them moved again in a few weeks. Let's move the docs (and just docs, not the proto file) back to where they were.

olafurpg:
Fixes #1630


xeno-by:
LGTM. Merging without waiting for CI, because no code is affected.

PR #1635, opened 2018-06-26
Update the spec regarding singleton types

xeno-by:
Looks like I forgot to update the spec when implementing singleton types 😅

### DIFF
--- a/semanticdb/semanticdb3/semanticdb3.md
+++ b/semanticdb/semanticdb3/semanticdb3.md
@@ -312,10 +312,13 @@ consumers.
 
 ```protobuf
 message Type {
-  reserved 1, 3, 4, 5, 6, 11, 12, 15;
+  reserved 1, 3, 4, 5, 6, 11, 12, 15, 16;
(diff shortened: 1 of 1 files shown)


semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -351,35 +354,46 @@ references to nested type definitions, e.g. path-dependent types in Scala,
...
+```
+
+`SuperType` is a singleton type that is inhabited only by the value
+of `super` reference to the definition represented by `symbol`,
+prefixed by an optional `prefix`, only visible inside the lexical extent
+of the enclosing definition.
+
+```protobuf
+message ConstantType {
+  Constant constant = 1;
```
olafurpg: The spec is missing the definition of `Constant`

xeno-by: Done. I meant to discuss whether the triviality of constants merits not including them in the spec, but it was simpler to just go ahead and add them.

PR #1638, opened 2018-06-27
Fix scalacp references to static inner Java classes

olafurpg:
This commit fixes a bug that resulted in scalacp producing references
external static Java inner classes as if they were Scala object symbols.
For example, previously scalacp produced the symbol `java.util.Map.Entry#`
while javacp produced the expected symbol `java.util.Map#Entry#`.

This issue turned out to be tricky to solve because Scala signatures
don't distinguish between references to external static Java inner
classes and Scala inner classes inside of objects. The solution in this
commit is to require the full dependency classpath in metacp and query
the classfiles whether they are Java or Scala defined.

The overhead of scanning the full classpath is somewhat costly but
manageable, for a classpath of ~200mb it takes around 200ms on my
laptop. I expected the overhead to be more significant.

While fixing this issue, I had to implement missing functionality to
handle classpaths:

- `PathIO.{basename,dirname}` to extract directory and base name of string paths
- `ClasspathEntry` as an abstraction over a single package, file or zip
  entry in the classpath.
- `ClasspathLookup` to scan a full classpath and resolve paths to
  `ClasspathEntry`.  I started by using `java.net.URLClassLoader` but it
  it not correctly handle jar files that are missing directory entries
  (which includes the JDK jars).  ClasspathLookup can handle classpaths
  entries that are directories, regular jars and manifest jars.

The default metacp behavior for missing symbols is to fail fast with a
helpful error message. However, some classpaths can be incomplete for
several reasons. For example, some libraries embed shaded dependencies
in which case the Scala signatures will point to non-existing
classfiles.  To be able to recover from missing symbols, metacp allows
users to provide the flags `--assume-java <package>` and
`--assume-scala <package>` to tell metacp to assume symbols in those
packages are Scala/Java defined. This option was required to keep ConvertSuite green, since
both akka-stream and spark-sql referenced symbols that were missing in
the classpath.


olafurpg:
I would like to add more tests since this PR introduced a lot of new functionality. Suggestions for tricky cases are welcome. I decided to open this PR right now to unblock review.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ClasspathEntry.scala
```diff
@@ -0,0 +1,46 @@
...
+import java.io.File
+import java.io.FilterInputStream
+import java.io.InputStream
+import java.nio.file.Files
+import java.util.zip.ZipEntry
+import java.util.zip.ZipFile
+import scala.collection.mutable
+import scala.meta.io.AbsolutePath
+
+/** Represents a entry in a classpath that is either a package, file on disk or zip entry. */
```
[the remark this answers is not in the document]
xeno-by: Given that, I suggest alternative names:
  * ClasspathEntry => ClasspathElement
  * ReadableClasspathEntry => Classfile
  * Package => Classdir (this class and classes below can become top-level, since their names become unique enough)
  * OnDisk => UncompressedClassfile
  * InJar => CompressedClassfile

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ClasspathLookup.scala
```diff
@@ -0,0 +1,134 @@
...
+import scala.meta.io.Classpath
+import scala.meta.internal.io.PathIO
+
+final class ClasspathLookup private (val dirs: collection.Map[String, ClasspathEntry.Package]) {
+
+  override def toString: String = {
+    s"ClasspathLookup(${dirs.size} entries)"
+  }
+
+  def getEntry(path: String): Option[ReadableClasspathEntry] = {
```
xeno-by: If you accept the naming suggestions from above, it would be consistent to rename this method to `getClassfile`.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ClasspathLookup.scala
```diff
@@ -0,0 +1,134 @@
...
+    getEntry(PathIO.dirname(path), PathIO.basename(path))
+  }
+
+  def getEntry(directory: String, filename: String): Option[ReadableClasspathEntry] = {
```
xeno-by: Looks like `getClassfile`, too.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ClasspathLookup.scala
```diff
@@ -0,0 +1,134 @@
...
+      case Some(pkg) =>
+        pkg.members.get(filename).collect {
+          case e: ReadableClasspathEntry => e
+        }
+      case _ =>
+        None
+    }
+  }
+
+  def isPackage(name: String): Boolean =
```
xeno-by: `isClassdir`

olafurpg:
The windows failures are legitimate, they seem to be caused by 
```scala
      def relpath(dir: Path): String =
        root.toNIO.relativize(dir).toString + "/"
```
in `ClasspathLookup.apply`

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ClasspathLookup.scala
```diff
@@ -0,0 +1,134 @@
...
+import java.nio.file.Path
+import java.nio.file.SimpleFileVisitor
+import java.nio.file.attribute.BasicFileAttributes
+import java.util.jar.JarFile
+import scala.collection.mutable
+import scala.meta.io.AbsolutePath
+import scala.meta.io.Classpath
+import scala.meta.internal.io.PathIO
+
+final class ClasspathLookup private (val dirs: collection.Map[String, ClasspathEntry.Package]) {
```
xeno-by: From the name `ClasspathLookup`, I'd expect the class to encapsulate the results of a single lookup in a classpath. I propose to call this class `ClasspathIndex` instead.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ToplevelClassfile.scala
```diff
@@ -1,9 +1,18 @@
...
 
-final case class ToplevelClassfile(base: AbsolutePath, path: AbsolutePath, node: ClassNode) {
+final case class ToplevelClassfile(
+    base: AbsolutePath,
+    path: AbsolutePath,
+    node: ClassNode,
+    lookup: ClasspathLookup,
+    settings: Settings,
+    reporter: Reporter
+) {
```
xeno-by: I believe that `lookup`, `settings` and `reporter` should be separate from `ToplevelClassfile` and be passed as parameters to `Scalacp` and `Javacp`. This is somewhat more hassle than the current arrangement, but this will prevent us from ending up with a Global/Context design, where we have a single data structure that is a grab bag of stuff from a bunch of different submodules.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
```diff
@@ -9,7 +9,7 @@ import scala.meta.internal.semanticdb.SymbolInformation.{Kind => k}
 import scala.meta.internal.semanticdb.Scala._
 import scala.tools.scalap.scalax.rules.scalasig._
 
-class Scalacp private (classfile: ToplevelClassfile)
+class Scalacp private (val classfile: ToplevelClassfile)
```
xeno-by: Why does this need to become a val?

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/package.scala
```diff
@@ -1,18 +1,37 @@
...
 import scala.tools.asm._
 import scala.tools.asm.ClassReader._
 import scala.tools.asm.tree._
 
 package object metacp {
+  private val SKIP_ALL = SKIP_CODE | SKIP_DEBUG | SKIP_FRAMES
+
+  implicit class XtensionAsmBytes(entry: ReadableClasspathEntry) {
+    def hasScalaSig: Boolean = {
+      val visitor = new HasScalaSigVisitor
```
xeno-by: Let's use `toClassNode` here instead of rolling our own visitor. This will get rid of a class (`HasScalaSigVisitor`) that's not used anywhere else.

xeno-by: Also, this logic is inconsistent with internal/metacp/Main.scala, which checks for both `ScalaSig` and `Scala` classfile attrs.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalalib.scala
```diff
@@ -87,7 +87,7 @@ object Scalalib {
     )
     val syntheticBase = PathIO.workingDirectory
     val syntheticPath = syntheticBase.resolve("scala/" + name + ".class")
-    val syntheticClassfile = ToplevelClassfile(syntheticBase, syntheticPath, null)
+    val syntheticClassfile = ToplevelClassfile(syntheticBase, syntheticPath, null, null, null, null)
```
xeno-by: See the comment to ToplevelClassfile.scala.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/SymbolOps.scala
```diff
@@ -1,16 +1,78 @@
...
-import scala.meta.internal.semanticdb.Scala.{Names => n}
 import scala.meta.internal.semanticdb.SymbolInformation.{Kind => k}
 import scala.tools.scalap.scalax.rules.scalasig._
 
 trait SymbolOps { self: Scalacp =>
   lazy val symbolCache = new HashMap[Symbol, String]
+  lazy val isJavaDefined = mutable.Map.empty[Symbol, Boolean]
   implicit class XtensionSymbolSSymbol(sym: Symbol) {
+
+    def isReallyPackage: Boolean = {
```
xeno-by: Let's move these methods to `XtensionSymbol`. This implicit class, as well as its neighbor right below, are focused on high-level concepts from the SemanticDB spec.

xeno-by: Also, it looks like none of the methods except `isJavaDefined` are used outside this class. If that's the case, I suggest to make these methods private to simplify understanding of the code.

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/ExpectSuite.scala
```diff
@@ -45,7 +46,7 @@ class ExpectSuite extends FunSuite with DiffAssertions {
         import MetacIndexExpect._
         assertNoDiff(loadObtained, loadExpected)
       }
-      test("metac-metacp.expect.diff") {
+      test("metac-metacp.diff") {
```
xeno-by: Why change the test name?

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/ExpectSuite.scala
```diff
@@ -133,6 +134,11 @@ trait ExpectHelpers extends FunSuiteLike {
         .Settings()
         .withCacheDir(AbsolutePath(target))
         .withClasspath(Classpath(AbsolutePath(in)))
+        .withDependencyClasspath(
+          Classpath(
+            Library.scalaLibrary.classpath().entries ++
+              Library.jdk.classpath().entries
+          ))
```
xeno-by: Can this be written as `Library.scalaLibrary.classpath() ++ Library.jdk.classpath()`?

tests/jvm/src/test/scala/scala/meta/tests/metacp/ClasspathLookupSuite.scala
```diff
@@ -0,0 +1,61 @@
...
+
+import org.scalatest.FunSuite
+import scala.meta.internal.io.PathIO
+import scala.meta.internal.metacp.ClasspathLookup
+import scala.meta.io.AbsolutePath
+import scala.meta.io.Classpath
+import scala.meta.tests.BuildInfo
+import scala.meta.tests.semanticdb.ManifestMetacp
+
+class ClasspathLookupSuite extends FunSuite {
```
xeno-by: Very thorough tests! 👍

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -132,6 +135,7 @@ class Main(settings: Settings, reporter: Reporter) {
                 infos.save(out)
               }
             } catch {
+              case e: ClasspathLookup.Error => throw e
```
xeno-by: Why give special treatment to this exception?

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/SymbolInformationOps.scala
```diff
@@ -66,7 +66,8 @@ trait SymbolInformationOps { self: Scalacp =>
             }
           }
           val isModuleClass = sym.entry.entryType == 10
-          if (hasTermName || isModuleClass) k.OBJECT
+          def isScalaObject: Boolean = isModuleClass && !sym.isJavaDefined
+          if (hasTermName || isScalaObject) k.OBJECT
```
xeno-by: 👍 👍 👍

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/ExpectSuite.scala
olafurpg: To play nicer with scalatest -z option that allows you to filter tests by regexp. I wasn't able to run metacp.expect standalone before.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ClasspathEntry.scala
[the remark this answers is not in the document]
olafurpg: I like these suggestions. Will update the names 👍

tests/jvm/src/test/scala/scala/meta/tests/metacp/ClasspathLookupSuite.scala
olafurpg: Was gratifying to see it catch a windows bug!

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/ExpectSuite.scala
xeno-by: The test is called like that after the file it's loading. Let's rename that file, then.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
xeno-by: Can we force `lookup` in `Main.process` then, as opposed to doing that in `convertClasspathEntry`? This way this catch clause will become unnecessary.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ToplevelClassfile.scala
xeno-by: No wrapper class, since these entities are unrelated. Just three independent parameters, in addition to `classfile`.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
xeno-by: Looks like that will no longer be necessary, once lookup becomes a standalone member of Scalacp.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
olafurpg: I made `Settings` more strict to avoid this catch

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ToplevelClassfile.scala
olafurpg: Sounds good 👍 much better!

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/package.scala
olafurpg: Are you OK with keeping it? I think it's fine to write custom visitor

olafurpg: I've synchronised the two

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
olafurpg: We still need access to `settings` and `index`, do you really think it's worth writing forwarder methods in Scalacp?

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/SymbolOps.scala
olafurpg: Done.

olafurpg: Only `ownerChain` can be privatized, we use the other ones via `parent.isReallyPackage`

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/ExpectSuite.scala
olafurpg: Done.

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/ExpectSuite.scala
olafurpg: Yes it can! So much nicer.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/HasScalaSigVisitor.scala
```diff
@@ -0,0 +1,15 @@
+package scala.meta.internal.metacp
+
+import scala.tools.asm.Attribute
+import scala.tools.asm.ClassVisitor
+import scala.tools.asm.Opcodes
+
+final class HasScalaSigVisitor extends ClassVisitor(Opcodes.ASM5) {
```
xeno-by: I'd prefer to remove this class and use `ClassNode` to: 1) be consistent with how things are done in Main, 2) shrink the conceptual surface of the module.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ToplevelClassfile.scala
```diff
@@ -4,6 +4,10 @@ import scala.meta.internal.io._
 import scala.meta.io._
 import scala.tools.asm.tree._
 
-final case class ToplevelClassfile(base: AbsolutePath, path: AbsolutePath, node: ClassNode) {
+final case class ToplevelClassfile(
+    base: AbsolutePath,
+    path: AbsolutePath,
+    node: ClassNode
+) {
```
xeno-by: I'm surprised that Scalafmt doesn't undo these changes to formatting, now that `ToplevelClassfile` has three parameters again.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ClasspathIndex.scala
```diff
@@ -0,0 +1,138 @@
...
+      new ClasspathIndex(dirs)
+    }
+
+    private def expandPath(path: AbsolutePath): Unit = {
+      if (path.isFile) expandJar(path)
+      else if (path.isDirectory) expandDirectory(path)
+      else throw new IllegalArgumentException(path.toString)
+    }
+
+    private def getPackage(name: String): Classdir = {
```
xeno-by: `getClassdir`?

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ClasspathIndex.scala
```diff
@@ -0,0 +1,138 @@
...
+        case _ =>
+          val parent = getPackage(PathIO.dirname(name))
+          val entry = Classdir(name)
+          parent.members(PathIO.basename(name)) = entry
+          dirs(name) = entry
+          entry
+      }
+    }
+
+    private def expandJar(jarpath: AbsolutePath): Unit = {
```
xeno-by: How about calling this method `expandJarEntry` and the method below `expandDirEntry`? That will disambiguate `directory` in `expandDirectory` between a directory entry and a directory element.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/SymbolOps.scala
```diff
@@ -1,16 +1,20 @@
...
 import scala.meta.internal.semanticdb.Scala._
 import scala.meta.internal.semanticdb.Scala.{Descriptor => d}
-import scala.meta.internal.semanticdb.Scala.{Names => n}
 import scala.meta.internal.semanticdb.SymbolInformation.{Kind => k}
 import scala.tools.scalap.scalax.rules.scalasig._
 
-trait SymbolOps { self: Scalacp =>
+trait SymbolOps { scalacp: Scalacp =>
   lazy val symbolCache = new HashMap[Symbol, String]
+  lazy val isJavaDefined = mutable.Map.empty[Symbol, Boolean]
```
xeno-by: `private lazy val javaDefinedCache`? This way, you won't have to disambiguate with the `javaDefined` method.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/SymbolOps.scala
```diff
@@ -144,6 +148,60 @@ trait SymbolOps { self: Scalacp =>
   }
 
   implicit class XtensionSymbol(sym: Symbol) {
+    def isJavaDefined: Boolean = scalacp.isJavaDefined.get(sym) match {
+      case Some(cachedResult) =>
+        cachedResult
+      case _ =>
+        val result =
+          if (sym.isReallyPackage) false
```
xeno-by: What does `really` mean here?

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/SymbolOps.scala
```diff
@@ -144,6 +148,60 @@ trait SymbolOps { self: Scalacp =>
...
+          else if (sym.isPackageObject) false
```
xeno-by: `sym.isPackageObject` is defined as `def isPackageObject: Boolean = sym.name == "package"`. Is there something in javac that prevents programmers from calling their classes `package`?

olafurpg:
Needed to rebase on top of master to make tests compile.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/HasScalaSigVisitor.scala
olafurpg: Removed.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ToplevelClassfile.scala
olafurpg: Scalafmt keeps "config style" if it's there https://scalameta.org/scalafmt/#Configstyle It's nice to be able to force breaks even when the class could fit in a single line.

I manually undid these changes.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/SymbolOps.scala
olafurpg: Done.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/SymbolOps.scala
olafurpg: Updated with comment and renamed to `isPackageAccordingToClasspath`
```scala
    // scalap Symbol.isPackage returns a constant false for external symbols so we have to
    // query the classpath to know which symbols are truly packages.
```

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/SymbolOps.scala
olafurpg: Good catch. I removed this guard, it was there from an early experiment that was obsolete once we gained the ability to query the classpath index.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ClasspathIndex.scala
olafurpg: Done.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ClasspathIndex.scala
olafurpg: Good idea, done.

PR #1639, opened 2018-06-27
add escape hatch allow building on JDK 9+

SethTisue:
this is needed for the Scala community build. we at least need every
build definition to be loadable on JDK 9+ (even if the project doesn't
actually build, tests pass, etc)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -28,8 +28,10 @@ version.in(ThisBuild) ~= { old =>
 }
 name := {
   println(s"[info] Welcome to scalameta ${version.value}")
-  val javaVersion = sys.props("java.specification.version")
-  if (javaVersion != "1.8") sys.error(s"Obtained Java version $javaVersion. Expected 1.8")
+  if (!sys.props.isDefinedAt("scalameta.allow-any-jdk")) { // escape hatch for Scala community build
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -28,8 +28,10 @@ version.in(ThisBuild) ~= { old =>
 }
 name := {
   println(s"[info] Welcome to scalameta ${version.value}")
-  val javaVersion = sys.props("java.specification.version")
-  if (javaVersion != "1.8") sys.error(s"Obtained Java version $javaVersion. Expected 1.8")
+  if (!sys.props.isDefinedAt("scalameta.allow-any-jdk")) {  // escape hatch for Scala community build
+    val javaVersion = sys.props("java.specification.version")
+    if (javaVersion != "1.8") sys.error(s"Obtained Java version $javaVersion. Expected 1.8")
+  }
```
xeno-by: @olafurpg Do you remember why we added this check?

olafurpg: I don't remember, it could be this assertion is to guard against errors on JDK 7. I prefer to keep the assertion by default since users on Java 9-10 are likely to encounter errors and I prefer to address the breaking changes once Java 11 is out.

SethTisue:
squashed, rebased.

Issue #1647, opened 2018-07-01
Use slash / for package symbols


olafurpg:
Currently, SemanticDB package symbols use `.` like for `scala` in `scala.Predef.String#`. This is problematic since it's not possible to distinguish packages from Scala objects like `Predef`. I propose we change the format to be `scala/Predef.String#` instead. The benefit of this strategy is that we can stop persisting package symbols altogether and we can implement lookup for packages with a stateless method like this

```scala
def info(symbol: String): SymbolInformation =
  if (symbol.endsWith("/")) SymbolInformation(symbol, k.PACKAGE)
  else ???
```

We currently have problems like https://github.com/scalameta/scalameta/issues/1420 where package symbols are attributed different languages in metac/scalacp/javacp.

A downside to this change is that it's less natural to write `scala/collection/immutable/List#` than `scala.collection.immutable.List#` and it's breaking with existing code that hardcodes strings. To smoothen the breaking change, it's possible to support `a.b.c.D#` symbols by automatically converting them to `a/b/c/D#` behind the scenes if you know which which symbols are packages.

xeno-by:
I'm tentatively assigning this ticket to RC1 while we're figuring out whether we want to tackle it at the 11th hour of the release.

PR #1656, opened 2018-07-02
Persist symbols even if they have no matching occurrence

### LINKED ISSUE #1495: Metac does not emit symbol information for higher kinded wildcard type parameter
```scala
$ cat wildcard.scala
class A[C[_]]
```
In metac, the wildcard type parameter seems to get a symbol ending with `[_]` but no SymbolInformation.

```
metac wildcard.scala && metap -proto META-INF/semanticdb/wildcard.scala.semanticdb | grep 'symbol:.*A#\[' -A 20
warning: there was one feature warning; re-run with -feature for details
one warning found
  symbols {
    symbol: "_empty_.A#[C]"
    kind: TYPE_PARAMETER
    properties: 0
    name: "C"
    tpe {
      tag: TYPE_TYPE
      typeType {
        type_parameters: "_empty_.A#[C][_]" // <-- missing SymbolInformation
        lower_bound {
          tag: TYPE_REF
          typeRef {
            symbol: "scala.Nothing#"
          }
        }
        upper_bound {
          tag: TYPE_REF
          typeRef {
            symbol: "scala.Any#"
          }
        }
      }
```
Metacp seems to do the right thing
```
$ scalac wildcard.scala
$ metap -proto $(metacp .) | grep 'symbol:.*A#\[' -A 20
    symbol: "_empty_.A#[C][_]"
    kind: TYPE_PARAMETER
    properties: 0
    name: "_"
    tpe {
      tag: TYPE_TYPE
      typeType {
        lower_bound {
          tag: TYPE_REF
          typeRef {
            symbol: "scala.Nothing#"
          }
        }
        upper_bound {
          tag: TYPE_REF
          typeRef {
            symbol: "scala.Any#"
          }
        }
      }
    }
```

For the record, it's good to look at the scala.meta.Tree structure for this input
```scala
@ q"class Foo[C[_]]"
res0: Defn.Class = Defn.Class(
...
Type.Param(Nil, Type.Name("C"), List(Type.Param(Nil, Name(""), Nil, Type.Bounds(None, None), Nil, Nil)
```

### LINKED ISSUE #1504: Metac should emit global symbol definitions even if they don't have occurrences
While testing the new hover implementation in metals with semanticdb v3.7, I hit on 
```
java.util.NoSuchElementException: org.langmeta.lsp.Hover#
```
which can be explained by the fact that `@JsonCodec class Hover` has a macro annotation

https://github.com/scalameta/metals/blob/fcac1cc31fedf0a2dd8a1c723cc97c59a398edca/lsp4s/src/main/scala/org/langmeta/lsp/Commands.scala#L201

The annotation is messing up with positions causing there to be no matching occurrence for the definition. This breaks both goto-definition as well as symbol lookups.

- Metac should emit SymbolInformation for ALL definitions, regardless if they have a matching occurrence or not
- Metac should emit a SymbolOccurrence for global definitions even if they have no matching position in g.Tree, this should accommodate macro annotations that mess up positions

In a nutshell, metac should syntactically emit symbol occurrences for global definitions like tags does.

olafurpg:
Fixes #1504.
Fixes #1495.

This PR changes the strategy metac emits symbol informations from "persist if there exists a matching occurrence" to "always persist unless the symbol is useless". This change is required to properly persist symbols that are generated by macro annotations. This change also fixed a number of "unknown" errors in the metap output for `metac.expect`. The diff is fairly large since the number of local symbols has increased.


tests/jvm/src/test/resources/metac.expect
```diff
@@ -1128,24 +1157,43 @@ classes.M.C5#`<init>`().(x) => param x: Int
...
-  local => local1
+  local => local9
   Nothing => scala.Nothing#
 classes.N.anonFun(). => val method anonFun: List[Int]
   List => scala.collection.immutable.List#
   Int => scala.Int#
-local0 => val method local: Nothing
-  Nothing => scala.Nothing#
-local2 => param i: Int
+local0 => val local C1$1: C1
```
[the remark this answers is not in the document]
olafurpg: Decided to push this to a separate ticket instead https://github.com/scalameta/scalameta/issues/1659

I think we need to come up with a consistent solution for how to handle these situations.

tests/jvm/src/test/resources/metac.expect
```diff
@@ -2550,23 +2639,30 @@ Schema => SemanticDB v4
...
 local1 => abstract method k: Int
   Int => scala.Int#
-local10 => typeparam T
-local2 => method k: Int
+local10 => abstract type T
+local12 => type L[T] = List[T]
+  T => local13
+  List => scala.package.List#
+local13 => typeparam T
+local14 => val local RepeatedType$1: RepeatedType
```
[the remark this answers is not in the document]
olafurpg: Moved to #1659

tests/jvm/src/test/scala-2.12/scala/meta/tests/semanticdb/IncrementalSuite.scala
```diff
@@ -35,6 +35,7 @@ class IncrementalSuite extends FunSuite with BeforeAndAfterEach with DiffAsserti
        |
        |Toplevels:
        |==========
+       |_empty_. => src/A.scala.semanticdb
```
olafurpg: Note, these symbols will be removed after #1647

xeno-by: Sounds good.

tests/jvm/src/test/resources/metac.expect
```diff
@@ -2377,6 +2458,9 @@ local1 => selfparam self: B
 local2 => selfparam self: B with Int {}
   B => selfs.B#
   Int => scala.Int#
+local3 => selfparam x$1
+local4 => selfparam x$2: B
+  B => selfs.B#
```
xeno-by: @olafurpg Can we open a ticket for these guys as well?

olafurpg: Done https://github.com/scalameta/scalameta/issues/1662

I didn't observe these were generated from the wildcards, I agree they should not be persisted.

Issue #1625, opened 2018-06-22
Metacp emits reference to non-existent term symbol for val forwarders


olafurpg:
Using v4.0.0-M4

```scala
$ cat Foo.scala && scalac Foo.scala && metap -proto $(metacp .)
object a {
  val x: Either.type = null
}
```
```diff
- obtained
+ expected
...
  symbols {
    symbol: "_empty_.a.x()."
    kind: METHOD
    properties: 1024
    name: "x"
    accessibility {
      tag: PUBLIC
      symbol: ""
    }
    language: SCALA
    signature {
      methodSignature {
        type_parameters {
        }
        return_type {
          singleType {
            prefix {
            }
-            symbol: "scala.package.Either."
+            symbol: "scala.package.Either()."
          }
        }
      }
    }
  }
  language: SCALA
  md5: ""
}
```

xeno-by:
@olafurpg Sorry for an unrelated question, but why is `md5` empty here?

olafurpg:
`md5` is empty metacp doesn't include `text`

xeno-by:
Would it make sense to hash the classfile bytes then?

olafurpg:
I don't have a use-case for that personally, we already cache metacp processed jars by a more lightweight fingerprint (the location of the cache jar is an md5 fingerprint of filename + timestamp + file size).

xeno-by:
@olafurpg Alright. No use case - no work to do :)

olafurpg:
This bug popped up `metac-metacp.diff` in #1660 which might give a hint this is related to false detection of a package symbol.

olafurpg:
The new `metacp.undefined` expect tests seems to have reproduced this issue https://github.com/scalameta/scalameta/blob/8fc2a4de80c934dcc13748bb4ebf5e52d5452cec/tests/jvm/src/test/resources/metacp.undefined#L2

olafurpg:
I'm encountering this error a lot due to `blackbox/Context#universe.` appearing in prefix position for def macros
```
typeRef {
  prefix {
  }
  symbol: "scala/package.List#"
  type_arguments {
    typeRef {
      prefix {
        singleType {
          prefix {
            singleType {
              prefix {
              }
              symbol: "org/scalatest/DiagrammedExprMacro#context()."
            }
          }
          symbol: "scala/reflect/macros/blackbox/Context#universe."
        }
      }
      symbol: "scala/reflect/api/Trees#Tree#"
    }
  }
}
[...]

PR #1660, opened 2018-07-03
Use slash / for package symbols

### LINKED ISSUE #1420: Decide on `language` for package symbols
It's annoying that for javacp, package symbols are `JAVA`, but for scalacp/metac package symbols are `SCALA`.

### LINKED ISSUE #1647: Use slash / for package symbols

### LINKED ISSUE #1655: Remove redundant targeted tests
We have accumulated a large number of test cases in `TargetedSuite` that we can remove because

- expect tests already cover those cases in more detailed way
- they're disorganized
- they're annoying to update when expected behavior changes

The targeted tests are still very handy for local development since they provide a tight edit/run/debug cycle (with 2-4 second turnarounds!). We should keep the testing infrastructure for this purpose alone.

olafurpg:
Fixes #1647.
Fixes #1655 (remove redundant unit tests)
Fixes #1420 (package symbol language)

Currently, SemanticDB package symbols use `.` like for `scala` in `scala.Predef.String#`. This is problematic since it's not possible to distinguish packages from Scala objects like `Predef`. This PR follows up on the proposal in #1647 to change the format to be `scala/Predef.String#` instead. The benefit of this strategy is that we no longer need to persist package symbols and we can compute their `SymbolInformation` with a stateless method like this

```scala
def info(symbol: String): SymbolInformation =
  if (symbol.endsWith("/")) SymbolInformation(symbol, k.PACKAGE)
  else ...
```


tests/jvm/src/test/resources/metac-metacp.diff
```diff
@@ -1078,122 +1054,147 @@ flags.p.package.S#[T]
...
 
 
+======================
+types/Test.C#Either().
+======================
+--- metac
++++ metacp
+       singleType {
+         prefix {
+-        symbol: "scala/util/Either."
```
olafurpg: This is the only regression I can see, and the issue is already tracked in https://github.com/scalameta/scalameta/issues/1625#issuecomment-402174890 I will fix that issue later this week.

olafurpg: Turns out #1625 was unrelated after all, this regression has been fixed now.

olafurpg:
Seems like my fix for `scala/util.Either.` by removing `hasTermName` introduced a regression

```
scala.meta.internal.scalacp.MissingSymbolException: scala.reflect.macros.blackbox 
```

I will investigate later.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/SymbolInformationOps.scala
```diff
@@ -67,7 +67,8 @@ trait SymbolInformationOps { self: Scalacp =>
           }
           val isModuleClass = sym.entry.entryType == 10
           def isScalaObject: Boolean = isModuleClass && !sym.isJavaDefined
-          if (hasTermName || isScalaObject) k.OBJECT
+          if (sym.isPackageAccordingToClasspath) k.PACKAGE
+          else if (hasTermName || isScalaObject) k.OBJECT
```
olafurpg: You can ignore the commit 0c5c84b03f2a6910376e624b828051afcfebbb72. I got greedy in 1e3a73bf326538ffa53db726b7465f815b097c50 and tried to fix an issue unrelated to this PR.

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -2377,8 +2377,8 @@ In this section, we describe the Java symbol format.
 
 **Descriptor** is:
   * For `LOCAL`, unsupported.
-  * For `FIELD` or `PACKAGE`, concatenation of its simple name [\[98\]][98]
-    and a dot (`.`).
+  * For `PACKAGE`, concatenation of its encoded name and a forward slash (`/`).
```
xeno-by: It doesn't look like the Java part of the spec defines what's an encoded name.

tests/jvm/src/test/resources/metac-metacp.diff
```diff
@@ -291,13 +291,13 @@ com.javacp.ClassSuffix#Inner#`<init>`().
 -  tag: PUBLIC
 -  symbol: ""
 +  tag: PRIVATE_WITHIN
-+  symbol: "com.javacp."
++  symbol: "com/javacp/"
```
xeno-by: Why did this change?

olafurpg: The change is expected, it used to be `com.javacp.` and is now `com/javacp/`. Observe that the diff is only one line, the `PRIVATE_WITHIN` diff is from the diff file itself, it's unchanged in this PR.

semanticdb/semanticdb3/semanticdb3.md
olafurpg: Done.

tests/jvm/src/test/resources/metac-metacp.diff
xeno-by: Ah right.

PR #1663, opened 2018-07-03
Transcribing scalac parsing logic into cantBeWrittenWithoutBackquotes

### LINKED ISSUE #1661: Unescaped names are emitted during codegen
I'm generating code from arbitrary user input. Sometimes I get names that are not valid Scala syntax; it would be nice if scalameta correctly handled this case.
```
scala> import scala.meta._
import scala.meta._

scala> q"val `foo-bar`: Long = 5L"
res0: meta.Defn.Val = val foo-bar: Long = 5L

scala> res0.structure
res1: String = Defn.Val(Nil, List(Pat.Var(Term.Name("foo-bar"))), Some(Type.Name("Long")), Lit.Long(5L))

scala> q"val ${Pat.Var(Term.Name("`foo-bar`"))}: Long = 5L"
res2: meta.Defn.Val = val `foo-bar`: Long = 5L
```

Previously I've sort of hacked around this using [`.transform`](https://github.com/twilio/guardrail/blob/cbf9acd9e8ff226cc0f4bbf2f278669071126d5e/modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala#L259-L267) and [some regex/string matching](https://github.com/twilio/guardrail/blob/cbf9acd9e8ff226cc0f4bbf2f278669071126d5e/modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala#L269-L333), but this was intended to be a short/medium term workaround and now I'm running into stack depth issues.

This is using scalameta 3.7.4 and Scala 2.11/2.12

blast-hardcheese:
Resolves #1661 

Motivation: There are many rules to govern whether or not a sequence of characters is legal for a term or type name in scalac, as defined by [scala/scala's spec/01-lexical-syntax.md](https://github.com/scala/scala/blob/2.13.x/spec/01-lexical-syntax.md#lexical-syntax).

This PR is an attempt at bringing `cantBeWrittenWithoutBackquotes` in line with scalac's expectations.

With this change, I can remove a workaround in [twilio/guardrail](https://github.com/twilio/guardrail/blob/cbf9acd9e8ff226cc0f4bbf2f278669071126d5e/modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala#L269-L333)

```
// 100k iterations
scala> benchmark(100000) { assert(cantBeWrittenWithoutBackquotes(Term.Name("this is a very long test"))) }
Average: 0.01088

// 1M iterations
scala> benchmark(1000000) { assert(cantBeWrittenWithoutBackquotes(Term.Name("this is a very long test"))) }
Average: 0.005877

// 100k iterations with a longer string
scala> benchmark(100000) { assert(cantBeWrittenWithoutBackquotes(Term.Name("this is a very long test foo bar baz blix '"))) }
Average: 0.00904
```

vs the original function:
```
// 100k iterations
scala> benchmark(100000) { assert(cantBeWrittenWithoutBackquotesO(Term.Name("this is a very long test foo bar baz blix '"))) }
Average: 7.0E-5
```

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
@@ -624,5 +624,28 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
     assert(q"val `a b` = 2".syntax == "val `a b` = 2")
   }
 
-}
+  test("#1661 Names outside ") {
+    // Must start with either a letter or an operator
+    assert(q"val `foo` = 2".syntax == "val foo = 2")
+    assert(q"val `++++` = 2".syntax == "val ++++ = 2")
+    assert(q"val `_+` = 2".syntax == "val `_+` = 2")
 
+    // Non-leading operators are accepted only after underscores
+    assert(q"val `a_+` = 2".syntax == "val a_+ = 2")
+    assert(q"val `a_a_+` = 2".syntax == "val a_a_+ = 2")
+
+    // Operators must not be followed by non-operators
+    assert(q"val `+_a` = 2".syntax == "val `+_a` = 2")
+    assert(q"val `a_++` = 2".syntax == "val a_++ = 2")
+    assert(q"val `a_++a` = 2".syntax == "val `a_++a` = 2")
+
+    // Lexical letters and digits can follow underscores
+    assert(q"val `_a` = 2".syntax == "val _a = 2")
+    assert(q"val `a_a` = 2".syntax == "val a_a = 2")
+
+    // Non-operators must not be followed by operators
+    assert(q"val `a+` = 2".syntax == "val `a+` = 2")
+    assert(q"val `a-b` = 2".syntax == "val `a-b` = 2")
+    assert(q"val `a:b` = 2".syntax == "val `a:b` = 2")
+  }
+}
--- a/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
@@ -647,5 +647,10 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
     assert(q"val `a+` = 2".syntax == "val `a+` = 2")
     assert(q"val `a-b` = 2".syntax == "val `a-b` = 2")
     assert(q"val `a:b` = 2".syntax == "val `a:b` = 2")
+
+    // Comments must be handled carefully
+    assert(q"val `/*` = 2".syntax == "val `/*` = 2")
+    assert(q"val `//` = 2".syntax == "val `//` = 2")
+    assert(q"val `a_//` = 2".syntax == "val `a_//` = 2")
   }
 }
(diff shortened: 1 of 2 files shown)


tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
```diff
@@ -624,5 +624,28 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
...
+    assert(q"val `a_++a` = 2".syntax == "val `a_++a` = 2")
+
+    // Lexical letters and digits can follow underscores
+    assert(q"val `_a` = 2".syntax == "val _a = 2")
+    assert(q"val `a_a` = 2".syntax == "val a_a = 2")
+
+    // Non-operators must not be followed by operators
+    assert(q"val `a+` = 2".syntax == "val `a+` = 2")
+    assert(q"val `a-b` = 2".syntax == "val `a-b` = 2")
+    assert(q"val `a:b` = 2".syntax == "val `a:b` = 2")
```
olafurpg: What happens in the following cases?
```scala
    assert(q"val `//` = 2".syntax == "val `//` = 2")
    assert(q"val `/*` = 2".syntax == "val `/*` = 2")
```

iirc, in akka-http they have backticked identifiers with `/*`

In scala-syntax we use https://github.com/scalacenter/scala-syntax/blob/7352163153066bf9594d63c8e78d62aaa9123656/format/src/main/scala/org/scalafmt/internal/Identifier.scala that we adapted from the Ammonite codebase.

blast-hardcheese:
Seems as though `` `//` `` turns into `//`, as you've pointed out -- investigating

blast-hardcheese:
It looks like `//` and `/*` are both (correctly) considered operator strings, though that raises some concerns for structures like

    val a_// = 2

perhaps comment characters should be checked outside of the analyzer, as they operate at a higher level

blast-hardcheese:
@olafurpg Resolved!

Additionally, I noticed there was no branch for 3.7.x bugfixes, what's the policy around this? I'd rather be able to upgrade to 4.x at a separate time to making other major changes. I've been maintaining a changeset [here](https://github.com/blast-hardcheese/scalameta/tree/backtick-bugfix-backport-v3.7.x), `deployLocal`-ing so I can continue iterating on guardrail during the PR review process.

olafurpg:
We don't backport for 3.x so this fix will only be available in 4.x. However, there are no breaking changes in the syntactic api (`Tree`, `Token`) so you should not encounter problems upgrading. The breaking changes only relate to SemanticDB.

Our CI releases a snapshot after merge, you can use `4.0.0-M4-127-8fc2a4de-SNAPSHOT` to enjoy this fix. Snapshots are effectively immutable since the version changes with every commit, and with 
```
    updateOptions := updateOptions.value.withLatestSnapshots(false)
```
the resolution times are reasonable. If you are using `publishLocal`, then we might want to update the build to `.aggregate(...)`  correctly so that `scalametaJVM/publishLocal` triggers a publish for all modules.

blast-hardcheese:
@olafurpg Ah, I ran into that problem; every time I tried to resolve there were more missing packages. Proper `aggregate` would be very much appreciated, if that would make `publishLocal` publish all deps.

Thank you!

olafurpg:
@blast-hardcheese I will update the build in #1682 so that `scalametaJVM/publishLocal` aggregates all modules.

Issue #1671, opened 2018-07-04
Make targetroot configurable in metac


olafurpg:
Metac should support `-P:semanticdb:targeroot:<path>` to override the default which matches `-d`. There are valid cases where users might not want to persist SemanticDB files alongside classfiles. If the custom targetroot ends with `.jar` then metac should compress the output files into a jar file instead of directory.

xeno-by:
After having implemented this functionality, we'll also need to update the Java compiler plugin to use the same UI.

olafurpg:
> After having implemented this functionality, we'll also need to update the Java compiler plugin to use the same UI.

The Java compiler plugin already requires targetroot to be provided as ```"-Xplugin:semanticdb <target-dir> --sourceroot <source-root>"``` since it doesn't have access to the compiler output directory.

xeno-by:
Fixed in https://github.com/scalameta/scalameta/pull/1743.

PR #1682, opened 2018-07-06
Add symbol table module

### LINKED ISSUE #1679: Introduce scala.meta.internal.symtab
Metacp represents a significant upfront requirement for virtually anyone who wants to use SemanticDB-based tools. Scalafix needs it, Metals needs it, etc etc.

Recent changes (https://github.com/scalameta/scalameta/pull/1638 and https://github.com/scalameta/scalameta/pull/1660) have made this even more frustrating. Now Metacp needs a dependency classpath, which means that it can now incur `missing symbol` errors if the dependency classpath is not in the shape that we expected (https://github.com/scalameta/scalameta/issues/1675 and probably more), meaning that the requirement on our users has become even stricter.

However, as our experience has shown, SemanticDB-based tools rarely need full SemanticDB representations of the entire classpaths they are working with. What many of the applications need instead of something that @olafurpg has implemented in Scalafix ([source code](https://github.com/scalacenter/scalafix/blob/c9933a4e2f651149644525507ea09df77ab3a5e8/scalafix-core/shared/src/main/scala/scalafix/internal/util/SymbolTable.scala)):

```scala
trait SymbolTable {
  def info(symbol: String): Option[s.SymbolInformation]
}
```

Based on these findings and our recent chat with @olafurpg, I propose the following:
  * Introduce `"org.scalameta" %% "symtab"` to host the aforementioned interface and its implementation based on lazy conversion of JVM metadata and ScalaSignatures to SemanticDB. 
  * Keep the `SymbolTable` interface in `scala.meta.internal.symtab`, since it depends on a data structure from `scala.meta.internal`.
  * Remove the requirement for Metacp-based classpaths in our SemanticDB-based applications and accept normal classpaths via `SymbolTable`.
  * Keep Metacp as a CLI tool, since it's useful for debugging purposes, but remove caching and parallelism that were only necessary because Metacp was a core part of our toolchain.

olafurpg:
Fixes #1679.

- Adds new `symtab` module that given a `scala.meta.io.Classpath` can
  compute `SymbolInformation` on the fly.
- New simple `ClassfileInfos` API to extract `SymbolInformation` from
  classfiles. This API can be used by a symbol table implementation.
- Remove semanticidx.proto because the index becomes irrelevant once we
  remove caching. Users can instead compute the index directly from
  SemanticDB payloads if they want or they will be able to use the
  upcoming symtab module to compute symbols on the fly from a classpath.
- Replace --cache-dir metacp flag with --out, which can either be a
  directory or a jar. Processed output is placed in the output
  unconditionally without caching.
- Parse `ScalaSig` directly from ASM `ClassNode` to simplify the
  API for Scalacp. Now we only need to pass around a `ClassNode`.
- Remove bloop/zinc testing infrastructure because we no longer
  need to test the index.
- Remove directories-jvm dependency since metacp no longer uses a global
  cache.
- Remove data structures around "top-level symbols".


olafurpg:
Opened this PR before tackling #1679 because this diff can be reviewed in isolation. The changes are already large enough and I suspect that it will take several iterations to get the symbol table interface+implementation right. 

I think it might make sense to merge this PR alongside the new symbol table APIs.

xeno-by:
`.semanticidx` is a foundational piece of Rsc's symbol table: https://github.com/twitter/rsc/blob/master/rsc/src/main/scala/rsc/classpath/Index.scala. Let's wait until the implementation of https://github.com/scalameta/scalameta/issues/1679 before making a decision to remove it.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/SymbolIndex.scala
```diff
@@ -6,6 +6,11 @@ import scala.meta.internal.metacp._
 import scala.tools.scalap.scalax.rules.scalasig._
 
 final class SymbolIndex private (classpathIndex: ClasspathIndex) {
+  lazy val syntheticsPaths: Set[String] =
+    Scalalib.synthetics.map { synthetic =>
+      synthetic.relativeUri.stripSuffix(".class").replace('/', '.')
+    }.toSet
```
xeno-by: Why is this a public API? Also, let's move this method to the end of the class, so that the first thing that the readers see is the main API of the class, not implementation details.

semanticdb/metacp/src/main/scala/scala/meta/metacp/Settings.scala
```diff
@@ -94,10 +98,8 @@ object Settings {
     loop(Settings(), allowOptions = true, args)
   }
 
-  def defaultCacheDir: AbsolutePath = {
-    val projectDirectories = ProjectDirectories.from("org.scalameta", "", "SemanticDB")
-    val cacheRoot = AbsolutePath(projectDirectories.cacheDir)
-    cacheRoot.resolve(BuildInfo.version)
+  def defaultOut: AbsolutePath = {
+    PathIO.workingDirectory
```
xeno-by: `PathIO.workingDirectory.resolve("out.jar")`?

build.sbt
```diff
@@ -198,6 +197,16 @@ lazy val metacp = project
...
   .dependsOn(semanticdbJVM, cliJVM, ioJVM)
 
+lazy val symtab = project
+  .in(file("semanticdb/symtab"))
+  .settings(
+    publishableSettings,
+    description := "SemanticDB symbol table for Scala 2.x and Java classpaths"
+  )
+  // NOTE: workaround for https://github.com/sbt/sbt-core-next/issues/8
+  .disablePlugins(BackgroundRunPlugin)
```
xeno-by: Why do wo we need this?

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
```diff
@@ -236,9 +232,12 @@ object Javacp {
     directInnerClasses.foreach { ic =>
       val innerClassSymbol = ssym(ic.name)
       decls += innerClassSymbol
-      val innerPath = asmNameToPath(ic.name, toplevel.base)
-      val innerClassNode = innerPath.toClassNode
-      buf ++= sinfos(toplevel, innerClassNode, ic.access, classScope)
+      val path = ic.name + ".class"
+      val classfile = classpathIndex.getClassfile(path).getOrElse {
+        throw MissingSymbolException(path)
```
xeno-by: The definition of the missing symbol exception says `final case class MissingSymbolException(symbol: String)`. I'm surprised that we're passing a path to it.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ClassfileInfos.scala
```diff
@@ -0,0 +1,67 @@
...
+    val semanticdbAbspath =
+      out.resolve("META-INF").resolve("semanticdb").resolve(relativeUri + ".semanticdb")
+    // First element wins in case of conflicts to match scala-compiler behavior:
+    // $ scala -classpath $(coursier fetch org.scalameta:metacp_2.12:4.0.0-M3 -p):$(coursier fetch org.scalameta:metacp_2.12:4.0.0-M4 -p)
+    // scala> scala.meta.internal.metacp.BuildInfo
+    // res0: meta.internal.metacp.BuildInfo.type = version: 4.0.0-M3
+    //
+    // $ scala -classpath $(coursier fetch org.scalameta:metacp_2.12:4.0.0-M4 -p):$(coursier fetch org.scalameta:metacp_2.12:4.0.0-M3 -p)
+    // scala> scala.meta.internal.metacp.BuildInfo
+    // res0: meta.internal.metacp.BuildInfo.type = version: 4.0.0-M4
```
xeno-by: Looks like a copy/paste?

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -30,81 +34,41 @@ class Main(settings: Settings, reporter: Reporter) {
...
-
     classpath.foreach { entry =>
       if (entry.isDirectory) {
-        val out = AbsolutePath(Files.createTempDirectory("semanticdb"))
-        val semanticdbIndex = new SemanticdbIndex
-        val res = convertClasspathEntry(entry, out, semanticdbIndex)
-        semanticdbIndex.save(out)
-        success.compareAndSet(true, res)
-        buffer.add(out)
+        success.compareAndSet(true, convertClasspathEntry(entry, outRoot))
```
xeno-by: I'd prefer if we removed parallel processing from metacp as well. Since metacp's performance is no longer a feature, I think it'd be great to remove all complications that are there solely for performance.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/ScalaSigAttribute.scala
```diff
@@ -0,0 +1,26 @@
...
+      codeOff: Int,
+      labels: Array[Label]
+  ): Attribute = {
+    val bytecode = new ByteCode(cr.b, off, len)
+    val scalaSig = ScalaSigAttributeParsers.parse(bytecode)
+    ScalaSigAttribute(scalaSig)
+  }
+}
+final case class ScalaSigAttribute(scalaSig: ScalaSig) extends BaseScalaSigAttribute
+object ScalaSigAttribute extends BaseScalaSigAttribute
```
xeno-by: This class/companion business still looks mysterious to me. If I understand correctly, it's a clever way to reuse some code by fusing multiple concepts into one entity?

If yes, how about having two different classes: 1) `final case class ScalaSigAttribute(scalaSig: ScalaSig) extends extends Attribute(Main.SCALA_SIG)`, 2) `object ScalaSigPrototype extends Attribute(Main.SCALA_SIG) { def read... }`?

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
```diff
@@ -66,12 +38,12 @@ final class Scalacp private (
 
 object Scalacp {
   def parse(
-      classfile: ToplevelClassfile,
-      settings: Settings,
+      node: ClassNode,
+      scalaSig: ScalaSig,
```
xeno-by: I like the new way to organize code, but it's still unsatisfying that we have to pass two things here (`node` and `scalaSig`) instead of one, especially since the latter is computed from the former.

Here's what I propose. How about `node.scalaSig` returns `ScalaSigAttribute`, which is in turn changed to also capture `ClassReader.getClassName`?

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/SymbolIndex.scala
```diff
@@ -69,6 +67,10 @@ final class SymbolIndex private (classpathIndex: ClasspathIndex) {
       buf.result()
     }
   }
+  private lazy val syntheticsPaths: Set[String] =
```
xeno-by: `scalalibSyntheticPaths`

semanticdb/metacp/src/main/scala/scala/meta/metacp/Settings.scala
```diff
@@ -1,21 +1,20 @@
...
 import scala.meta.cli._
+import scala.meta.internal.io.PathIO
 import scala.meta.io.AbsolutePath
 import scala.meta.io.Classpath
-import scala.meta.internal.metacp.BuildInfo
 
 final class Settings private (
-    val cacheDir: AbsolutePath,
+    val out: AbsolutePath,
     val classpath: Classpath,
```
xeno-by: Given what we discussed earlier today, how about changing `classpath: Classpath` to `in: AbsolutePath`?

semanticdb/metacp/src/main/scala/scala/meta/metacp/Settings.scala
```diff
@@ -69,8 +73,8 @@ object Settings {
       args match {
         case "--" +: rest =>
           loop(settings, false, rest)
-        case "--cache-dir" +: cacheDir +: rest if allowOptions =>
-          loop(settings.copy(cacheDir = AbsolutePath(cacheDir)), true, rest)
```
xeno-by: Shall we maybe emit a more informative error here?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SemanticdbOps.scala
```diff
@@ -19,5 +18,4 @@ trait SemanticdbOps
     with TypeOps {
   val global: Global
   var config: SemanticdbConfig = SemanticdbConfig.default
-  val semanticdbIndex: SemanticdbIndex = new SemanticdbIndex
 }
```
xeno-by: Am I right in thinking that you didn't plan to address https://github.com/scalameta/scalameta/issues/1671 in this pull request?

semanticdb/semanticdb3/guide.md
```diff
@@ -436,18 +436,6 @@ nor `Diagnostics` sections are present, because they both require source
...
-following scheme:
-  * Jar files. Always cached in the OS cache directory keyed by the application
-    name `semanticdb`, Metacp version and the fingerprint of the jar.
-    The default cache location depends on the OS and is computed using
-    [directories-jvm](https://github.com/soc/directories-jvm).
-    For example, on macOS the location for scala-library would be
-    `$HOME/Library/Caches/org.scalameta.SemanticDB/3.7.4/scala-library-FINGERPRINT.jar`.
-  * Class directories. Never cached to avoid overflowing the cache with
-    transient results.
-
```
xeno-by: The mention of `--cache-dir` above also needs to be updated.

semanticdb/symtab/src/main/scala/scala/meta/internal/symtab/GlobalSymbolTable.scala
```diff
@@ -0,0 +1,72 @@
...
+import scala.meta.io._
+import scala.meta.internal.metacp._
+import scala.meta.internal.scalacp.Scalalib
+import scala.meta.internal.semanticdb.Scala._
+import scala.reflect.NameTransformer
+
+/** A lazy symbol table that returns global symbols on-the-fly from disk. */
+final class GlobalSymbolTable(classpathIndex: ClasspathIndex) extends SymbolTable {
+
+  private val symbolCache = TrieMap.empty[String, SymbolInformation]
```
xeno-by: What's the difference between a trie map and a usual map?

semanticdb/symtab/src/main/scala/scala/meta/internal/symtab/GlobalSymbolTable.scala
```diff
@@ -0,0 +1,72 @@
...
+        }
+      case _ =>
+        ()
+    }
+  }
+
+  override def info(symbol: String): Option[SymbolInformation] =
+    if (symbol.isPackage) {
+      if (classpathIndex.isClassdir(symbol)) {
+        Some(SymbolInformation(symbol, kind = SymbolInformation.Kind.PACKAGE))
```
xeno-by: SemanticDB specification also requires package infos to have a non-empty name.

semanticdb/symtab/src/main/scala/scala/meta/internal/symtab/GlobalSymbolTable.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  @tailrec private def getToplevel(symbol: String): String = {
+    val owner = symbol.owner
+    if (owner.isPackage) symbol
+    else getToplevel(owner)
+  }
+
+  private def loadSymbol(symbol: String): Unit = {
+    val toplevel = getToplevel(symbol)
+    val classdir = toplevel.owner
```
xeno-by: `symbol.ownerChain.dropWhile(!_.isPackage).head`?

semanticdb/symtab/src/main/scala/scala/meta/internal/symtab/GlobalSymbolTable.scala
```diff
@@ -0,0 +1,72 @@
...
+          Some(x)
+        case None =>
+          loadSymbol(symbol)
+          symbolCache.get(symbol)
+      }
+    }
+}
+
+object GlobalSymbolTable {
+  def fromClasspath(classpath: Classpath): GlobalSymbolTable = {
```
xeno-by: `def apply`?

build.sbt
olafurpg: Cargo culted from other project definitions. If we remove it we should do it for all projects in a separate PR.

semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala
olafurpg: Good catch, updated to `ssym(ic.name)`.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/ClassfileInfos.scala
olafurpg: I removed the repl session but kept "first element wins" comment.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
olafurpg: The most important user of --par is in fact ConvertSuite because it shaves off ~35s for `slow:test` on my machine (just tested right now). That's a significant productivity difference when iterating on changes locally. I vote to keep --par.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/ScalaSigAttribute.scala
olafurpg: That simplified a lot, great suggestion 👍

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/Scalacp.scala
olafurpg: Done, I introduced a new `case class ScalaSigNode(relativeUri: String, scalaSig: ScalaSig)` to match `ClassNode` and since `ScalaSigAttribute` is already used.

semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/SymbolIndex.scala
olafurpg: Done.

semanticdb/metacp/src/main/scala/scala/meta/metacp/Settings.scala
olafurpg: I vote for "no" because

- we use `Classpath` in `ConvertSuite`
- running `Classpath.entries.map(Metacp.process)` has a big performance penalty because `ClasspathIndex` needs to be re-created every time

I would prefer to come up with a scheme so that metacp produces one jar per classpath entry.

semanticdb/metacp/src/main/scala/scala/meta/metacp/Settings.scala
olafurpg: Good idea, done. Metacp now reports "--cache-dir is deprecated, use --out instead"

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SemanticdbOps.scala
olafurpg: Correct, this PR is already large enough and the required changes in metac are mostly unrelated.

semanticdb/semanticdb3/guide.md
olafurpg: Good catch, I've updated the flag table.

Note to self, we should automatically generate --help to prevent these errors in the future. Metaconfig supports this (https://olafurpg.github.io/metaconfig/#settingstoclihelp) and 100 other command-line parsing libraries.

semanticdb/symtab/src/main/scala/scala/meta/internal/symtab/GlobalSymbolTable.scala
olafurpg: `TrieMap` is the only concurrent Map in the Scala stdlib I'm aware of, I wanted to make `info().` thread-safe and opted to not use Java collections.

semanticdb/symtab/src/main/scala/scala/meta/internal/symtab/GlobalSymbolTable.scala
olafurpg: That doesn't read as nicely IMO

semanticdb/symtab/src/main/scala/scala/meta/internal/symtab/GlobalSymbolTable.scala
olafurpg: Good point, updated.

semanticdb/symtab/src/main/scala/scala/meta/internal/symtab/GlobalSymbolTable.scala
olafurpg: Done.

build.sbt
xeno-by: Let's open a ticket for that.

semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
xeno-by: As discussed, I'm fine with keeping --par for now, but we'll need to return to the original behavior of multiple inputs => multiple outputs.

semanticdb/metacp/src/main/scala/scala/meta/metacp/Settings.scala
xeno-by: Alright, let's have metacp produce one jar per classpath entry.

semanticdb/symtab/src/main/scala/scala/meta/internal/symtab/GlobalSymbolTable.scala
```diff
@@ -0,0 +1,77 @@
...
+import scala.meta.internal.metacp._
+import scala.meta.internal.scalacp.Scalalib
+import scala.meta.internal.semanticdb.Scala._
+import scala.reflect.NameTransformer
+
+/** A lazy symbol table that returns global symbols on-the-fly from disk. */
+final class GlobalSymbolTable(classpathIndex: ClasspathIndex) extends SymbolTable {
+
+  private val symbolCache = TrieMap.empty[String, SymbolInformation]
+  Scalalib.synthetics.foreach(enter)
```
xeno-by: This takes care of scalalib synthetics, but doesn't take RootPackage and EmptyPackage into account.

semanticdb/symtab/src/main/scala/scala/meta/internal/symtab/GlobalSymbolTable.scala
xeno-by: As it currently stands, `getToplevel` freezes on symbols that aren't owned by packages (e.g. locals, nosymbol and root). I'd prefer if we used a standard way to recur on owners.

xeno-by:
I have migrated to GlobalSymbolTable from this PR in Rsc: https://github.com/xeno-by/rsc/commits/topic/symtab. Predictably, the performance went down significantly: RscSemanticdb degraded from 136ms to 2772ms.

build.sbt
```diff
@@ -184,10 +184,9 @@ lazy val metacp = project
   .in(file("semanticdb/metacp"))
   .settings(
     publishableSettings,
-    description := "Scala 2.x classpath to SemanticDB converter",
+    description := "Scala 2.x and Java classpath to SemanticDB converter",
     libraryDependencies ++= List(
-      "org.scala-lang" % "scalap" % scalaVersion.value,
-      "io.github.soc" % "directories" % "10"
```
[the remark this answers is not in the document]
dwijnand: What kind of grief?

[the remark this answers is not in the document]
xeno-by: We were forced to upgrade from v6 to v10 when some of our dependees upgraded, because otherwise we were getting NoSuchMethodErrors.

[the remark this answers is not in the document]
dwijnand: Gotcha.

Yeah, that was my reservation on using it (for example, in a future version of sbt).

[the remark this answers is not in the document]
olafurpg: If you want to use it in sbt just inline the sources under a new namespace with attribution, it's a tiny project. The functionality provided by the library is great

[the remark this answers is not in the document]
dwijnand: Yup, I would probably look to shading it in.

olafurpg:
Merging :shipit: 🔥

Issue #1659, opened 2018-07-03
Exclude synthetic local symbols in value class equals method


olafurpg:
Values classes
```scala
class C1(val x: Int) extends AnyVal
```
result in the following trees after typer
```scala
  final class C1 extends scala.AnyVal {
    ...
    override <synthetic> def equals(x$1: Any): Boolean = x$1 match {
  case (_: classes.C1) => true
  case _ => false
}.&&({
      <synthetic> val C1$1: classes.C1 = x$1.asInstanceOf[classes.C1]; // <-- synthetic local symbol
      C1.this.x1.==(C1$1.x1)
    })
  };
```

After #1656, we persist the symbol information for `C1$1` in metac.expect

```
 local0 => val local C1$1: C1 
   C1 => classes.C1# 
```

Since it's not possible to reference these symbols we should consider not persisting their information.

olafurpg:
This opens up a more general question of what symbols we should persist and how the spec should document this.

I personally care most about being able to lookup information about all symbols that are visible in source code. This includes term wildcards (#1657) even if they are synthetic and local but excludes local variables inside generated macro annotation code.

The strategy so far in semanticdb-scalac has been to only persist what we can associate by position with `scala.meta.Tree` nodes. However, this strategy doesn't cut it for symbols since we need to be able to lookup information about for example macro annotation generated public members (related #1504). The PR #1656 changes the strategy to persist every non-useless symbol definitions, which guarantees we capture all symbols but risks exposing too many implementation details of the Scala compiler.

xeno-by:
I've added the RC1 milestone because I'd like us to discuss this in-depth before the feature freeze. Unlike things like https://github.com/scalameta/scalameta/issues/1651, this ticket is fundamental enough and recent enough to warrant careful consideration.

olafurpg:
I think the first problem is to solve is defining the `SYNTHETIC` property https://github.com/scalameta/scalameta/issues/1492 In the JVMS https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.1 the definition for `ACC_SYNTHETIC` is

> Declared synthetic; not present in the source code. 

I think this is a good definition, but in practice it seem the Java and Scala compilers are not compliant with it. For example, Java enums have generated `values()` and `valueOf()` methods that don't have the synthetic flag and scalac generates a lot of code without setting the synthetic flag.

[...]

PR #1690, opened 2018-07-12
Create new metai command-line tool

### LINKED ISSUE #1523: RelativePath.toURI should return RelativeUri
```diff
import $ivy.$                                 , scala.meta._, testkit._, contrib._
@ RelativePath("a/b/c").toURI
- res1: java.net.URI = file:/Users/ollie/dev/scalafix/a/b/c // obtained
+ res1: java.net.URI = a/b/c // expected
```

Seems we need to take care of producing the URI ourselves to keep it relative
```scala
    def toUri: URI = {
      val reluri =
        path.toNIO.asScala.iterator.map(_.getFileName.toString).mkString("/")
      URI.create(URLEncoder.encode(reluri, StandardCharsets.UTF_8.name()))
    }
```

Discovered while loading `*.semanticdb` files and comparing `s.TextDocument.uri` with `RelativePath.toURI.toString`

We should introduce the following new types
```
case class RelativeURI(uri: URI)
case class AbsoluteURI(uri: URI)
```

and make sure that `AbsolutePath.toURI: AbsoluteURI` and `RelativePath.toURI: RelativeURI`.

### LINKED ISSUE #1689: Implement metai
https://github.com/scalameta/scalameta/pull/1682 removes support for `.semanticidx` to simplify the codebase and streamline SemanticDB generation for the common case. 

However, this is critical for our Symtab in Rsc (https://github.com/scalameta/scalameta/pull/1682#issuecomment-403993748), which is why we decided to keep the semantic index fuctionality, spinning index generation into a standalone tool called `metai`. This way common case would stay simple, and more advanced cases don't regress.

Here is the schema that I propose for `.semanticidx` (Note the `PackageEntry` message - we need this functionality in Rsc):

```protobuf
message Indexes {
  repeated Index indexes = 1;
}
message Index {
  repeated map<string, Entry> entries = 1;
}
message Entry {
  oneof sealed_value {
    PackageEntry packageEntry = 1;
    ToplevelEntry toplevelEntry = 2;
  }
}
message PackageEntry {
}
message ToplevelEntry {
  string uri = 1; // relative URI to semanticdb file containing full information
}
```

olafurpg:
Fixes #1689.

The tool can be used like this
```
metai [<classpath> ...]
```
and it will write in-place the file `META-INF/semanticdb.semanticidx`
with a lookup index from toplevel symbols to the relative URIs where
those files are contained.

This commit

- fixes #1523: `RelativePath.toURI` to return a proper relative URI
- introduces `Classpath.foreach(AbsolutePath => Unit)` helper
  to traverse a classpath handling jar files and manifests.

The tests pass except that metai.expect does not contain entries from
manifest.jar.


build.sbt
```diff
@@ -198,7 +198,22 @@ lazy val metacp = project
   .enablePlugins(BuildInfoPlugin)
   // FIXME: https://github.com/scalameta/scalameta/issues/1688
   .disablePlugins(BackgroundRunPlugin)
-  .aggregate(scalametaJVM, cliJVM, ioJVM)
+  .aggregate(semanticdbJVM, cliJVM, ioJVM)
```
xeno-by: This looks pretty tedious, but unfortunately I don't know of a better way.

build.sbt
```diff
@@ -198,7 +198,22 @@ lazy val metacp = project
...
+  .dependsOn(semanticdbJVM, cliJVM, ioJVM)
+
+lazy val metai = project
+  .in(file("semanticdb/metai"))
+  .settings(
+    publishableSettings,
+    description := "SemanticDB classpath indexer",
+    mainClass := Some("scala.meta.cli.Metai"),
+    buildInfoKeys := Seq[BuildInfoKey](version),
+    buildInfoPackage := "scala.meta.internal.metai"
```
xeno-by: Why do we need to persist Scalameta version for Metai?

xeno-by: Same question for Metacp.

scalameta/io/shared/src/main/scala/scala/meta/internal/io/package.scala
```diff
@@ -10,23 +11,45 @@ package object io {
...
+                  finally in.close()
+                  val classpathAttr = manifest.getMainAttributes.getValue("Class-Path")
+                  if (classpathAttr != null) {
+                    classpathAttr.split(" ").foreach { relativePath =>
+                      val jar = base.resolveSibling(_ => relativePath)
+                      processEntry(jar)
+                    }
+                  }
+                }
+              }
```
xeno-by: 👍 Thank you for deduplicating this!

scalameta/io/shared/src/main/scala/scala/meta/io/RelativePath.scala
```diff
@@ -6,12 +6,25 @@ import java.nio.file.Path
...
 import java.nio.file.Paths
 import scala.meta.internal.io.PathIO
+import scala.collection.JavaConverters._
 
 /** Wrapper around a relative nio.Path. */
 sealed abstract case class RelativePath(toNIO: Path) {
   require(!toNIO.isAbsolute, s"$toNIO is not relative!")
   def toFile: File = toNIO.toFile
-  def toURI: URI = toFile.toURI
+  def toURI: URI = {
```
xeno-by: It's pretty annoying that we have to deal with `isDirectory`. Is there absolutely no way around this?

xeno-by: If not, I propose that we drop the no-parameter overload, since it seems that `isDirectory` is a core part of the API that doesn't have a universally useful default value.

semanticdb/metai/src/main/scala/scala/meta/internal/metai/Main.scala
```diff
@@ -0,0 +1,100 @@
...
+import scala.meta.internal.semanticdb.Scala._
+import scala.meta.internal.io._
+import scala.meta.io.AbsolutePath
+import scala.meta.io.Classpath
+import scala.meta.io.RelativePath
+import scala.meta.metai.Settings
+import scala.util.control.NonFatal
+
+final class Main(settings: Settings, reporter: Reporter) {
+  def process(): Option[Classpath] = {
```
xeno-by: If Metai modifies the classpath in place, I suggest that we don't pretend that it's `Classpath => Classpath` and just return a boolean for success or failure. If you agree, I suggest we also don't print anything to the console upon successful completion.

semanticdb/metai/src/main/scala/scala/meta/internal/metai/Main.scala
```diff
@@ -0,0 +1,100 @@
...
+    if (success) {
+      Some(settings.classpath)
+    } else {
+      None
+    }
+  }
+
+  private def processEntry(entry: AbsolutePath): Unit = {
+    val semanticdb = entry.resolve("META-INF").resolve("semanticdb")
+    if (!semanticdb.isDirectory) return
```
xeno-by: Let's error out here to avoid situation when we silently do the wrong thing.

semanticdb/metai/src/main/scala/scala/meta/internal/metai/Main.scala
```diff
@@ -0,0 +1,100 @@
...
+  private def updateIndex(
+      index: mutable.Map[String, i.Entry],
+      reluri: String,
+      docs: s.TextDocuments
+  ): Unit = {
+    val PackageEntry = i.PackageEntry()
+    for {
+      doc <- docs.documents
+      toplevel <- doc.symbols
+      if toplevel.symbol.owner.isPackage
```
xeno-by: Am I right in thinking that the "Symbols" section can no longer include package symbols? If yes, it would be good to update the spec to reflect that.

scalameta/io/shared/src/main/scala/scala/meta/io/RelativePath.scala
```diff
@@ -6,12 +6,26 @@ import java.nio.file.Path
...
 import scala.meta.internal.io.PathIO
+import scala.collection.JavaConverters._
 
 /** Wrapper around a relative nio.Path. */
 sealed abstract case class RelativePath(toNIO: Path) {
   require(!toNIO.isAbsolute, s"$toNIO is not relative!")
   def toFile: File = toNIO.toFile
-  def toURI: URI = toFile.toURI
+  @deprecated("Use toURI(isDirectory = false) instead", "4.0.0")
+  def toURI: URI = {
```
xeno-by: I'm not sure whether it's worth keeping this method around. As far as I can see, it has potential to produce incorrect results, so I think that breaking binary compatibility is the right thing to do here. Compatibility with incorrect behavior doesn't sound like a good idea.

semanticdb/metai/src/main/scala/scala/meta/internal/metai/Main.scala
xeno-by: Agreed.

semanticdb/metai/src/main/scala/scala/meta/internal/metai/Main.scala
```diff
@@ -0,0 +1,88 @@
...
+      path,
+      indexes,
+      StandardOpenOption.CREATE,
+      StandardOpenOption.TRUNCATE_EXISTING
+    )
+  }
+
+  private def getEntriesForPath(root: AbsolutePath, relpath: RelativePath): i.Index = {
+    val abspath = root.resolve(relpath)
+    val reluri = relpath.toURI(isDirectory = false).toString
```
xeno-by: Why immediately do toString here? It would seem that it'd be more typesafe to pass around the most specific type.

semanticdb/metai/src/main/scala/scala/meta/internal/metai/Main.scala
```diff
@@ -0,0 +1,88 @@
...
+    getEntriesForDocs(reluri, docs)
+  }
+
+  private def getEntriesForDocs(reluri: String, docs: s.TextDocuments): i.Index = {
+    val PackageEntry = i.PackageEntry()
+    val entries = Map.newBuilder[String, i.Entry]
+    for {
+      doc <- docs.documents
+      toplevel <- doc.symbols
+      if !toplevel.symbol.isPackage
```
xeno-by: Let's open a ticket for this?

semanticdb/metai/src/main/scala/scala/meta/internal/metai/Main.scala
olafurpg: Done, metai now errors in case no semanticdb files are found unless the entry is a manifest jar.

semanticdb/metai/src/main/scala/scala/meta/internal/metai/Main.scala
olafurpg: We don't buy safety by keeping this as `URI`, we don't do any further processing on this value.

semanticdb/metai/src/main/scala/scala/meta/internal/metai/Main.scala
olafurpg: I left a comment in https://github.com/scalameta/scalameta/issues/1659#issuecomment-404789629

olafurpg:
Metai handles manifest jars thanks to the new `Classpath.foreach` helper but there is no `ManifestMetai` expect test because the current `manifest.jar` is not suited for metai

- part1.jar doesn't have semanticdb files
- part0.jar and part2.jar have conflicting symbols so the produced index gets two relative toplevel URIs for the same symbol

I leave it to you @xeno-by to create a suitable manifest jar for testing metai.

PR #1697, opened 2018-07-16
Exceptionally remove disambiguator for VAL METHOD

### LINKED ISSUE #1625: Metacp emits reference to non-existent term symbol for val forwarders

olafurpg:
Fixes #1625.

This PR removes the disambiguator for `VAL METHOD` symbols.

```scala
class A {
  val x = 2
  // before: A#x().
  // now: A#x.
}
```

This fixes a bug in metacp where it would incorrectly use `A#x.` for external symbol references, which happened for example frequently for def macro signatures that have parameters with type ref prefixes that point to `blackbox.Context#universe.`.


xeno-by:
Looks like AppVeyor is stuck. I'll be merging.

PR #1705, opened 2018-07-16
Upgrade to Scala Native 0.3.8

xeno-by:


xeno-by:
@densh Looks like some of the 0.3.8 artifacts are missing from both Maven Central (https://repo1.maven.org/maven2/org/scala-native/tools_2.10/) and Sonatype (https://oss.sonatype.org/content/repositories/releases/org/scala-native/tools_2.10/). Are we doing something wrong?

PR #1709, opened 2018-07-19
Semanticdb synthetics refactor

maxov:
This PR refactors current synthetics handling into a tree ADT format as I worked out with @xeno-by.
The new API is easier to work with compared to the old API which relied on nested `TextDocuments` and occurrences to track what synthetics looked like.

The PR also makes the necessary refactors to the semanticdb-scalac plugin and metap tools to emit and consume these synthetics.

Here is a high-level description of the additions to the semanticdb protobuf schema:
```
message Synthetic(range: Range, tree: Tree)
message Tree
message ApplyTree(fn: Tree, args: List[Tree]) extends Tree
message FunctionTree(params: List[Id], term: Tree) extends Tree
message IdTree(sym: String) extends Tree
message LiteralTree(const: Constant) extends Tree
message MacroExpansionTree(expandee: Tree, tpe: Type) extends Tree
message OriginalTree(range: Range) extends Tree
message SelectTree(qual: Tree, id: Id) extends Tree
message TypeApplyTree(fn: Tree, targs: List[Type]) extends Tree
```

I will document the schema changes to the SemanticDB spec in an upcoming PR.

Things to note:
- Added [ConstantPrinter](https://github.com/scalameta/scalameta/pull/1709/files#diff-6307c098e885f721d827019fa7b2265b) for shared behavior between synthetic and symbol info printer
- Made `InfoPrinter` and `InfoNotes` API protected so that [SyntheticPrinter can access them](https://github.com/scalameta/scalameta/pull/1709/files#diff-5adf5cd710f37d180fc22a086cee3a9bR40)
- The printer now wraps quotes from original source in `orig()`, like [here](https://github.com/scalameta/scalameta/pull/1709/files#diff-f90bb281d4f293db21adab2fef7daef1R3291)
- Added [this method](https://github.com/scalameta/scalameta/pull/1709/files#diff-5adf5cd710f37d180fc22a086cee3a9bR12) to shorten quotes from original source for pretty printing, but I'm not using it. Not sure if I should use it or just remove it completely
- I [removed a lot of stuff](https://github.com/scalameta/scalameta/pull/1709/files#diff-dcc084ede384784c08203eb73e2848d0) related to the old-style synthetics in the scalac plugin. Not sure if I should remove `PrinterOps`, is it used anywhere else?
- For consumers of this API, trying to map synthetics to the original trees can be nontrivial when they are nested. For example:
    ```
    [16:46..16:47): orig(x) => ArrowAssoc[Int](orig(x))
      ArrowAssoc => scala/Predef.ArrowAssoc().
      Int => scala/Int#
    [16:46..16:50): orig(x ->) => orig(x ->)[Int]
      Int => scala/Int#
    ```
  The first synthetic is "nested" inside the second one. A consumer of the API, if they wanted to inline the synthetic trees, would have to do the second expansion before the first one to keep the originals the same.
    A potential fix could be to get rid of nested synthetics and include them where the current original trees are. However, it would require us to basically duplicate all scalameta trees in semanticdb to be able to model all the possible cases where synthetics could be nested in each other. The current way avoids that, with the only cost coming to consumers. This cost could be mitigated with an API scalameta provides to "apply" the synthetic transformations onto scalameta trees.


semanticdb/metap/src/main/scala/scala/meta/internal/metap/SymbolInformationPrinter.scala
```diff
@@ -9,7 +9,21 @@ import scala.meta.internal.semanticdb.SymbolInformation._
...
+    infos.foreach { info =>
+      if (!printed(info.symbol)) {
+        printed += info.symbol
+        out.print("  ")
+        out.print(info.name)
+        out.print(" => ")
+        out.println(info.symbol)
+      }
+    }
+  }
```
xeno-by: I don't think that sharing 10 lines of straightforward code is worth: 1) increased coupling between `SymbolInformationPrinter` and `SyntheticPrinter`, 2) loss of flexibility to modify `Symbols:` and `Synthetics:` printers independently, 3) the hassle to come up with a name. I suggest we inline this method into its callers.

semanticdb/metap/src/main/scala/scala/meta/internal/metap/SyntheticPrinter.scala
```diff
@@ -1,28 +1,84 @@
...
+    if (newline == -1) {
+      if (s.length > 100) s.takeRight(100) + "..."
+      else s
+    } else s.substring(0, newline) + "..."
+  }
+
+  def pprint(synth: s.Synthetic): Unit = {
     opt(synth.range)(pprint)
-    opt(": ", doc.substring(synth.range))(out.print)
+    opt(": orig(", doc.substring(synth.range), ")")(out.print)
```
xeno-by: Let's drop the `orig` thing here, since there's no `OriginalTree` involved.

semanticdb/metap/src/main/scala/scala/meta/internal/metap/SyntheticPrinter.scala
```diff
@@ -1,28 +1,84 @@
...
+    out.println()
+
+    if (settings.format.isDetailed) {
+      printInfosVisited(infoNotes.visited)
     }
   }
 
-  implicit def synthOrder: Ordering[Synthetic] = {
-    Ordering.by(s => (s.range, s.text.map(_.text)))
+  implicit def newSynthOrder: Ordering[s.Synthetic] = Ordering.by(_.range)
```
xeno-by: Why is this called `newSynthOrder`?

semanticdb/metap/src/main/scala/scala/meta/internal/metap/SyntheticPrinter.scala
```diff
@@ -1,28 +1,84 @@
 package scala.meta.internal.metap
 
-import scala.math.Ordering
-import scala.meta.internal.semanticdb._
+import scala.meta.internal.{semanticdb => s}
```
xeno-by: Other printers use the wildcard `import scala.meta.internal.semanticdb._`, so I suggest we do the same here for consistency.

semanticdb/metap/src/main/scala/scala/meta/internal/metap/SyntheticPrinter.scala
```diff
@@ -1,28 +1,84 @@
...
+  implicit def newSynthOrder: Ordering[s.Synthetic] = Ordering.by(_.range)
+
+  private class TreePrinter(notes: InfoNotes) extends InfoPrinter(notes) {
+
+    def pprint(tree: s.Tree): Unit = {
+      tree match {
+        case tree: s.ApplyTree =>
+          pprint(tree.fn)
+          out.print("(")
+          rep(tree.args, ",")(pprint)
```
xeno-by: The Scalameta tree printer (as well as related printers in other Scala compilers) use `, ` (comma and whitespace) to separate arguments. Let's do that too.

semanticdb/metap/src/main/scala/scala/meta/internal/metap/SyntheticPrinter.scala
```diff
@@ -1,28 +1,84 @@
...
+          out.print(")")
+        case tree: s.FunctionTree =>
+          out.print("{")
+          rep("(", tree.params, ",", ") => ")(pprint)
```
xeno-by: Same suggestion regarding comma and whitespace.

semanticdb/metap/src/main/scala/scala/meta/internal/metap/SyntheticPrinter.scala
```diff
@@ -1,28 +1,84 @@
...
+          opt(doc.substring(tree.range))(out.print)
+          out.print(")")
+        case tree: s.SelectTree =>
+          pprint(tree.qual)
+          out.print(".")
+          opt(tree.id)(pprint)
+        case tree: s.TypeApplyTree =>
+          pprint(tree.fn)
+          out.print("[")
+          rep(tree.targs, ",")(pprint)
```
xeno-by: Same suggestion regarding comma and whitespace.

semanticdb/metap/src/main/scala/scala/meta/internal/metap/SyntheticPrinter.scala
```diff
@@ -1,28 +1,84 @@
...
+          out.print("]")
+        case s.Tree.Empty =>
```
xeno-by: To be consistent with `NoType`, `NoConstant` and `NoSignature`, please add the `NoTree` alias.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/PrinterOps.scala
```diff
@@ -11,27 +11,6 @@ import scala.reflect.internal.ModifierFlags._
 trait PrinterOps { self: SemanticdbOps =>
```
xeno-by: Let's remove this.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TreeOps.scala
```diff
@@ -0,0 +1,44 @@
+package scala.meta.internal.semanticdb.scalac
+
+import scala.meta.internal.{semanticdb => s}
+import scala.meta.internal.semanticdb.Scala._
+import scala.meta.internal.inputs._
+
+trait TreeOps { self: SemanticdbOps =>
```
xeno-by: `SyntheticOps`?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TreeOps.scala
```diff
@@ -0,0 +1,44 @@
...
+      case gTree: g.Select => gTree.toSemanticId
+      case gTree: g.Ident => gTree.toSemanticId
+      case gTree: g.This => gTree.toSemanticId
+      case gTree: g.Typed =>
+        s.MacroExpansionTree(
+          expandee = gTree.expr.toSemanticTree,
+          tpe = gTree.tpt.tpe.toSemanticTpe
+        )
+      case _ =>
+        println(s"No match on: $gTree ${gTree.getClass}")
```
xeno-by: Commonly used error message in this situation is something along the lines of: `s"unsupported tree: ..."`.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TreeOps.scala
```diff
@@ -0,0 +1,44 @@
+package scala.meta.internal.semanticdb.scalac
+
+import scala.meta.internal.{semanticdb => s}
+import scala.meta.internal.semanticdb.Scala._
+import scala.meta.internal.inputs._
+
+trait TreeOps { self: SemanticdbOps =>
+  import g._
+
+  implicit class XtensionGTreeSyntheticTerm(gTree: g.Tree) {
```
xeno-by: `XtensionGTreeSTree`?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TreeOps.scala
```diff
@@ -0,0 +1,44 @@
...
+          tpe = gTree.tpt.tpe.toSemanticTpe
+        )
+      case _ =>
+        println(s"No match on: $gTree ${gTree.getClass}")
+        s.Tree.Empty
+    }
+
+    def toSemanticId: s.IdTree = s.IdTree(sym = gTree.symbol.toSemantic)
+
+    def toSemanticOriginalTree: s.Tree = s.OriginalTree(
```
xeno-by: `toSemanticOriginal` to be consistent with `toSemanticId`?

semanticdb/semanticdb/semanticdb.proto
```diff
@@ -320,5 +320,55 @@ message Diagnostic {
 
 message Synthetic {
   Range range = 1;
-  TextDocument text = 2;
+  Tree tree = 2;
+}
+
+message Tree {
+  oneof sealed_value {
+    ApplyTree apply = 1;
```
xeno-by: Common convention here would be to say `applyTree`, `functionTree`, etc.

semanticdb/semanticdb/semanticdb.proto
```diff
@@ -320,5 +320,55 @@ message Diagnostic {
...
+message IdTree {
+  string sym = 1;
+}
+
+message LiteralTree {
+  Constant const = 1;
+}
+
+message MacroExpansionTree {
+  Tree expandee = 1;
```
xeno-by: Expandee can be an arbitrary term, so this is not going to work unless we model all Scalac terms. I suggest we just drop this field.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TreeOps.scala
xeno-by: Also, I don't think we should be issuing unpositioned and unconditional messages in production. I suggest we remove this println unless we're willing to introduce dedicated configuration option for this (personally, I don't think that's worth it).

tests/jvm/src/test/resources/metac.expect
```diff
@@ -1065,9 +1065,8 @@ Occurrences:
 [42:21..42:25): head => scala/collection/IterableLike#head().
 
 Synthetics:
-[39:9..39:9):  => *[Unit]
-  [0:0..0:1): * => _star_.
-  [0:2..0:6): Unit => scala/Unit#
+[39:2..39:9): orig(locally) => orig(locally)[Unit]
```
xeno-by: I think that we should optimize for the common case and print `*` instead of `orig(<the original tree on the lhs>)`.

xeno-by: Also, as mentioned above, I don't think we should print `orig()` around the original tree on the lhs.

semanticdb/semanticdb/semanticdb.proto
```diff
@@ -320,5 +320,55 @@ message Diagnostic {
 
 message Synthetic {
   Range range = 1;
-  TextDocument text = 2;
+  Tree tree = 2;
+}
```
xeno-by: I agree that the current API is suboptimal when it comes to reconstructing a fully desugared tree. Let's open a ticket for that.

tests/jvm/src/test/resources/metac.expect
```diff
@@ -3214,228 +3205,146 @@ Occurrences:
...
+  apply => scala/collection/immutable/Stream.apply().
+  Int => scala/Int#
+[10:6..10:10): orig(Name) => orig(Name).unapplySeq
+  unapplySeq => scala/util/matching/Regex#unapplySeq().
+[11:4..11:26): orig(#:: 2 #:: Stream.empty) => orig(#:: 2 #:: Stream.empty)[Int]
+  Int => scala/Int#
+[11:8..11:26): orig(2 #:: Stream.empty) => consWrapper[Int](orig(2 #:: Stream.empty))
+  consWrapper => scala/collection/immutable/Stream.consWrapper().
+  Int => scala/Int#
+[11:10..11:26): orig(#:: Stream.empty) => orig(#:: Stream.empty)[Int]
```
xeno-by: Looks like the old logic didn't emit this synthetic. Nice improvement.

tests/jvm/src/test/resources/metac.expect
```diff
@@ -3214,228 +3205,146 @@ Occurrences:
...
+  i => local6
+  withFilter => scala/collection/TraversableLike#withFilter().
+  j => local7
+  map => scala/collection/generic/FilterMonadic#map().
+  j => local8
+  canBuildFrom => scala/collection/immutable/IndexedSeq.canBuildFrom().
+[18:12..18:13): orig(1) => intWrapper(orig(1))
+  intWrapper => scala/LowPriorityImplicits#intWrapper().
+[18:26..18:27): orig(0) => intWrapper(orig(0))
+  intWrapper => scala/LowPriorityImplicits#intWrapper().
```
xeno-by: For loops look great!

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/Inferred.scala
```diff
@@ -1,85 +0,0 @@
...
-      range = Some(pos.toRange),
-      text = Some(
-        s.TextDocument(
-          text = synthetic.text,
-          occurrences = occurrences
-        )
-      )
-    )
-  }
-}
```
xeno-by: I like how the new data model makes this unnecessary!

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -53,8 +53,9 @@ trait TextDocumentOps { self: SemanticdbOps =>
       val binders = mutable.Set[m.Position]()
       val occurrences = mutable.Map[m.Position, String]()
       val symbols = mutable.Map[String, s.SymbolInformation]()
-      val synthetics = mutable.Map[m.Position, Inferred]().withDefaultValue(Inferred())
+      val synthetics = mutable.ListBuffer[s.Synthetic]()
       val isVisited = mutable.Set.empty[g.Tree] // macro expandees can have cycles, keep tracks of visited nodes.
+      val visitedSyntheticParent = mutable.Set.empty[g.Tree] // synthetics we have already visited the parents of
```
xeno-by: `isVisitedParent`?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -371,11 +372,6 @@ trait TextDocumentOps { self: SemanticdbOps =>
           private def tryFindInferred(gtree: g.Tree): Unit = {
```
xeno-by: `tryFindSynthetic`?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -385,62 +381,165 @@ trait TextDocumentOps { self: SemanticdbOps =>
...
+                case gtree if isForSynthetic(gtree) =>
+                  val range = gtree.pos.toMeta.toRange
+                  val synthTree = forSyntheticOrOrig(gtree)
+                  synthetics += s.Synthetic(
+                    range = Some(range),
+                    tree = synthTree
+                  )
+                case _ =>
+                // do nothing
+              }
```
xeno-by: This looks plausible as far as my mental debugging goes. I can't guarantee that this is 100% correct, since the for loop desugaring in Scalac is way too involved, but we have tests for that.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TreeOps.scala
```diff
@@ -0,0 +1,44 @@
...
+          targs = gTree.args.map(_.tpe.toSemanticTpe)
+        )
+      case gTree: g.Select => gTree.toSemanticId
+      case gTree: g.Ident => gTree.toSemanticId
+      case gTree: g.This => gTree.toSemanticId
+      case gTree: g.Typed =>
+        s.MacroExpansionTree(
+          expandee = gTree.expr.toSemanticTree,
+          tpe = gTree.tpt.tpe.toSemanticTpe
+        )
```
xeno-by: This should be something along the following lines:

```
case gtree if gtree.hasAttachment[g.analyzer.MacroExpansionAttachment] =>
  s.MacroExpansionTree(tpe = gtree.tpt.tpe.toSemanticTpe)
```

semanticdb/metap/src/main/scala/scala/meta/internal/metap/SyntheticPrinter.scala
maxov: I like the symmetry it has on the printouts, and it is still a quote from the original source.

semanticdb/semanticdb/semanticdb.proto
maxov: Not sure how to phrase the ticket

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
maxov: It is not correct with `=` statements, but I decided that this was out of scope as it was not tested in current tests. For example:

```scala
for {
  x <- 1 to 10
  y = x
} yield x + y
```

The following is produced:
```scala
for (x <- 1 to 10; y = x) yield x + y =>
orig(1 to 10)
.map[Tuple2[Int, Int], IndexedSeq[Tuple2[Int, Int]]]({ (x) =>
    orig(y = x)
})(canBuildFrom[Tuple2[Int, Int]])
.map[Int, IndexedSeq[Int]]({ (local10) =>
    orig(x + y)
})(canBuildFrom[Int])
```

Which is not terrible, but is different from expected:

```scala
orig(1 to 10)
.map[(Int, Int), scala.collection.immutable.IndexedSeq[(Int, Int)]](
    ((x: Int) => {
    val y = x;
    scala.Tuple2.apply[Int, Int](x, y)
    }))(scala.collection.immutable.IndexedSeq.canBuildFrom[(Int, Int)])
.map[Int, Any](((x$1: (Int, Int)) =>
    (x$1: @scala.unchecked) match {
    case scala.Tuple2((x @ _), (y @ _)) => orig(x + y)
    }))(scala.collection.immutable.IndexedSeq.canBuildFrom[Int])
```
Although to fully support this, we'd need to add a match ADT node or something to the synthetics tree..

tests/jvm/src/test/resources/metac.expect
maxov: I would like to push against this, it is extremely helpful to see the full context when looking at synthetics, it was invaluable for me to debug for statements and I think it would be useful for users of metap.

xeno-by: I should've been more precise in this formulation. 

Having parts of the original tree wrapped in `orig(...)` is super helpful. That's one of the very important features of this PR - both datamodel-wise and prettyprinting-wise.

However, when the original tree is used verbatim in the synthetic, this introduces extra verbosity that I'm not sure is worth the cost. For example, currently:

```
[13:14..13:36): #:: 2 #:: Stream.empty => orig(#:: 2 #:: Stream.empty)[Int]
  Int => scala/Int#
```

Here the right-hand side basically duplicates whatever is written on the left-hand side, completely dominating the otherwise pretty small synthetic. Compare this with the proposed:

```
[13:14..13:36): #:: 2 #:: Stream.empty => *[Int]
  Int => scala/Int#
```

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
xeno-by: Please submit a ticket.

semanticdb/semanticdb/semanticdb.proto
xeno-by: This may be an indication that the problem is too abstract for now. In that case, the ticket would probably be not worth it.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
maxov: Created #1711

PR #1722, opened 2018-07-27
A few specification cleanups

xeno-by:
/cc @maxov


xeno-by:
The PR only modifies the spec, so I'll be merging without waiting for CI.

PR #1723, opened 2018-07-27
Assorted improvements

xeno-by:


semanticdb/metacp/src/main/scala/scala/meta/internal/classpath/ClasspathIndex.scala
```diff
@@ -40,6 +36,9 @@ final class ClasspathIndex private (val dirs: collection.Map[String, Classdir])
   def isClassdir(path: String): Boolean =
     dirs.contains(path)
 
+  override def toString: String = {
+    s"ClasspathIndex($classpath)"
```
olafurpg: Is this helpful? In my experience the full classpath is too verbose and is not helpful in debugging a problem. I would prefer to expose the `Classpath` so that users can drill into `.entries` to investigate.

semanticdb/metacp/src/main/scala/scala/meta/internal/classpath/ClasspathIndex.scala
```diff
@@ -13,11 +13,7 @@ import scala.meta.io.Classpath
 import scala.meta.internal.io.PathIO
 
 /** An index to lookup class directories and classfiles by their JVM names. */
-final class ClasspathIndex private (val dirs: collection.Map[String, Classdir]) {
-
-  override def toString: String = {
-    s"ClasspathIndex(${dirs.size} entries)"
-  }
+final class ClasspathIndex private (classpath: Classpath, dirs: collection.Map[String, Classdir]) {
```
olafurpg: I propose we expose the ability to query into `dirs` somehow, either make it a val or add a method `getClassdir(path: String): Option[Classdir]`. Making it private makes it impossible to investigate problems such as the bugs I discovered in https://github.com/scalameta/scalameta/pull/169

scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Position.scala
```diff
@@ -41,4 +41,17 @@ object Position {
...
   }
+
+  object Range {
+    def apply(
+        input: Input,
+        startLine: Int,
+        startColumn: Int,
+        endLine: Int,
+        endColumn: Int): Position.Range = {
+      val start = input.lineToOffset(startLine) + startColumn
```
olafurpg: 👍 I already implemented this in Scalafix and I agree it should be in the public API. 

However, we should decide what to do when the column numbers exceed the column width in the actual input. This implementation is not LSP compliant, where `column=Int.MaxValue` will translate to the largest possible column for the corresponding line number. I opened https://github.com/scalameta/scalameta/issues/1728 to discuss this better.

semanticdb/metap/src/main/scala/scala/meta/internal/metap/RangePrinter.scala
```diff
@@ -17,26 +18,19 @@ trait RangePrinter extends BasePrinter {
...
-              i += 1
-            }
-            if (buf.last != chars.length) buf += chars.length
-            buf.toArray
-          })
-          val startOffset = lineToOffset(range.startLine) + range.startCharacter
-          val endOffset = lineToOffset(range.endLine) + range.endCharacter
-          Some(doc.text.substring(startOffset, endOffset))
+          val input = inputCache.getOrElseUpdate(doc, Input.String(doc.text))
+          val pos = Position.Range(
```
olafurpg: 🙏 Thank you!

semanticdb/metacp/src/main/scala/scala/meta/internal/classpath/ClasspathIndex.scala
xeno-by: That was helpful to debug what classpath Scalafix receives from our Pants integration. Exposing classpath would've achieved the same goal, so I'm totally up for that.

semanticdb/metacp/src/main/scala/scala/meta/internal/classpath/ClasspathIndex.scala
xeno-by: Discussion in https://github.com/scalameta/scalameta/issues/1729.

scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Position.scala
xeno-by: Great catch!! Thank you for following up!

PR #1726, opened 2018-07-30
Assorted improvements to metacp and metai that make everything work on a case study project

xeno-by:
@olafurpg This is the most significant of the pull requests that I merged while you were away. You'll want to take a look at this one, since I imagine there may be a few controversial decisions in here.

### DIFF
--- a/semanticdb/cli/src/main/scala/scala/meta/cli/Reporter.scala
+++ b/semanticdb/cli/src/main/scala/scala/meta/cli/Reporter.scala
@@ -1,6 +1,7 @@
 package scala.meta.cli
 
 import java.io._
+import scala.meta.internal.cli._
 
 final class Reporter private (val out: PrintStream, val err: PrintStream) {
   private def this() = {
@@ -25,11 +26,6 @@ final class Reporter private (val out: PrintStream, val err: PrintStream) {
 
   private def copy(out: PrintStream = out, err: PrintStream = err): Reporter = {
     new Reporter(out = out, err = err)
-  }
-
-  private def devnull: PrintStream = {
-    val os = new OutputStream { override def write(b: Int) = () }
-    new PrintStream(os)
   }
 }
 
--- a/semanticdb/cli/src/main/scala/scala/meta/cli/Reporter.scala
+++ b/semanticdb/cli/src/main/scala/scala/meta/cli/Reporter.scala
@@ -11,12 +11,25 @@ final class Reporter private (val out: PrintStream, val err: PrintStream) {
     copy(out = out)
   }
 
+  def silenceOut(): Reporter = {
+    withOut(devnull)
+  }
+
   def withErr(err: PrintStream): Reporter = {
     copy(err = err)
   }
 
+  def silenceErr(): Reporter = {
+    withErr(devnull)
+  }
+
   private def copy(out: PrintStream = out, err: PrintStream = err): Reporter = {
     new Reporter(out = out, err = err)
+  }
+
+  private def devnull: PrintStream = {
+    val os = new OutputStream { override def write(b: Int) = () }
+    new PrintStream(os)
   }
 }
 
--- a/semanticdb/metac/src/main/scala/scala/meta/cli/Metac.scala
+++ b/semanticdb/metac/src/main/scala/scala/meta/cli/Metac.scala
@@ -1,6 +1,7 @@
 package scala.meta.cli
 
 import java.io._
+import scala.meta.internal.cli.Args
 import scala.meta.internal.metac._
 import scala.meta.metac._
 
@@ -20,7 +21,8 @@ object Metac {
   }
 
   private def process(args: Array[String], reporter: Reporter): Int = {
-    Settings.parse(args.toList, reporter) match {
+    val expandedArgs = Args.expand(args)
+    Settings.parse(expandedArgs, reporter) match {
       case Some(settings) =>
         if (process(settings, reporter)) 0
         else 1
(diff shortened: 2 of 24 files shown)


semanticdb/metacp/src/main/scala/scala/meta/internal/metacp/Main.scala
```diff
@@ -178,7 +177,7 @@ class Main(settings: Settings, reporter: Reporter) {
                   reporter.out.println(e.getMessage)
                   success = false
                 }
-              case NonFatal(ex) =>
+              case ex: Throwable =>
```
olafurpg: https://github.com/scalameta/scalameta/issues/1730

semanticdb/metai/src/main/scala/scala/meta/metai/Result.scala
```diff
@@ -0,0 +1,13 @@
+package scala.meta.metai
+
+import scala.meta.io.AbsolutePath
+
+final class Result private (val status: Map[AbsolutePath, Boolean]) {
+  def success: Boolean = status.forall(_._2)
```
olafurpg: Nitpick: I would name this `def isSuccess`

I try to start method names with `isX` when possible if they return boolean responses.

semanticdb/metai/src/main/scala/scala/meta/internal/metai/Main.scala
```diff
@@ -47,12 +47,11 @@ final class Main(settings: Settings, reporter: Reporter) {
   }
 
   private def processEntry(file: ClasspathFile): Boolean = {
-    var hasSemanticdb = false
+    val hasSemanticdb = file.path.resolve("META-INF").resolve("semanticdb").isDirectory
```
olafurpg: iirc, this did not work when I first implemented this because we did not call `Files.createDirectories` before creating semanticdb files. Keep that in mind for other tools that produce SemanticDB files.

semanticdb/metai/src/main/scala/scala/meta/internal/metai/Main.scala
```diff
@@ -47,12 +47,11 @@ final class Main(settings: Settings, reporter: Reporter) {
...
     val ls = FileIO.listAllFilesRecursively(file.path)
```
olafurpg: If you can assume the directory zip entry `META-INF/semanticdb` exists then you can simplify this logic with

```scala
val root = file.path.resolve("META-INF").resolve("semanticdb")
val ls = FileIO.listAllFilesrecursively(root)
ls.files.foreach { ... }
```

semanticdb/cli/src/main/scala/scala/meta/cli/Reporter.scala
```diff
@@ -11,13 +11,26 @@ final class Reporter private (val out: PrintStream, val err: PrintStream) {
     copy(out = out)
   }
 
+  def silenceOut(): Reporter = {
```
olafurpg: Nitpick: I find it weird to use the builder pattern with methods that are not named "withX". I would have named this method `withSilentOut`

semanticdb/metac/src/main/scala/scala/meta/cli/Metac.scala
```diff
@@ -20,7 +21,8 @@ object Metac {
   }
 
   private def process(args: Array[String], reporter: Reporter): Int = {
-    Settings.parse(args.toList, reporter) match {
+    val expandedArgs = Args.expand(args)
```
olafurpg: Why expand the args before calling `Settings.parse`? For me, this logic is the responsibility of `Settings.parse`. I don't have a strong opinion however as long as it's consistent.

semanticdb/metai/src/main/scala/scala/meta/internal/metai/Main.scala
```diff
@@ -25,19 +25,18 @@ final class Main(settings: Settings, reporter: Reporter) {
...
-            processEntry(entry)
-          } catch {
-            case e: Throwable =>
-              reporter.err.println(s"Error indexing ${entry.pathOnDisk}:")
-              e.printStackTrace(reporter.err)
-              false
-          }
+      var success = true
+      try {
+        Classpath(entry).foreach { entry =>
```
olafurpg: Note that this changes the behaviour from error accumulation to failing on first error.

semanticdb/metai/src/main/scala/scala/meta/metai/Result.scala
xeno-by: Will fix.

semanticdb/metai/src/main/scala/scala/meta/internal/metai/Main.scala
xeno-by: Yeah, I had to fix this in metacp.

semanticdb/cli/src/main/scala/scala/meta/cli/Reporter.scala
xeno-by: Will fix.

semanticdb/metac/src/main/scala/scala/meta/cli/Metac.scala
xeno-by: What I had in mind here is that the `@args` expansion is only relevant to CLI invocations, since it's only the CLI invocations that are limited by the max size of a shell command. As you said, this is consistent - both within Scalameta and within Rsc.

PR #1732, opened 2018-07-31
upgrade sbt-jdi-tools to a JDK 9+ friendly version

### LINKED ISSUE #1719: build fails on Java 9+ (sbt-jdi-tools isn't compatible)
f9c534f93c630f21cff4dee03cb96e77e01364a8 added `addSbtPlugin("org.scala-debugger" % "sbt-jdi-tools" % "1.0.0")` but this breaks the build (as in, you can't even start sbt) on Java 9 and 10, because of https://github.com/chipsenkbeil/sbt-jdi-tools/issues/5

yes, Scalameta is normally built on Java 8, but we want to build it on Java 9 and 10 as part of the Scala community build. or at least, we want the build definition to be loadable, so we can use the same project list regardless of JDK version

SethTisue:
fixes #1719

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -36,5 +36,4 @@ addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.22")
 
 addSbtPlugin("org.scala-native" %% "sbt-scala-native" % "0.3.8")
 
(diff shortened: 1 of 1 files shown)


xeno-by:
LGTM

PR #1738, opened 2018-08-04
Fix my part of the remaining 4.0.0-RC1 tickets

xeno-by:


xeno-by:
Changeset that upgrades Scalafix: https://github.com/xeno-by/scalafix/pull/7.

xeno-by:
Changeset that upgrades Rsc: https://github.com/xeno-by/rsc/commit/9d698528fb964240390ff2502162388374c3d407.

xeno-by:
I also did a successful dry run of an internal upgrade, which discovered three oversights in the original pull requests. After those three fixes, it's all good now.

semanticdb/javac/src/main/scala/scala/meta/internal/semanticdb/javac/semantics/Elements.scala
```diff
@@ -178,6 +178,7 @@ trait Elements {
             }
           }
           s.TypeSignature(
+            typeParameters = Some(s.Scope()),
```
olafurpg: I think we should consider using the `scalapb.no_box = true` option, that would avoid further problems like this. Worth squeezing into v4?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SemanticdbPipeline.scala
```diff
@@ -28,7 +29,7 @@ trait SemanticdbPipeline extends SemanticdbOps { self: SemanticdbPlugin =>
   }
 
   def handleCrash(unit: Option[g.CompilationUnit]): PartialFunction[Throwable, Unit] = {
-    case ex: Throwable =>
+    case NonFatal(ex) =>
```
olafurpg: Your raised a good question about why metac defaults to warnings instead of errors. I think it would be good to change the default to error with v4. You in favor of that change?

xeno-by: Yes. I won't have time to get this through myself this week, but if you do, that'd be great.

semanticdb/javac/src/main/scala/scala/meta/internal/semanticdb/javac/semantics/Elements.scala
xeno-by: Let's open a ticket and defer to post-v4? If we do no_box here, maybe we'd want to do no_box in other places, too, and it'd be great to do that kind of change at once.

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -823,7 +823,7 @@ languages map onto these properties.
   <tr>
     <td><code>0x8000</code></td>
     <td><code>DEFAULT</code></td>
-    <td>Is a default parameter?</td>
+    <td>Is a default parameter or a default method?</td>
```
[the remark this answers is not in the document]
xeno-by: We could separate these flags, especially since there may be confusion between default getters and default methods.

Do you have any suggestions how to call the flags? `DEFAULT_PARAM` and `DEFAULT_METHOD` unfortunately don't work due to the reason mentioned above.

semanticdb/javac/src/main/scala/scala/meta/internal/semanticdb/javac/semantics/Elements.scala
olafurpg: I opened https://github.com/scalameta/scalameta/issues/1741

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SemanticdbPipeline.scala
olafurpg: I may take a stab at it, opened https://github.com/scalameta/scalameta/issues/1742

olafurpg:
Merging to unblock my work on the latest tickets and it's unclear what design is best for `p.DEFAULT`

PR #1747, opened 2018-08-16
Last-minute changes before the 4.0 code freeze

xeno-by:


xeno-by:
Tentative Rsc upgrade to this pull request: https://github.com/xeno-by/rsc/commits/topic/upgrade.

semanticdb/semanticdb/semanticdb.proto
```diff
@@ -270,7 +270,7 @@ message SymbolInformation {
   Language language = 16;
   Kind kind = 3;
   int32 properties = 4;
-  string name = 5;
+  string display_name = 5;
```
olafurpg: It would make upgrades easier if we add a deprecated extension method `SymbolInformation.name` that forwards to `displayName`

xeno-by: On the other hand, compilation errors may force the upgrader to consider whether they want `SymbolInformation.name` or `sym.desc.name`. This was super valuable to me when I was upgrading Scalameta and Rsc.

I don't mind either way though. Let me know your preference, and I'll submit a follow-up pull request if you'd prefer to have an extension method.

PR #1756, opened 2018-08-27
Add pretty-printers for synthetic/signature/tree/constant

olafurpg:
These avoid the need for scalafix to poke further into internals.


olafurpg:
Ignoring the appveyor error since the build failed to trigger due to GitHub being down when I pushed the commit.

Issue #1749, opened 2018-08-21
Missing symbol occurrences when enabling synthetics


fikovnik:
There are missing symbol occurrences when I use an custom sbt plugin to generate semanticdb  (the code is based on @olafurpg code - just minor changes from my side). When using metac all is good. Happens with both 4.0.0-M8 and M9.

For example, the semanticdb for the following file (a simplified version of [commons.scala](https://github.com/ensime/ensime-server/blob/3.0/api/src/main/scala/org/ensime/api/common.scala#L47) where I hit the problem first):
```scala
import java.io.File

sealed trait T1 extends Ordered[T1] {
  def file: File
  def text: String
  def from: Int
  def to: Int

  import scala.math.Ordered.orderingToOrdered

  def compare(that: T1): Int =
    (file, from, this.to, this.text)
      .compare((that.file, that.from, that.to, that.text))
}
```

will be missing all symbol occurrences from the line 11 (0-based numbering) `(file, from, this.to, this.text)`: 

```
Occurrences:
[0:7..0:11): java => java/
[0:12..0:14): io => java/io/
[0:15..0:19): File => java/io/File#
[2:13..2:15): T1 <= _empty_/T1#
[2:24..2:31): Ordered => scala/package.Ordered#
[2:32..2:34): T1 => _empty_/T1#
[3:6..3:10): file <= _empty_/T1#file().
[3:12..3:16): File => java/io/File#
[4:6..4:10): text <= _empty_/T1#text().
[4:12..4:18): String => scala/Predef.String#
[5:6..5:10): from <= _empty_/T1#from().
[5:12..5:15): Int => scala/Int#
[6:6..6:8): to <= _empty_/T1#to().
[6:10..6:13): Int => scala/Int#
[8:9..8:14): scala => scala/
[8:15..8:19): math => scala/math/
[8:20..8:27): Ordered => scala/math/Ordered.
[8:28..8:45): orderingToOrdered => scala/math/Ordered.orderingToOrdered().
[10:6..10:13): compare <= _empty_/T1#compare().
[10:14..10:18): that <= _empty_/T1#compare().(that)
[10:20..10:22): T1 => _empty_/T1#
[10:25..10:28): Int => scala/Int#
[12:7..12:14): compare => scala/math/Ordered#compare().
[12:16..12:20): that => _empty_/T1#compare().(that)
[12:21..12:25): file => _empty_/T1#file().
[12:27..12:31): that => _empty_/T1#compare().(that)
[...]

PR #1758, opened 2018-08-28
Fix for missing symbol occurrences in synthetics implicit views (cf. #1749)

fikovnik:


semanticdb/integration/src/main/scala/example/Synthetic.scala
```diff
@@ -44,4 +44,10 @@ class Synthetic {
     if a < b
   } yield a
 
+  // See https://github.com/scalameta/scalameta/issues/1749
+  import scala.math.Ordered.orderingToOrdered
+  val x1 = new F
+  val x2 = new F
+  (x1, x1)
```
olafurpg: Please `./bin/scalafmt --diff` to make CI green.

semanticdb/integration/src/main/scala/example/Synthetic.scala
```diff
@@ -44,4 +44,10 @@ class Synthetic {
     if a < b
   } yield a
 
+  // See https://github.com/scalameta/scalameta/issues/1749
+  import scala.math.Ordered.orderingToOrdered
```
olafurpg: Just curious, can this import and `new F` be replaced with `val x = 42; (x, x)`? Would be nice to minimize the test case as much as possible

fikovnik:
I could do the example without import, but in that case I would have to define more terms (unless you can think of a better way) which will congest the example. I think that in this form it is fairly obvious what the intention is and it is visible from the `metac.expect` what has been missing.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -460,7 +460,7 @@ trait TextDocumentOps { self: SemanticdbOps =>
                 case gimpl: g.ApplyToImplicitArgs =>
                   gimpl.fun match {
                     case gview: g.ApplyImplicitView =>
-                      isVisited += gview
+                      isVisitedParent += gview
```
olafurpg: Impressive you tracked this down 😍

PR #1763, opened 2018-09-09
Reformat quasiquote spec according to commonmark

olafurpg:
This commit formats the quasiquote markdown document with Prettier
(https://prettier.io/) to make it compatible with commonmark so that it
renders normally in Docusaurus. To make the tables render correctly, we
alsoy escape pipe characters `|` for `Pat.Alternative` and `Type.Or`


olafurpg:
Before this PR 

screen shot 2018-09-09 at 20 28 46

After this PR

screen shot 2018-09-09 at 20 30 53

PR #1767, opened 2018-09-12
Introduce --stub-broken-signatures for Metacp

xeno-by:
Many JVM-related tools like Javac, Scalac (and the Hotspot JVM itself)
perform lazy linking, which means that missing dependencies aren't
a problem unless they are explicitly used.

At the moment, Metacp is uncharacteristically strict about this.
Since it performs eager conversion to SemanticDB, any missing dependencies
result in an immediate error.

While GlobalSymbolTable solves this problem by being lazy, there are
still some use cases (one of them being Rsc) that rely on eager
conversion. That doesn't work too well in a number of scenarios
including shading and `<dependency><optional>true</optional></dependency>`.

This commit provides a workaround for this problem by introducing the
ability to silence missing symbol errors and emit empty signatures instead.


baroquebobcat:
Nice!

PR #1775, opened 2018-09-19
Update guide to use v4.0.0 [CI SKIP]

olafurpg:


olafurpg:
Merging right away so the website can pick up these changes.

PR #1778, opened 2018-09-24
Add documentation for Trees.scala

MasseGuillaume:
See https://github.com/scalameta/scalameta/pull/1380

### DIFF
--- a/notes/trees.md
+++ b/notes/trees.md
@@ -1,0 +1,264 @@
+# Trees.scala
+
+## Top level (`meta.Source`, `meta.Pkg`, `meta.Pkg.Object`)
+
+```scala
+"package a".parse[Source].get.structure
+
+Source(stats = List(Pkg(ref = Term.Name("a"), stats = Nil)))
+```
+
+## Declarations (`meta.Decl`)
+
+```scala
+"val a: Int".parse[Stat].get.structure
+
+Decl.Val(
+  mods = Nil,
+  pats = List(Pat.Var(name = Term.Name("a"))),
+  decltpe = Type.Name("Int")
+)
+```
+
+Decl            | Examples
+----------------|-----------
+`Decl.Val`      | `val a: Int`
+`Decl.Var`      | `var a: Int`
+`Decl.Def`      | `def f: Int`
+`Decl.Type`     | `type T`
+
+
+## Definitions (`meta.Defn`)
+
+```scala
+"val a = 1".parse[Stat].get.structure
+
+Defn.Val(
+  mods = Nil,
+  pats = List(Pat.Var(name = Term.Name("a"))),
+  decltpe = None,
+  rhs = Lit.Int(value = 1)
+)
+```
+
+Defn         | Examples
+-------------|-----------
+`Defn.Val`   | `val a = 1`
+`Defn.Var`   | `var a = 1`
+`Defn.Def`   | `def f = 1`
+`Defn.Macro` | `def f = macro impl`
+`Defn.Type`  | `type T = Int`
+`Defn.Class` | `class A`
+`Defn.Trait` | `trait A`
+`Defn.Object`| `object A`
+
+## Terms (`meta.Term`)
+
+```scala
+"1 + 1".parse[Term].get.structure
+
+Term.ApplyInfix(
+  lhs = Lit.Int(value = 1),
+  op = Term.Name("+"),
+  targs = Nil,
+  args = List(Lit.Int(value = 1))
+)
+```
+
+Ref               | Examples
+------------------|-----------
+`Term.This`       | `this`, `a.this`
+`Term.Super`      | `super`, `a.super`, `super[a]`, `a.super[b]`
+`Term.Name`       | `a`
+`Term.Select`     | `a.b`
+`Term.ApplyUnary` | `!a`, `~a`, `-a`, `a`
+
+Term                    | Examples
+------------------------|------------------------------
+`Term.Apply`            | `f(42)`
+`Term.ApplyType`        | `implicitly[Ordering[Int]]`
+`Term.ApplyInfix`       | `a + a`
+`Term.Assign`           | `a = 1`
+`Term.Return`           | `return a`
+`Term.Throw`            | `throw e`
+`Term.Ascribe`          | `a: Int`
+`Term.Annotate`         | `(x: @annot)`
+`Term.Tuple`            | `(1, 1)`
+`Term.Block`            | `{ f1(); f2() }`
+`Term.If`               | `if (p) t else f`
+`Term.Match`            | `x match { case _ => }`
+`Term.Try`              | `try f catch { case _ => }`
+`Term.TryWithHandler`   | `try f catch (h)`
+`Term.Function`         | `(a, b) => a + b`
+`Term.PartialFunction`  | `{ case _ =>  }`
+`Term.While`            | `while(p){ f() }`
+`Term.Do`               | `do { f() } while(p)`
+`Term.For`              | `for{x <- xs} f(x)`
+`Term.ForYield`         | `for{x <- xs} yield f(x)`
+`Term.New`              | `new A`
+`Term.NewAnonymous`     | `new A { }`
+`Term.Placeholder`      | `_`
+`Term.Eta`              | `f _`
+`Term.Repeated`         | `f(x: _*)`
+`Term.Param`            | `x: Int` in `def f(x: Int)`
+`Term.Interpolate`      | `s"Hello $name"`
+`Term.Xml`              | `<h1>Hello {name}</h1>`
+
+## Types (`meta.Type`)
+
+```scala
+"A with B".parse[Type].get.structure
+
+Type.With(lhs = Type.Name("A"), rhs = Type.Name("B"))
+```
+
+Ref              | Examples
+-----------------|---------
+`Type.Name`      | `B`
+`Type.Select`    | `a.B`
+`Type.Project`   | `a#B`
+`Type.Singleton` | `a.type`
+
+Type                    | Examples
+------------------------|--------------------
+`Type.Apply`            | `F[T]`
+`Type.ApplyInfix`       | `K Map V`
+`Type.Function`         | `A => B`
+`Type.ImplicitFunction` | `implicit A => B`
+`Type.Tuple`            | `(A, B)`
+`Type.With`             | `A with B`
+`Type.And`              | `A & B`
+`Type.Or`               | <code> A &#124; B </code>
+`Type.Refine`           | `A { def f: Int }`
+`Type.Existential`      | `A forSome { type T }`
+`Type.Annotate`         | `T @annot`
+`Type.Lambda`           | `[T] => (T, T)` see below (only for supported dialects)
+`Type.Method`           | `(x: T): T` see below (only for supported dialects)
+`Type.Placeholder`      | `_` in `T[_]`
+`Type.Bounds`           | `T >: Lower <: Upper` in `def F[T >: Lower <: Upper] = 1`
+`Type.ByName`           | `=>T` in `def f(x: => T) = x`
+`Type.Repeated`         | `T*` in `def f(x: T*): Unit`
+`Type.Var`              | `t` in `case _: List[t] =>`
+`Type.Param`            | `X` in `trait A[X]`
+
+## Patterns (meta.Pat) and Cases (meta.Case)
+
+```scala
+"_: A | _: B".parse[Pat].get.structure
+
+Pat.Alternative(
+  Pat.Typed(Pat.Wildcard(), Type.Name("A")),
+  Pat.Typed(Pat.Wildcard(), Type.Name("B"))
+)
+```
+
+Pat                  | Examples
+---------------------|----------------------------
+`Lit                 | `"literal"`
+`Pat.Wildcard`       | `_`
+`Pat.SeqWildcard`    | `_*` in `case List(xs @ _*) =>`
+`Pat.Var`            | `a` in `case a =>`
+`Pat.Bind`           | `a @ A()`
+`Pat.Alternative`    | `1 | 2`
+`Pat.Tuple`          | `(a, b)`
+`Pat.Extract`        | `A(a, b)`
+`Pat.ExtractInfix`   | `a E b`
+`Pat.Interpolate`    | `r"Hello (.+)$name"`
+`Pat.Xml`            | `<h1>Hello, World!</h1>`
+`Pat.Typed`          | `a: Int`
+`Term.Select`        | `a.b`
+
+
+```scala
+"case a => ()".parse[Case].get.structure
+
+Case(Pat.Var(Term.Name("a")), None, Lit.Unit(()))
+```
+
+Case    | Examples
+--------|----------------------
+`Case`  | `case a if p => f()`
+
+## Name (`meta.Name`)
+
+`meta.Name.Anonymous()`
+
+```scala
+"class B { }".parse[Source].get.structure
+
+Source(
+  stats = List(
+    Defn.Class(
+      mods = Nil,
+      name = Type.Name("B"),
+      tparams = Nil,
+      ctor = Ctor.Primary(
+        mods = Nil,
+        name = Name.Anonymous(),
+        paramss = Nil
+      ),
+      templ = Template(
+        early = Nil,
+        inits = Nil,
+        self = Self(
+          name = Name.Anonymous(),
+          decltpe = None
+        ),
+        stats = Nil
+      )
+    )
+  )
+)
+```
+
+`meta.name.Indeterminate()`
+
+Is it a type, or is it a term? Indeterminate.
+
+```scala
+"import a.b".parse[Source].get.structure
+
+Source(
+  stats = List(
+    Import(
+      importers = List(
+        Importer(
+          ref = Term.Name("a"),
+          importees = List(
+            Importee.Name(
+              name = Name.Indeterminate(
+                value = b
+              )
+            )
+          )
+        )
+      )
+    )
+  )
+)
+```
+
+## Literals (`meta.Lit`)
+
+```scala
+"null".parse[Term].get.structure
+
+Lit.Null()
+```
+
+Lit        | Example
+-----------|-------
+`Null`     |  `null`
+`Boolean`  |  `true`
+`Unit`     |  `()`
+`Int`      |  `1`
+`Double`   |  `1.0`
+`Float`    |  `1.0F`
+`Long`     |  `1L`
+`Byte`[1]  |  `1Z`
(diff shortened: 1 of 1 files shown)


notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Declarations (`meta.Decl`)
+
+```scala
+"val a: Int".parse[Stat].get.structure
+```
+
+Decl            | Examples
+----------------|-----------
+Decl.Val        | val a
```
olafurpg: Can we include a `: Int` type annotations here? The ability to leave out `: Unit` here is a deprecated feature

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Terms (`meta.Term`)
+
+```scala
+"1 + 1".parse[Term].get.structure
+```
+
+Ref             | Examples
+----------------|-----------
+Term.This       | `this`, `foo.this`
```
olafurpg: Can we use something else than `foo`? For example `a.this`, which is consistent with `a` above

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Term.Super      | `super`, `foo.super`, `super[bar]"`, `foo.super[bar]`
+Term.Name       | `foo`
+Term.Select     | `bar.foo`
+Term.ApplyUnary | `!foo`, `~foo`, `-foo`, `foo`
+
+Term                  | Examples
+----------------------|------------------------------
+Term.Apply            | `buzz(42)`
```
olafurpg: `f(42)`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Term.ApplyType        | `implicitly[Ordering[Int]]`
+Term.ApplyInfix       | `a + a`
+Term.Assign           | `a = 1`
+Term.Return           | `return bar`
+Term.Throw            | `throw exception`
+Term.Ascribe          | `a: Int`
+Term.Annotate         | `(x: @unchecked)`
+Term.Tuple            | `(1, 1)`
+Term.Block            | `{ bar; buzz }`
```
olafurpg: `f1(); f2();`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Term.If               | `if (p) t else f`
+Term.Match            | `x match { case _ => }`
+Term.Try              | `try foo catch { case _ => }`
```
olafurpg: `try f()`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Term.TryWithHandler   | `try foo catch (h)`
+Term.Function         | `(a, b) => a + b`
+Term.PartialFunction  | `{ case _ =>  }`
+Term.While            | `while(p){ d }`
```
olafurpg: `while (p) { f() }`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Term.Do               | `do { d } while(p)`
+Term.For              | `for{x <- xs} f(x)`
+Term.ForYield         | `for{x <- xs} yield f(x)`
+Term.New              | `new A`
+Term.NewAnonymous     | `new A { }`
+Term.Placeholder      | `_`
+Term.Eta              | `foo _`
```
olafurpg: `f _`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Term.Repeated         | `foo(x: _*)`
+Term.Param            | `def foo(x: Int)`
```
olafurpg: `x: Int` in `def f(x: Int)`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Term.Interpolate      | `s"foo ~ $bar ~ ${42}"`
```
olafurpg: `s"Hello $name"`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Term.Xml              | `<h1>Hello ~ {bar} ~ {42}</h1>`
```
olafurpg: `Hello {name}`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
+# Trees.scala
+
+## Top level (`meta.Source`, `meta.Pkg`, `meta.Pkg`)
+
+```scala
+"package a".parse[Source].get.structure
```
olafurpg: Can we include the output from `.structure` here? Same for the examples below

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+```scala
+"A with B".parse[Type].get.structure
+```
+
+Ref            | Examples
+--------------------------
+Type.Name      | `B`
+Type.Select    | `a.B`
+Type.Project   | `a#B`
+Type.Singleton | `t.type`
```
olafurpg: `a.type`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+```
+
+Defn        | Examples
+------------|-----------
+Defn.Val    | val a = 1
+Defn.Var    | var a = 1
+Defn.Def    | def f = 1
+Defn.Macro  | def f = macro impl
+Defn.Type   | type T = Int
+Defn.Class  | class A
```
olafurpg: can we use `class T` and `trait T` but `object A`? Classes and traits create types while objects create terms.

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Types (`meta.Type`)
+
+```scala
+"A with B".parse[Type].get.structure
+```
+
+Ref            | Examples
+--------------------------
+Type.Name      | `B`
```
olafurpg: `T`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Type.Select    | `a.B`
```
olafurpg: `a.T`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Type.Project   | `a#B`
```
olafurpg: `a#T`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Type                  | Examples
+----------------------|--------------------
+Type.Apply            | `F[T]`
+Type.ApplyInfix       | `K Map V`
+Type.Function         | `A => B`
+Type.ImplicitFunction | `implicit A => B`
+Type.Tuple            | `(A, B)`
+Type.With             | `A with B`
+Type.And              | `A & B`
+Type.Or               | `A \| B`
```
olafurpg: This does not work in commonmark I believe, at least the remarkable JS parser doesn't support it. Use `<code> A &#124; B </code>` instead

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Type.Refine           | `A { def f: Int }`
```
olafurpg: `T { def f1: Int; def f2: Int }`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Term.Select     | `bar.foo`
+Term.ApplyUnary | `!foo`, `~foo`, `-foo`, `foo`
+
+Term                  | Examples
+----------------------|------------------------------
+Term.Apply            | `buzz(42)`
+Term.ApplyType        | `implicitly[Ordering[Int]]`
+Term.ApplyInfix       | `a + a`
+Term.Assign           | `a = 1`
+Term.Return           | `return bar`
```
olafurpg: `return a`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Term.Throw            | `throw exception`
+Term.Ascribe          | `a: Int`
+Term.Annotate         | `(x: @unchecked)`
```
olafurpg: `a: @unchecked`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Type.Apply            | `F[T]`
+Type.ApplyInfix       | `K Map V`
+Type.Function         | `A => B`
+Type.ImplicitFunction | `implicit A => B`
+Type.Tuple            | `(A, B)`
+Type.With             | `A with B`
+Type.And              | `A & B`
+Type.Or               | `A \| B`
+Type.Refine           | `A { def f: Int }`
+Type.Existential      | `A forSome { type T }`
```
olafurpg: `T1 forSome { type T2 }`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Type.Annotate         | `T @A`
+Type.Lambda           | `[X] => (X, X)`
+Type.Method           | see bellow
```
olafurpg: `(x: T): T` (only for supported dialects)

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Type.Function         | `A => B`
+Type.ImplicitFunction | `implicit A => B`
+Type.Tuple            | `(A, B)`
+Type.With             | `A with B`
+Type.And              | `A & B`
+Type.Or               | `A \| B`
+Type.Refine           | `A { def f: Int }`
+Type.Existential      | `A forSome { type T }`
+Type.Annotate         | `T @A`
+Type.Lambda           | `[X] => (X, X)`
```
olafurpg: `[T] => (T, T)` (only for supported dialects)


NOTE: needs same treatment as `Type.Method`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Type.Method           | see bellow
+Type.Placeholder      | `T[_ >: Nothing <: Any]`
+Type.Bounds           | `def F[T >: Nothing <: Any] = 1`
```
olafurpg: `def f[T >: Lower <: Upper]`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Type.ByName           | `def f(v: => T) = v`
```
olafurpg: `=>T` in `def f(x: =>T)`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Type.Repeated         | `def print(v: Any*): Unit`
```
olafurpg: `x: Int*` in `def f(x: Int*)`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Type.Var              | `case _: List[t] =>`
```
olafurpg: `t` in `case _: List[t] =>`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Type.Param            | `trait A[X]`
```
olafurpg: `X` in `trait T[X]`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Type.ApplyInfix       | `K Map V`
+Type.Function         | `A => B`
+Type.ImplicitFunction | `implicit A => B`
+Type.Tuple            | `(A, B)`
+Type.With             | `A with B`
+Type.And              | `A & B`
+Type.Or               | `A \| B`
+Type.Refine           | `A { def f: Int }`
+Type.Existential      | `A forSome { type T }`
+Type.Annotate         | `T @A`
```
olafurpg: `T @annot`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Type.Lambda           | `[X] => (X, X)`
+Type.Method           | see bellow
+Type.Placeholder      | `T[_ >: Nothing <: Any]`
+Type.Bounds           | `def F[T >: Nothing <: Any] = 1`
+Type.ByName           | `def f(v: => T) = v`
+Type.Repeated         | `def print(v: Any*): Unit`
+Type.Var              | `case _: List[t] =>`
+Type.Param            | `trait A[X]`
+
+`Type.Method`
```
olafurpg: This takes up unnecessary real-estate IMO considering `Type.Method` is not used in any application I'm aware of (it was created for a use-case in Scalafix that no longer exists). For consistency with the quasiquote spec https://scalameta.org/docs/trees/quasiquotes.html, I propose to keep "(only for supported dialects)" in the tables and drop this here.

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Patterns (meta.Pat) and Cases (meta.Case)
+
+```scala
+"""_: Foo | _: Bar""".parse[Pat].get.structure
+```
+
+Pat                | Examples
+-------------------|----------------------------
+Lit                | `1`
```
olafurpg: How about "`1` in `case 1 =>`"? It's not clear that this is in a pattern. Same for the examples below.

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Pat.Wildcard       | `_`
+Pat.SeqWildcard    | `_*`
```
olafurpg: `_*` in `case List(xs @ _*) =>`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Pat.Var            | `foo`
```
olafurpg: `a` in `case a =>`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Pat.Bind           | `foo @ Ext`
+Pat.Alternative    | `1 | 2`
+Pat.Tuple          | `(foo, bar)`
+Pat.Extract        | `Ext(foo, bar)`
+Pat.ExtractInfix   | `foo Ext bar`
+Pat.Interpolate    | `r"example (.+)$foo"`
+Pat.Xml            | `<h1>Hello, World!</h1>`
+Pat.Typed          | `foo: Int`
+Term.Name          | `foo`
```
olafurpg: Remove from this section? It's not clear from the document what is the different from `Pat.Var`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+"""_: Foo | _: Bar""".parse[Pat].get.structure
+```
+
+Pat                | Examples
+-------------------|----------------------------
+Lit                | `1`
+Pat.Wildcard       | `_`
+Pat.SeqWildcard    | `_*`
+Pat.Var            | `foo`
+Pat.Bind           | `foo @ Ext`
```
olafurpg: `case a @ A() =>`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Pat.Alternative    | `1 | 2`
+Pat.Tuple          | `(foo, bar)`
+Pat.Extract        | `Ext(foo, bar)`
```
olafurpg: `case A(a1, a2) =>`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+
+Pat                | Examples
+-------------------|----------------------------
+Lit                | `1`
+Pat.Wildcard       | `_`
+Pat.SeqWildcard    | `_*`
+Pat.Var            | `foo`
+Pat.Bind           | `foo @ Ext`
+Pat.Alternative    | `1 | 2`
+Pat.Tuple          | `(foo, bar)`
```
olafurpg: `case (a, b) =>`

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+Case               | `p"case $pat if $expropt => $expr"`
+
+## Name (`meta.Name`)
+
+`meta.name.Anonymous()`
+
+```scala
+"class B { }".parse[Source].get.structure
+
+Defn.Class(
```
olafurpg: The `width=1` formatting is not helping readability in my opinion, something like width=40 or 60 might be better.

olafurpg: Might be good to include label names as well.

notes/trees.md
```diff
@@ -0,0 +1,234 @@
...
+  Template(
+    Nil,
+    Nil,
+    Self(Name.Anonymous(), None),
+    Nil
+  )
+)
+```
+
+`meta.name.Indeterminate()`
```
olafurpg: `meta.Name.Indeterminate()`

notes/trees.md
MasseGuillaume: T is for type parameters.

notes/trees.md
MasseGuillaume: Why two stats ?

olafurpg: It can be a single stat `def f: Int`. It we want to be consistent with `Term.Block` however, we should update every place where a list of  arguments is expected to use `f1(); f2(); ...`. Overkill?

MasseGuillaume:
@olafurpg I updated the PR to address your comments.

notes/trees.md
olafurpg: `s/n/N`

notes/trees.md
```diff
@@ -0,0 +1,311 @@
...
+`Case`  | `case a if p => f()`
+
+## Name (`meta.Name`)
+
+`meta.Name.Anonymous()`
+
+```scala
+"class B { }".parse[Source].get.structure
+
+"Source(List(Defn.Class(Nil, Type.Name(\"B\"), Nil, Ctor.Primary(Nil, Name(\"\"), Nil), Template(Nil, Nil, Self(Name(\"\"), None), Nil))))"
```
olafurpg: use `println(Tree.structure)` to avoid the `\"` escapes in the ammonite repl

olafurpg: it would be good to not wrap in a string

notes/trees.md
```diff
@@ -0,0 +1,311 @@
...
+`Type.Function`         | `A => B`
+`Type.ImplicitFunction` | `implicit A => B`
+`Type.Tuple`            | `(A, B)`
+`Type.With`             | `A with B`
+`Type.And`              | `A & B`
+`Type.Or`               | <code> A &#124; B </code>
+`Type.Refine`           | `A { def f: Int }`
+`Type.Existential`      | `A forSome { type T }`
+`Type.Annotate`         | `T @A`
+`Type.Lambda`           | `[T] => (T, T)` see bellow (only for supported dialects)
```
olafurpg: s/bellow/below/

notes/trees.md
olafurpg: wdyt?

notes/trees.md
```diff
@@ -0,0 +1,311 @@
...
+```
+
+Defn         | Examples
+-------------|-----------
+`Defn.Val`   | `val a = 1`
+`Defn.Var`   | `var a = 1`
+`Defn.Def`   | `def f = 1`
+`Defn.Macro` | `def f = macro impl`
+`Defn.Type`  | `type T = Int`
+`Defn.Class` | `class A`
```
olafurpg: `class T`, I think it's important to use different characters for `Term.Name` and `Type.Name`.

ShaneDelmore:
This is incredibly useful, thanks @MasseGuillaume!

olafurpg:
This document is live on the website https://scalameta.org/docs/trees/examples.html

PR #1786, opened 2018-10-02
Update synthetics default value

olafurpg:


xeno-by:
Looks like our mima infrastructure is unhappy because it can't find a 4.0.0-RC1 plugin for 2.12.7.

olafurpg:
aha, that explains it, RC1 is hardcoded in build.sbt. I've updated the mima checks to compare against 4.0.0 instead

PR #1793, opened 2018-10-07
Faster tree traversal

marcelocenerine:
While profiling Scalafix I found out that the main hotspot is `scala.meta.Traverser`. It accounts for ~60% of the running time (see details here: https://github.com/scalacenter/scalafix/issues/894). It turns out that the code generated by the `scala.meta.internal.transversers.TraverserMacros` macro annotation contains a large amount of pattern matching making traversal terribly slow.

This PR introduces a new traverser which relies on `Tree.children`. JMH benchmarks show that it makes tree traversal ~20-70 times faster than the existing one:

>As input for the benchmark, I picked some random *.scala files with 3 (XS), 50 (S), 500 (M), 1000 (L) and 5800 (XL) lines of code and measured the cost of traversing the entire tree. FYR, the source code can be found here https://github.com/marcelocenerine/scalameta-benchmark. 

```
[info] Benchmark                            (size)  Mode  Cnt      Score     Error  Units
[info] TraverserBenchmark.simple_traverser      XS  avgt   10      0.389 ±   0.021  us/op
[info] TraverserBenchmark.simple_traverser       S  avgt   10      8.444 ±   0.134  us/op
[info] TraverserBenchmark.simple_traverser       M  avgt   10    248.154 ±  38.820  us/op
[info] TraverserBenchmark.simple_traverser       L  avgt   10    320.614 ±   2.501  us/op
[info] TraverserBenchmark.simple_traverser      XL  avgt   10   1958.657 ±  22.550  us/op
[info] TraverserBenchmark.traverser             XS  avgt   10     26.749 ±   0.320  us/op
[info] TraverserBenchmark.traverser              S  avgt   10    361.805 ±  28.225  us/op
[info] TraverserBenchmark.traverser              M  avgt   10   5192.518 ± 114.954  us/op
[info] TraverserBenchmark.traverser              L  avgt   10   6553.171 ± 125.726  us/op
[info] TraverserBenchmark.traverser             XL  avgt   10  36190.514 ± 585.949  us/op
```

Methods in `scala.meta.contrib.TreeOps` and `scala.meta.transversers.Api` were updated to use the new traverser. Other utilities under `.internal.` remains as is (same for the alias in `scala.meta.transversers.Aliases`). `scala.meta.transversers.Transformer` (and everything that relies on it) also remains as is.

`scala.meta.internal.transversers.TraverserMacros` (and the supertype `scala.meta.internal.transversers.TransverserMacros`) contains some additional logic to ignore non-relevant fields as well as to sort fiels by priority. For the sake of simplicity, those were not implemented in the new `SimpleTraverser`. Despite that, the traversal order appears to have been preserved (tested on the following repos: spark, akka, scala, kafka, scalafix and scalameta). If there is any edge case that requires that special logic, please let me know.

/cc @olafurpg

### DIFF
--- a/scalameta/testkit/src/test/scala/scala/meta/tests/transversers/TraverserOrder.scala
+++ b/scalameta/testkit/src/test/scala/scala/meta/tests/transversers/TraverserOrder.scala
@@ -1,0 +1,46 @@
+package scala.meta.tests
+package transversers
+
+import org.scalatest.FunSuite
+
+import scala.collection.mutable
+import scala.meta.Tree
+import scala.meta.testkit.SyntaxAnalysis
+import scala.meta.tests.contrib.ContribSuite
+import scala.meta.transversers.{SimpleTraverser, Traverser}
+
+class TraverserOrder extends FunSuite {
+  test("traversal order is preserved") {
+    val errors = SyntaxAnalysis.onParsed[Tree](ContribSuite.corpus) { ast =>
+      val trees1 = asList_Traverser(ast)
+      val trees2 = asList_SimpleTraverser(ast)
+
+      if (trees1.size != trees2.size) List(ast)
+      else {
+        val firstDiff =
+          trees1.zip(trees2)
+            .collectFirst { case (a, b) if a != b => a }
+        firstDiff.toList
+      }
+    }
+    assert(errors.isEmpty)
+  }
+
+  private def asList_Traverser(tree: Tree): List[Tree] = {
+    val buf = mutable.ListBuffer[Tree]()
+    object traverser extends Traverser {
+      override def apply(tree: Tree): Unit = { buf += tree; super.apply(tree) }
+    }
+    traverser(tree)
+    buf.toList
+  }
+
+  private def asList_SimpleTraverser(tree: Tree): List[Tree] = {
+    val buf = mutable.ListBuffer[Tree]()
+    object traverser extends SimpleTraverser {
+      override def apply(tree: Tree): Unit = { buf += tree; super.apply(tree) }
+    }
+    traverser(tree)
+    buf.toList
+  }
+}
--- a/scalameta/transversers/shared/src/main/scala/scala/meta/transversers/SimpleTraverser.scala
+++ b/scalameta/transversers/shared/src/main/scala/scala/meta/transversers/SimpleTraverser.scala
@@ -1,0 +1,6 @@
+package scala.meta
+package transversers
+
+private[meta] class SimpleTraverser {
+  def apply(tree: Tree): Unit = tree.children.foreach(apply)
+}
--- a/tests/shared/src/test/scala/scala/meta/tests/transversers/TransformerSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/transversers/TransformerSuite.scala
@@ -1,0 +1,56 @@
+package scala.meta.tests.transversers
+
+import org.scalatest._
+
+import scala.compat.Platform.EOL
+import scala.meta._
+
+class TransformerSuite extends FunSuite {
+
+  test("Transformer Ok") {
+    val tree0 = q"""
+      def foo(x: x)(x: Int) = x + x
+      class C(x: x) {
+        def bar(x: x) = ???
+      }
+    """
+    val log = scala.collection.mutable.ListBuffer[String]()
+    object transformer extends Transformer {
+      override def apply(tree: Tree): Tree = tree match {
+        case Term.Name("x") => Term.Name("y")
+        case Type.Name("x") => Type.Name("y")
+        case _ => super.apply(tree)
+      }
+    }
+    val tree1 = transformer(tree0)
+    assert(tree1.toString === """
+      |{
+      |  def foo(y: y)(y: Int) = y + y
+      |  class C(y: y) { def bar(y: y) = ??? }
+      |}
+    """.trim.stripMargin.split('\n').mkString(EOL))
+  }
+
+  test("Transformer Fail") {
+    val tree0 = q"""
+      def foo(x: x)(x: Int) = x + x
+      class C(x: x) {
+        def bar(x: x) = ???
+      }
+    """
+    val log = scala.collection.mutable.ListBuffer[String]()
+    object transformer extends Transformer {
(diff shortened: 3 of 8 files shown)


scalameta/transversers/shared/src/main/scala/scala/meta/transversers/SimpleTraverser.scala
```diff
@@ -0,0 +1,6 @@
+package scala.meta
+package transversers
+
+private[meta] class SimpleTraverser {
+  def apply(tree: Tree): Unit = tree.children.foreach(apply)
```
marcelocenerine: Neither traversers are stack safe. It turns out that there is a file in the Kafka repo that causes the existing traverser to blow up the stack (bug?): https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/server/KafkaConfig.scala:
```
Exception in thread "main" java.lang.StackOverflowError
	at scala.meta.Term$Apply$TermApplyImpl.args(Trees.scala:81)
	at scala.meta.Term$Apply$TermApplyImpl.args(Trees.scala:81)
	at scala.meta.Term$Apply$TermApplyImpl.args(Trees.scala:81)
```

I also benchmarked a stack-safe traverser but turned out to be 30-50% slower than this one.

tests/shared/src/test/scala/scala/meta/tests/transversers/TransformerSuite.scala
```diff
@@ -0,0 +1,56 @@
+package scala.meta.tests.transversers
```
marcelocenerine: no changes to the code below... I just separated the transformer's tests from the traverser's

tests/shared/src/test/scala/scala/meta/tests/transversers/TraverserSuite.scala
```diff
@@ -0,0 +1,56 @@
+package scala.meta.tests
```
marcelocenerine: same here

scalameta/transversers/shared/src/main/scala/scala/meta/transversers/SimpleTraverser.scala
xeno-by: How about we get rid of the autogenerated Traverser altogether? Then SimpleTraverser would simply become Traverser.

marcelocenerine: done

marcelocenerine: this causes Mima to complain because the macro used to generate some additional `apply` methods which are no longer needed:
```
[error]  * method apply(scala.Option)
[error]  * method apply(scala.collection.immutable.List)
[error]  * method apply(scala.Option,org.scalameta.overload.Hack1)
[error]  * method apply(scala.collection.immutable.List,org.scalameta.overload.Hack2)
[error]  * object scala.meta.transversers.Traverser does not have a correspondent in current version
```
Are you ok in adding exceptions in Mima for those or maybe have a better suggestion since it breaks backwards compatibility?

marcelocenerine:
deleting the autogenerated traverser causes Mima to complain about some missing `apply` methods which used to be generated by the macro but are no longer needed:
```
[error]  * method apply(scala.Option)
[error]  * method apply(scala.collection.immutable.List)
[error]  * method apply(scala.Option,org.scalameta.overload.Hack1)
[error]  * method apply(scala.collection.immutable.List,org.scalameta.overload.Hack2)
[error]  * object scala.meta.transversers.Traverser does not have a correspondent in current version
```
Are you ok in adding exceptions in Mima for those or maybe have a better suggestion since it breaks backwards compatibility?

scalameta/transversers/shared/src/main/scala/scala/meta/transversers/Traverser.scala
```diff
@@ -1,5 +1,6 @@
 package scala.meta
 package transversers
 
-@scala.meta.internal.transversers.traverser
-abstract class Traverser
+abstract class Traverser {
```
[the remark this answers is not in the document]
marcelocenerine: ok, I just had to drop the last commit

marcelocenerine:
>How did you test this?

I did that using a scalafix rule which transformed the tree into a list using both the old and new traversers and compared the sizes as well as the order of each node.

> Even if it's not a unit test we run on every PR, it might be good to have commit the analysis using SyntaxAnalysis.onParsed

Done. Please see if that's correct. I did not manage to run the test using the sbt command you suggested though... (nor I figured a different way)

scalameta/transversers/shared/src/main/scala/scala/meta/transversers/SimpleTraverser.scala
olafurpg: It's normal that users need to increase the stack size to use scalameta, we probably don't do a good enough job at documenting it however 🤔

scalameta/testkit/src/test/scala/scala/meta/tests/transversers/TraverserOrder.scala
```diff
@@ -0,0 +1,46 @@
...
+  test("traversal order is preserved") {
+    val errors = SyntaxAnalysis.onParsed[Tree](ContribSuite.corpus) { ast =>
+      val trees1 = asList_Traverser(ast)
+      val trees2 = asList_SimpleTraverser(ast)
+
+      if (trees1.size != trees2.size) List(ast)
+      else {
+        val firstDiff =
+          trees1.zip(trees2)
+            .collectFirst { case (a, b) if a != b => a }
```
olafurpg: 👍

scalameta/transversers/shared/src/main/scala/scala/meta/transversers/SimpleTraverser.scala
```diff
@@ -0,0 +1,6 @@
+package scala.meta
+package transversers
+
+private[meta] class SimpleTraverser {
```
olafurpg: I will follow up with a PR to make this class public so that users can extend it to implement custom traversals like documented here https://scalameta.org/docs/trees/guide.html#custom-traversals

PR #1801, opened 2018-10-25
Shade fastparse v1 under scala.meta.internal.fastparse

### LINKED ISSUE #1800: Upgrade to fastparse 2.0
Found this while trying to use scalameta with ammonite-repl 1.3.2 which uses fastparse 2.0 .

olafurpg:
To avoid binary conflicts with fastparse v2. This commit doesn't fully
solve our problem since scalapb still brings in fastparse, we may have
to shade scalapb in a separate PR.

Fix #1800

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -360,7 +360,7 @@ lazy val tokenizers = crossProject(JSPlatform, JVMPlatform, NativePlatform)
     description := "Scalameta APIs for tokenization and their baseline implementation",
     // NOTE(olafur): use shaded version of fastparse because the latest version v2.0.0
     // has binary incompatibilites with the v1.0.0 version used by Scalameta.
-    libraryDependencies += "com.geirsson" %%% "fastparse" % "1.0.0",
+    libraryDependencies += "org.scalameta" %%% "fastparse" % "1.0.0",
     enableMacros
   )
   .nativeSettings(nativeSettings)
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -358,7 +358,9 @@ lazy val tokenizers = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .settings(
     publishableSettings,
     description := "Scalameta APIs for tokenization and their baseline implementation",
-    libraryDependencies += "com.lihaoyi" %%% "fastparse" % "1.0.0",
+    // NOTE(olafur): use shaded version of fastparse because the latest version v2.0.0
+    // has binary incompatibilites with the v1.0.0 version used by Scalameta.
+    libraryDependencies += "com.geirsson" %%% "fastparse" % "1.0.0",
```
olafurpg: I can publish this under `org.scalameta` for that matter, I realised com.geirsson wasn't so suitable after the release was finished and I'm too lazy to do another release.

xeno-by: If you don't mind, let's fork fastparse under the scalameta organization and publish under `org.scalameta`.

olafurpg: Done, forked to scalameta/fastparse and updated the default branch to the shaded commit https://github.com/scalameta/fastparse

PR #1804, opened 2018-11-07
Remove typer instrumentation

olafurpg:
Instrumentation of typer has caused problems in the past

- difficult to upgrade compiler versions
- incompatibilities with other compiler plugins

While testing metals in the spotify/scio and typelevel/cats repos, I
encountered infinite loops during scalamacros/paradise expansion caused
by typer instrumentation.  This problem was the last drop in the bucket
in my opinion and enough justification to remove typer instrumentation.

If it turns out we will need it then we can add it back later.


olafurpg:
Unfortunately, the `OriginalTreeAttachment` is not available in 2.11 so we'll have to fall back to `None` in those cases. For applications where the typer instrumentation is critical the `-Dsemanticdb.instrument-typer=true` option is available to enable instrumentation. I believe that the best default is to disable instrumentation however since infinite while loops the worse than missing symbol occurrences.

More information on infinite while loop

- it happens here https://github.com/scalamacros/paradise/blob/a6137aac8f6f3c9f67455fd2c003256772b2af2c/plugin/src/main/scala/org/scalamacros/paradise/typechecker/Namers.scala#L501-L509
- my theory for why it happens is that reference equality in  `cx != g.analyzer.NoContext` always returns false after instrumentation

olafurpg:
Another important bit of information on the infinite loop

- I am only able to reproduce it with a recent version of bloop. I am unable to reproduce from sbt or with bloop v1.0.0. 

I still feel uncomfortable with the typer hijacking and I am eager to disable it by default.

olafurpg:
Regression bug reported to bloop https://github.com/scalacenter/bloop/issues/705

The typer instrumentation triggers "missing implicit" compile errors without scalamacros/paradise enabled.

I still propose we go ahead with this change since the development cost of keeping it around is higher than the benefits it provides. I think it's OK keep `-Dsemanticdb.instrument-typer` around for a while longer in case we discover that the instrumentation is more critical than we think.

olafurpg:
I published a temporary release for semanticdb-scalac v4.0.0-163560a8 (latest commit in this PR) with the to maven central to unblock my work in metals. I can confirm that it not only fixes the infinite loop with macroparadise but also fixes spurious "missing implicit" compile errors when compiling scalafix.

xeno-by:
Superseded by https://github.com/scalameta/scalameta/pull/1810.

Issue #1817, opened 2018-12-21
Term.ApplyInfix prettyprint removes neccessary brackets


hugo-vrijswijk:
```scala
val add = q"def add(a: Int, b: Int) = a + b"
val list = q"val list = List(1, 2, 3)"
val added = q"list map (add(_, 1))"

added.syntax // list map add(_, 1)
```

The printed result gives a compile error as the brackets around `add(_, 1)` are needed while Scalameta removes them

hugo-vrijswijk:
I assume this is another issue with the tree printer, similar to #1792. I know @legopiraat had a go at fixing the previous issue I posted, but did not get very far due to the complexity of it

hugo-vrijswijk:
I've just run into another case where this goes wrong. When giving a `Seq` to a function that requires `varargs` it's common to add `foo: _*` (a `Term.Repeated`). 

Calling `.syntax` on the following expression `foo op (bar: _*)` returns `foo op bar: _*` which is invalid Scala code.

olafurpg:
good catch @hugo-vrijswijk I added a commit to #1818 fixing the repeated case.

hugo-vrijswijk:
I think we should re-open this issue. I've just found another case that this doesn't fix: `foo map (_.bar)` syntaxes to `foo map _.bar`, which is incorrect. Also see [this issue for Stryker4s](https://github.com/stryker-mutator/stryker4s/issues/148)

olafurpg:
@hugo-vrijswijk please open a separate issue. FWIW, the prettyprinter has a decent number of known issues. If I had a few weeks to work on it I would finish https://github.com/scalacenter/scala-syntax to replace the current printer but I estimate that won't happen in the near future. I will do my best to review fixes to the current printer in the meantime.

hugo-vrijswijk:
Alright, will do. I'll also have a look at a fix 👍

PR #1818, opened 2018-12-21
Attempted fix for #1817

Wmaarts:
Fix for #1817

### What this PR does
This implementation checks if an argument of an ApplyInfix Term is an Apply Term. If the Apply Term has a Placeholder Term as an argument, parentheses are added, because they are necessary.

These examples are tested: 
``` 
"list map (println)".syntax => "list map println".syntax  // Parentheses not necessary
"list map (add(1))".syntax == "list map add(1)"         // Parentheses not necessary
"list map (add(_, 1))".syntax == "list map (add(_, 1))" // Parentheses necessary
```

### Considerations
I'm not at all sure if all cases of this problem are solved with this. Also I have no experience with the scala-meta project, so I could very well be overlooking things. 😅


Wmaarts:
Hadn't signed the Scala CLA yet, but I have now.

olafurpg:
I messed up while trying to push a commit to this PR, will open separate PR. @Wmaarts I recommend using a different branch than `master` for opening a PR

PR #1827, opened 2019-02-01
Fix for #1826 ApplyInfix parentheses syntax

### LINKED ISSUE #1826: ApplyInfix syntax removes needed parentheses in Tuple and Placeholder
As discussed in #1817. 

Pretty printing syntax on both `Term.Select(Term.Placeholder, _)` and `Term.Tuple` remove parentheses which are needed.

I'm working on a fix ATM.

hugo-vrijswijk:
Fixes #1826 syntax printer bugs.

~One case still fails: `list map (_ diff foo)`. It adds an extra set of parentheses around the argument. I am unsure why that fails but another case `list map (_ -> foo)` does work, as they seem to have the same AST.~

### DIFF
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
@@ -299,8 +299,8 @@ object TreeSyntax {
                     case _ => false
                   } =>
                   true
-                // Include parentheses for `a op (bar: _*)`
-                case repeated: Term.Repeated =>
+                // Include parentheses for `a op (b: _*)`, tuple `a op ((b, c))`, `a op (_.bar), a op (_.opTwo(b))`, `a op (_ opTwo b)`
+                case _: Term.Repeated | _: Term.Tuple |  Term.Select(_: Term.Placeholder, _) | Term.Apply(Term.Select(_: Term.Placeholder, _), _) | Term.ApplyInfix(_: Term.Placeholder, _, _, _) =>
                   true
                 case _ =>
                   false
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
@@ -293,17 +293,20 @@ object TreeSyntax {
               s("(())")
             case (arg: Term) :: Nil =>
               val needsParens = arg match {
-                // Include parentheses for `a op (apply(b, _))`
-                case apply: Term.Apply if apply.args.exists {
+                case apply: Term.Apply if apply.args.exists { // `a op (apply(b, _))`
                     case _: Term.Placeholder => true
                     case _ => false
-                  } =>
+                  } => true
+                case Term.Select(_: Term.Placeholder, _) => // `a op (_.b)`
                   true
-                // Include parentheses for `a op (b: _*)`, tuple `a op ((b, c))`, `a op (_.bar), a op (_.opTwo(b))`, `a op (_ opTwo b)`
-                case _: Term.Repeated | _: Term.Tuple |  Term.Select(_: Term.Placeholder, _) | Term.Apply(Term.Select(_: Term.Placeholder, _), _) | Term.ApplyInfix(_: Term.Placeholder, _, _, _) =>
+                case Term.Apply(Term.Select(lhs: Term.Placeholder, _), _) => // `a op (_.b(c))`
                   true
-                case _ =>
+                case Term.ApplyInfix(lhs, _, _, _) if !lhs.isInstanceOf[Term.Placeholder] => // Not `a op (_ b c)`
                   false
+                case _: Lit | _: Term.Ref | _: Term.Function | _: Term.Apply | _: Term.If | _: Term.Match =>
+                  false
+                case _ =>
+                  true
               }
               if (needsParens) s("(", p(InfixExpr(t.op.value), arg, right = true), ")")
               else s(p(InfixExpr(t.op.value), arg, right = true))
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
@@ -301,14 +301,22 @@ object TreeSyntax {
                     case _: Term.Placeholder => true
                     case _ => false
                   }
-                case Term.ApplyInfix(lhs, _, _, _) if !lhs.isInstanceOf[Term.Placeholder] => // Not `a op (_ b c)`
-                  false
-                case _: Lit | _: Term.Ref | _: Term.Function | _: Term.If | _: Term.Match =>
+                case Term.ApplyInfix(_: Term.Placeholder, _, _, _) =>  // `a op (_ b c)`
+                  true
+                case _: Lit | _: Term.Ref | _: Term.Function | _: Term.If | _: Term.Match | _: Term.ApplyInfix =>
                   false
                 case _ =>
                   true
               }
-              if (needsParens) s("(", p(InfixExpr(t.op.value), arg, right = true), ")")
+              val checkPrecedence = arg match {
+                case Term.ApplyInfix(_: Term.Placeholder, _, _, _) =>
+                  false
+                case _ =>
+                  true
+              }
+              if (needsParens) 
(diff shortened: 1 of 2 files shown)


scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -292,8 +292,8 @@ object TreeSyntax {
                     case _ => false
                   } =>
                   true
-                // Include parentheses for `a op (bar: _*)`
-                case repeated: Term.Repeated =>
+                // Include parentheses for `a op (b: _*)`, tuple `a op ((b, c))`, `a op (_.bar), a op (_.opTwo(b))`, `a op (_ opTwo b)`
+                case _: Term.Repeated | _: Term.Tuple |  Term.Select(_: Term.Placeholder, _) | Term.Apply(Term.Select(_: Term.Placeholder, _), _) | Term.ApplyInfix(_: Term.Placeholder, _, _, _) =>
```
olafurpg: Would it make sense to flip the logic around and not introduce parentheses for a whitelist of trees? Maybe it's enough to only skip parentheses for `_: Term.Ref`

hugo-vrijswijk: Hmm, good idea. I'll try it

olafurpg:
Please ping me when this is ready for a review.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
hugo-vrijswijk: Ok, so I've (mostly) flipped it around. It's a little cleaner and should err on the side of adding parentheses in most cases. However, there's still some weird behaviour:

`list map (_ -> foo)` -> ` list map (_ -> foo)`
`list map (_ op foo)` -> ` list map ((_ op foo))` (Note the double parentheses)

I have no idea where this comes from, as far as I can tell [the AST's are the same](https://astexplorer.net/#/gist/38a1965132e6d507f974bd345c71caf1/df1fc34cbaa19b2c87e2087fd2856af24f141c62).

hugo-vrijswijk:
@olafurpg It's ready now. Have a look at my comment

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
olafurpg: Those parentheses could be caused by the pretty-printer for `Term.ApplyInfix`

hugo-vrijswijk: Isn't `_ -> foo` also a `Term.ApplyInfix` though? It seems to work there. With some testing it looks to me like the difference comes from using symbols or not

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
```diff
@@ -660,4 +660,28 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
     assert(q"list map (add(_, 1))".syntax == "list map (add(_, 1))")
     assert(q"list map (bar:_*)".syntax == "list map (bar: _*)")
   }
+  test("#1826 ApplyInfix parentheses on Select") {
+    assert(q"list map (_.bar)".syntax == "list map (_.bar)")
```
olafurpg: Can we introduce a helper method to run those assertions and verify the AST is unchanged?
```
def check(original: Tree, expected: String): Unit = {
  assertNoDiff(original.syntax == expected)
  assertNoDiff(original.structure = parseStat(expected).structure)
}
```

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -286,17 +286,20 @@ object TreeSyntax {
...
+                case Term.Select(_: Term.Placeholder, _) => // `a op (_.b)`
                   true
-                // Include parentheses for `a op (bar: _*)`
-                case repeated: Term.Repeated =>
+                case Term.Apply(Term.Select(lhs: Term.Placeholder, _), _) => // `a op (_.b(c))`
                   true
-                case _ =>
+                case Term.ApplyInfix(lhs, _, _, _) if !lhs.isInstanceOf[Term.Placeholder] => // Not `a op (_ b c)`
+                  false
+                case _: Lit | _: Term.Ref | _: Term.Function | _: Term.Apply | _: Term.If | _: Term.Match =>
```
olafurpg: Is `Term.Match` correct here?
```scala
scala> List(1) map _ match { case 1 => println("BOOM") }
<console>:12: error: missing parameter type for expanded function
The argument types of an anonymous function must be fully known. (SLS 8.5)
Expected type was: ?
       List(1) map _ match { case 1 => println("BOOM") }
                   ^

scala> List(1) map (_ match { case 1 => println("BOOM") })
BOOM
res1: List[Unit] = List(())
```

hugo-vrijswijk: I am not entirely sure, but it seems to want to add double parentheses without it. I've added a test for that case though (and there were some others)

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
hugo-vrijswijk: Done

olafurpg:
There seems to be a failing test case
```
- #1826 ApplyInfix parentheses on ApplyInfix function *** FAILED ***
[info]   "list map ([(_ diff foo)])" did not equal "list map ([_ diff foo])" (SyntacticSuite.scala:692)
[info]   Analysis:
[info]   "list map ([(_ diff foo)])" -> "list map ([_ diff foo])"
```

hugo-vrijswijk:
Yeah, I can't see why it fails though. It adds two parentheses on this case, but only one on a test with the same structure but different operator (`list map (_ -> foo)`)

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
olafurpg: It depends on how the pretty-printer interprets those trees, they can be printed differently even if the AST is identical. The code quality is quite bad in the pretty-printer, it was written before my time and I haven't touched it much.

RCMartins:
If you change line 304 from
`if (needsParens) s("(", p(InfixExpr(t.op.value), arg, right = true), ")")`
to
`if (needsParens) s("(", arg, ")")`
 it makes the `list map (_ diff foo)` test pass

The problem here is operator precedence. the `p` function checks if the parens are required based on the operators precedence. Because we already know it needs parens (and we don't want multiple) we should skip the call to `p`.

> Yeah, I can't see why it fails though. It adds two parentheses on this case, but only one on a test with the same structure but different operator (`list map (_ -> foo)`)

In the `list map (_ -> foo)` test `p` doesn't put parens because the operator `->` has greater precedence than `map`, you can see the precedence list it uses here: https://github.com/scalameta/scalameta/blob/232d56a457f92582ca2cc1ac5a8f96780a8415f4/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/package.scala#L41

---

But the next test in the same block: `list map (a diff foo)  --->  list map a diff foo` will still fail (it was already failing), but I think here the problem is with the test itself.
Because the operators `map` and `diff` have the same precedence you can't remove the parens.
`list map a diff foo` is the same as `list.map(a).diff(foo)`
`list map (a diff foo)` is the same as `list.map(a.diff(foo))`
So they are not equivalent expressions

olafurpg:
@RCMartins nice job tracking down the problem! Your analysis sounds correct 👍

hugo-vrijswijk:
@RCMartins thanks for the explanation! Very insightful. I've learned a lot about Scala syntax this week haha. The `list map a diff foo` was indeed incorrect (structure comparison also failed). The test cases and code should now be correct

olafurpg:
@hugo-vrijswijk I can try to cut a new release in the coming week with the pretty-printer fixes. Are there other blocking issues you'd like to get in first?

hugo-vrijswijk:
@olafurpg Not that I know of, but there could be 😉. But that'd be great!

PR #1836, opened 2019-02-25
Fix #1792

### LINKED ISSUE #1792: Incorrect syntax result of Term.Interpolate
Hi, when I have a `Term.Interpolate` that interpolates a variable with characters in the string right after it, Scalameta does not add brackets around the expression, resulting in an incorrect syntax.

Example:

```scala
val s = Term.Interpolate(Term.Name("s"),
                         List(Lit.String(""), Lit.String("bar")),
                         List(Term.Name("foo")))

s.syntax
```

Actual result: `s"$foobar"` (can cause compile error)

Expected result: `s"${foo}bar"`

Having a `Term.Select` or `Term.Apply` as the interpolated variable causes the brackets to appear properly (`s"${foo.baz()}bar"`).

Tried on Scala 2.12.7 with Scalameta versions 3.7.3 and 4.0.0

RCMartins:
This PR fixes #1792 and also fixes another problem with interpolation that I found about using `$$` in interpolation strings. I didn't found an existing issue for that problem. I added the test `interpolator with $ character` for it.


olafurpg:
Nice job tracking this down!

hugo-vrijswijk:
Also LGTM!

PR #1838, opened 2019-02-28
Fix #1837

### LINKED ISSUE #1837: Implicit val moved to start of class constructor (Prettyprinting)
```scala
val tree = q"class foo(number1: Int, implicit val number2: Int = 4)"
tree.syntax == "class foo(number1: Int, implicit val number2: Int = 4)" // false
tree.syntax == "class foo(implicit number1: Int, val number2: Int = 4)" // true
```

Also see https://github.com/stryker-mutator/stryker4s/issues/162

RCMartins:
I tried not to go too crazy on the test cases because there are so many combinations of modifiers.


hugo-vrijswijk:
Awesome! I don't fully see why it went wrong in the first place, but it looks good to me!

PR #1841, opened 2019-03-06
Use @JSImport to load native "fs" and "path" modules

gabro:
This effectively reverts https://github.com/scalameta/scalameta/commit/032dc56318bc81d3ecd4aa39040a15a74bf4fd8c#diff-e392e6315fa9beedb70580bda57714c0

According to [a discussion with Sébastien on Gitter](https://gitter.im/scala-js/scala-js?at=5c7fc7f1f895944c084d730b), accessing `require` from the Global scope is invalid. The reason it works is that tests are run into a sort-of node REPL, where the global scope has a `require` function. However, when running Scalameta as a script, `require` is `undefined`, causing a runtime failure.

It seems we have to use `@JSImport` to reference the native `path` and `fs` modules.

It used to be that way, but it was changed by @jonas at some point in 2017 (see the commit below), but I couldn't find any further context for that change.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -539,6 +539,9 @@ lazy val tests = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .jvmConfigure(
     _.dependsOn(testkit, interactive, metac, metacp, metai, symtab, semanticdbIntegration)
   )
+  .jsSettings(
+    scalaJSModuleKind := ModuleKind.CommonJSModule
+  )
   .nativeSettings(
     nativeSettings,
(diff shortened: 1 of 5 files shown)


olafurpg:
If I remember correctly, we needed to dynamically load the modules to make scalameta load in the browser for metabrowse. Have you validated how this change works in the browser?

gabro:
I've published this locally and I can confirm it fixes the issue I've seen happening in metarpheus

olafurpg:
Let's give this a chance and try to fix the webpack setup in metabrowse instead to mock out the node modules.

PR #1846, opened 2019-03-19
LanguageOps: Detect Typelevel compiler

tindzk:
This allows to use Typelevel-specific features like literal types.


tindzk:
I reformatted the diff.

PR #1862, opened 2019-04-25
Enable full cross-build release from CI.

olafurpg:


olafurpg:
Merging since this can only be tested on master.

PR #1880, opened 2019-06-23
Scala 2.13 supports by-name implicits

travisbrown:
Resubmitting this PR (originally #1879) as requested by the CLA error message in CI. I'm sure I've signed these CLAs before but I did them both again.

I've just copied the relevant tests from the Dotty dialect test suite—I'm happy to expand them if people want.


travisbrown:
This time the failure looks definitely spurious instead of just probably spurious.

gabro:
The CI failure is definitely spurious, I've cleaned the cache and rested the failing job. Let's see

gabro:
Ok, all ✅ now!

PR #1886, opened 2019-07-30
Modify processing of method invocations w/ default arguments

### LINKED ISSUE #1883: Incorrect local symbol occurrence for class instantiation with default parameters
```scala
object Metals {
  case class Msg(body: String, head: String = "Test", tail: String)

  val bodyText = "Hello"
  val msg = Msg(bodyText, tail = "tail") // bodyText shown as a local symbol
  println(bodyText) 
}
```

Looks like a similar issue to https://github.com/scalameta/scalameta/issues/1878

Default arguments desugar into blocks with synthetic local variables with the synthetic and artifact flags.

jameskoch:
New contributor; apologies in advance. :)

Fixes #1883.

I think I've understood where the current impl goes wrong w/r/t methods w/ default values. I've described this in a detailed comment in TextDocumentOps.scala.

Detecting the generated code pattern in Scala 2.11 required a heuristic. I took the liberty of repurposing the single cross-compiling source file (AttachmentOps.scala) as a single injection point for future cross-compiling sources (so renamed to VersionSpecificOps.scala).

Actual fix resides in TextDocumentOps. It is slightly kludgy, but minimally disruptive to the surrounding code. Curious what others think.


jameskoch:
Added the expect case. Notably we're still generating symbols in the doc for several of the compiler artifacts; not sure if anyone holds that as a blocker:

https://github.com/scalameta/scalameta/pull/1886/files#diff-f90bb281d4f293db21adab2fef7daef1R2758-R2783

jameskoch:
@olafurpg I finally got around to signing the CLA. Would love to get my two CRs reviewed if you can find the time, or point me to another maintainer.

PS - Where will you be working post-holiday?

Issue #1875, opened 2019-06-16
Consider memoizing DescriptorParser for performance


wiwa:
Hacked it up over [here](https://github.com/wiwa/rsc/commit/10f9bb3d6cb26bf6f5e906b5965d680057bf83ca), it showed a useful speedup in outlining a large target. Unsure of memory implications, though.

Issue #1068, opened 2017-08-15
Get rid of megaCache


olafurpg:
- https://github.com/lucidsoftware/neo-sbt-scalafmt/blob/e79e0fc76c8d6416b891f23597ce4c7528bed5e1/scalafmt-impl/src/shared/main/scala/com/lucidchart/scalafmt/impl/Scalafmtter.scala#L24
- #778 
- `Input.read; Files.write(Input); Input.read // stale!`

It would be really nice to get rid of this global mutable cache.

olafurpg:
I propose we leave this out of the v2.0 milestone to unblock a release.

xeno-by:
👍

olafurpg:
I have a proposal for how to solve this without breaking existing code, instead of always using the global megaCache, https://github.com/scalameta/scalameta/blob/1af15eb96c30c437d1b4aa57aca1740a36ed44a4/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala#L13

we can make it a parameter to `ScalametaTokenizer`. We could then make it easier for applications to construct parsers/tokenizers with custom megaCache. For example, in the case of scalafmt/scalafix, it would be little trouble for me change from `InputLike.parse[T]` to a slightly more verbose syntax that uses custom megaCache. Something in this direction would be enough to solve my problems.

netvl:
Copied from gitter (per @olafurpg request):

> I'm currently playing with scalameta and encountered quite a strange thing.
> 
> ```scala
> scala> res31.syntax
> res42: String = override val key: String = "teradata"
> 
> scala> res31.structure
> res43: String = Defn.Def(List(Mod.Override()), Term.Name("key"), Nil, Nil, Some(Type.Name("String")), Lit.String("teradata"))
> 
> scala> Defn.Def(List(Mod.Override()), Term.Name("key"), Nil, Nil, Some(Type.Name("String")), Lit.String("teradata"))
> res44: scala.meta.Defn.Def = override def key: String = "teradata"
> 
> scala> res31
> res45: scala.meta.Stat = override val key: String = "teradata"
> ```
> 
> The next to last expression is that I just copied res43 and pasted it to repl. How is it possible that a statement is printed as val (and it is actually loaded from a file where it is a val), but its structure says it's a def?

> this is really strange - apparently I cannot reproduce it anymore! it now shows correct structure if I execute exactly the same code in sbt's consoleProject repl.
> 
[...]

PR #1891, opened 2019-08-01
Memoize DescriptorParser (#1875)

wiwa:
Not sure how to bench, so just did `sbt benchAll`:

`master`:
```
[info] Result "scala.meta.internal.bench.MetacpScalaLibrary.run":
[info]   N = 15
[info]   mean =   2245.071 ±(99.9%) 80.041 ms/op

[info] Result "scala.meta.internal.bench.ScalacBaseline.run":
[info]   N = 67
[info]   mean =    789.296 ±(99.9%) 30.497 ms/op

[info] Result "scala.meta.internal.bench.ScalacRangepos.run":
[info]   N = 63
[info]   mean =    813.046 ±(99.9%) 25.182 ms/op
```

With this PR:
```
[info] Result "scala.meta.internal.bench.MetacpScalaLibrary.run":
[info]   N = 15
[info]   mean =   2222.981 ±(99.9%) 53.298 ms/op

[info] Result "scala.meta.internal.bench.ScalacBaseline.run":
[info]   N = 65
[info]   mean =    798.708 ±(99.9%) 36.026 ms/op

[info] Result "scala.meta.internal.bench.ScalacRangepos.run":
[info]   N = 67
[info]   mean =    770.030 ±(99.9%) 18.157 ms/op
```

Don't know if these are adequate nor how this change affects memory usage if that's okay.

It seems that everything is within margin of error. However, I've observed a significant performance improvement for Rsc while outlining larger targets.

### DIFF
--- a/semanticdb/semanticdb/src/main/scala/scala/meta/internal/semanticdb/Scala.scala
+++ b/semanticdb/semanticdb/src/main/scala/scala/meta/internal/semanticdb/Scala.scala
@@ -282,9 +282,15 @@ object Scala {
   }
 
   private[meta] object DescriptorParser {
-    def apply(symbol: String): (Descriptor, String) = {
+    private final val _cache = new java.util.HashMap[String, (Descriptor, String)]()
+
+    private def compute(symbol: String): (Descriptor, String) = {
       val parser = new DescriptorParser(symbol)
       parser.entryPoint()
+    }
+
+    def apply(symbol: String): (Descriptor, String) = {
+      _cache.computeIfAbsent(symbol, compute)
     }
   }
 }


semanticdb/semanticdb/src/main/scala/scala/meta/internal/semanticdb/Scala.scala
```diff
@@ -282,9 +282,15 @@ object Scala {
   }
 
   private[meta] object DescriptorParser {
-    def apply(symbol: String): (Descriptor, String) = {
+    private final val _cache = new java.util.HashMap[String, (Descriptor, String)]()
```
olafurpg: I'm on the fence about introducing global caches in library code like here, we have had a lot of problems with a similar cache that is used by the tokenizer https://github.com/scalameta/scalameta/issues/1068. I don't believe this cache will have the same correctness issues as the tokenizer cache but there is still a risk of growing memory usage.

I believe it's safer to introduce caches in applications like rsc/scalafmt/scalafix/metals where there are clear phases for cleaning up acquired resources.

Is this descriptor parser used by rsc?

wiwa: That's fair. This is used by rsc. I would then consider a way to easily override the included descriptor parser if the user wanted to have a cache, how does that sound?

olafurpg:
Closing due to inactivity. We can revisit this in the future if performance of the symbol parsers becomes a bottleneck again.

PR #1893, opened 2019-08-07
Add support for 2.12.9

gabro:


olafurpg:
I can look into it next week. We’ll also need a couple changes in Metals to support 2.12.9

gabro:
@olafurpg I wish it would be that simple :D Tests don't pass due to 

```
java.lang.RuntimeException: unsupported value class scala.tools.scalap.scalax.rules.scalasig.Externa
lSymbol java.nio.file.LinkOption.NOFOLLOW_LINKS
        at scala.sys.package$.error(package.scala:30)
        at scala.meta.internal.semanticdb.package$XtensionSemanticdbConstantCompanion.apply(package.
scala:128)
```

in a few places

gabro:
The root cause seems to be this change https://github.com/scala/scala/commit/0b28d2fb9e84e21ea3744a3f28a258661859de07

gabro:
Ok, more in details: scalap now correctly handles `ConstantType`s whose underlying symbol is `ExternalSymbol` or `Type`.

Scalacp already handles the latter case in 

https://github.com/scalameta/scalameta/blob/2035fd691312fed264acd4aa29b7d16ba1bffd40/semanticdb/metacp/src/main/scala/scala/meta/internal/scalacp/TypeOps.scala#L48-L55

but it doesn't handle the `ExternalSymbol` case (which happens, for example, in case the constant is a Java enum) here:

https://github.com/scalameta/scalameta/blob/1cbb19f4c1a1020189a461ac4cea981a322b73d1/semanticdb/semanticdb/src/main/scala/scala/meta/internal/semanticdb/package.scala#L96-L112

I'm not sure of what's the correct fix fo this.

PR #1897, opened 2019-08-20
Support methods with default args invoked via unstable identifi…

jameskoch:
I overlooked this case in my prior PR.

What I'm observing is that given input:
```
package k
class target {
  def foo(a: A, b: Int = 1, c: Int = 2): Int = ???
}
object consumer {
  def unstableQual = new target
  unstableQual.<<foo>>(1, <<c>> = 1)
}
```

scalac is not generating any Tree w/ position corresponding to the invocation of`<<foo>>`, rather there is only one corresponding to `<<.foo>>` (note leading period).

This Tree has its Position set by this logic: https://github.com/scala/scala/blob/7c354fc56df7c72ee829417264c2a2f97e65f274/src/compiler/scala/tools/nsc/typechecker/NamesDefaults.scala#L198-L200

I can't immediately spot why it would produce an off-by-one issue like this.

Some debug output corresponding to this test case:
```
Syntactic Positions:
indexName [8, 9] => k
indexName [16, 22] => target
indexName [31, 34] => foo
indexName [35, 36] => a
indexName [38, 41] => Int
indexName [43, 44] => b
indexName [46, 49] => Int
indexName [55, 56] => c
indexName [58, 61] => Int
indexName [68, 71] => Int
indexName [74, 77] => ???
indexName [87, 95] => consumer
indexName [104, 116] => unstableQual
indexName [123, 129] => target
indexName [132, 144] => unstableQual
indexName [145, 148] => foo
indexName [152, 153] => c

scalac output:
<132:158>Block(
  List(
    <132:144>ValDef(Modifiers(ARTIFACT), TermName("qual$1"), [132]TypeTree(), [132:144]Select([132]This(TypeName("consumer")), TermName("unstableQual"))), 
    [149:150]ValDef(Modifiers(ARTIFACT), TermName("x$1"), [149]TypeTree(), [149:150]Literal(Constant(1))),
    [156:157]ValDef(Modifiers(ARTIFACT), TermName("x$2"), [156]TypeTree(), [156:157]Literal(Constant(1))),
    [145]ValDef(Modifiers(ARTIFACT), TermName("x$3"), [145]TypeTree(), [145]Select([132]Ident(TermName("qual$1")), TermName("foo$default$2")))
  ),
  <132:158>Apply([144:148]Select([132]Ident(TermName("qual$1")), TermName("foo")), List([149]Ident(TermName("x$1")), [145]Ident(TermName("x$3")), [156]Ident(TermName("x$2"))))
)
```

### DIFF
--- a/semanticdb/scalac/library/src/main/scala-2.11/scala/meta/internal/semanticdb/scalac/VersionSpecificOps.scala
+++ b/semanticdb/scalac/library/src/main/scala-2.11/scala/meta/internal/semanticdb/scalac/VersionSpecificOps.scala
@@ -12,13 +12,17 @@ trait VersionSpecificOps { self: SemanticdbOps =>
     */
   object NamedApplyBlock {
     def unapply(block: g.Block): Option[Option[g.analyzer.NamedApplyInfo]] =
-      if (block.stats.forall(stat =>
-            stat.symbol != null &&
-              stat.symbol.isArtifact &&
-              (stat match {
-                case g.ValDef(_, name, _, _) if name.startsWith(g.termNames.NAMEDARG_PREFIX) => true
-                case _ => false
-              }))) {
+      if (block.stats.forall(
+            stat =>
+              stat.symbol != null &&
+                stat.symbol.isArtifact &&
+                (stat match {
+                  case g.ValDef(_, name, _, _)
+                      if name.startsWith(g.termNames.NAMEDARG_PREFIX) ||
+                        name.startsWith(g.termNames.QUAL_PREFIX) =>
+                    true
+                  case _ => false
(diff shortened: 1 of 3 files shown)


olafurpg:
I just pushed a v4.2.2 tag to kickstart a new release.

jameskoch:
A release of scalafix as well would be appreciated. Thanks. Sorry for the churn. If it's a pain I can work-around in other ways.

jameskoch:
> A release of scalafix as well would be appreciated. Thanks. Sorry for the churn. If it's a pain I can work-around in other ways.

I was talking nonsense; no need for a scalafix release.

PR #1890, opened 2019-07-31
Fixes mishandling of URI -> nio.file.Path conversions and splicing.

jameskoch:
Previously this caused an input file "filename with spaces.scala" to briefly generate
"filename%20%with%20spaces.scala.semanticdb" only to have it subsequently deleted by
RemoveOrphanSemanticdbFiles (which attempts to remove stale semanticdb outputs).

Fixes #1888

PR #1899, opened 2019-08-26
Guard against null in `URI.getPath`.

olafurpg:
Previously, calling `AbsolutePath.toURI` could throw a null pointer
exception for directories inside jars. Now, we guard against null
so calling `.toURI` succeeds.

This fixes a regression from #1890. Could you take a look @jameskoch?

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/io/NIOPathTest.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/io/NIOPathTest.scala
@@ -4,8 +4,10 @@ package io
 import java.io.File
 import java.nio.file.Path
 import java.nio.file.Paths
+import scala.meta.io._
 import scala.meta.internal.io._
 import org.scalatest.FunSuite
+import java.nio.file.Files
 
 class NIOPathTest extends FunSuite {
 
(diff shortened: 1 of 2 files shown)


tests/shared/src/test/scala/scala/meta/tests/io/NIOPathTest.scala
```diff
@@ -79,13 +81,25 @@ class NIOPathTest extends FunSuite {
...
     // assert(project.toUri.getPath.endsWith("project/"))
   }
+
+  test("jar.toUri") {
+    if (scala.meta.internal.platform.isJVM) {
+      val jar = Files.createTempDirectory("scalameta").resolve("foo.jar")
+      FileIO.withJarFileSystem(AbsolutePath(jar), true, true) { root =>
+        val hello = root.resolve("hello")
+        Files.createDirectories(hello.toNIO)
+        assert(hello.toURI.toString.endsWith("hello"))
```
olafurpg: I'm not sure if the expected behavior here is `endsWith("hello/")` or `endsWith("hello")`. I don't mind either way however, as long as `toURI` doesn't crash.

olafurpg:
Merging to unblock a new release.

PR #1901, opened 2019-09-14
Add support for 2.13.1

gabro:
I've tested this locally against 2.13.1-bin-a3791d4 (the current release candidate) and tests work.

I'll leave this here for when 2.13.1 is out (supposedly, next week)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -508,19 +508,13 @@ lazy val isScala213 = Def.setting {
   scalaVersion.value.startsWith("2.13")
 }
 
-lazy val paradiseVersion = Def.setting {
-  scalaVersion.value match {
-    case "2.12.10" => "2.12.9"
-    case v => v
-  }
-}
 lazy val sharedSettings = Def.settings(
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -508,19 +508,13 @@ lazy val isScala213 = Def.setting {
...
-  }
-}
 lazy val sharedSettings = Def.settings(
   scalaVersion := LanguageVersion,
   crossScalaVersions := LanguageVersions,
   organization := "org.scalameta",
   libraryDependencies ++= {
     if (isScala213.value) Nil
-    else List(compilerPlugin("org.scalamacros" % s"paradise_${paradiseVersion.value}" % "2.1.0"))
+    else List(compilerPlugin("org.scalamacros" % s"paradise" % "2.1.1" cross CrossVersion.full))
```
gabro: This is not strictly related, but since 2.1.1 is now published for 2.12.10, we don't need to fall back to 2.12.9 anymore

PR #1902, opened 2019-09-18
Save occurence of anonymous class symbol

tgodzik:
Not entirely sure it's the way to go, but the occurrence is saved for sure, the only problem is that it looks to be saved as reference and not definition.

@olafurpg Is this the way to go? Will try to look at how to make it def tomorrow.

Edit: Ok, added it to `binders` set - that's seems to define what is a definition

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
@@ -203,6 +203,11 @@ trait TextDocumentOps { self: SemanticdbOps =>
 
             saveSymbol(gsym)
             if (gsym.isClass && !gsym.isTrait) {
+              if (gsym.isAnonymousClass) {
+                val pos = gsym.pos.focus.toMeta
+                binders += pos
+                occurrences(pos) = gsym.toSemantic
+              }
               val gprim = gsym.primaryConstructor
               saveSymbol(gprim)
               gprim.info.paramss.flatten.foreach(saveSymbol)
--- a/tests/jvm/src/test/resources/example/Advanced.scala
+++ b/tests/jvm/src/test/resources/example/Advanced.scala
@@ -10,8 +10,8 @@ class C/*<=advanced.C#*/[T/*<=advanced.C#[T]*/] {
 
 class Structural/*<=advanced.Structural#*/ {
   def s1/*<=advanced.Structural#s1().*/: { val x/*<=local0*/: Int/*=>scala.Int#*/ } = ???/*=>scala.Predef.`???`().*/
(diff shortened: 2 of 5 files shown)


tests/jvm/src/test/resources/example/Advanced.scala
```diff
@@ -10,8 +10,8 @@ class C/*<=advanced.C#*/[T/*<=advanced.C#[T]*/] {
 
 class Structural/*<=advanced.Structural#*/ {
   def s1/*<=advanced.Structural#s1().*/: { val x/*<=local0*/: Int/*=>scala.Int#*/ } = ???/*=>scala.Predef.`???`().*/
-  def s2/*<=advanced.Structural#s2().*/ = new { val x/*<=local1*/: Int/*=>scala.Int#*/ = ???/*=>scala.Predef.`???`().*/ }
-  def s3/*<=advanced.Structural#s3().*/ = new { def m/*<=local3*/(x/*<=local4*/: Int/*=>scala.Int#*/): Int/*=>scala.Int#*/ = ???/*=>scala.Predef.`???`().*/ }
+  def s2/*<=advanced.Structural#s2().*/ = new { val x: Int = ??? }/*<=local2*/
```
tgodzik: Seems it ignores the stuff inside now... need to fix that somehow.

tests/jvm/src/test/resources/example/Traits.scala
```diff
@@ -6,7 +6,7 @@ trait T/*<=traits.T#*/ {
 
 sealed trait U/*<=traits.U#*/
 object U/*<=traits.U.*/ {
-  def u/*<=traits.U.u().*/: U/*=>traits.U#*/ = new U/*=>traits.U#*/ /*=>java.lang.Object#`<init>`().*/{}
+  def u/*<=traits.U.u().*/: U/*=>traits.U#*/ = new U/*<=local0*/ /*=>java.lang.Object#`<init>`().*/{}
```
tgodzik: Now it seems to override the occurrence of `U`

tgodzik: Fixed that, but not 100% sure about the result :/

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -203,6 +203,14 @@ trait TextDocumentOps { self: SemanticdbOps =>
 
             saveSymbol(gsym)
             if (gsym.isClass && !gsym.isTrait) {
+              if (gsym.isAnonymousClass) {
+                val pos = gsym.pos.toMeta
```
olafurpg: There is a “focus” method on compiler positions to make the start and end point to the same offset. That would avoid the need for startPosition below.

tgodzik: Fixed

PR #1906, opened 2019-09-30
Run CLA check in GitHub Actions

olafurpg:


olafurpg:
Seems to be working OK https://github.com/scalameta/scalameta/pull/1906/checks?check_run_id=241951159

PR #1907, opened 2019-10-30
fix: remove NotImplementedError in SymbolInformationOps

ingarabr:
Causes scalameta/scalafix to fail when synthetic is turned on.


olafurpg:
Ignoring unrelated CI failure on Windows.

Issue #1909, opened 2019-11-12
Type in classOf[T] is not added to occurrences in semanticDB


tgodzik:
```scala
object Main{
  val a = classOf[Int]
}
```

In the example `Int` symbol will not be added to occurrences. @sswistun-vl is taking a look at it.

tgodzik:
Should we also add the occurence for `classOf` ? @olafurpg what do you think?

olafurpg:
I think we should add occurrences to `classOf`

tanishiking:
I looked into this a bit, and I found that we have the following trees for `classOf[...]`.

```scala
classOf[Int]

Literal(
  value = Constant(
    value = TypeRef(pre = ThisType(sym = package scala), sym = class Int, args = List())
  )
)
```

```scala
classOf[List[Int]]

Literal(
  value = Constant(
    value = TypeRef(
      pre = SingleType(
        pre = SingleType(pre = ThisType(sym = package <root>), sym = package scala),
        sym = package object scala
      ),
      sym = type List,
      args = List(TypeRef(pre = ThisType(sym = package scala), sym = class Int, args = List()))
    )
  )
)
```

This is because

- type-checker generate and replace `classOf[...]` with `Literal(Constatnt(...))` 
  - https://github.com/scala/scala/blob/5cc8ebbd712ca5a0e051a1075af55bc64ad7a2da/src/compiler/scala/tools/nsc/typechecker/Typers.scala#L4312-L4314
  - https://github.com/scala/scala/blob/5cc8ebbd712ca5a0e051a1075af55bc64ad7a2da/src/reflect/scala/reflect/internal/TreeGen.scala#L281-L294
- While it is possible to extract symbols from types, unlike type tree (like `AppliedTypeTree`) we can't get position information from `Type`s.

PR #1910, opened 2019-11-13
Add occurrences for classOf

sswistun-vl:
Before, classOf didn't generate occurrences, both for `classOf` method.
Now, it generates them as `scala.Predef.classOf()`. Tests were adjusted to reflect the changes.

Related to https://github.com/scalameta/scalameta/issues/1909

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
@@ -644,10 +644,10 @@ trait TextDocumentOps { self: SemanticdbOps =>
                 traverse(gtree.check())
 
               case gtree: g.Literal
-                  if isClassOf(gtree.pos.toMeta) &&
-                    gtree.tpe != null &&
+                  if gtree.tpe != null &&
                     gtree.tpe.typeSymbol == g.definitions.ClassClass &&
-                    gtree.pos.isRange =>
+                    gtree.pos.isRange &&
+                    isClassOf(gtree.pos.toMeta) =>
                 val coLen = "classOf".length
                 val mpos = gtree.pos.toMeta
                 val mposFix = new m.Position.Range(mpos.input, mpos.start, mpos.start + coLen)
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
@@ -604,6 +604,14 @@ trait TextDocumentOps { self: SemanticdbOps =>
               case _ =>
             }
 
+            def isClassOf(pos: m.Position): Boolean = {
+              val chars = pos.input.chars
+              val classOfChars = "classOf".toCharArray
+
+              chars.length >= (classOfChars.length + pos.start) &&
+              (0 until classOfChars.length).forall(i => chars(i + pos.start) == classOfChars(i))
+            }
+
             gtree match {
               case OriginalTreeOf(original) =>
                 traverse(original)
@@ -634,6 +642,17 @@ trait TextDocumentOps { self: SemanticdbOps =>
                 traverse(gtree.original)
               case gtree: g.TypeTreeWithDeferredRefCheck =>
                 traverse(gtree.check())
+
+              case gtree: g.Literal
+                  if isClassOf(gtree.pos.toMeta) &&
+                    gtree.tpe != null &&
+                    gtree.tpe.typeSymbol == g.definitions.ClassClass &&
+                    gtree.pos.isRange =>
+                val coLen = "classOf".length
+                val mpos = gtree.pos.toMeta
+                val mposFix = new m.Position.Range(mpos.input, mpos.start, mpos.start + coLen)
+
+                occurrences(mposFix) = "scala/Predef.classOf()."
 
               case gtree: g.MemberDef =>
                 gtree.symbol.annotations.foreach(ann => traverse(ann.original))
--- a/tests/jvm/src/test/resources/example/InstrumentTyper.scala
+++ b/tests/jvm/src/test/resources/example/InstrumentTyper.scala
@@ -17,11 +17,12 @@ class InstrumentTyper/*<=example.InstrumentTyper#*/ { self/*<=local0*/: AnyRef =
     Literal/*=>types.Test.Literal.*/.bool/*=>types.Test.Literal.bool.*/,
     Literal/*=>types.Test.Literal.*/.unit/*=>types.Test.Literal.unit.*/,
     Literal/*=>types.Test.Literal.*/.javaEnum/*=>types.Test.Literal.javaEnum.*/,
-    Literal/*=>types.Test.Literal.*/.clazzOf/*=>types.Test.Literal.clazzOf.*/,
+    Literal/*=>types.Test.Literal.*/.clazzOfInt/*=>types.Test.Literal.clazzOfInt.*/,
+    Literal/*=>types.Test.Literal.*/.clazzOfOption/*=>types.Test.Literal.clazzOfOption.*/,
     List/*=>scala.collection.immutable.Nil.*/()
   )
   def existential/*<=example.InstrumentTyper#existential().*/: U/*=>local1*/[Int/*=>scala.Int#*/] forSome { type U/*<=local1*/[T/*<=local2*/ <: Int] } = ???/*=>scala.Predef.`???`().*/
   type AnnotatedType/*<=example.InstrumentTyper#AnnotatedType#*/ = Int/*=>scala.Int#*/ @param
   def singletonType/*<=example.InstrumentTyper#singletonType().*/(x/*<=example.InstrumentTyper#singletonType().(x)*/: Predef.type) = ???/*=>scala.Predef.`???`().*/
-  final val clazzOf/*<=example.InstrumentTyper#clazzOf.*/ = classOf[Option[Int]]
+  final val clazzOf/*<=example.InstrumentTyper#clazzOf.*/ = classOf/*=>scala.Predef.classOf().*/[Option[Int]]
 }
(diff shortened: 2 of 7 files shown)


tgodzik:
@olafurpg it seems the classOf[A] is a Literal inside the compiler and there is no way to get the possitions of symbols aside from doing it manually. Let us know if there is any better solution for this!

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -636,7 +636,8 @@ trait TextDocumentOps { self: SemanticdbOps =>
                 traverse(gtree.check())
 
               case gtree: g.Literal
-                  if gtree.tpe != null && gtree.tpe.typeSymbol.name.decode.contentEquals("Class") =>
+                  if gtree.tpe != null && gtree.tpe.typeSymbol.name.decode
```
olafurpg: The idiomatic way to compare symbols is via `==`, you can get a reference to the `java.lang.Class` symbol in `g.definitions.ClassClass`

Screenshot 2019-11-15 at 09 03 40

olafurpg: Can we add an additional guard that the source code actually contains `classOf`?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -635,6 +635,21 @@ trait TextDocumentOps { self: SemanticdbOps =>
...
+                val mpos = gtree.pos.toMeta
+                val mposFix = new m.Position.Range(mpos.input, mpos.start, mpos.start + coLen)
+
+                occurrences(mposFix) = "scala/Predef.classOf()."
+                binders += mposFix
+
+                val cpos =
+                  new m.Position.Range(mpos.input, mpos.start + coLen + 1, mpos.end - 1)
+                occurrences(cpos) = gtree.value.typeValue.typeSymbol.toSemantic
+                binders += cpos
```
olafurpg: Is this line needed? `classOf` is not a definition

olafurpg: same for `mposFix`

tests/jvm/src/test/resources/example/InstrumentTyper.scala
```diff
@@ -23,5 +23,5 @@ class InstrumentTyper/*<=example.InstrumentTyper#*/ { self/*<=local0*/: AnyRef =
   def existential/*<=example.InstrumentTyper#existential().*/: U/*=>local1*/[Int/*=>scala.Int#*/] forSome { type U/*<=local1*/[T/*<=local2*/ <: Int] } = ???/*=>scala.Predef.`???`().*/
   type AnnotatedType/*<=example.InstrumentTyper#AnnotatedType#*/ = Int/*=>scala.Int#*/ @param
   def singletonType/*<=example.InstrumentTyper#singletonType().*/(x/*<=example.InstrumentTyper#singletonType().(x)*/: Predef.type) = ???/*=>scala.Predef.`???`().*/
-  final val clazzOf/*<=example.InstrumentTyper#clazzOf.*/ = classOf[Option[Int]]
+  final val clazzOf/*<=example.InstrumentTyper#clazzOf.*/ = classOf/*<=scala.Predef.classOf().*/[Option[Int]/*<=scala.Option#*/]
```
olafurpg: Is there a way to also capture the reference to `Int` inside the type argument?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
sswistun-vl: Fixed.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
sswistun-vl: Fixed.

tgodzik:
We can either merge this (will work only with the first class not ones with type params, so not ideal) and work on fixing the compiler separately OR drop it entirely for know. I am on the edge for making it partially here, but it would help with most use cases in Metals.

@olafurpg what do you think?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -604,6 +604,13 @@ trait TextDocumentOps { self: SemanticdbOps =>
               case _ =>
             }
 
+            def isClassOf(pos: m.Position): Boolean = {
+              val (chars, classOfChars) = (pos.input.chars, "classOf".toCharArray)
```
olafurpg: nit: a more idiomatic way to declare two variables
```suggestion
              val chars = pos.input.chars
              val classOfChars = "classOf".toCharArray
```

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -604,6 +604,13 @@ trait TextDocumentOps { self: SemanticdbOps =>
...
+
+              chars.length >= (classOfChars.length + pos.start) && (0 until classOfChars.length)
```
olafurpg: Nit: please insert manual line breaks around infix operators to make it clearer to read the conditions 
```suggestion
              chars.length >= (classOfChars.length + pos.start) &&
                (0 until classOfChars.length)
```

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -635,6 +642,18 @@ trait TextDocumentOps { self: SemanticdbOps =>
               case gtree: g.TypeTreeWithDeferredRefCheck =>
                 traverse(gtree.check())
 
+              case gtree: g.Literal
+                  if isClassOf(gtree.pos.toMeta) && gtree.tpe != null && gtree.tpe.typeSymbol == g.definitions.ClassClass && gtree.pos.isRange =>
```
olafurpg: nit: multiple lines make it easier to read the multiple conditions
```suggestion
                  if isClassOf(gtree.pos.toMeta) &&
                    gtree.tpe != null &&
                    gtree.tpe.typeSymbol == g.definitions.ClassClass &&
                    gtree.pos.isRange =>
```

tests/jvm/src/test/resources/example/Types.scala
```diff
@@ -111,6 +111,6 @@ object Test/*<=types.Test.*/ {
     final val bool/*<=types.Test.Literal.bool.*/ = true
     final val unit/*<=types.Test.Literal.unit.*/ = ()
     final val javaEnum/*<=types.Test.Literal.javaEnum.*/ = java.nio.file.LinkOption/*=>java.nio.file.LinkOption#*/.NOFOLLOW_LINKS/*=>java.nio.file.LinkOption#NOFOLLOW_LINKS.*/
-    final val clazzOf/*<=types.Test.Literal.clazzOf.*/ = classOf[Option[Int]]
+    final val clazzOf/*<=types.Test.Literal.clazzOf.*/ = classOf/*=>scala.Predef.classOf().*/[Option[Int]/*=>scala.Option#*/]
```
olafurpg: I would expect the occurrence position to enclose the `Option` identifier 🤔 
```suggestion
    final val clazzOf/*<=types.Test.Literal.clazzOf.*/ = classOf/*=>scala.Predef.classOf().*/[Option/*=>scala.Option#*/[Int]]
```

tests/jvm/src/test/resources/metac.expect
```diff
@@ -1801,6 +1801,8 @@ Occurrences:
 [24:20..24:21): x <= example/InstrumentTyper#singletonType().(x)
 [24:38..24:41): ??? => scala/Predef.`???`().
 [25:12..25:19): clazzOf <= example/InstrumentTyper#clazzOf.
+[25:22..25:29): classOf => scala/Predef.classOf().
+[25:30..25:41): Option[Int] => scala/Option#
```
olafurpg: This means that renaming `Option` to `Foobar` would replace `Option[Int]` with `Foobar`

tests/jvm/src/test/resources/example/Types.scala
olafurpg: Another test case to handle: qualifiers
```scala
classOf[scala.Option[Int]]
```

olafurpg: Another test case: renamed imports
```scala
import scala.concurrent.{Future => Fut}

classOf[Fut[Int]]
```

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
sswistun-vl: Fixed.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
sswistun-vl: Fixed.

tests/jvm/src/test/resources/metac.expect
sswistun-vl: Fixed. You might want to check out this solution as it's based on character occurrence in string - I'm not sure if it's a proper way to do it.

tests/jvm/src/test/resources/example/Types.scala
sswistun-vl: I've added tests, but not sure if they are in the right place.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
sswistun-vl: Fixed.

tgodzik:
I think this is a bit of a dead end - we should revert to just including the classOf occurence and aim to fix the exact positions in the compiler. @sswistun-vl

sswistun-vl:
I've removed occurrences for classOf arguments as they can't be resolved properly right now.

tgodzik:
@olafurpg we reverted the part reposible for the type param and I think this is mergable now. It will at least fix the classOf occurences and provide an entry point when we fix the possitions for the type params inside classOf. What do you think?

tgodzik:
> LGTM

Thanks for taking a look at it!

PR #1913, opened 2019-11-25
Fix span methods on Tokens

travisbrown:
If the predicate passed to `span` is true for all of the elements in a collection, the first element of the result should be the original collection and the second should be empty (and the reverse for `spanRight`).

The current implementation is broken because it doesn't handle the case where `indexWhere` returns -1. I've included some failing tests that show the issue, and a fix. I took a quick look around for similar bugs but didn't see anything.


olafurpg:
I don't think the `Tokens` methods have seen a lot of usage, but I personally find them quite helpful when writing Scalafix rewrites 😊

olafurpg:
I have kickstarted a v4.3.0 release with this fix, may the CI/Sonatype gods be nice today

PR #1918, opened 2019-12-02
sbt 1.x migration

ghik:


build.sbt
```diff
@@ -32,7 +31,7 @@ name := {
 }
 nonPublishableSettings
 crossScalaVersions := Nil
-unidocSettings
+enablePlugins(ScalaUnidocPlugin)
```
olafurpg: we can remove unidoc if it's causing problems, the plugin is not really used

ghik:
It seems that one test is failing, `scala.meta.tests.semanticdb.ExpectSuite.metac.expect`, no idea why. Can you please take a look, @olafurpg ?

olafurpg:
I'll take a look!

PR #1957, opened 2020-01-24
Replace ScalaTest with MUnit

olafurpg:
Docs https://scalameta.org/munit/


olafurpg:
cc/ @SethTisue we may need to add MUnit to the community build after this PR https://github.com/scalameta/munit 

I'm hoping to port most of my projects to MUnit as I find the time to work on the migration. It should be easy to cross-build MUnit, it supports most Scala compilers versions (2.11 until Dotty) and platforms (JS,Native).

SethTisue:
> we may need to add MUnit to the community build

done: https://github.com/scala/community-build/pull/1067

I didn't actually advance the scalameta SHA to a point where it needs munit; I'll likely get to it in my next monthly-ish sweep

PR #1963, opened 2020-01-31
Scala.js source maps are now pointing to github

ghik:
By default, source maps generated by Scala.js compiler point to local files.
This PR changes this so that for non-snapshot versions, source maps point to github raw content.


ghik:
Builds are failing because `org.scalameta:munit_sjs0.6_2.13:0.4.1` cannot be found, which is unrelated to this PR.

PR #1973, opened 2020-02-05
Move documentation to the Scalameta repo

olafurpg:
Fixes https://github.com/scalameta/tutorial/issues/66

Previously, we maintained the scalameta docs in a separate repo. This
setup relied on a fairly complex setup downloading markdown files from
the scalameta repo. The documentation in the Scalameta repo was also
lying in several different directories

* `notes/`
* `semanticdb/semanticdb3/`

Now, all the docs for scalameta.org lie in the scalameta/scalameta repo
under the `docs/` repository, using the same convention as we have in
other repositories.


olafurpg:
Let's merge this and see how it runs in master

.github/workflows/release.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: olafurpg/setup-scala@v2
+      - uses: olafurpg/setup-gpg@v2
+      - run: git fetch --unshallow
+      - name: Publish
+        run: |
+          sbt ci-release
+          sbt docs/docusaurusPublishGhpages
+        env:
```
ckipp01: Does scalameta-parsers also get published here? If so, should we also include that here?

olafurpg: To npm? That module has so far been published manually I think. cc/ @gabro it would be nice to automate that publish step but we can do it separately

ckipp01: Yup, to NPM. And sounds good! Was just curious.

gabro: Yeah, I've always been lazy and I've never got around automating that step, mostly because I'm probably the only one using that package (in AST Explorer) :P 

I agree it would be good to automate it

PR #1980, opened 2020-02-07
Parse `try` whose body is just a tuple

Duhemm:
The following was incorrectly rejected by the parser:

    try (1, 2)

But it should be accepted and parsed as a `try` whose body is the tuple
`(1, 2)`.


olafurpg:
I'm curious, did you hit on this issue somewhere? I haven't seen it before 🤔

Issue #1889, opened 2019-07-31
testkit's own tests don't compile on 2.13


SethTisue:
what's Latin for "who will test the testkit?"

I'm wondering if this is somehow known/intended, or if it fell through the cracks:

```
% sbt
> ++2.13.0
> testkit/test
...
[error] /Users/tisue/scalameta/scalameta/testkit/src/test/scala/scala/meta/testkit/ScalametaParserProperties.scala:10: not found: type Observation
[error]   type ParserBug = Observation[BugKind]
[error]                    ^
...
```

there is a `scalameta/testkit/src/main/scala-2.12` directory, but no 2.13 counterpart

this came up during work on scala/community-builds#946 (but no worries, I can work around it easily)

PR #1997, opened 2020-03-04
Fix testkit for 2.13 after MUnit migration

poslegm:
Fixes https://github.com/scalameta/scalameta/issues/1889

It seems like after MUnit migration `ScalametaParserProperties.scala` doesn't compiles with any version of Scala. And 0 tests run after `testkit/test`. Fixed it by `testFrameworks` setting.


kpbochenek:
@poslegm Thanks for taking a look at it!

One of actions: CI / Mima still fails with `unresolved dependency: org.scalameta#testkit_2.12;4.3.2: not found` 

Could you take a look if it is something easy to fix?

I also found those `testkit` problems a little annoying it could be good to clean that up and bring all tests green again :+1:

tgodzik:
> @poslegm Thanks for taking a look at it!
> 
> One of actions: CI / Mima still fails with `unresolved dependency: org.scalameta#testkit_2.12;4.3.2: not found`
> 
> Could you take a look if it is something easy to fix?
> 
> I also found those `testkit` problems a little annoying it could be good to clean that up and bring all tests green again

I think that mima error is unrelated, it fails on on other PRs too.

poslegm:
@kpbochenek mima fails on all PRs because scalameta 4.3.2 isn't published to Maven Central

PR #2000, opened 2020-03-12
Fix documentation for arguments for metap

kpbochenek:
This is a code parsing params, it is expected to provide them with single '-' but in website documentation there is '--' which throws unknown parameters error

```scala
        case "-compact" +: rest if allowOptions =>
          loop(settings.copy(format = Format.Compact), allowOptions = true, rest)
        case ("-detailed" | "-pretty") +: rest if allowOptions =>
          loop(settings.copy(format = Format.Detailed), allowOptions = true, rest)
        case "-proto" +: rest if allowOptions =>
          loop(settings.copy(format = Format.Proto), allowOptions = true, rest)
```

Also added to .gitignore 'metals.sbt' (it should not be commited similary to as it is in metals, right?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ lib_managed/
 src_managed/
 project/boot/
 project/plugins/project/
+project/metals.sbt
 .scala_dependencies
 .worksheet
 *.actual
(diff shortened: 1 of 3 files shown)


kpbochenek:
I have no idea why CLA check fails, it checks there is true in signed here:
https://www.lightbend.com/contribute/cla/scala/check/kpbochenek
which is, but it says false << what can you do >>

PR #2003, opened 2020-03-16
Upgrade to 2.12.11

olafurpg:


tests/jvm/src/test/resources/metacp.expect
```diff
@@ -1318,15 +1318,15 @@ com/javacp/Test#InnerShadowTypeParam# => private[javacp] class InnerShadowTypePa
 com/javacp/Test#InnerShadowTypeParam#[A] => typeparam A
 com/javacp/Test#InnerShadowTypeParam#`<init>`(). => private[javacp] ctor <init>()
   javacp => com/javacp/
-com/javacp/Test#InnerShadowTypeParam#move(). => method move(a: A, b: B): Unit
-  a => com/javacp/Test#InnerShadowTypeParam#move().(a)
+com/javacp/Test#InnerShadowTypeParam#move(). => method move(param0: A, param1: B): Unit
```
olafurpg: @lrytz Congrats on the release! We fall back to `paramN` names when `scala.tools.asm.tree.MethodNode.node().parameters() == null`, see https://github.com/scalameta/scalameta/blob/2ef927539dde5d3e53a403e260b6bd85b3898f92/semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala#L171-L174

Any idea what we can do to get the 2.12.10 behavior back? It looks related to the ASM upgrade

olafurpg: NOTE: @lrytz I'm sorry I forgot to include context, there seems to be a regression when upgrading to 2.12.11

lrytz: Hmm, if I take a trivial example, `class C { def f(meta: Int) = 0 }` the classfiles are identical when compiling with 2.12.10 or 2.12.11, and javap shows me

```
  public int f(int);
  ...
    MethodParameters:
      Name                           Flags
      meta                           final
```

Can you try to minimize the change?

olafurpg: I investigated closer and it looks like there has been a change in how `SKIP_DEBUG` is interpreted when building a `scala.tools.asm.tree.ClassNode` for javac emitted signatures with the `-parameters` compiler option.

In 2.12.10 we're using `SKIP_DEBUG` and MethodParameters are read from Java signatures. In 2.12.11, the Java parameter names no longer appear unless we remove the usage of `SKIP_DEBUG`. Any idea what might cause this change?

lrytz: Seems to be this: https://gitlab.ow2.org/asm/asm/-/merge_requests/278 / https://gitlab.ow2.org/asm/asm/issues/317885

olafurpg: Indeed. Thank you for the help, we’ll stop using SKIP_DEBUG :)

PR #2005, opened 2020-03-17
Fix parsing package with leftbrace in newline

kpbochenek:
Fixes issue `sbt/A.scala#L2`
Fixes issue `breeze/package.scala#L43`
(they are both the same issues)
https://github.com/scalameta/scalameta/issues/312#issuecomment-267081928

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -328,11 +328,13 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
     accept[RightBrace]
     ret
   }
-  @inline final def inBracesOrError[T](body: => T, alt: T): T =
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -328,11 +328,13 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
     accept[RightBrace]
     ret
   }
-  @inline final def inBracesOrError[T](body: => T, alt: T): T =
+  @inline final def inBracesOrError[T](body: => T, alt: T): T = {
+    if (token.is[LF]) { accept[LF] }
```
olafurpg: What happens if there are multiple `LF`?

kpbochenek: it will generate `LFLF` which is invalid in this scenario

PR #2008, opened 2020-03-23
Fix *-parameter check for method definition

### LINKED ISSUE #2006: vararg-argument parsed but error
Also a quick overview about situation with *-parameters
It is related to issue reported in #312 (lila/projection.scala#L30) and  #1063
They will both need more work but this one should be fixed first.

### Situation in scala 2.12 / 2.13
`*-parameter` can only be used as a last parameter to function. Examples below explain what is possible and what is not:

Code | Scala | Scalameta
------------ | ------------- | -------------
def obj(fields: Int*): Boolean = true | okay | okay
def obj2(a: String, fields: Int*): Boolean = true = true | okay | okay
def obj3(fields: Int*, f2: Int*): Boolean = true | error: *-parameter must come last | __okay__
def obj4(fields: Int*, a: String): Boolean = true | error: *-parameter must come last  | __okay__
def obj5(fields: Int*)(a: String): Boolean = true | okay | okay
def obj6(fields: Int*)(f2: Int*): Boolean = true | okay | okay
def obj7(fields: Int*)(implicit f2: Int*): Boolean = true | okay | okay
def obj8(fields: Int*): Boolean* = true | error | error
def obj9(fields: => Int*): Boolean = true | error | error

There is one mismatch where compiler throws error but ScalametaParser accepts it.
By-name vararg-argument is supported in dotty but as talked to @tgodzik it will probably use another parser which means `obj9` as error is okay in both cases.

How vararg parameter is represented by scalameta structure for reference:
```
Term.Param(Nil,
                     Term.Name("fields"),
                     Some(Type.Repeated(Type.Name("Int"))),
                     None)
```

Fixing this is a first step towards fixing issue #312 (lila/projection.scala#L30)

I would like to take a look at it but one thing is, we agree this is an issue and it should be fixed?
If so there is only one simple test in `DeclSuite`, I could all mentioned cases there and try to fix it.
Also it looks like a structure is okay, correct?
There is a `Defn.Def` class that holds `paramss`, it would probably be enough to add a check that only last item on list can be `Type.Repeated` ?
I will investigate it more but just want to know if this is good direction.

kpbochenek:
Fixes #2006

One check is added in `trees` hierarchy to ensure structure is consistent and if anyone uses trees somewhere else this bad state will throw exception.
Let's say "duplicated check" is done in ScalametaParser to provide proper feedback for the user(same as compiler would generate).
(that's why 2 similar checks, I don't know have vision how to have only one check)

Also I added 1 test that should throw error for scala 2 but should be valid for scala 3, do we have a way to determine in test specific scala version?

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2997,6 +2997,17 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
       )
     val tparams = typeParamClauseOpt(ownerIsType = false, ctxBoundsAllowed = true)
     val paramss = paramClauses(ownerIsType = false).require[List[List[Term.Param]]]
+
+    def onlyLastParameterCanBeRepeated(params: List[Term.Param]): Unit = {
+      params.reverse.tail.foreach(p =>
+        if (p.decltpe.exists(_.isInstanceOf[Type.Repeated])) {
+          syntaxError("*-parameter must come last", p)
+        }
+      )
+    }
+
+    paramss.foreach(onlyLastParameterCanBeRepeated)
+
     newLineOptWhenFollowedBy[LeftBrace]
     val restype = fromWithinReturnType(typedOpt())
     if (token.is[StatSep] || token.is[RightBrace]) {
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -3002,7 +3002,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
       params.iterator
         .take(params.length - 1)
         .filter(p => !p.is[Term.Param.Quasi] && p.decltpe.exists(_.is[Type.Repeated]))
-        .foreach(syntaxError("*-parameter must come last", _))
+        .foreach(p => syntaxError("*-parameter must come last", p))
     }
 
     paramss.foreach(onlyLastParameterCanBeRepeated)
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2999,13 +2999,10 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
     val paramss = paramClauses(ownerIsType = false).require[List[List[Term.Param]]]
 
     def onlyLastParameterCanBeRepeated(params: List[Term.Param]): Unit = {
-      params
+      params.iterator
         .take(params.length - 1)
-        .foreach(p =>
-          if (p.decltpe.exists(_.isInstanceOf[Type.Repeated])) {
-            syntaxError("*-parameter must come last", p)
-          }
-        )
+        .filter(p => !p.is[Term.Param.Quasi] && p.decltpe.exists(_.is[Type.Repeated]))
+        .foreach(syntaxError("*-parameter must come last", _))
     }
 
     paramss.foreach(onlyLastParameterCanBeRepeated)
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2999,11 +2999,13 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
     val paramss = paramClauses(ownerIsType = false).require[List[List[Term.Param]]]
 
     def onlyLastParameterCanBeRepeated(params: List[Term.Param]): Unit = {
-      params.reverse.tail.foreach(p =>
-        if (p.decltpe.exists(_.isInstanceOf[Type.Repeated])) {
-          syntaxError("*-parameter must come last", p)
-        }
-      )
+      params
+        .take(params.length - 1)
+        .foreach(p =>
+          if (p.decltpe.exists(_.isInstanceOf[Type.Repeated])) {
+            syntaxError("*-parameter must come last", p)
+          }
+        )
     }
 
     paramss.foreach(onlyLastParameterCanBeRepeated)
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/package.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/package.scala
@@ -217,6 +217,6 @@ package object trees {
   }
 
   def onlyLastParamCanBeRepeated(params: List[Term.Param]): Boolean = {
-    params.reverse.tail.forall(!_.decltpe.exists(_.isInstanceOf[Type.Repeated]))
+    params.take(params.length - 1).forall(!_.decltpe.exists(_.isInstanceOf[Type.Repeated]))
   }
 }
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/package.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/package.scala
@@ -215,4 +215,8 @@ package object trees {
     if (rank == 0) clazz
     else arrayClass(ScalaRunTime.arrayClass(clazz), rank - 1)
   }
+
+  def onlyLastParamCanBeRepeated(params: List[Term.Param]): Boolean = {
+    params.reverse.tail.forall(!_.decltpe.exists(_.isInstanceOf[Type.Repeated]))
+  }
 }
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/package.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/package.scala
@@ -217,7 +217,7 @@ package object trees {
   }
 
   def onlyLastParamCanBeRepeated(params: List[Term.Param]): Boolean = {
-    params
+    params.iterator
       .take(params.length - 1)
       .forall(p => p.is[Term.Param.Quasi] || !p.decltpe.exists(_.is[Type.Repeated]))
   }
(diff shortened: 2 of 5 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2997,6 +2997,17 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
       )
     val tparams = typeParamClauseOpt(ownerIsType = false, ctxBoundsAllowed = true)
     val paramss = paramClauses(ownerIsType = false).require[List[List[Term.Param]]]
+
+    def onlyLastParameterCanBeRepeated(params: List[Term.Param]): Unit = {
+      params.reverse.tail.foreach(p =>
```
olafurpg: `params.init` is equivalent to `reverse.tail`. However, either approach allocates a lot of memory. We should avoid re-creating the list to do a check, we can do it like this
```scala
def checkRepeatedParams(params: List[Term.Param]): Unit =  params match {
  case Nil | _ :: Nil => // OK
  case param :: tail =>
    checkNotRepeated(param)
    checkRepeatedParams(params)
}
def checkRepeatedParamss(sparamss: List[List[Term.Param]]): Unit = paramss match {
  case Nil => // OK
  case params :: Nil => checkRepeatedParams(params)
  case params :: nonEmptyTail =>
    params.foreach(p => checkNonRepeated(p))
    onlyLastParameterCanBeRepeated(nonEmptyTail)
}
```

olafurpg: Alternatively, 
```scala
val repeatedParam = paramss.iterator.map(_.iterator).flatten.find(isRepatedParam)
val lastParam = paramss.lastOption.flatMap(_.lastOption))
if (repeatedParam.nonEmpty && lastParam != lastOption) {
  boom()
}
```

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2999,11 +2999,13 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
     val paramss = paramClauses(ownerIsType = false).require[List[List[Term.Param]]]
 
     def onlyLastParameterCanBeRepeated(params: List[Term.Param]): Unit = {
-      params.reverse.tail.foreach(p =>
-        if (p.decltpe.exists(_.isInstanceOf[Type.Repeated])) {
-          syntaxError("*-parameter must come last", p)
-        }
-      )
+      params
+        .take(params.length - 1)
```
olafurpg: This will also allocate a new list, but you can avoid that by making it an iterator first

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kpbochenek: I changed `reverse.tail` because it throws exception for empty list (similar to `list.init`)

kpbochenek: @olafurpg to be clear what is expected here, we have `List[List[Param]]` and we want to ensure for each list only last Param in each list can be RepeatedType. There might be multiple RepeatedType parameters in the end e.g.:
`def fx(x1: Boolean, x2: Int*)(y1: String, y2: Int*)`
we will get: 
`List(List(Boolean, Int*), List(String, Int*))` - which is correct

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kpbochenek: good catch with this allocation, I will try use `iterator`

kpbochenek:
To show you this in action, unsaved file is parsed by ScalametaParser, saved file error is reported by scala compiler.
Both report the same error in the same place. You can see how it looks with your current metals.

starparameterlast

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kpbochenek: changed to iterator :+1:

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
olafurpg: I misunderstood the expected behavior. I thought the following program was invalid but it's actually OK
```
object Foo {
  def foo(a: Int, b: Int*)(c: Int, d: Int*) = ???
}


```

scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/package.scala
```diff
@@ -215,4 +215,10 @@ package object trees {
     if (rank == 0) clazz
     else arrayClass(ScalaRunTime.arrayClass(clazz), rank - 1)
   }
+
+  def onlyLastParamCanBeRepeated(params: List[Term.Param]): Boolean = {
+    params
+      .take(params.length - 1)
```
tgodzik, line 221: Shouldn't we also have an iterator here in order not to allocate a new list?

tests/shared/src/test/scala/scala/meta/tests/parsers/StarParameterSuite.scala
```diff
@@ -0,0 +1,136 @@
+package scala.meta.tests
+package parsers
+
+import scala.meta._, Defn.Class
+import scala.meta.dialects.Scala211
+
+class StarParameterSuite extends ParseSuite {
+
+  test("star parameter single argument") {
```
tgodzik, line 9: `vararg parameter single argument` maybe? And with all the other tests.

tests/shared/src/test/scala/scala/meta/tests/parsers/StarParameterSuite.scala
```diff
@@ -0,0 +1,136 @@
...
+  test("error on return type star parameters") {
+    val definition = "def obj(f: Int*): Boolean* = true"
+
+    val error = intercept[parsers.ParseException] {
+      source(s"object Test { ${definition} }")
+    }
+    assert(error.getMessage.contains("error: = expected but identifier found"))
+  }
+
+  // TODO: This case should be correct in dotty(no error)!
```
tgodzik, line 99: We can remove this and change in Dotty with needed, or add to the checklist: https://github.com/scalameta/scalameta/issues/2022

scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/package.scala
kpbochenek: correct, changed :+1:

tests/shared/src/test/scala/scala/meta/tests/parsers/StarParameterSuite.scala
kpbochenek: right, replaced 'star' with 'vararg', also renamed a file(that's why this comment might get lost sorry)

tests/shared/src/test/scala/scala/meta/tests/parsers/StarParameterSuite.scala
kpbochenek: removed and added check to #2022

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kpbochenek: Not really a syntax you use everyday :)

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2997,6 +2997,16 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
       )
     val tparams = typeParamClauseOpt(ownerIsType = false, ctxBoundsAllowed = true)
     val paramss = paramClauses(ownerIsType = false).require[List[List[Term.Param]]]
+
+    def onlyLastParameterCanBeRepeated(params: List[Term.Param]): Unit = {
+      params.iterator
+        .take(params.length - 1)
+        .filter(p => !p.is[Term.Param.Quasi] && p.decltpe.exists(_.is[Type.Repeated]))
+        .foreach(syntaxError("*-parameter must come last", _))
```
tgodzik, line 3005:
```suggestion
        .foreach(syntaxError("vararg parameter must come last", _))
```

tgodzik: And I think it should be really `varargs` instead of `vararg` everywhere - sorry!

kpbochenek: But this is an error taken from scala compiler, do you want to have different error message from what scala reports?

olafurpg: We should use the same error message as the compiler, even if it's cryptic. Unrelated, I think the following reads better with a named parameter

```scala
.foreach(p => syntaxError("varargs parameter must come last", p))
```

Term wildcards are great in cases like `foreach(_.foo)` but otherwise it's almost always clearer with a named parameter

kpbochenek: to be fair @tgodzik must have checked dotty error, because they have different messages:
```
scala> def x(x: Int*, y: Int*): Unit = ???
             ^
       error: *-parameter must come last
```
```
scala> def x(x: Int*, y: Int*): Unit = ???
1 |def x(x: Int*, y: Int*): Unit = ???
  |         ^^^^
  |         varargs parameter must come last
```

kpbochenek: Updated to not have anonymous lambda in foreach.

I cannot really do `if dotty: errorA else errorB` because we operate on dialects not scala versions and this error not related to any features in dialect. I could use one that I know is only enabled in dotty but that would go against idea of dialects. I would keep current error and maybe change later when dotty gets more mature.

PR #2019, opened 2020-04-09
Fix symbols of nullary method eta-expansions

dwijnand:
When the Scala parser encounters an underscore while parsing expressions it
discards that token's position and creates a `MethodValue` of the target
method:

    case USCORE =>
      atPos(t.pos.start, in.skipToken()) { MethodValue(stripParens(t)) }

Then the Scala typechecker creates a `Function0` out of that, defining its
position as the position of the underlying method value:

    val pos = methodValue.pos
    // must create it here to change owner (normally done by typed's typedFunction)
    val funSym = context.owner.newAnonymousFunctionValue(pos)
    new ChangeOwnerTraverser(context.owner, funSym) traverse methodValue

    val result = typed(Function(List(), methodValue) setSymbol funSym setPos pos, mode, pt)

The result is that `XtensionCompilationUnitDocument#toTextDocument`, which
uses positions, incorrectly matches the `Function0` (Global) tree
(e.g. `(() => this.prop)`) to the underlying method expression (`prop`) 
Scalameta tree, and then converts that anonymous class (Global) symbol into
a local Semanticdb symbol.

So, instead of that, we skip the outer function when it has no params
(by flipping from `exists` to `forall`) so now the Scalameta tree `prop` is
correctly associated to the method's symbol.

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
@@ -632,7 +632,7 @@ trait TextDocumentOps { self: SemanticdbOps =>
                 traverse(original)
               case SelectOf(original) =>
                 traverse(original)
-              case g.Function(params, body) if params.exists { param =>
+              case g.Function(params, body) if params.forall { param =>
                     param.symbol.isSynthetic ||
                     param.name.decoded.startsWith("x$")
                   } =>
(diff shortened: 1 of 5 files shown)


semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -632,7 +632,7 @@ trait TextDocumentOps { self: SemanticdbOps =>
                 traverse(original)
               case SelectOf(original) =>
                 traverse(original)
-              case g.Function(params, body) if params.exists { param =>
+              case g.Function(params, body) if params.forall { param =>
```
olafurpg, line 635: That was a small change! Nice catch

olafurpg:
Kickstarted 4.3.8 release with this fix!

dwijnand:
Nice, thank you, Olaf!

PR #2028, opened 2020-04-23
Upgrade to Scala 2.13.2

olafurpg:


kpbochenek:
Looks like unused imports stopped being reported.

olafurpg:
I'm looking into the issue. We run the compiler in a special way and there seems to be a change in when reporting happens during the compiler phases.

kpbochenek:
Yeah it looks like adding `-Wconf:cat=unused-imports:w` could do the trick?
here: `    val warnUnusedImports = if (isScala213) "-Wunused:imports" else "-Ywarn-unused-import"`

Just a guess :crossed_fingers:

olafurpg:
It looks like we're not capturing any diagnostics. The targeted tests manually apply compiler phases, which I believe is the cause. The diagnostics are captured when running the sbt integration tests.

kpbochenek:
ahh ok that makes more sense, only in those 2 tests we check diagnostics :+1:

olafurpg:
Looks like it's best to stop supporting tests against diagnostics in SemanticdbSuite. There are only two tests that use this feature and it seems to rely on unstable internal implementation details of how the compiler reports diagnostics (which has changed in 2.13.2). We increasingly use ExpectSuite instead because those tests are more representative for how the SemanticDB compiler plugin runs in production.

PR #2032, opened 2020-04-25
ScalametaParser: fix try of partial enclosed expr

### LINKED ISSUE #2009: Parse error on try { 1 }.toString
```
try { 1 }.toString
```
Scalameta says 

> ; expected but . found

But scalac accepts that code.

Real life example: https://github.com/scala/scala/blob/d33518992f13dc609383229e6de9ffb412ab561d/src/library/scala/collection/immutable/Vector.scala#L261

### LINKED ISSUE #2031: `Try` fails to parse partially parenthesized expression without braces
The following fails parsing (4.3.7 and 4.3.10):
```
  try (a.b).c
  finally d
```
while these work:
```
  try { (a.b).c }
  finally d
```
and
```
  try (a.b.c)
  finally d
```
@olafurpg i can try to fix it, wanted to include the fix in `scalafmt 2.5.0`, but don't know where to start.

kitbellew:
Previously, if try was followed by `{` or `(`, the code assumed the
entire body is enclosed. This is an attempt to fix this.

Fixes #2031. Fixes #2009.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1606,17 +1606,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
         val body: Term = token match {
           case _ if dialect.allowTryWithAnyExpr => expr()
           case LeftBrace() => autoPos(inBracesOrUnit(block()))
-          case p @ LeftParen() =>
-            val term = inParensOrTupleOrUnit(location, allowRepeated)
-            term match {
-              case _: Lit.Unit | _: Term.Tuple =>
-                // NOTE: the position needs to be adapted to include parentheses
-                // when parsing a tuple or unit.
-                // See comments to makeTupleType for discussion
-                atPos(p, auto)(term)
-              case _ =>
-                term
-            }
+          case LeftParen() => inParensOrUnit(expr())
           case _ => expr()
         }
         val catchopt =
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1604,6 +1604,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
       case KwTry() =>
         next()
         val body: Term = token match {
+          case _ if dialect.allowTryWithAnyExpr => expr()
           case LeftBrace() => autoPos(inBracesOrUnit(block()))
           case p @ LeftParen() =>
             val term = inParensOrTupleOrUnit(location, allowRepeated)
--- a/tests/shared/src/test/scala-2.12/scala/meta/tests/tokenizers/TokenizerCoverageSuite.scala
+++ b/tests/shared/src/test/scala-2.12/scala/meta/tests/tokenizers/TokenizerCoverageSuite.scala
@@ -98,8 +98,13 @@ class TokenizerCoverageSuite() extends BaseTokenizerCoverageSuite {
   check[Term.Try]("try (→f←) catch { →case x => x;← →case y => y← } finally →{ }←")
   check[Term.Try]("try →()←")
   check[Term.Try]("try →(true, false)←")
+  check0[Term.Try]("try →{1 + 2}.toString←")(dialect = dialects.Scala213)
+  check0[Term.Try]("try →(a.b).c←")(dialect = dialects.Scala213)
   check[Term.TryWithHandler]("try (→f←) catch →(h)← finally →{ }←")
   check[Term.TryWithHandler]("try →(true, false)← catch →(h)← finally →(1, 2)←")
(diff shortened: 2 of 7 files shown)


kitbellew:
@olafurpg please take a look? not sure if `simpleExprRest` is sufficient but it's the only one that seems to continue building. also, not sure what those arrows in `TokenizerCoverageSuite` are checking.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1520,19 +1520,23 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
         }
       case KwTry() =>
         next()
+        val startToken = token
+        def getRestIfNeeded(x: Term) =
```
olafurpg, line 1524: It's easier to review when the type is explicitly annotated
```suggestion
        def getRestIfNeeded(x: Term): Term =
```

tests/shared/src/test/scala/scala/meta/tests/parsers/TermSuite.scala
```diff
@@ -337,6 +337,28 @@ class TermSuite extends ParseSuite {
...
+
+  test("try { 1 + 2 }.toString") {
+    val Try(
+      Select(
+        Term.Block(ApplyInfix(Lit(1), TermName("+"), Nil, Lit(2) :: Nil) :: Nil),
+        TermName("toString")
+      ),
+      Nil,
+      None
+    ) = term("try { 1 + 2 }.toString")
```
olafurpg, line 359: It would be good to additionally test infix operators
```scala
        term("try (1 :: Nil) map fn")
```

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1520,19 +1520,23 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
...
+          if (token.is[KwCatch] || token.is[KwFinally]) x
+          else simpleExprRest(atPos(startToken, auto)(x), true)
         val body: Term = token match {
```
olafurpg, line 1527: Judging by the diff in scalac, we need to change how the body is parsed https://github.com/scala/scala/pull/8071/files#diff-e059ce64777235990044f1624042e5d0R1518

Screenshot 2020-04-27 at 08 27 21

tests/shared/src/test/scala-2.12/scala/meta/tests/tokenizers/TokenizerCoverageSuite.scala
```diff
@@ -98,8 +98,11 @@ class TokenizerCoverageSuite() extends BaseTokenizerCoverageSuite {
   check[Term.Try]("try (→f←) catch { →case x => x;← →case y => y← } finally →{ }←")
   check[Term.Try]("try →()←")
   check[Term.Try]("try →(true, false)←")
+  check[Term.Try]("try →{1 + 2}.toString←")
```
olafurpg, line 101: These arrows are to show what tokens are owned by which tree node. I can't remember the exact meaning of the arrows, but if we change how the parser encloses which tokens then these tests should fail so that we can better understand the implications.

olafurpg:
The Scalameta parser error is consistent with 2.12.10
```scala
❯ cat foo.scala
object a {
  val fn: Int => Int = _ + 1
  def bar22 = try (1 :: Nil).map(fn) catch { case e: Throwable => e }
  def bar1 = try (1 :: Nil) map fn  catch { case e: Throwable => e }
}
❯ cs launch scalac:2.13.2 -- foo.scala # OK
❯ cs launch scalac:2.12.10 -- foo.scala # Error
foo.scala:3: error: ';' expected but '.' found.
  def bar22 = try (1 :: Nil).map(fn) catch { case e: Throwable => e }
                            ^
foo.scala:4: error: ';' expected but identifier found.
  def bar1 = try (1 :: Nil) map fn  catch { case e: Throwable => e }
                            ^
two errors found
```

Duhemm:
@olafurpg I don't have much more context. I just got asked why some code was crashing scalafmt 😄

olafurpg:
@Duhemm were you using 2.13 or 2.12?

Duhemm:
@olafurpg I think we were on 2.13.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: revised

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: done. added link to dialect.

tests/shared/src/test/scala/scala/meta/tests/parsers/TermSuite.scala
kitbellew: done

kitbellew:
@olafurpg PTAL. also, should we revert #1980? perhaps tuple was just a special case of _any_ expression and works only in 2.13.1+, hence would be covered by new code?

PR #2042, opened 2020-05-01
Dialect: allow copying with all internal vars

kitbellew:
For now, there's only one; but others are added in the future, copying
will not be possible as the number of `.withXxx` methods will grow.

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
@@ -101,11 +101,15 @@ import scala.compat.Platform.EOL
   override def canEqual(that: Any): Boolean = this eq that.asInstanceOf[AnyRef]
   override def equals(other: Any): Boolean = this eq other.asInstanceOf[AnyRef]
   override def hashCode: Int = System.identityHashCode(this)
+
   // Create new reference that copies over internal var settings from this instance.
   private def copyWithInternalVars(): Dialect = {
+    copyWithInternalVarsFrom(this)
+  }
+  def copyWithInternalVarsFrom(from: Dialect): Dialect = {
     val result = copy()
     result.internalAllowNumericLiteralUnderscoreSeparators =
-      this.internalAllowNumericLiteralUnderscoreSeparators
+      from.internalAllowNumericLiteralUnderscoreSeparators
     result
   }
 


scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -101,9 +101,12 @@ import scala.compat.Platform.EOL
   override def canEqual(that: Any): Boolean = this eq that.asInstanceOf[AnyRef]
   override def equals(other: Any): Boolean = this eq other.asInstanceOf[AnyRef]
   override def hashCode: Int = System.identityHashCode(this)
+
   // Create new reference that copies over internal var settings from this instance.
   private def copyWithInternalVars(): Dialect = {
-    val result = copy()
+    copyInternalVarsTo(copy())
+  }
+  def copyInternalVarsTo(result: Dialect): Dialect = {
```
olafurpg, line 109: This method performs side effects on `result`, can you elaborate on why this needs to be a public method?

kitbellew: That's the whole purpose. In scalafmt, I needed to change some field and obtain a modified dialect, which today means a .copy followed by the custom method copying the numeric underscores.

But your comment inside this class says that any new fields must be added as internal vars only. Meaning, if a new field is added, that sequence I described above will not work anymore.

Hence, if copying a dialect is potentially a valid operation we want to allow, then you need to provide a public copy method. People already have access to the regular case class copy method, and they have access to the custom set for each custom field, but if they don't need to modify the custom fields, what then?

olafurpg: The problem with exposing this method is that users can pass in the global `scala.meta.dialects.Scala211` instance as an argument and mutate it. I agree the current situation is not ideal since it's easy to forget to copy over new fields like is done in https://github.com/scalameta/scalafmt/pull/1940


The best long-term solution is probably to remove the `@data` annotation, manually implement the generated methods (`copy`, `apply`, ...) with `@deprecated` annotation and write new `withX()` builders for all fields. While the boilerplate is not fun to write, this approach has been working well in my experience (example https://github.com/scalacenter/scalafix/blob/6cfc5733c2e40cf0ad299300932fad06439a99d8/scalafix-core/src/main/scala/scalafix/v1/Configuration.scala)

olafurpg: If we manually write out `copy()` then we can make sure that the implementation copies over all fields.

kitbellew: let me propose a different solution :)

kitbellew: please see the last patch. the proposed usage will be:
```
  Scala211.copy(...).copyWithInternalVarsFrom(Scala211)
```
it guarantees that mutation will only happen on a new instance.

olafurpg:
I took a stab at refactoring `Dialect` so that we can keep on evolving this class in a binary compatible way while still making `copy()` work as expected https://github.com/scalameta/scalameta/pull/2044

PR #2044, opened 2020-05-02
Make it possible to evolve `Dialect` in a binary compatible way. 

olafurpg:
Supersedes #2042 

Previously, we used a hacky solution based on private vars to introduce
new fields to the `Dialect` class in a binary compatible way. This hacky
solution leaked to users when calling the auto-generated `copy()` method
because it would not copy over the private mutable fields.

Now, the `Dialect` class no longer uses the `@data` annotation and
instead this commit manually implements the macro expansion. This allows
us to introduce new constructor parameters without breaking the binary
signature of the previously auto-generated `copy()` method. This change
also fixes the situation so that users no longer need to worry about the
order in which they customize fields in `Dialect.copy(...).withX(...)`.


olafurpg:
FYI @kpbochenek if you rebase on top of master once this is merged then you can add new fields to `Dialect` for the Dotty PR without breaking the binary API

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -0,0 +1,427 @@
...
+    val allowWithTypes: Boolean,
+    // Are XML literals supported by this dialect?
+    // We plan to deprecate XML literal syntax, and some dialects
+    // might go ahead and drop support completely.
+    val allowXmlLiterals: Boolean,
+    // What kind of separator is necessary to split top-level statements?
+    // Normally none is required, but scripts may have their own rules.
+    val toplevelSeparator: String,
+    // Are numeric literal underscore separators, i.e. `1_000_000` legal or not?
+    val allowNumericLiteralUnderscoreSeparators: Boolean
```
kitbellew, line 71: why not use `= false` here and remove the other ctor?

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -0,0 +1,427 @@
...
+  }
+  def withAllowNumericLiteralUnderscoreSeparators(
+      allowNumericLiteralUnderscoreSeparators: Boolean
+  ): Dialect = {
+    privateCopy(allowNumericLiteralUnderscoreSeparators = allowNumericLiteralUnderscoreSeparators)
+  }
+
+  // NOTE(olafur): add the next `withX()` method above this comment.
+
+  private[this] def privateCopy(
```
kitbellew, line 222: interesting idea... why not make this public, and get rid of all those `withXxx` methods? this would look exactly like a `case class` copy, and if anyone ever used `.copy` elsewhere, would continue to work unchanged.

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -0,0 +1,427 @@
...
+  override def hashCode: Int = System.identityHashCode(this)
+  // Smart prettyprinting that knows about standard dialects.
+  override def toString = {
+    Dialect.standards.find(_._2 == this) match {
+      case Some((name, _)) => name
+      case None => s"Dialect(/* with custom parameters */)"
+    }
+  }
+
+  @deprecated("Use withX method instead", "4.3.11")
```
kitbellew, line 299: why? just curious...

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
kpbochenek: If you used `copy` method that defines all parameters, on every new parameter added you would break binary compatibility. If you only use `withX` those methods don't change with addition of new parameters, you are safe. (My reasoning for this change :P also binary compatibility sucks :< )

kitbellew: a, rozumiem...

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
kitbellew: i am guessing, because for binary compatibility, this ctor is private.

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -0,0 +1,427 @@
...
+      toplevelSeparator,
+      allowNumericLiteralUnderscoreSeparators = false
+      // NOTE(olafur): declare the default value for new fields above this comment.
+    )
+  }
+
+  // Are unquotes ($x) and splices (..$xs, ...$xss) allowed?
+  def allowUnquotes: Boolean = allowTermUnquotes || allowPatUnquotes
+
+  def withAllowAndTypes(allowAndTypes: Boolean): Dialect = {
```
kitbellew, line 140: optional: why not call the parameter as simply `flag` or even `f`, `b` (for bool), `v` (for value), to avoid repetition/clutter and to shorten (although i remember that you prefer `unfold`)? there's clear semantic meaning.
```
  def withAllowAndTypes(v: Boolean): Dialect = privateCopy(allowAndTypes = v)
```

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -0,0 +1,427 @@
...
+  // because sometimes dialects representing distinct Scala versions
+  // can be structurally equal to each other.
+  override def canEqual(that: Any): Boolean = this eq that.asInstanceOf[AnyRef]
+  override def equals(other: Any): Boolean = this eq other.asInstanceOf[AnyRef]
+  override def hashCode: Int = System.identityHashCode(this)
+  // Smart prettyprinting that knows about standard dialects.
+  override def toString = {
+    Dialect.standards.find(_._2 == this) match {
+      case Some((name, _)) => name
+      case None => s"Dialect(/* with custom parameters */)"
```
kitbellew, line 295: `productIterator` using reflection? :)

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -0,0 +1,427 @@
...
+      allowViewBounds,
+      allowWithTypes,
+      allowXmlLiterals,
+      toplevelSeparator
+    )
+  }
+}
+
+object Dialect extends InternalDialect {
+  def apply(
```
kitbellew, line 356: should this also be deprecated?

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
olafurpg: Default parameter values auto-generate synthetic `$default..` methods on the companion object that leak into the public API. It's easiest to reason about binary compatibility when you type out all the methods by hand, for better or worse.

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
olafurpg: Good suggestion. I renamed the parameter to `newValue` so that it's easier to copy-paste new methods.

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
olafurpg: @kpbochenek Correct, by keeping this method private we don't break the public API when we add new parameters.

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
olafurpg: I made `productIterator` return empty list to avoid redundant boilerplate. `Dialect` should not be a `Product`, I added `@deprecated()` annotations on the `def product*` methods.

I changed this `toString` to return `Dialect()` instead. I never like the old `Dialect.toString`, it was very verbose for not benefit because it's hard to make sense of output like `Dialect(true, false, true, true, true, ....)`

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
olafurpg: Because we can't safely evolve this method in a binary compatible way. Case class `copy()` methods are easy to evolve in a source-compatible way but every new field addition is a binary breaking change even if you provide a default value.

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
olafurpg: We may want to deprecate this method in the future but I think it's OK to keep it unchanged for now. I tried to introduce a default constructor `def apply(): Dialect` but couldn't find out a way to implement it without having to declare the default values in two different places, which I want to avoid. Most users adjust existing dialects like `Scala213`.

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
olafurpg: @kpbochenek I agree binary compatibility can be annoying, but it could be worse. My understanding is that evolving APIs in a binary compatible way is a lot harder and sometimes near impossible in other languages/ecosystems. It's at least **possible** with Scala/JVM, even if it's cumbersome.

PR #2048, opened 2020-05-21
internal.Scaladoc: another way to parse a scaladoc

kitbellew:
The existing `DocToken`/`ScaladocParser` pair is missing support for list blocks and has incomplete support for tags.

To avoid breaking any existing usage, propose creating a new pair of classes, to add the missing support. Also, as the code is already using a sophisticated parsing library, let's add more structure into the tokens.

Needed in order to address:
https://github.com/scalameta/scalafmt/issues/891
https://github.com/scalameta/scalafmt/issues/1387

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
@@ -1,0 +1,452 @@
+package scala.meta.tests.parsers
+
+import scala.meta.internal.Scaladoc
+import scala.meta.internal.Scaladoc._
+import scala.meta.internal.parsers.ScaladocParser
+
+import munit.FunSuite
+
+/**
+ * Test for [[ScaladocParser]]
+ */
+class ScaladocParserSuite extends FunSuite {
+
+  private def parseString(comment: String) =
+    ScaladocParser.parse(comment.trim)
+
+  private def generateTestString(tagType: TagType): String = {
+    val sb = new StringBuilder
+    sb.append(tagType.tag)
+    if (tagType.hasLabel) sb.append(" TestLabel")
+    if (tagType.hasDesc) sb.append("\n *  Test Description")
+    sb.result()
+  }
+
+  test("example usage") {
+    assertEquals(
+      parseString("/** Example scaladoc */"),
+      Some(Scaladoc(Paragraph(Text(Word("Example"), Word("scaladoc")))))
+    )
+  }
+
+  test("indentation checks") {
+
+    val expectedBody: String = "BODY"
+    val expectedBodyToken = Text(Word(expectedBody))
+
+    assertEquals(
+      parseString(s"/** $expectedBody*/"),
+      Some(Scaladoc(Paragraph(expectedBodyToken)))
+    )
+    assertEquals(
+      parseString(
+        s"""
+         /** $expectedBody
+          */
+         """
+      ),
+      Some(Scaladoc(Paragraph(expectedBodyToken)))
+    )
+    assertEquals(
+      parseString(
+        s"""
+         /**       $expectedBody
+          */
+         """
+      ),
+      Some(Scaladoc(Paragraph(expectedBodyToken)))
+    )
+    assertEquals(
+      parseString(
+        s"""
+         /**
+          *$expectedBody
+          */
+         """
+      ),
+      Some(Scaladoc(Paragraph(expectedBodyToken)))
+    )
+  }
+
+  test("paragraph parsing") {
+    val descriptionBody = "Description Body"
+    val words = descriptionBody.split("\\s+").map(Word.apply)
+    assertEquals(
+      parseString(
+        s"""
+         /**
+          *
+          *$descriptionBody
+          *
+          *$descriptionBody
+          *
+          */
+         """
+      ),
+      Option(
+        Scaladoc(
+          Paragraph(Text(words: _*)),
+          Paragraph(Text(words: _*))
+        )
+      )
+    )
+  }
+
+  test("paragraph parsing with leading space") {
+    val descriptionBody = "Description Body"
+    val words = descriptionBody.split("\\s+").map(Word.apply)
+    assertEquals(
+      parseString(
+        s"""
+         /**
+          *
+          * $descriptionBody
+          *
+          * $descriptionBody
+          *
+          */
+         """
+      ),
+      Option(
+        Scaladoc(
+          Paragraph(Text(words: _*)),
+          Paragraph(Text(words: _*))
+        )
+      )
+    )
+  }
+
+  test("paragraph parsing with references") {
+    val descriptionBody = "Description Body"
+    val words = descriptionBody.split("\\s+").toSeq.map(Word.apply)
+    val ref = Seq(Link("Description", "Body"))
+    assertEquals(
+      parseString(
+        s"""
+         /**
+          *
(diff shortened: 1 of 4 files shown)


kitbellew:
@xavier-fernandez can you please take a look?

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/CommentOps.scala
```diff
@@ -1,46 +1,32 @@
...
 
-  @tailrec
-  private def dropRightWhile(str: String, predicate: (Char) => Boolean): String =
-    if (str.isEmpty)
-      ""
-    else if (predicate(str.last))
-      dropRightWhile(str.init, predicate)
-    else
-      str
+  @inline
```
olafurpg, line 9: Let's drop `@inline` annotations even if they were there before, we have no indication these help with performance bottlenecks

olafurpg:
The mima checks are failing because the 4.3.11 artifacts don't exist. We can tweak `mimaPreviousArtifacts` in build.sbt to use 4.3.0 instead

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/CommentOps.scala
kitbellew: will do

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/ScaladocToken.scala
```diff
@@ -0,0 +1,197 @@
+package scala.meta.contrib
```
kitbellew, line 1: a few comments, after reading your review of the other implementation several years ago (and some online, in subsequent code).

while this does look a tree, i didn't understand your suggestion referring to @root and @leaf, so i didn't implement that; i looked at the Adt code but didn't really understand what it does, and have zero experience with macros

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/ScaladocToken.scala
```diff
@@ -0,0 +1,197 @@
...
+      val hasLabel: Boolean = false,
+      val hasDesc: Boolean = false
+  )
+
+  /**
+   * Represents a tagged documentation remark
+   * @param label set iff [[tag.hasLabel]]
+   * @param desc set iff [[tag.hasDesc]]
+   */
+  final case class Tag(tag: TagType, label: Word = null, desc: Text = null) extends ScaladocToken
```
kitbellew, line 66: tags are implemented using a single type, referring to a tag type as a field, rather than as separate types, and this part resembles the existing approach. it was easier to instantiate them like that when parsing, let me know if you prefer to have separate types and have a simple way to construct.

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/ScaladocToken.scala
```diff
@@ -0,0 +1,197 @@
...
+  /** A blank line */
+  final case object Break extends ScaladocToken
+
+  /** A paragraph which failed parsing */
+  final case class Unknown(text: String) extends ScaladocToken
+
+  // Text
+
+  /** An inline part of a text block */
+  trait TextPart {
```
kitbellew, line 21: i didn't implement the reference type as a subclass of this. primarily because I'm not sure what the syntax is (some examples on the web include multiple words between the bracket pairs).

another aspect i sort of struggled with... in my case, the goal is to be able to format this, but this is not the place to decide on formatting, so i only added this `syntax` method to inline tokens, those which don't span multiple lines, to reduce the number of case expressions in the match statement in the formatter, listing only a subset of all possible tokens.

so if reference could potentially be split across lines, with multiple words, this complicates things.

there are also other examples, such as markdown of italics, fixed-width font, etc. they are not well documented, so I skipped them, as well.

scalameta/contrib/shared/src/main/scala/scala/meta/contrib/ScaladocTokenizer.scala
```diff
@@ -0,0 +1,184 @@
...
+            res ++= p.value
+            p.value.nonEmpty
+          case _ =>
+            res += Unknown(paragraph)
+            true
+        }
+      }
+
+    var flushed = false
+    comment.linesIterator.foreach { line =>
```
kitbellew, line 167: one other comment you made in that old pr is to avoid text manipulation outside of fastparse. while i did that, compared to the previous implementation, by combining all parsers into one, there was one thing that i couldn't get fastparse to do.

i couldn't write a parser that would be able to identify paragraph breaks (one or more black lines) and at the same time not consume the last newline. keeping that newline was critical to implementing the list block parser, which is sensitive to indentation at the beginning of a line.

so i decided to split code into paragraphs, stripping leading space before asterisk and trailing space, and applying fastparse to the result.

olafurpg:
I'm OK with not promising binary compatibility for this new API. I think mima has settings to ignore for example everything in a certain package. Alternatively, we can keep this new functionality under `scala.meta.internal` while it's incubating.

I think it's a good call not to use the `@ast` and `@leaf` macros, I wish we didn't have any macro annotations in the codebase.

One relevant background, in Metals we use another scaladoc parser that is derived from the compiler docstring parser. I'm not sure if that AST is suitable for a code formatter, but it might be worth looking at how it does things:

https://github.com/scalameta/metals/blob/c169b4ce667d731c34851b8a0dbb85ad8adb1231/mtags/src/main/scala-2/scala/meta/internal/docstrings/ScaladocParser.scala 

I'm afraid I won't have time to review the implementation details. I wish I could help more since I think this is a really exciting new feature.

kitbellew:
> I'm OK with not promising binary compatibility for this new API. I think mima has settings to ignore for example everything in a certain package. Alternatively, we can keep this new functionality under `scala.meta.internal` while it's incubating.

Done, moved to `internal`.

> One relevant background, in Metals we use another scaladoc parser that is derived from the compiler docstring parser. I'm not sure if that AST is suitable for a code formatter, but it might be worth looking at how it does things:

Looked at it. It seems it is more suited to extracting information from the scaladoc, but it seems to lose ordering of elements. Also, for some reason, it doesn't have lists (perhaps it's not a commonly used feature, https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html#other-formatting-notes).

Perhaps, if metals uses scalameta, it might make sense to move the metals parser over, and expand it to support the formatter use case.

I will merge it for now, will use in scalafmt.

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
```diff
@@ -0,0 +1,425 @@
...
+          *    - $list21
+          *       I. $list31
+          *        - $list41
+          *        - $list42
+          *       I. $list32
+          *    - $list22
+          */
+         """
+      )
+    val expected = Option(
```
olafurpg, line 274: This value is a bit too big to produce helpful diffs on test failures. I would consider using a smaller example

kitbellew: done. split into several tests.

kitbellew:
@olafurpg please let me know when you planning to tag a new release, i will get it into `scalafmt`. for now, however, i will build it locally and test it with `scalafmt`, to make sure it actually works :)

PR #2059, opened 2020-05-30
Scaladoc: implement parsing of tables

kitbellew:
https://www.scala-lang.org/blog/2018/10/04/scaladoc-tables.html

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
@@ -37,6 +37,10 @@ object ScaladocParser {
 
   private val listPrefix = "-" | CharIn("1aiI") ~ "."
 
+  private val escape = P("\\")
+  private val tableSep = P("|")
+  private val tableSpaceSep = P(hspaces0 ~ tableSep)
+
   private val codePrefix = P("{{{")
   private val codeSuffix = P(hspaces0 ~ "}}}")
 
@@ -83,8 +87,8 @@ object ScaladocParser {
   }
 
   private val textParser: Parser[Text] = {
-    val anotherBeg = P(hspaces0 ~/ (CharIn("@=") | (codePrefix ~ nl) | listPrefix))
-    val end = P(End | nl ~/ anotherBeg)
+    val anotherBeg = P(CharIn("@=") | (codePrefix ~ nl) | listPrefix | tableSep | "+-")
+    val end = P(End | nl ~/ hspaces0 ~/ anotherBeg)
     val part: Parser[TextPart] = P(codeExprParser | linkParser | wordParser)
     val sep = P(!end ~ nlHspaces1)
     val text = hspaces0 ~ part.rep(1, sep = sep)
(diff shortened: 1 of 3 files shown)


scalameta/trees/shared/src/main/scala/scala/meta/internal/Scaladoc.scala
```diff
@@ -48,6 +48,45 @@ object Scaladoc {
...
   final case class Heading(level: Int, title: String) extends Term
 
+  /**
+   * A table
+   * [[https://www.scala-lang.org/blog/2018/10/04/scaladoc-tables.html]]
+   */
+  final case class Table(
+      header: Table.Row,
+      align: Seq[Table.Align],
+      row: Table.Row*
```
olafurpg, line 58: Would it make sense to use Seq here instead of varargs? My personal rule of thumb is to avoid varargs where possible because they make it harder to introduce new fields in the future.

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
```diff
@@ -496,4 +496,47 @@ class ScaladocParserSuite extends FunSuite {
     )
   }
 
+  test("table") {
```
olafurpg, line 499: Might be worth having test cases for parsing incomplete tables, for example missing pipes, multiple alignment rows, wrong characters in the alignment row, is it possible to escape a pipe character?

scalameta/trees/shared/src/main/scala/scala/meta/internal/Scaladoc.scala
kitbellew: will done. chose earlier for brevity. based on this comment, submitted #2062.

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
kitbellew: will do.

* there's a already a test for escaped pipe char (line 504)
* the code already allows a different number of columns in each row
  * a row is essentially a line starting and ending with a pipe char
  * there's a test in `scalafmt` but will add it here as well
  * but doesn't allow a missing ending pipe char (will add a test)
* the code doesn't check that alignment characters are '-' or ':'; it simply checks if the cell starts or ends with ':'; should it be "more careful"?

olafurpg: That sounds great! The biggest risk i can think of with formatting docstrings is that Scalafmt will ruin somebody’s ascii art or something else that should be left alone verbatim (even if it doesn’t render correctly in the Scalafix html).

Unrelated to this PR, is there a way for users to disable this feature for a specific docstring?

kitbellew: > That sounds great! The biggest risk i can think of with formatting docstrings is that Scalafmt will ruin somebody’s ascii art or something else that should be left alone verbatim (even if it doesn’t render correctly in the Scalafix html).

We could parse out any lines which don't start with a letter as an opaque line, rather than splitting into text with words.

> Unrelated to this PR, is there a way for users to disable this feature for a specific docstring?

I am guessing `// format: off` should work, will test that in `scalafmt`.

kitbellew: added a few more tests, as requested, and expanded the parser a bit.

kitbellew:
@olafurpg any other modifications you could suggest? otherwise, I'd merge, tag and add to the formatter.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -130,13 +134,73 @@ object ScaladocParser {
...
+      }
+    }
+
+    val cellChar = escape ~ AnyChar | !(nl | escape | tableSep) ~ AnyChar
+    val row = (cellChar.rep.! ~ tableSpaceSep).rep(1)
+    // non-standard but frequently used delimiter line, e.g.: +-----+-------+
+    val delimLine = hspaces0 ~ CharsWhileIn("+-")
+    val sep = nl ~ (delimLine ~ nl).rep ~ tableSpaceSep
+    val table = row.rep(2, sep = sep).map { x =>
+      val rest = x.tail.map(_.map(_.trim))
```
olafurpg, line 162: Do we know `x` is always non-empty?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -130,13 +134,73 @@ object ScaladocParser {
...
+  private val tableParser: Parser[Table] = {
+    def toRow(x: Iterable[String]): Table.Row = Table.Row(x.toSeq)
+    def toAlign(x: String): Option[Table.Align] = {
+      def isEnd(y: Char) = y match {
+        case ':' => Some(true)
+        case '-' => Some(false)
+        case _ => None
+      }
+      for {
+        isLeft <- isEnd(x.head)
```
olafurpg, line 146: Is it guaranteed that `x` is non-empty?

kitbellew: let me test this scenario

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
kitbellew: i will add a test but if i read the `.rep(2)` above correctly, it should have at least 2 elements.

PR #2062, opened 2020-05-31
Scaladoc: use Seq, not varargs, for extensibility

kitbellew:


kitbellew:
@olafurpg based on your comment to #2059, i am changing all varargs to `Seq`. let me know if i understood your point correctly.

PR #2065, opened 2020-06-04
keep a reference to the semanticdb reporter

ingarabr:
Keep a reference to the reporter we're overriding. Other compiler plugins (silencer compiler plugin) can change the reporter and we're not guaranteed that it will be a StoreReporter. We can end up in a situation where the reporter is replaced and the messages aren't passed to the underlying reporter or ignored.

---
We (@laughedelic and myself) discovered this issue when I tried to use the silencer plugin in combination with scalafix. Both compiler plugins swap out the reporter with there own but pass the messages to the underlying reporter.

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/HijackReporter.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/HijackReporter.scala
@@ -2,14 +2,15 @@ package scala.meta.internal.semanticdb.scalac
 
 trait HijackReporter { self: SemanticdbPlugin =>
 
-  def hijackReporter(): Unit = {
-    if (!isSupportedCompiler) return
+  def hijackReporter(): Option[SemanticdbReporter] = {
+    if (!isSupportedCompiler) return None
 
     g.reporter match {
-      case _: SemanticdbReporter => // do nothing, already hijacked
+      case reporter: SemanticdbReporter => Some(reporter) // do nothing, already hijacked
       case underlying =>
         val semanticdbReporter = new SemanticdbReporter(underlying)
         g.reporter = semanticdbReporter
+        Some(semanticdbReporter)
     }
   }
 }
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/ReporterOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/ReporterOps.scala
@@ -5,12 +5,14 @@ import scala.tools.nsc.reporters.StoreReporter
 import scala.reflect.internal.util.{Position => gPosition}
 
 trait ReporterOps { self: SemanticdbOps =>
+  val semanticdbReporter: Option[SemanticdbReporter] = None
+
   // Hack, keep track of how many messages we have returns for each path to avoid
   // duplicate messages. The key is System.identityHashCode to keep memory usage low.
   private val returnedMessagesByPath = mutable.Map.empty[g.CompilationUnit, Int]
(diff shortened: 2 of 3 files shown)


semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/HijackReporter.scala
```diff
@@ -2,14 +2,15 @@ package scala.meta.internal.semanticdb.scalac
 
 trait HijackReporter { self: SemanticdbPlugin =>
 
-  def hijackReporter(): Unit = {
-    if (!isSupportedCompiler) return
+  def hijackReporter(): Option[SemanticdbReporter] = {
+    if (!isSupportedCompiler) return null
```
tgodzik, line 6:
```suggestion
    if (!isSupportedCompiler) return None
```
?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/ReporterOps.scala
```diff
@@ -5,12 +5,14 @@ import scala.tools.nsc.reporters.StoreReporter
 import scala.reflect.internal.util.{Position => gPosition}
 
 trait ReporterOps { self: SemanticdbOps =>
+  var semanticdbReporter: Option[SemanticdbReporter] = None
```
tgodzik, line 8:
```suggestion
  val semanticdbReporter: Option[SemanticdbReporter] = None
```

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SemanticdbPlugin.scala
```diff
@@ -13,7 +12,8 @@ class SemanticdbPlugin(val global: Global)
   val name = SemanticdbPlugin.name
   val description = SemanticdbPlugin.description
 
-  hijackReporter()
+  semanticdbReporter = hijackReporter()
```
tgodzik, line 15:
```suggestion
  override val semanticdbReporter = hijackReporter()
```

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/HijackReporter.scala
ingarabr: heh, good catch :)

ingarabr:
@tgodzik, I have applied the changes and force pushed them. I expected some pushback on the var but wanted to get it out there to see if it was in the right direction :-)

Thank you for the fast review.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SemanticdbPlugin.scala
```diff
@@ -13,7 +12,8 @@ class SemanticdbPlugin(val global: Global)
   val name = SemanticdbPlugin.name
   val description = SemanticdbPlugin.description
 
-  hijackReporter()
+  override val semanticdbReporter = hijackReporter()
```
olafurpg, line 15: FYI the `override` here shouldn't be necessary, I think you can remove `val semanticdbReporter = None` in `ReporterOps` and it will still work

ingarabr: We need to have a reference to pick it up in `ReporterOps`. Your proposal can be done if we change `val semanticdbReporter:Option[SemanticdbReporter] = None` to `def semanticdbReporter: Option[SemanticdbReporter]` in `ReporterOps`. By doing so it's not an override and it can be removed.

PR #2073, opened 2020-06-26
Fix issue with a match expression in an if statement

tgodzik:
Found the issue while testing inside Metals. The code is a bit convoluted around the separate regions part, but in essence  the condition was wrong.

`sepRegions1.head != '}' || sepRegions1.head != '$'` will always be true, which was the issue.


tgodzik:
Checked with Metals and this fixes the issues we've had.

PR #2093, opened 2020-07-13
Add support for Scala 2.12.12

olafurpg:


.github/workflows/ci.yml
```diff
@@ -16,6 +16,7 @@ jobs:
           - "++2.12.10 testsJVM/test"
           - "++2.12.11 testsJVM/test"
           - "++2.12.11 testsJS/test"
+          - "++2.12.12 testsJS/test"
           - "++2.13.1 testsJVM/test"
           - "++2.13.2 testsJVM/test"
```
kpbochenek, line 21: 3 lines below:
`download-scala-library ++2.12.11 testsJVM/slow:test` -> `download-scala-library ++2.12.12 testsJVM/slow:test`

.github/workflows/ci.yml
```diff
@@ -16,6 +16,7 @@ jobs:
           - "++2.12.10 testsJVM/test"
           - "++2.12.11 testsJVM/test"
           - "++2.12.11 testsJS/test"
+          - "++2.12.12 testsJS/test"
```
kpbochenek, line 19: looks like we run JVM tests for all versions and JS only for latest.
Suggestion if that is the case:
remove `"++2.12.11 testsJS/test"`
add `"++2.12.12 testsJVM/test"`

.github/workflows/ci.yml
```diff
@@ -16,6 +16,7 @@ jobs:
...
           - "++2.13.1 testsJVM/test"
           - "++2.13.2 testsJVM/test"
           - "++2.13.3 testsJVM/test"
```
kpbochenek, line 22: a few lines below is:
`run: sbt ++2.12.11 testsJVM/test` 
this should also probably be updated to `2.12.12`?

kpbochenek:
Still something doesn't compile Oo
```
home/runner/work/scalameta/scalameta/target/scala-library/scala-2.12.12/src/library/scala/collection/immutable/TreeMap.scala:333: error: type arguments [_] do not conform to class SortedMapFactory's type parameter bounds [CC[A, B] <: scala.collection.SortedMap[A,B] with scala.collection.SortedMapLike[A,B,CC[A,B]]]
      case cbf: SortedMapFactory[_]#SortedMapCanBuildFrom[_,_] => {
                ^
/home/runner/work/scalameta/scalameta/target/scala-library/scala-2.12.12/src/library/scala/collection/immutable/TreeSet.scala:210: error: type arguments [_] do not conform to class SortedSetFactory's type parameter bounds [CC[A] <: scala.collection.SortedSet[A] with scala.collection.SortedSetLike[A,CC[A]]]
      case cbf: SortedSetFactory[_]#SortedSetCanBuildFrom[_] => {
                ^
two errors found
```

olafurpg:
I changed to compile the 2.13.3 sources instead of 2.12.12, which works around the issue. I unable to reproduce the issue by disabling the SemanticDB compiler plugin when compiling the sources of 2.12.12 so I suspect the root cause is on our side instead of a regression in 2.12.12. I don't have time to investigate further so I'm gonna go ahead and merge anyways.

We should keep an eye out for related issues where enabling SemanticDB causes compile errors in 2.12.12. I haven't seen this happen before, the plugin doesn't mess with the type-checker anymore (it used to replace the type-checker with runtime reflection).

PR #2094, opened 2020-07-20
InteractiveSemanticdb doesn't get diagnostics for unused imports in 2.13

mlachkar:
When compiling on-demand with `new compiler`, even if I set scalacOptions, I cannot get diagnostics at least for Unused imports. 
The test shows that diagnostics is empty. 
I need some help on this issue. We use it in Scalafix, and I noticed that since 2.13, it's not working correctly anymore. Do you have any hints on how to pass scalacOptions correctly to the compiler on 2.13.


tgodzik:
This might also have something with the SemanticdbReporter class. We pack the default reporter into there and maybe the warnings don't get forwarded?

mlachkar:
I agree it may be the reporter, but I don't know how to validate it's the root cause and /fix it.

tgodzik:
I can try to help out later today.

tgodzik:
Tried digging into it today and it seems running `askLoadedTyped` is no longer reporting any warnings. If I try to run the compilation via Global they seem to be reported properly. Might be an issue in the compiler?

mlachkar:
This PR has already been merged in 2.13.2, and the tests here at least are still not passing in 2.13.2 nor 2.13.3.
In this test, `new StoreReporter()` is used as a reporter. Do you advise another reporter?

tgodzik:
> This PR has already been merged in 2.13.2, and the tests here at least are still not passing in 2.13.2 nor 2.13.3.
> In this test, `new StoreReporter()` is used as a reporter. Do you advise another reporter?

I am not sure if the reporter is the issue. I am able to get warnings if I try to compile the file not just get type information.

tgodzik:
Fixed by https://github.com/scalameta/scalameta/pull/2216

PR #2105, opened 2020-07-30
Fix object names as unary operators

### LINKED ISSUE #2095: Parse error with a minus (dash) token
Coming from https://github.com/scalameta/scalafmt/issues/2104
The following code causes a parsing error.
```scala
sealed trait Foo {
  def negate : - = -
}
trait -
case object - extends -
```

kpbochenek:
Fixes #2095

When we encounter token looking like 'unary operator' we always parse it this way.
But as in example in #2095 it can be also an ident(name of object/class etc). 

Instead of throwing exception when we cannot parse it as unary operator we can try and assume '-' is ident and parse it this way.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -30,6 +30,9 @@ import scala.meta.Defn.Given
 import SoftKeyword._
 import scala.meta.Defn.ExtensionGroup
 import scala.meta.Defn.ExtensionMethod
+import scala.util.Try
+import scala.util.Success
+import scala.util.Failure
 
 class ScalametaParser(input: Input, dialect: Dialect) { parser =>
   require(Set("", EOL).contains(dialect.toplevelSeparator))
@@ -422,7 +425,20 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
     val pos = TokenStreamPosition(startTokenPos, endTokenPos + 1)
     result.withOrigin(Origin.Parsed(input, dialect, pos)).asInstanceOf[T]
   }
+
+  def atPosTry[T <: Tree](start: Pos, end: Pos)(body: => Try[T]): Try[T] = {
+    val startTokenPos = start.startTokenPos
+    val result = body
+    var endTokenPos = end.endTokenPos
+    if (endTokenPos < startTokenPos) endTokenPos = startTokenPos - 1
+    val pos = TokenStreamPosition(startTokenPos, endTokenPos + 1)
+    result.map(_.withOrigin(Origin.Parsed(input, dialect, pos)).asInstanceOf[T])
+  }
+
   def autoPos[T <: Tree](body: => T): T = atPos(start = auto, end = auto)(body)
+
+  def autoPosTry[T <: Tree](body: => Try[T]): Try[T] =
+    atPosTry(start = auto, end = auto)(body)
 
   /* ------------- ERROR HANDLING ------------------------------------------- */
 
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2074,58 +2086,72 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
...
+            // we would fail here anyway, let's try to treat it as ident
+            simpleExprRest(op, canApply = true)
+        }
+      }
     }
 
   def simpleExpr(allowRepeated: Boolean): Term = autoPos {
+    simpleExpr0(allowRepeated) match {
+      case Right(result) => result
+      case Left(error) => throw error
```
tgodzik, line 2103: Maybe let's add a test case with this failing?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -422,8 +422,20 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
...
     result.withOrigin(Origin.Parsed(input, dialect, pos)).asInstanceOf[T]
   }
+
+  def atPosE[E, T <: Tree](start: Pos, end: Pos)(body: => Either[E, T]): Either[E, T] = {
+    val startTokenPos = start.startTokenPos
+    val result = body
+    var endTokenPos = end.endTokenPos
+    if (endTokenPos < startTokenPos) endTokenPos = startTokenPos - 1
+    val pos = TokenStreamPosition(startTokenPos, endTokenPos + 1)
+    result.map(_.withOrigin(Origin.Parsed(input, dialect, pos)).asInstanceOf[T])
```
tgodzik, line 432: Will this work in Scala 2.11.12? I think Either is right biased starting from 2.12.0

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2074,58 +2086,72 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
...
+          Right(macroSplice())
         case MacroQuotedIdent(_) =>
-          macroQuotedIdent()
+          Right(macroQuotedIdent())
         case MacroSplicedIdent(_) =>
-          macroSplicedIdent()
+          Right(macroSplicedIdent())
         case _ =>
-          syntaxError(s"illegal start of simple expression", at = token)
+          Left(new ParseException(token.pos, "illegal start of simple expression"))
```
tgodzik, line 2151: since we are throwing it anyway, would it make sense to just throw it here? There would be no need for Either

kpbochenek: this either is used in 2 places, in one it is thrown, in second where I fix mentioned thing it is not thrown

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kpbochenek: yeah test failed, will explicitly map right value

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kpbochenek: added in test "simple-expression-parse-error"

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -422,8 +422,21 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
     val pos = TokenStreamPosition(startTokenPos, endTokenPos + 1)
     result.withOrigin(Origin.Parsed(input, dialect, pos)).asInstanceOf[T]
   }
+
+  def atPosE[E, T <: Tree](start: Pos, end: Pos)(body: => Either[E, T]): Either[E, T] = {
```
kitbellew, line 426: should we just use `Try` instead of `Either`? do you envision a use case when something other than an exception would be used?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: if you had used `Try`, this would be just `.get`

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kpbochenek: Check `simpleExpr0` where I create `Left` in one specific case when parsing didn't fall into any match case.
It is hardly doable to get the same results with Try without some weird stuff.
Because if e.g. parsing of `case Underscore()` fails I still want it to throw exception that was raised and not return `Left(...)`

Try could definitely replace Either but how to use that in `simpleExpr0` would be problematic. I don't want to wrap everything in Try because that would change a logic.

Ideally I would like something like:
`def atPosE[F, T <: Tree](start: Pos, end: Pos)(body: => F[T]): F[T]`

but Either is not right biased in 2.11 - what can you do? :/

To sum up what I want:
- in `simpleExpr0` if fall into `case _ =>` notify parent method about this! 
- in `simpleExpr0` if exception is thrown -> keep it throwing to parent methods
- parent of `simpleExpr0` if notified about `case _ =>` continue parsing assuming unary_op is actually defined object name.

kitbellew: Panie Krzyśku, my bad, I should have been clearer.

I am _not_ suggesting to wrap everything in `Try`. I am simply suggesting to replace `Either` with `Try`, `Right` with `Success`, and `Left` with `Failure`. In cases where you currently throw an exception, you keep that and not wrap in `Failure`.

Would that work?

kpbochenek: Total respect for perfect conversion "Krzysztof" -> "Krzyśku" even with ś!

Yes using explicitly `Left` and `Right` and `.get` where I want walue or want to throw exception will work.
It might be misleading that a function returns `Try[X]` and can also additionally throw exception but if already 2 people don't like Either here there must be something in it :) Will change to Try :+1:

kitbellew: > Total respect for perfect conversion "Krzysiek" -> "Krzyśku" even with ś!

Well, I grew up w Litwie, so I still have some rudimentary knowledge języka polskiego. Clearly, I had to look up the nasal "e" and the accented soft "sh" as my keyboard doesn't have them. :) Fun exercise.

> Yes using explicitly `Left` and `Right` and `.get` where I want walue or want to throw exception will work.
> It might be misleading that a function returns `Try[X]` and can also additionally throw exception but if already 2 people don't like Either here there must be something in it :) Will change to Try 👍

You're right; if you believe that it will be confusing to use `Try` which lets some exceptions sneak through, let's not use it. We use it routinely, so I didn't think twice about that.

kpbochenek: changed to `Try`

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kpbochenek: changed to `Try`as Albert suggested.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2074,58 +2090,67 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
       val op = termName()
       if (op.value == "-" && token.is[NumericLiteral])
         simpleExprRest(atPos(op, auto)(literal(isNegated = true)), canApply = true)
-      else
-        atPos(op, auto)(Term.ApplyUnary(op, simpleExpr(allowRepeated = true)))
+      else {
+        simpleExpr0(allowRepeated = true) match {
```
kitbellew, line 2094: if we hadn't had to build it for 2.11, `fold` would have been ideal here, although our usage is also equivalent to `.map(...).getOrElse(...)`.

PR #2115, opened 2020-09-22
Fix #2106 - properly escape non-letter non-ascii characters

### LINKED ISSUE #2106: Scalameta removes Unicode escapes in String literals
Scalameta escapes backslashes when printing String literals, but this prevents generating Unicode escapes for special characters that Scalameta doesn't already escape, like control characters.

For example, this compiles in a Scala repl: `val str = "xy\u001az"` (`\u001a` is a substitution control character)

However, Scalameta can't generate this. Including the control character with `Lit.String("xy\u001az")` causes it to be printed directly, generating invalid code. Escaping it with `Lit.String("xy\\u001az")` also doesn't work because Scalameta escapes the backslash.

I think the problem is that [`enquote`](https://github.com/scalameta/scalameta/blob/master/scalameta/common/shared/src/main/scala/scala/meta/internal/prettyprinters/Enquote.scala) escapes backslashes but also doesn't escape all characters not allowed in Scala source. I'm not sure if it's better to escape more characters in `enquote` or to allow users to include their own arbitrary Unicode escapes.

Scalameta version: 4.3.20

olafurpg:
Previously, the pretty printer incorrectly escaped characters in
string literals like `Lit.String("xy\u001az")`. Now, we format
these characters with the correct unicode escapes.

### DIFF
--- a/scalameta/common/shared/src/main/scala/scala/meta/internal/prettyprinters/Enquote.scala
+++ b/scalameta/common/shared/src/main/scala/scala/meta/internal/prettyprinters/Enquote.scala
@@ -25,7 +25,9 @@ object enquote {
         case '\'' if style eq SingleQuotes =>
           sb.append("\\\'")
         case c =>
-          sb.append(c)
+          val isNonReadableAscii = c < ' ' || c > '~'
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM!

Issue #2116, opened 2020-09-24
Wrong occurence when using method with default arguments in child class


tgodzik:
```scala
import scala.concurrent.ExecutionContext
abstract class A {

  def check(
      includeDocs: Boolean = false,
      includeCommitCharacter: Boolean = false
  )(implicit loc: ExecutionContext): Unit = {}
}

class B extends A {

  implicit val ec = scala.concurrent.ExecutionContext.global

  check(
    includeCommitCharacter = true
  )

}
```

check function in this case references the class B.

PR #2117, opened 2020-09-24
ThisType should only occur when actually in the code

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/2116

Looks like we are getting ThisType sometimes on class members, but I am no 100% sure where is that coming from or sure about the solution. IS it possible to better check if ThisType is not actually synthetic and actually exits in the code?

### DIFF
--- a/tests/jvm/src/test/resources/example/Issue2116.scala
+++ b/tests/jvm/src/test/resources/example/Issue2116.scala
@@ -1,0 +1,22 @@
+// See https://github.com/scalameta/scalameta/issues/2116
+package example
+
+import scala.concurrent.ExecutionContext/*=>scala.concurrent.ExecutionContext.*//*=>scala.concurrent.ExecutionContext#*/
+
+abstract class Issue2116/*<=example.Issue2116#*/ {
+
+  def check/*<=example.Issue2116#check().*/(
+      includeDocs/*<=example.Issue2116#check().(includeDocs)*/: Boolean/*=>scala.Boolean#*/ = false,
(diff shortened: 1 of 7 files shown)


tests/jvm/src/test/resources/example/Issue2116.scala
```diff
@@ -0,0 +1,22 @@
...
+      includeCommitCharacter/*<=example.Issue2116#check().(includeCommitCharacter)*/: Boolean/*=>scala.Boolean#*/ = false
+  )(implicit loc/*<=example.Issue2116#check().(loc)*/: ExecutionContext/*=>scala.concurrent.ExecutionContext#*/): Unit/*=>scala.Unit#*/ = {}
+}
+
+class Issue2116_2/*<=example.Issue2116_2#*/ extends Issue2116/*=>example.Issue2116#*/ {
+
+  implicit val ec/*<=example.Issue2116_2#ec.*/ = scala.concurrent.ExecutionContext/*=>scala.concurrent.ExecutionContext.*/.global/*=>scala.concurrent.ExecutionContext.global().*/
+
+  check/*=>example.Issue2116#check().*/(
+    includeCommitCharacter/*=>example.Issue2116#check().(includeCommitCharacter)*/ = true
```
kpbochenek, line 19: previously it linked to: `example.Issue2116_2#check().(includeCommitCharacter)` ?

tgodzik: Previously, the line above was:
`check/*=>example.Issue2116#*/(`

kpbochenek (approved):
:+1:

PR #2125, opened 2020-10-01
Fix #1119 (Type.Param with Name.Anonymous is not supported in Dotty)

### LINKED ISSUE #1119: Type.Param with Name.Anonymous is not supported in Dotty
```scala
trait F[_]
```

results in
```scala
-- [E039] Syntax Error: foo.scala:1:8 ------------------------------------------
1 |trait F[_]
  |        ^
  |        an identifier expected, but '_' found

Explanation
===========

If you necessarily want to use '_' as identifier, you may put it in backticks.

```

kpbochenek:
Fix "Type.Param with Name.Anonymous is not supported in Dotty"

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -3296,13 +3296,21 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
   def entrypointTermParam(): Term.Param =
     param(ownerIsCase = false, ownerIsType = true, isImplicit = false, isUsing = false)
 
-  def typeParamClauseOpt(ownerIsType: Boolean, ctxBoundsAllowed: Boolean): List[Type.Param] = {
+  def typeParamClauseOpt(
+      ownerIsType: Boolean,
+      ctxBoundsAllowed: Boolean,
+      allowUnderscore: Boolean = true
+  ): List[Type.Param] = {
     newLineOptWhenFollowedBy[LeftBracket]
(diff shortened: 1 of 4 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3800,7 +3831,11 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
 
   def enumSingleCaseDef(mods: List[Mod]): Defn.EnumCase = {
     val name = termName()
-    val tparams = typeParamClauseOpt(ownerIsType = true, ctxBoundsAllowed = true)
+    val tparams = typeParamClauseOpt(
+      ownerIsType = true,
+      ctxBoundsAllowed = true,
+      allowUnderscore = dialect.allowTypeParamUnderscore
```
tgodzik, line 3837: Why not have it in one place only? Why do we need to forward it through all the functions?

tgodzik (approved):
LGTM with just one small question. Thanks for yet another improvement!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kpbochenek: That's because it is not in all functions, even example in issue shows this:
`F[_]` <- should fail
`F[X[_]]` <- valid
and this function is called recursively.

tgodzik: Ok, thanks for the explanation!

PR #2126, opened 2020-10-02
Fix #494

### LINKED ISSUE #494: Parse repeated by-name parameters in Dotty dialect
Repeated by-name parameters:

``` scala
class Foo(bars: => Int*)
```

Reference: http://docs.scala-lang.org/sips/pending/repeated-byname.html

Scala meta has the concept of dialect for different Scala syntax (think sbt, scala or dotty). The dotty dialect does not yet support repeated by-name parameters.

A good place to start is https://github.com/scalameta/scalameta/blob/master/scalameta/parsers/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L2680.

kpbochenek:
fixes #494

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/VarargParameterSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/VarargParameterSuite.scala
@@ -95,17 +95,21 @@ class VarargParameterSuite extends ParseSuite {
     )
   }
 
-  test("error on vararg parameter by name") {
-    checkError(
-      "def obj(f: => Int*): Boolean = true",
-      "error: ) expected but identifier found"
-    )
-  }
-
   test("error on multiple parameters vararg not last") {
     checkError(
       "def obj(fa: Int*, fb: String): Boolean = true",
       "error: *-parameter must come last"
+    )
+  }
+
+  test("error on repeated byname parameter") {
+    checkError(
+      "def fx(x: => Int*): Int = 3",
+      ") expected but identifier found"
+    )
+    checkError(
+      "class Foo(bars: => Int*)",
+      ") expected but identifier found"
     )
   }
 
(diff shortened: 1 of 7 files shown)


tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MinorDottySuite.scala
```diff
@@ -372,4 +372,46 @@ class MinorDottySuite extends BaseDottySuite {
       )
     )
   }
+
+  test("error on repeated byname class parameter") {
```
tgodzik, line 376: This is not an error though?

tests/shared/src/test/scala/scala/meta/tests/parsers/VarargParameterSuite.scala
```diff
@@ -126,7 +130,7 @@ class VarargParameterSuite extends ParseSuite {
     val error = intercept[parsers.ParseException] {
       templStat(definition)
     }
-    assert(error.getMessage.contains(expected))
+    assert(clue(error).getMessage.contains(clue(expected)))
```
tgodzik, line 133: We shouldn't need to change this if it was working before.

tgodzik (changes requested):

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MinorDottySuite.scala
kpbochenek: right, changed :+1:

tests/shared/src/test/scala/scala/meta/tests/parsers/VarargParameterSuite.scala
kpbochenek: reverted

tgodzik (approved):
LGTM!

PR #2127, opened 2020-10-04
Dotty support for lazy abstract values

### LINKED ISSUE #1202: Dotty seems to support lazy abstract values
```scala
scalac> trait Foo {   protected[this] lazy val from: Int }
(console):1: lazy values may not be abstract
trait Foo {   protected[this] lazy val from: Int }
                                       ^
dotc> trait Foo {   protected[this] lazy val from: Int }
defined trait Foo
```

kpbochenek:
Fixes #1202


tgodzik (approved):
LGTM

PR #2128, opened 2020-10-05
Anonymous method with using

kpbochenek:
- bumping dotty hash in community tests
- handle `val fn = (using x: Int) => 1 + x`
- handle looking like old while yet with do block
- add tests for ? type `val x: List[?]` (implementation was already done)
-  add test for extension + using, this will require changing `Extension` as I did not include this initially, test ignored, will fix in next PR.


tgodzik (approved):
LGTM

PR #2133, opened 2020-10-07
Fix indent after arrow in braces

kpbochenek:
- fix handling comment after ColonEOL
- bump community-dotty to include 2 fixes
- handle `=>` in match/catch { } should not trigger Indentation.


tgodzik (approved):
LGTM

tgodzik:
Just run scalafmt formatting.

PR #2135, opened 2020-10-08
Allow match to chain with other operators

tgodzik:


tgodzik:
Closing for now, I need to rework the approach.

PR #2136, opened 2020-10-08
fix-new-catch-case-as-parameter

kpbochenek:
- try catch as parameter of method :)
- context functions (they are syntactic sugar for given/using that's why hidden under this flag)

ContextFunctions:
https://github.com/lampepfl/dotty/commit/576e535ac5314da253d2a0abb7887b1a02fd4ef9
https://dotty.epfl.ch/docs/reference/contextual/context-functions.html

Sooo unrewarding to implement this and only improve parsing dotty codebase with 1 more file ech

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -432,12 +432,6 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
     loop(-1, 0, Nil, -1, false)
     val underlying = parserTokens.result
 
-    // underlying.foreach { t =>
-    //   try {
-    //   println(s"TOKEN ${t.name} :: ${t.text}")
-    //   } catch { case e => println(s"TOKEN ERR ${t.name}") }
-    // }
-
     (Tokens(underlying, 0, underlying.length), parserTokenPositions.result)
   }
 
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -252,6 +252,13 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
         parserTokens += new Indentation.Outdent(curr.input, curr.dialect, curr.start, curr.end)
         parserTokenPositions += currPos
       }
+      def shouldCloseCaseOnNonCase(regions: List[SepRegion]): Boolean = {
+        val closeCaseOnNonCase = regions.headOption.exists {
+          case RegionIndent(indent, true) => indent == currentIndent
+          case _ => false
+        }
+        closeCaseOnNonCase && !curr.is[KwCase]
+      }
 
       lazy val currentIndent = countIndent(currPos)
       var shouldCloseIndent = false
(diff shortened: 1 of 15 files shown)


tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/ControlSyntaxSuite.scala
```diff
@@ -1246,6 +1246,58 @@ class ControlSyntaxSuite extends BaseDottySuite {
...
+      Defn.Def(
+        Nil,
+        Term.Name("fx"),
+        Nil,
+        Nil,
+        Some(Type.Name("String")),
+        Term.Block(
+          List(
+            Term.Match(Term.Name("x"), List(Case(Lit.Int(2), None, Lit.String("ERROR")))),
+            Term.EndMarker(Term.Name("match"))
```
tgodzik, line 1274: Any way we can say that it's the keyword actually? Reprinting it might add `` always

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -112,6 +112,18 @@ object Term {
...
+  @ast class ContextFunction(params: List[Term.Param], body: Term) extends Term {
+    checkFields(
+      params.forall(param =>
+        param.is[Term.Param.Quasi] ||
+          (param.name.is[scala.meta.Name.Anonymous] ==> param.default.isEmpty)
+      )
+    )
+    checkFields(
+      params.exists(_.is[Term.Param.Quasi]) ||
+        params.exists(_.mods.exists(_.is[Mod.Implicit])) ==> (params.length == 1)
```
tgodzik, line 124: Should it not be Mod.Using ?

tgodzik (changes requested):
Thanks for more fixes! We are getting closer to finishing!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -436,6 +432,12 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
     loop(-1, 0, Nil, -1, false)
     val underlying = parserTokens.result
 
+    // underlying.foreach { t =>
```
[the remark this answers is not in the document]
kpbochenek: adding, commenting, removing, full circle over and over again :sweat_smile:

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
kpbochenek: copy pasted it from `Term.Function` it should actually not have implicit/using mods at all as `?=>` implicitly says it is using (explicitly?)

kpbochenek: from docs:
`g(ExecutionContext ?=> f(3))  // is expanded to g((using ev: ExecutionContext) => f(3)(using ev))`

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/ControlSyntaxSuite.scala
kpbochenek: Yes good catch, it printed match surrounded with ``
The thing is it needs to handle normal name: `end myCusfomFunc` but also keywords `end if`
Now I just changed treeprinting to print back `ternname.value` because printing whole TermName added `` if it was keyword which is probably correct way in rest of the code.

One way would be to keep it as is.
Other way would be to simply keep raw `String` in EndMarker?
Third way would be to introduce sth. like `Either[Term.Name, Token]`

tgodzik: Let's leave it as is for now.

tgodzik (approved):
LGTM!

PR #2138, opened 2020-10-15
Allow `type T` inside block for Dotty

tgodzik:


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -4384,16 +4384,16 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
...
-      (defOrDclOrSecondaryCtor(mods) match {
-        case stat if stat.isBlockStat => stat
-        case other => syntaxError("is not a valid block statement", at = other)
-      })
-    else
-      (tmplDef(mods) match {
-        case stat if stat.isBlockStat => stat
-        case other => syntaxError("is not a valid block statement", at = other)
-      })
+      }) {
```
kpbochenek, line 4387: took me a minute to realize but very nice extraction of match to be used only once :+1:

kpbochenek (approved):
small and nice one, LGTM :+1:

PR #2143, opened 2020-10-19
Fix remaining issues in Dotty codebase

tgodzik:
- fix issue with operators with literal on the next line
- fix parsing of extension methods with a comment and no `:`

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -3713,7 +3713,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
     } else if (token.is[LeftBrace]) {
       inBraces(templateStats())
     } else {
-      newLineOpt()
+      newLinesOpt()
       List(nonLocalDefOrDcl())
     }
     val body: Stat = if (methodsAll.size == 1) methodsAll.head else Term.Block(methodsAll)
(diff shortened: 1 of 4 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -295,7 +295,9 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
           if (indentedRegion(sepRegionsParameter) && currentIndent >= 0) {
             var sepRegionsProcess = sepRegionsParameter
             while ((indentedRegion(sepRegionsProcess)
-                && sepRegionsProcess.head.indent > currentIndent) ||
+                && sepRegionsProcess.head.indent > currentIndent && !isLeadingInfixOperator(
```
kpbochenek, line 298: good catch! :+1:

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3710,7 +3713,8 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
     } else if (token.is[LeftBrace]) {
       inBraces(templateStats())
     } else {
-      newLineOpt()
+      while (token.is[LF] || token.is[LFLF])
```
kpbochenek, line 3716: tokenizer should never produce 2 consecutive LF / LFLF combinations:
2x LF => LFLF
2x LFLF => LFLF

`while` should not be needed, `newLinesOpt()` should be enough

kpbochenek (approved):
Nice work fixing those issues! :+1:

(one minor cosmetic sugestion - can be ignored :) )

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Changed and added more newlines, thanks!

Issue #2144, opened 2020-10-19
Missing named parameters symbols for constructor


tgodzik:
In this case:

```scala

package example
  object Example extends App {
    class Test(isRenameAwesome: Boolean)
    val x = new Test(isRenameAwesome = true)
}
```
we are missing the occurence for `<<isRenameAwesome>> = true`

tgodzik:
Connected to https://github.com/scalameta/metals/issues/1086

tgodzik:
Unfortunately this is not entirely fixed yet:
```
package example
object Issue2144/*<=example.Issue2144.*/ {
  class Test/*<=example.Issue2144.Test#*/(a/*<=example.Issue2144.Test#a.*/: Boolean/*=>scala.Boolean#*/, b/*<=example.Issue2144.Test#b.*/: Int/*=>scala.Int#*/ = 1, c/*<=example.Issue2144.Test#c.*/: Int/*=>scala.Int#*/ = 2)
  val x/*<=example.Issue2144.x.*/ = new Test/*=>example.Issue2144.Test#*/(a/*=>example.Issue2144.Test#`<init>`().(a)*/ = true, c/*=>example.Issue2144.Test#`<init>`().(c)*/ = 1)
}
```

`a/*<=example.Issue2144.Test#a.*/ ` doesn't seem correct as `a` is only a constructor argument. I need to dig in a bit more here.

tgodzik:
So I was thinking about changing the occurrence for the named parameter to a field in case it's a primary constructor. The change however was quite involved, so I decided against it. The current status is not really wrong, both `example.Issue2144.Test#a.` and `example.Issue2144.Test#<init>().(a).` make sense here.

PR #2145, opened 2020-10-19
Fix missing constructor named arguments

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/2144


kpbochenek (approved):
not sure what `metac-metacp.diff ` is, but other LGTM! :+1:

PR #2146, opened 2020-10-21
Fix if continuation as dot and using lambda

kpbochenek:
- Fix condition checking in if for "." to continue `if (cond).rest then ...`
- Fix lambda with unnamed using `val lbda = (using _: Context) => ...`
- Fix catch case, new-style catch-case supports only expr: `catch case ex => singleExprOnly()`


tgodzik (approved):
LGTM!

PR #2147, opened 2020-10-23
Add additional logic to only add indent on correct colons

tgodzik:
This removes colonEol token, since we can better determine it later while preparing tokens for parsing.

This might cause Mima errors, but will be neccessary for now.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -164,7 +164,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
         val scannerToken = scannerTokens(roughIndex)
         def exactMatch = scannerToken eq token
         def originMatch =
-          (token.is[LFLF] || token.is[Indentation.Outdent]) &&
+          (token.is[LFLF] || token.is[Indentation.Outdent] || token.is[Indentation.Indent]) &&
             scannerToken.start == token.start && scannerToken.end == token.end
         if (exactMatch || originMatch) roughIndex
         else lurk(roughIndex - 1)
@@ -180,6 +180,12 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
 
     def nextSafe: Token = scannerTokens.apply(Math.min(token.index + 1, scannerTokens.length - 1))
 
+    def nextEnd: Token = {
+      val next = nextSafe
+      if ((next.is[Whitespace] && !next.is[LineEnd]) || next.is[Comment]) next.nextEnd
+      else next
+    }
+
     def next: Token = {
       val next = nextSafe
       if (next.is[Whitespace] || next.is[Comment]) next.next
@@ -190,6 +196,10 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
       if (next.is[Space] || next.is[Tab] || next.is[Comment]) next.strictNext
       else next
     }
+  }
+
+  def isColonEol(token: Token) = {
+    dialect.allowSignificantIndentation && token.is[Colon] && token.nextEnd.is[LineEnd]
   }
 
   /* ------------- PARSER-SPECIFIC TOKENS -------------------------------------------- */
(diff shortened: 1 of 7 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -211,6 +221,44 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
   // NOTE: Special case for Enum region is needed because parsing of 'case' statement is done differently
   case object RegionEnumArtificialMark extends SepRegion
 
+  case class AllowColonEol() {
```
kpbochenek, line 224: `case class` => `class`

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -211,6 +221,44 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
...
+    private var unclosed = 0
+    def unclosedParen = unclosed
```
kpbochenek, line 226: `unclosedParensCount: Int` ?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -211,6 +221,44 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
...
+   */
+  def canBeFollowedByColonEol(
+      colonEolCanStartIndent: Option[AllowColonEol],
+      curr: Token
+  ) = {
+    if (dialect.allowSignificantIndentation)
+      colonEolCanStartIndent match {
+        case _ if curr.is[CanStartColonEol] =>
+          Some(AllowColonEol())
+        // This means we closed more parents than opened, which means we are outside of a region where colonEol is allows
```
kpbochenek, line 244: `.. is allowed`

kpbochenek: I think comment for particular case should be inside this case, maybe subjective, [suggestion]

e.g.:
https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/parsing/Scanners.scala#L194

kpbochenek (approved):
just cosmetic comments, LGTM!
And I understood what is going on so not that much complicated as initially thought :D 

For future:
It might be tricky to handle (https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html#variant-indentation-marker-) later but I agree ColonEol was not a way to go and current code can be later adjusted as we want without breaking changes :+1:

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -463,7 +523,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
   @tailrec
   private def isAheadNewLine(currentPosition: Int): Boolean = {
     val nextPos = currentPosition + 1
-    if (scannerTokens(currentPosition).is[ColonEol]) true
+    if (isColonEol(scannerTokens(currentPosition))) true
```
kpbochenek, line 526: I think this is no longer needed, as ColonEol was consuming LF, but not it is not so it should be emited correctly

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
```diff
@@ -194,16 +193,8 @@ class ScalametaTokenizer(input: Input, dialect: Dialect) {
       def nextToken() = legacyIndex += 1
       if (legacyIndex >= legacyTokens.length) return legacyIndex
 
-      if (dialect.allowSignificantIndentation && curr.token == COLON && aheadIsNewLine(
-          legacyIndex
-        )) {
-        curr.token = COLONEOL
-        emitToken()
-        legacyIndex += 2
```
kpbochenek, line 202: here we jumped +2 and consumed ':' and LF, but because we no longer do that, LF will be emitted correctly(related to comment above)

tgodzik:
> For future:
> It might be tricky to handle (https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html#variant-indentation-marker-) later but I agree ColonEol was not a way to go and current code can be later adjusted as we want without breaking changes

Och damn, I I think that would be even more problems to handle, but since flags are still experimental, let's leave it for now.  That would be bunch of other problems :fearful: .

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Removed! Seems to work now.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Changed to `is allowed`, but kept it here, I think it makes sense.

kpbochenek (approved):
LGTM! :+1:

PR #2153, opened 2020-10-27
Disallow `given as <ident>` syntax

tgodzik:
Based on https://github.com/lampepfl/dotty/pull/10089

### DIFF
--- a/community-test/src/test/scala/CommunityDottySuite.scala
+++ b/community-test/src/test/scala/CommunityDottySuite.scala
@@ -171,15 +171,16 @@ class CommunityDottySuite extends FunSuite {
   )
 
   final def scala3DocExclusionList = List(
+    // extension will most likely become a keyword
     // [scalameta] extension.getParameters.asScala(extension.get(MethodExtension).parametersListSizes(0))
     "main/scala/dotty/dokka/translators/ScalaSignatureProvider.scala",
-    // [scalameta] issues already fix with colonEol
-    "main/scala/example/level2/Documentation.scala",
-    // Below syntax is no longer valid
-    // val result: List[Int] = fmap[F = List, A = Int, B = Int](List(1,2,3))(i => i + 1)
+    // Below syntax is no longer valid:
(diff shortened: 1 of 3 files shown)


community-test/src/test/scala/CommunityDottySuite.scala
```diff
@@ -171,15 +171,16 @@ class CommunityDottySuite extends FunSuite {
...
-    // Below syntax is no longer valid
-    // val result: List[Int] = fmap[F = List, A = Int, B = Int](List(1,2,3))(i => i + 1)
+    // Below syntax is no longer valid:
+    // - val result: List[Int] = fmap[F = List, A = Int, B = Int](List(1,2,3))(i => i + 1)
     "main/scala/NamedTypeArguments.scala",
-    // def (s: String).doesntStartWithAnyOfThese(c: Char*) = c.forall(char => !s.startsWith(char.toString))
-    "test/scala/dotty/dokka/DottyTestRunner.scala"
+    // - def (s: String).doesntStartWithAnyOfThese(c: Char*) = c.forall(char => !s.startsWith(char.toString))
+    "test/scala/dotty/dokka/DottyTestRunner.scala",
+    // - given as Context = ctx
```
kpbochenek, line 182: changes were merged to dotty, bumping community-dotty should fix it

kpbochenek (approved):
LGTM! 
(if bumping dotty-codebase only fix this one exclusion we could include that change in this PR, if it adds other problems then what you prefer, can be handled another time)

community-test/src/test/scala/CommunityDottySuite.scala
tgodzik: It doesn't seem it was merged yet: https://github.com/lampepfl/dotty/pull/10081

tgodzik:
Also checked and bumping dotty codebase causes the new given imports to break too. I will merge this one and follow up with a bump later on.

community-test/src/test/scala/CommunityDottySuite.scala
kpbochenek: I referred to `lampepfl/dotty#10089` :) nvm, all good :+1:

PR #2159, opened 2020-11-09
Remove scala3doc community build - merged into dotty

kpbochenek:
- bump dotty hash commit to contain scala3doc
- remove scala3doc as separate community-build


tgodzik (approved):
LGTM

PR #2160, opened 2020-11-13
Add type match

tgodzik:
http://dotty.epfl.ch/docs/reference/new-types/match-types.html

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1423,33 +1423,31 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
           case RightArrow() => next(); Type.Function(List(t), typ())
           case ContextArrow() => next(); Type.ContextFunction(List(t), typ())
           case KwForsome() => next(); Type.Existential(t, existentialStats())
-          case KwMatch() if dialect.allowTypeMatch =>
-            next(); typeMatch(t)
+          case KwMatch() if dialect.allowTypeMatch => next(); Type.Match(t, typeCaseClauses())
           case _ => t
         }
       }
     }
 
-    def typeMatch(t: Type): Type.Match = autoPos {
+    def typeCaseClauses(): List[TypeCase] = {
+      def cases() = {
+        val allCases = new ListBuffer[TypeCase]
+        while (token.is[KwCase])
+          allCases += typeCaseClause()
+        allCases.toList
+      }
       if (token.is[LeftBrace]) {
         inBraces {
-          Type.Match(t, typeCaseClauses())
+          cases()
         }
       } else if (token.is[Indentation.Indent])
         indented {
-          Type.Match(t, typeCaseClauses())
+          cases()
         }
       else {
         syntaxError("Expected braces or indentation", at = token.pos)
       }
 
-    }
-
-    def typeCaseClauses(): List[TypeCase] = {
-      val allCases = new ListBuffer[TypeCase]
-      while (token.is[KwCase])
-        allCases += typeCaseClause()
-      allCases.toList
     }
 
     def typeCaseClause(): TypeCase = atPos(in.prevTokenPos, auto) {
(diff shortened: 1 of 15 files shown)


scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -373,7 +375,7 @@ object Defn {
       bounds: scala.meta.Type.Bounds,
       body: scala.meta.Type
   ) extends Defn with Member.Type {
-    checkFields(mods.exists(_.is[Mod.Opaque]))
+    // checkFields(mods.exists(_.is[Mod.Opaque]))
```
tgodzik, line 378: So, since type match can have type bounds, I thinking of renaming the class to something like `TypeAliasWithBounds`

I was thinking of adding an optional field to Type, but it's impossible to say if we should add `bounds`, which are also used in the body of Type.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3966,6 +3966,14 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
     }
   }
 
+  def indentedTypeOpt() = {
```
kpbochenek, line 3969: It suggests that indentation is always, but type is optional.
`def typeIndentedOpt` ?

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
```diff
@@ -0,0 +1,161 @@
...
+          Type
+            .Param(Nil, Type.Name("X"), Nil, Type.Bounds(None, Some(Type.Name("Tuple"))), Nil, Nil)
+        ),
+        Type.Match(
+          Type.Name("X"),
+          List(
+            TypeCase(
+              Type.Tuple(
+                List(Type.Name("x1"), Type.Placeholder(Type.Bounds(None, None)))
+              ), // TODO should be Pat.TypeVar and Wildcard()
```
kpbochenek, line 54: .

kpbochenek (changes requested):
Good work :+1: 

Just added a few minor comments, and also I think one exclusion can be removed from community-build :D

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
tgodzik: Forgot about that, I was thinking of adding additional nodes here, but might not actually make sense.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1416,25 +1416,65 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
...
-          else if (token.is[LeftBracket] && dialect.allowTypeLambdas) typeLambdaOrPoly()
-          else infixType(InfixMode.FirstOp)
+          if (token.is[LeftBracket] && dialect.allowTypeLambdas) typeLambdaOrPoly()
+          else infixTypeOrTuple()
 
         token match {
           case RightArrow() => next(); Type.Function(List(t), typ())
           case ContextArrow() => next(); Type.ContextFunction(List(t), typ())
           case KwForsome() => next(); Type.Existential(t, existentialStats())
+          case KwMatch() if dialect.allowTypeMatch =>
```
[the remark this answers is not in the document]

.dependebot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
```
tgodzik, line 1: Added similar to how @gabro updated some of the other repos.

gabro: cool, but beware: you have a typo in the file name :-)

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
tgodzik: Removed TODO, it doesn't make sens to add new trees here. Dotty syntax specifies it all as types not type pattern.

.dependebot.yml
tgodzik: :(

tgodzik: Fixed! Thanks for noticing!

kpbochenek (approved):
Great Work! :+1:

PR #2161, opened 2020-11-17
Use type in givens, which makes working on the tree easier

tgodzik:
Just updating this, since Type.Param is better used to statements like  `[A,B]` and I overdid it with Type.Param.


kpbochenek (approved):
LGTM :+1:

PR #2163, opened 2020-11-17
Move dependabot config file to right location

gabro:


gabro:
Merging since this seems to work ok in other repos

PR #2167, opened 2020-11-19
bump sbt 1.4.3

kpbochenek:
Supersedes: https://github.com/scalameta/scalameta/pull/2149

This will work? :crossed_fingers:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ repos.tar.gz*
 .settings
 .cache*
 .project
+.bsp/
 
 website/translated_docs
 website/build/
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.3.13
+sbt.version=1.4.3


tgodzik:
Seems there is an issue when compiling under 2.11.12 :/

I meant to report it, but haven't had the time.

kpbochenek:
Created issue in sbt, we have to start somewhere https://github.com/sbt/sbt/issues/6124

adpi2:
Surely something has changed in sbt causing this bug.

But I have found that there is 2 `SkipWindows` symbols defined in the same package:
- one `SkipWindows` object [here](https://github.com/scalameta/scalameta/blob/b286cdbdc4bca88bc81932efd29ef70c6f87f289/tests/jvm/src/main/scala/scala/meta/tests/SkipWindows.scala#L5)
- one `SkipWindows` val [here](https://github.com/scalameta/scalameta/blob/b286cdbdc4bca88bc81932efd29ef70c6f87f289/tests/jvm/src/test/scala/scala/meta/tests/package.scala#L5)

So I guess if we remove one of those symbols it will be easier for Zinc. I'll give a try.

adpi2:
New PR #2185

PR #2170, opened 2020-11-20
Fix error reporting of unclosed slice interpolation

kpbochenek:
Fixes https://github.com/scalameta/scalameta/issues/2169

### DIFF
--- a/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
+++ b/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
@@ -10,13 +10,12 @@ import scala.annotation.tailrec
 import scala.meta.inputs._
 import scala.meta.tokens._
 import scala.meta.tokenizers._
+import scala.meta.internal.tokenizers.ScalametaTokenizer.UnexpectedInputEndException
 
 class ScalametaTokenizer(input: Input, dialect: Dialect) {
   def tokenize(): Tokens = {
     input.tokenCache.getOrElseUpdate(dialect, uncachedTokenize())
   }
-
-  private final var interpolationStart = 0
 
   private def uncachedTokenize(): Tokens = {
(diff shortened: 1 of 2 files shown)


kpbochenek:
Error is thrown in `      def curr = legacyTokens(legacyIndex)`

I did not want to be too invasive, but it is actually used in multiple places, maybe better would be to check in `def curr` we are not outside of tokens array and throw more generic error "unclosed statement" or sth. similar?

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
```diff
@@ -324,6 +329,9 @@ class ScalametaTokenizer(input: Input, dialect: Dialect) {
     tokens.toArray(underlying)
     Tokens(underlying, 0, underlying.length)
   }
+
+  private def syntaxError(startPos: Int, endPos: Int, message: String): Nothing =
```
tgodzik, line 333: It's a bit weird to have a syntax error in the tokenizer, since it doesn't deal with syntax. This would still be better than the previous implementation, but I wonder if we could just skip the interpolation if it doesn't have an end? What do you think?

kpbochenek: as it is kind of messy recursive flow it is not easy to just simply stop. 
But if it is okay to skip not fully closed token, I propose v0.2 in my commit, WDYT?

kpbochenek: aaaaand we are hitting on JS https://github.com/scala/bug/issues/11920
Cannot throw AIOOB as it causes UndefinedBehaviour and you cannot catch it ...

kpbochenek: aaaaand worked around like a boss :sunglasses:

tgodzik (approved):
LGTM!

PR #2181, opened 2020-11-27
Fix regression with positions in blocks

tgodzik:
Because autoPos was now done inside `{}` this caused a regression in scalafmt and most likely parser positions changed.


kpbochenek (approved):
good catch! :+1:

PR #2183, opened 2020-11-30
Don't create block on indentation with a single statement

tgodzik:
After working a bit on scalafmt it seems that most problematic is the case with a single statement, which now is sometimes be treated as an indented block. So reformatting could sometimes cause a new block to be created even if there was none before.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2062,7 +2062,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
 
   private def exprMaybeIndented(): Term = {
     if (token.is[Indentation.Indent]) {
-      indented(block())
+      block()
     } else {
       expr()
     }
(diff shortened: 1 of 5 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2087,42 +2087,55 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
       atPos(t, auto)(Term.Match(t, inBracesOrNil(caseClauses())))
     }
   }
+
+  def ifClause(isInline: Boolean = false) = {
```
tgodzik, line 2091: This is moved out from `expr`

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2087,42 +2087,55 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
...
+    ifExpr
+  }
+
   // FIXME: when parsing `(2 + 3)`, do we want the ApplyInfix's position to include parentheses?
   // if yes, then nothing has to change here
   // if no, we need eschew autoPos here, because it forces those parentheses on the result of calling prefixExpr
   // see https://github.com/scalameta/scalameta/issues/1083 and https://github.com/scalameta/scalameta/issues/1223
   def expr(location: Location, allowRepeated: Boolean): Term =
     autoPos(token match {
+      case SkInline() if ahead(token.is[KwIf]) =>
```
tgodzik, line 2130: Turns out there was one more issue with ifs, so a much safer place to check it is here.

The problem was with
`def a = inline if cond then a`

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3825,14 +3844,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
...
       case KwIf() if mods.size == 1 && mods.head.is[Mod.Inline] =>
-        expr() match {
-          case ifExpr: Term.If =>
-            ifExpr.setMods(mods)
-            ifExpr
-          case expr => syntaxError("Unexpected expressions after inline modifier", at = expr.pos)
-        }
-      case KwIf() if mods.size > 0 =>
-        syntaxError("If can only contain inline modifier", at = token.pos)
+        ifClause(isInline = true)
```
tgodzik, line 3847: I removed the additional case with non-inline mods - it will be reported anyway as error.

kpbochenek (approved):
LGTM! :+1:

PR #2184, opened 2020-12-01
Fix PolyFunction with context param

adpi2:
Not sure if this `type F0 = [T] => List[T] ?=> Option[T]` will be valid in Scala 3.0.0 but it compiles in 3.0.0-M1.

So here is the fix for that.


tgodzik (approved):
LGTM! Thanks for the fix!

PR #2185, opened 2020-12-01
Bumpt to sbt 1.4.4

adpi2:


adpi2:
Following the discussion in #2167 I removed the duplicated `SkipWindows` symbol to make Zinc happier.

tgodzik (approved):
OMG! Good catch! Thanks a lot!

PR #2186, opened 2020-12-01
Fix newly found Scala 3 parsing issues

tgodzik:
This includes:
- bump community test commits
- fix issue with interpolation being on next line with an operator:
```scala
val a =
  ""
  + s""
```
- make sure poly functions can also exist in TypeApply
- update Dotty dialect to reflect current status
- migrate relevant tests from DottySuite


kpbochenek (approved):
LGTM! :+1:

PR #2187, opened 2020-12-02
Add given pattern and add all new Scala 3 structures to docs

tgodzik:


scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -261,7 +261,10 @@ object Pat {
     checkFields(lhs.is[Pat.Wildcard] || lhs.is[Pat.Var] || lhs.is[Pat.Quasi] || lhs.is[Pat.Bind])
     checkFields(!rhs.is[Type.Var] && !rhs.is[Type.Placeholder])
   }
-  @ast class Macro(body: Term) extends Pat
+  @ast class Macro(body: Term) extends Pat {
+    checkFields(body.is[Term.QuotedMacroExpr])
```
tgodzik, line 265: I also added the check to make sure that no one will try to add a different term here. We can always change the check, but might be harder to fix if we change the type.

kpbochenek (approved):
LGTM :+1:

PR #2189, opened 2020-12-03
Fix product for binaryCompat fields and add additional apply methods

tgodzik:


scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
```diff
@@ -120,7 +120,11 @@ class AstNamerMacros(val c: Context) extends AstReflection with CommonNamerMacro
         val binaryCompatStats = binaryCompatVars.flatMap { case vr: ValDef =>
           List(
             q"def ${getterName(vr)} = this.${vr.name} ",
-            q"def ${setterName(vr)}(${vr.name} : ${vr.tpt}) = this.${vr.name} = ${vr.name} ",
+            q"""def ${setterName(vr)}(${vr.name} : ${vr.tpt}) = {
+                   val node = this
+                   $CommonTyperMacrosModule.storeField(this.${vr.name}, ${vr.name}, ${vr.name
```
tgodzik, line 125: This actually copies the entire subtree for the field, which is normally done for other fields, but was not for binaryCompat ones.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
```diff
@@ -263,18 +266,30 @@ class AstNamerMacros(val c: Context) extends AstReflection with CommonNamerMacro
...
+        stats1 += q"override def productArity: $IntClass = ${productParamss.head.length + binaryCompatNum}"
         val pelClauses = ListBuffer[Tree]()
+        val fieldsNum = productParamss.head.length
         pelClauses ++= 0
-          .to(productParamss.head.length - 1)
+          .to(fieldsNum - 1)
           .map(i => cq"$i => this.${productParamss.head(i).name}")
+
+        // generate product elements for @binaryCompat fields
+        pelClauses ++= fieldsNum
```
tgodzik, line 281: This is actual the fix for product, which is used in structure or in pprint.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
```diff
@@ -336,7 +350,39 @@ class AstNamerMacros(val c: Context) extends AstReflection with CommonNamerMacro
...
+        """
+
+        // step 13a: generate additional binary compat Companion.apply
+        // generate all possible applies so with field A, B and additional binary compat ones C and D, we generate:
+        // apply(C, D, A), apply(C, D, B), apply(C, D, A, B)
+        val sortedBinaryCompatName = binaryCompatVars
+          .collect { case vr: ValDef => vr }
+          .sortBy(_.name.toString())
+        1.until(binaryCompatNum + 1).foreach { size =>
+          sortedBinaryCompatName.combinations(size).foreach { fields =>
```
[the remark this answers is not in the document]
olafurpg: Will this compile successfully if you have two or more binary compact fields with the same type?

```scala
@ast class Template(a: A, c: C) {
  @bincompat var b1: B
  @bincompat var b2: B
  @bincompat var b3: B
}
// generates
def apply(a: A, c: C, b1: B)
def apply(a: A, c: C, b2: B) // Error: same erasure as previous method
def apply(a: A, c: C, b3: B) // Error: same erasure as previous method
def apply(a: A, c: C, b1: B, b2: B)
...
```

To support adding new bincompat fields without breaking changes, I think we need to generate a new `apply` method for every bincompat field.

```
def apply(a: A, c: C, b1: B)
def apply(a: A, c: C, b1: B, b2: B)
def apply(a: A, c: C, b1: B, b2: B, b3: B)
```

This requires the ordering of the `apply` methods to match the declaration order. Otherwise, we would have to make sure to name new `@bincompat` fields to appear last in lexicographical order.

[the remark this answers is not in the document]
olafurpg: Every new bincompat field should have the equivalent of a `@since` annotation from https://github.com/alexarchambault/data-class#adding-fields

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
```diff
@@ -120,7 +120,11 @@ class AstNamerMacros(val c: Context) extends AstReflection with CommonNamerMacro
         val binaryCompatStats = binaryCompatVars.flatMap { case vr: ValDef =>
           List(
             q"def ${getterName(vr)} = this.${vr.name} ",
-            q"def ${setterName(vr)}(${vr.name} : ${vr.tpt}) = this.${vr.name} = ${vr.name} ",
+            q"""def ${setterName(vr)}(${vr.name} : ${vr.tpt}) = {
+                   val node = this
+                   $CommonTyperMacrosModule.storeField(this.${vr.name}, ${vr.name}, ${vr.name
+              .toString()})
```
olafurpg, line 126: Nit, I normally break up the quasiquote into multiple lines to avoid the line break here from Scalafmt.

olafurpg (changes requested):
This looks like a nice ergonomic improvement for bincompat fields. Thank you for taking binary compatibility so seriously, it's annoying that we have to do this ceremony to add new fields but I think downstream Scalameta users will appreciate the stability of the library.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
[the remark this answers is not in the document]
tgodzik: Ach, you're right. I will need to just generate a single new apply for each new added field. Should be a small change. I went overboard with the combinations.

Not sure if we need `since` here, it's needed when we add multiple fields at the same time and we want a single new apply. I think we should be fine with adding one new apply for each new compat field.  What do you think?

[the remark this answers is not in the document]
tgodzik: Ach no, you are right. We need something like since.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
```diff
@@ -336,7 +350,39 @@ class AstNamerMacros(val c: Context) extends AstReflection with CommonNamerMacro
...
-      """
+        """
+
+        // step 13a: generate additional binary compat Companion.apply
+        // generate all possible applies so with field A, B and additional binary compat ones C and D, we generate:
+        // apply(C, D, A), apply(C, D, B), apply(C, D, A, B)
+        val sortedBinaryCompatName = binaryCompatVars
+          .collect { case vr: ValDef => vr }
+          .sortBy(_.name.toString())
+        1.until(binaryCompatNum + 1).foreach { size =>
```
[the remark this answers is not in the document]
tgodzik: Fixed! Thanks!

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
tgodzik: I extracted the values and line break is no longer there.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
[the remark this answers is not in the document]
tgodzik: @olafurpg I added `since` field to `@binaryCompatField` and order the fields using it. I also switched to just generating one new apply per field.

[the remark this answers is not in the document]
olafurpg: If you implemented `@since` then that's great! I was thinking that we would infer `@since` for every bincompat field. I suspect it will be rare that we add two bincompat fields and don't want to add two new apply methods.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
```diff
@@ -336,7 +350,54 @@ class AstNamerMacros(val c: Context) extends AstReflection with CommonNamerMacro
...
-      """
+        """
+
+        // step 13a: generate additional binary compat Companion.apply
+        // generate new applies for each new field added
+        // with field A, B and additional binary compat ones C, D and E, we generate:
+        // apply(A, B, C), apply(A, B, C, D), apply(A, B, C, D, E)
+        val sortedBinaryCompatName = binaryCompatVars
+          .collect { case vr: ValDef => vr }
+          .sortBy { vr =>
```
olafurpg, line 361: Performance nitpick: the lambda to `sortBy` gets evaluated many times per element. You can cache the results by doing a method like here https://github.com/twitter/bazel-multideps/blob/db19c92d13bce01bcd0189309564c79213995917/multideps/src/main/scala/multideps/diagnostics/MultidepsEnrichments.scala#L58

olafurpg: The size of the list will always be small here so I suspect performance is not a concern. Just an FYI, feel free to ignore

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
```diff
@@ -336,7 +350,54 @@ class AstNamerMacros(val c: Context) extends AstReflection with CommonNamerMacro
...
+            try {
+              vr.mods.annotations.collectFirst { case q"new binaryCompatField($version)" =>
+                val since = version.toString().stripPrefix("\"").stripSuffix("\"")
+                val versions = since.split("\\.").map(_.toInt)
+                (versions(0), versions(1), versions(2))
```
olafurpg, line 366: Does this provide a stable sort when we have two bincompat fields with the same name?

olafurpg (approved):
LGTM 👍 Thank you!

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
tgodzik: Currently it's mostly only a single element list, so it should not be an issue. But I haven't really thought about the performance here. 

I think I will leave as is, because I am afraid of really adding any more complexity here.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
tgodzik: `sortBy` is stable judging by the docs so this should also be stable. So if we ever need to add two elements at the same time it should work ok.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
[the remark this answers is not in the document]
tgodzik: It's not really `@since`. It will still add two new applies. The only problem I see is if we add two new fields at the same version in different PRs and add the new one before the previous one.  Although I expect `@since` doesn't work in that case either :sweat_smile:

PR #2190, opened 2020-12-07
Introduce `ExportGiven`, remove `Export.given: Boolean`.

olafurpg:
Previously, the `scala.meta.Export` tree had a `given: Boolean` field.
This goes against a design principle in the Scalameta AST, which tries
to represent as much as possible with syntax. There's no other AST node that
has a boolean field to distinguish between two cases. This commit
introduces a new `ExportGiven` class to represent `export given a.b`
syntax.

This is a binary breaking change so the next release should strictly
speaking be Scalameta v5. However, I think it's fair to change the
compatibility policy to state that Dotty-specific tree nodes are
experimental for time being and subject to breaking changes. This policy
buys us time to refine the Scala 3 AST nodes so that they are well
designed and consistent with the rest of the Scalameta AST nodes.
Normal Scalameta users should not be affected by this change as long as
they don't use Scalameta to parse Scala 3 programs.

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -4,4 +4,4 @@
 -XX:ReservedCodeCacheSize=1024m
 -XX:+TieredCompilation
 -XX:+CMSClassUnloadingEnabled
-
+-Dsbt.server.autostart=false
--- a/project/Mima.scala
+++ b/project/Mima.scala
@@ -26,5 +26,11 @@ object Mima {
       true
   }
 
-  val apiCompatibilityExceptions: Seq[ProblemFilter] = Seq()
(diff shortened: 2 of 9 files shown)


tgodzik (approved):
LGTM!

tgodzik:
Typically RefelctionSuite and SurfaceSuite is failing:

```
[error] Failed: Total 2435, Failed 2, Errors 0, Passed 2424, Skipped 9
[error] Failed tests:
[error] 	scala.meta.tests.trees.ReflectionSuite
[error] 	scala.meta.tests.api.SurfaceSuite
```

They are only run under 2.12, but we default to 2.13, maybe it would make sense to move those tests to 2.13?

PR #2192, opened 2020-12-08
Add variable to track if we are inside quotes/splices 

### LINKED ISSUE #2191: Scala 3: Dollar identifier is incorrectly tokenized as `MacroSplicedIndent`
Steps to reproduce:
```
val dotty = {
  implicit val D = dialects.Dotty
  "map($ => $.a)".tokenize.get.map { t =>
    t.productPrefix -> t.syntax
  }
}

val scala2 = {
  implicit val D = dialects.Scala213
  "map($ => $.a)".tokenize.get.map { t =>
    t.productPrefix -> t.syntax
  }
}

```

In the Dotty example, `$` is tokenized as `MacroSplicedIndent` while it's `Ident` in the Scala 2 example. I expected the tokenizer to emit `Ident` in both examples. cc/ @kpbochenek

tgodzik:
In order to figure out if something is a spliced or quoted variable we need to know what is the current context. I haven't found a better way to actually figure out if something is an identifier or spliced/quoted identifier. Anything else would require us to add an additional parameter to all of the methods and I figured that since we are already in a mutable context this is a much easier solution.

We could possibly remove `MacroSplicedIdent`, since we can just check if identifier starts with $ inside the context, but it does work together with `MacroQuotedIdent` - kind of looks symetrical. But it does mean we have to handle `MacroSplicedIdent` specially.

Fix #2191

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
@@ -78,7 +78,8 @@ final class Dialect private (
     val allowExtensionMethods: Boolean,
     // Open modifier for classes introduced in dotty
     val allowOpenClass: Boolean,
-    // Whitebox macros introduced in dotty (splices/quotes)
+    // Scala 3 splices/quotes
+    @deprecated("Replaced with allowSpliceAndQuote", "4.4.1")
     val allowWhiteboxMacro: Boolean,
     // Top level statements introduced in dotty.
     // differs from ToplevelTerms because here you can define packages
@@ -122,7 +123,11 @@ final class Dialect private (
     // Dotty allows `match` on type
     val allowTypeMatch: Boolean,
     // Dotty allows to define types and methods with an `infix` soft keyword modifier
-    val allowInfixMods: Boolean
+    val allowInfixMods: Boolean,
+    // Scala 3 splices/quotes
+    val allowSpliceAndQuote: Boolean,
+    // Scala 3 disallowed symbol literals
+    val allowSymbolLiterals: Boolean
 ) extends Product with Serializable {
 
   // NOTE(olafur) checklist for adding a new dialect field in a binary compatible way:
@@ -208,7 +213,9 @@ final class Dialect private (
       allowPolymorphicFunctions = false,
       allowMatchAsOperator = false,
       allowTypeMatch = false,
-      allowInfixMods = false
+      allowInfixMods = false,
+      allowSpliceAndQuote = false,
+      allowSymbolLiterals = true
       // NOTE(olafur): declare the default value for new fields above this comment.
     )
   }
@@ -236,7 +243,7 @@ final class Dialect private (
   def withAllowImplicitByNameParameters(newValue: Boolean): Dialect = {
     privateCopy(allowImplicitByNameParameters = newValue)
   }
-  @deprecated("Implicit functions are not supported in any dialect")
+  @deprecated("Implicit functions are not supported in any dialect", "4.4.1")
   def withAllowImplicitFunctionTypes(newValue: Boolean): Dialect = {
     privateCopy(allowImplicitFunctionTypes = newValue)
   }
@@ -306,6 +313,7 @@ final class Dialect private (
   def withAllowOpenClass(newValue: Boolean): Dialect = {
     privateCopy(allowOpenClass = newValue)
   }
+  @deprecated("Replaced by withAllowSplicesAndQuotes", "4.4.1")
   def withAllowWhiteboxMacro(newValue: Boolean): Dialect = {
     privateCopy(allowWhiteboxMacro = newValue)
   }
@@ -322,7 +330,7 @@ final class Dialect private (
   def withAllowInterpolationDolarQuoteEscape(newValue: Boolean): Dialect = {
     privateCopy(allowInterpolationDolarQuoteEscape = newValue)
   }
-  @deprecated("Super traits are not supported in any dialect")
+  @deprecated("Super traits are not supported in any dialect", "4.4.1")
   def withAllowSuperTrait(newValue: Boolean): Dialect = {
     privateCopy(allowSuperTrait = newValue)
   }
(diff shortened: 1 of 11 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1755,18 +1762,24 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
...
+      res
+    }
     token match {
       case Ident(value) =>
-        val name = value.stripPrefix("`").stripSuffix("`")
-        val res = atPos(in.tokenPos, in.tokenPos)(ctor(name))
-        if (advance) next()
-        res
+        nameOfIdent(value)
+      case MacroSplicedIdent(value) =>
```
tgodzik, line 1775: If we remove `MacroSplicedIdent` this part will not be needed.

tgodzik (commented):
Honestly, I am not super sure about this solution, but I haven't been super sure about a lot of things when it comes to adjusting the parser to fit Scala 3.

tgodzik:
@olafurpg I removed both `MacroSplicedIdent` and `MacroQuotesIdent` tokens. It seems to be working ok now. I also deprecated symbol literals, so they can be tokenized, but will be rejected later by the parser (we could retain MacroQuoteIdent here, but it seems to work similar to how Dotty parser now). They will only work if used in the macro context. I also renamed the `allowWhiteboxMacros` to `allowSplicesAndQuotes`.

Let me know what you think!

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -122,7 +123,11 @@ final class Dialect private (
     // Dotty allows `match` on type
     val allowTypeMatch: Boolean,
     // Dotty allows to define types and methods with an `infix` soft keyword modifier
-    val allowInfixMods: Boolean
+    val allowInfixMods: Boolean,
+    // Scala 3 splices/quotes
+    val allowSplicesAndQuotes: Boolean,
```
olafurpg, line 128:
```suggestion
    val allowSpliceAndQuote: Boolean,
```

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -78,7 +78,8 @@ final class Dialect private (
     val allowExtensionMethods: Boolean,
     // Open modifier for classes introduced in dotty
     val allowOpenClass: Boolean,
-    // Whitebox macros introduced in dotty (splices/quotes)
+    // Scala 3 splices/quotes
+    @deprecated("Replaced with allowSplicesAndQuotes", "4.4.1")
     val allowWhiteboxMacro: Boolean,
```
olafurpg, line 83: I think you can turn this into `def allowWhiteboxMacro = allowSpliceAndQuote`

olafurpg (changes requested):
A few minor comments, otherwise this looks good

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -375,6 +383,12 @@ final class Dialect private (
   def withAllowInfixMods(newValue: Boolean): Dialect = {
     privateCopy(allowInfixMods = newValue)
   }
+  def withAllowSplicesAndQuotes(newValue: Boolean): Dialect = {
```
olafurpg, line 386:
```suggestion
  def withAllowSpliceAndQuote(newValue: Boolean): Dialect = {
```

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MacroSuite.scala
```diff
@@ -224,6 +219,38 @@ class MacroSuite extends BaseDottySuite {
...
+          Term.Function(
+            List(Term.Param(Nil, Term.Name("$d"), None, None)),
+            Term.Select(Term.Name("$d"), Term.Name("a"))
+          )
+        )
+      )
+    )
+  }
+
+  test("non-macro-dolar-type") {
```
olafurpg, line 237:
```suggestion
  test("non-macro-dollar-type") {
```

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MacroSuite.scala
```diff
@@ -224,6 +219,38 @@ class MacroSuite extends BaseDottySuite {
     )
   }
 
+  test("non-macro-dolar-ident") {
```
olafurpg, line 222:
```suggestion
  test("non-macro-dollar-ident") {
```

olafurpg (approved):
Some minor comments, feel free to merge after changing 👍

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
tgodzik: Fixed!

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
tgodzik: `allowWhiteboxMacro` no longer is used, so I think there is no need to do it.

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
tgodzik: Fixed!

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MacroSuite.scala
tgodzik: Fixed

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MacroSuite.scala
tgodzik: Fixed

PR #2193, opened 2020-12-09
Move soft keyword definitions outside the parser

tgodzik:
Based on https://github.com/scalameta/scalafmt/pull/2284#discussion_r538886558

This way they can be reused inside scalafmt.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -33,11 +33,9 @@ import scala.util.Try
 import scala.util.Success
 import scala.util.Failure
 
-class ScalametaParser(input: Input, dialect: Dialect) { parser =>
+class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   require(Set("", EOL).contains(dialect.toplevelSeparator))
-  implicit val currentDialect: Dialect = dialect
-  val softKeywords = new SoftKeywords(currentDialect)
-  import softKeywords._
+  private val soft = new SoftKeywords(dialect)
 
   /* ------------- PARSER ENTRY POINTS -------------------------------------------- */
 
@@ -811,12 +809,8 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
 
   /* -------------- TOKEN CLASSES ------------------------------------------- */
 
-  def isIdentAnd(pred: String => Boolean): Boolean = isIdentAnd(token, pred)
+  def isIdentAnd(pred: String => Boolean): Boolean = ScalametaParser.isIdentAnd(token, pred)
 
-  def isIdentAnd(token: Token, pred: String => Boolean): Boolean = token match {
-    case Ident(value) if pred(value.stripPrefix("`").stripSuffix("`")) => true
-    case _ => false
-  }
   def isUnaryOp: Boolean = isIdentAnd(name => Term.Name(name).isUnaryOp)
   def isIdentExcept(except: String) = isIdentAnd(_ != except)
   def isIdentOf(name: String) = isIdentAnd(_ == name)
@@ -824,9 +818,6 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
   def isStar: Boolean = isIdentOf("*")
   def isBar: Boolean = isIdentOf("|")
   def isAmpersand: Boolean = isIdentOf("&")
-
-  def isSoftKw(token: Token, name: String): Boolean = isIdentAnd(token, _ == name)
-
   def isColonWildcardStar: Boolean = token.is[Colon] && ahead(token.is[Underscore] && ahead(isStar))
   def isSpliceFollowedBy(check: => Boolean): Boolean =
     token.is[Ellipsis] && ahead(token.is[Unquote] && ahead(token.is[Ident] || check))
@@ -903,7 +894,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
   }
 
   private def inlineDefOrOpaque(token: Token): Boolean = {
-    (token.is[SkInline] || token.is[SkOpaque] || token.is[SkInfix]) &&
+    (token.is[soft.KwInline] || token.is[soft.KwOpaque] || token.is[soft.KwInfix]) &&
     (DclIntro.unapply(token.next) || Modifier.unapply(token.next))
   }
 
@@ -951,8 +942,8 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
       token.is[KwSealed] || token.is[KwImplicit] ||
       token.is[KwLazy] || token.is[KwPrivate] ||
       token.is[KwProtected] || token.is[KwOverride] ||
-      token.is[SkOpaque] || token.is[SkOpen] || token.is[SkTransparent] ||
-      token.is[SkInfix] || token.is[SkInline]
+      token.is[soft.KwOpaque] || token.is[soft.KwOpen] || token.is[soft.KwTransparent] ||
+      token.is[soft.KwInfix] || token.is[soft.KwInline]
     }
   }
 
@@ -964,13 +955,14 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
       )
 
     def unapply(token: Token): Boolean =
-      token.is[SkInline] && noIdentAhead()
+      token.is[soft.KwInline] && noIdentAhead()
   }
 
   @classifier
   trait NonParamsModifier {
     def unapply(token: Token): Boolean = {
-      token.is[SkOpen] || token.is[SkOpaque] || token.is[SkTransparent] || token.is[SkInfix]
+      token.is[soft.KwOpen] || token.is[soft.KwOpaque] || token.is[soft.KwTransparent] || token
+        .is[soft.KwInfix]
     }
   }
 
@@ -978,7 +970,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
   trait NonlocalModifier {
     def unapply(token: Token): Boolean = {
       token.is[KwPrivate] || token.is[KwProtected] ||
-      token.is[KwOverride] || token.is[SkOpen]
+      token.is[KwOverride] || token.is[soft.KwOpen]
     }
   }
 
@@ -1169,7 +1161,7 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
                 allowPatUnquotes = false,
                 allowMultilinePrograms = true
               )
-              val unquoteParser = new ScalametaParser(unquoteInput, unquoteDialect)
+              val unquoteParser = new ScalametaParser(unquoteInput)(unquoteDialect)
               if (dialect.allowTermUnquotes) unquoteParser.parseUnquoteTerm()
               else if (dialect.allowPatUnquotes) unquoteParser.parseUnquotePat()
               else unreachable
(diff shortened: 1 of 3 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
```diff
@@ -0,0 +1,87 @@
...
+import scala.meta.internal.classifiers.classifier
+
+class SoftKeywords(dialect: Dialect) {
+
+  def isIdentAnd(token: Token, pred: String => Boolean): Boolean = token match {
+    case Ident(value) if pred(value.stripPrefix("`").stripSuffix("`")) => true
+    case _ => false
+  }
+
+  def isSoftKw(token: Token, name: String): Boolean = isIdentAnd(token, _ == name)
```
olafurpg, line 15: Can this be `private`?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
```diff
@@ -0,0 +1,87 @@
+package scala.meta.internal.parsers
+
+import scala.meta.Dialect
+import scala.meta.tokens.Token
+import scala.meta.tokens.Token.Ident
+import scala.meta.internal.classifiers.classifier
+
+class SoftKeywords(dialect: Dialect) {
+
+  def isIdentAnd(token: Token, pred: String => Boolean): Boolean = token match {
```
olafurpg, line 10: Make private?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -837,77 +840,6 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
...
-  }
-
-  @classifier
-  trait SkInfix {
-    val name = "infix"
-    def unapply(token: Token): Boolean = {
-      isSoftKw(token, name) && dialect.allowInfixMods
-    }
-  }
-
```
kitbellew, line 910: why not extract all the other classifiers, too? lots of them are obviously useful to like whitespace, trivia etc.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
```diff
@@ -0,0 +1,87 @@
...
+    case Ident(value) if pred(value.stripPrefix("`").stripSuffix("`")) => true
+    case _ => false
+  }
+
+  def isSoftKw(token: Token, name: String): Boolean = isIdentAnd(token, _ == name)
+
+  @classifier
+  trait SkAs {
```
olafurpg, line 18: Nitpick: I'm not a fan of the "Sk" prefix. Instead of using wildcard imports, how do you like this syntax?
```suggestion
class SoftKeywords {
  trait KwAs { ... }
}

val soft = new SoftKeywords()
x match {
  case soft.KwAs
```

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
```diff
@@ -0,0 +1,87 @@
...
+    val name = "as"
+    def unapply(token: Token): Boolean = {
+      isSoftKw(token, name) && dialect.allowAsPatternBinding
```
olafurpg, line 21: Nit: can we move `dialect.allowAsPatternBinding` before the `isSoftKw`? The `isSoftKw` test is a lot more expensive and will be `false` in most cases

olafurpg (changes requested):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
tgodzik: Sure! Fun fact, this was added with M1 and will actually be removed with M3.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
```diff
@@ -0,0 +1,87 @@
...
+
+import scala.meta.Dialect
+import scala.meta.tokens.Token
+import scala.meta.tokens.Token.Ident
+import scala.meta.internal.classifiers.classifier
+
+class SoftKeywords(dialect: Dialect) {
+
+  def isIdentAnd(token: Token, pred: String => Boolean): Boolean = token match {
+    case Ident(value) if pred(value.stripPrefix("`").stripSuffix("`")) => true
```
kitbellew, line 11: very optional:
```
  case Ident(value) => pred(value.stripPrefix("`").stripSuffix("`"))
```

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -837,77 +840,6 @@ class ScalametaParser(input: Input, dialect: Dialect) { parser =>
       (isCaseClassOrObject || (token.is[KwCase] && token.next.is[Ident] && dialect.allowEnums))
   }
 
-  @classifier
```
kitbellew, line 840: (stupid github only allows commenting on red/green lines) should we remove `isSoftKw` above?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
kitbellew: actually, this duplicates the method in `ScalametaParser` above. should we do this instead:

```
private[parsers] object ScalametaParser {
  def isIdentAnd(token: Token, pred: String => Boolean): Boolean = token match {
    case Ident(value) => pred(value.stripPrefix("`").stripSuffix("`"))
    case _ => false
  }
}
```
and then import it in both classes?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
kitbellew: also, could add dialect check as well, so it's always in the right order of checking:
```
  private matches(token: Token, name: String, isEnabled: Dialect => Boolean): Boolean =
    isEnabled(dialect) && isIdentAnd(token, _ == name)
```
(in this class, everything is about soft keywords, so i also renamed the method)

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
```diff
@@ -0,0 +1,87 @@
+package scala.meta.internal.parsers
+
+import scala.meta.Dialect
+import scala.meta.tokens.Token
+import scala.meta.tokens.Token.Ident
+import scala.meta.internal.classifiers.classifier
+
+class SoftKeywords(dialect: Dialect) {
```
kitbellew, line 8: do we need to wrap this in an instance? wouldn't this work, with implicit:
```
  object SoftKw {
    @classifier
    trait Using {
      def unapply(token: Token)(implicit dialect: Dialect): Boolean = {
        dialect.allowGivenUsing && isSoftKw(token, "using")
      }
    }
  }

  // somewhere else
  token.is[SoftKw.Using]
```

kitbellew (changes requested):
what started as a simple refactor...

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
tgodzik: Tried it and it does not recognize it as an unapply method :/

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Most of it is internal to parser, so to be honest I am not sure what would make sense to extract. I will start with the soft keywords now, but let me know if you find anything that might be particularily useful in scalafmt.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
tgodzik: Looks good, I will change it!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
tgodzik: Moved it actually to ScalametParserObject and is now reused in SoftKeywords and ScalametaParser.

tgodzik: Ach no, not that one :sweat_smile: this one is renamed to `matches` and made private.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
tgodzik: Moved it actually to ScalametParserObject and is now reused in SoftKeywords and ScalametaParser.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
tgodzik: Reafctored!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
tgodzik: Used the advice from @kitbellew and now we have a matches method that first always checks if a particular dialect feature is supported.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
tgodzik: Fixed, looks much nicer.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Ach right! Removed it!

tgodzik:
> also, could add dialect check as well, so it's always in the right order of checking:
>
>  private matches(token: Token, name: String, isEnabled: Dialect => Boolean): Boolean =
>    isEnabled(dialect) && isIdentAnd(token, _ == name)
> (in this class, everything is about soft keywords, so i also renamed the method)

Fixed it, great suggestion! 

> actually, this duplicates the method in ScalametaParser above. should we do this instead:

Moved it, thanks! @kitbellew

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -36,6 +36,8 @@ import scala.util.Failure
 class ScalametaParser(input: Input, dialect: Dialect) { parser =>
   require(Set("", EOL).contains(dialect.toplevelSeparator))
   implicit val currentDialect: Dialect = dialect
+  val soft = new SoftKeywords(currentDialect)
```
kitbellew, line 39: somewhat optional: 1. `private` 2. use `dialect` directly, and not the intermediate `currentDialect`?

also, for a completely different, unrelated change: we could change the class to
```
class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   ...
```
and remove `currentDialect`.

kitbellew (approved):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Good idea! Changed!

kitbellew (approved):

kitbellew:
@tgodzik should we release and use in `scalafmt`?

tgodzik:
Sure, makes sense. We have two more PRs lined up, but it shouldn't really influence scalafmt. I will run it now.

PR #2194, opened 2020-12-10
Make sure enum cases can only be created within an enum

tgodzik:
Previously, parser would allow enum cases to be created anywhere, which is not allowed in the language itself. Now, we make sure that enums are the only ones that can own enum cases. 

This has added a number of parameters to a bunch of methods, but I think it's the cleanest solution that we can have.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -215,15 +215,20 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     def indent = -1
     def closeOnNonCase = false
     def indentOnArrow = true
+    def isIndented = false
   }
   case class RegionIndent(override val indent: Int, override val closeOnNonCase: Boolean)
-      extends SepRegion
+      extends SepRegion {
+    override def isIndented: Boolean = true
+  }
   case object RegionParen extends SepRegion
   case object RegionBracket extends SepRegion
   case class RegionBrace(override val indent: Int, override val indentOnArrow: Boolean)
       extends SepRegion
   case class RegionEnum(override val indent: Int) extends SepRegion
-  case class RegionIndentEnum(override val indent: Int) extends SepRegion
+  case class RegionIndentEnum(override val indent: Int) extends SepRegion {
+    override def isIndented: Boolean = true
+  }
   case object RegionArrow extends SepRegion
   // NOTE: Special case for Enum region is needed because parsing of 'case' statement is done differently
   case object RegionEnumArtificialMark extends SepRegion
@@ -278,10 +283,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   }
   lazy val (parserTokens, parserTokenPositions) = {
     def indentedRegion(regions: List[SepRegion]): Boolean =
-      regions.headOption.exists {
-        case RegionIndent(_, _) | RegionIndentEnum(_) => true
-        case _ => false
-      }
+      regions.headOption.exists(_.isIndented)
 
     val parserTokens = mutable.ArrayBuilder.make[Token]
     val parserTokenPositions = mutable.ArrayBuilder.make[Int]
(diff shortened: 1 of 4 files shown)


olafurpg (approved):
This is a nice improvement. LGTM 👍

Two optional suggestions

- should we add test cases for the parse error when enum cases appear outside enums?
- should we add a "checkFields" test to ensure that `Defn.EnumCase` only appears inside enums?

tgodzik:
> This is a nice improvement. LGTM 
> 
> Two optional suggestions
> 
> * should we add test cases for the parse error when enum cases appear outside enums?

It's already in `enum-case-toplevel` test - I can add it for a class etc. but it's basically I think the same code branch.

> * should we add a "checkFields" test to ensure that `Defn.EnumCase` only appears inside enums?

I will check! the problem is that the immediate parent is Template.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -411,7 +413,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
             while (!sepRegionsProcess.isEmpty &&
               (!sepRegionsProcess.head.isInstanceOf[RegionBrace] &&
                 !sepRegionsProcess.head.isInstanceOf[RegionEnum])) {
-              if (dialect.allowSignificantIndentation) insertOutdent()
+              if (dialect.allowSignificantIndentation && sepRegionsProcess.head.isIndented)
```
tgodzik, line 416: Added a condition to if, since it caused outdent to appear where none is needed - the parser should have failed at that point also, but the error was misleading.

scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/ParentChecks.scala
```diff
@@ -105,6 +105,15 @@ object ParentChecks {
     tree.tpe.is[Type.Singleton] ==> (parent.is[Ctor.Secondary] && destination == "init")
   }
 
+  def EnumCase(tree: Tree, parent: Tree, destination: String): Boolean = {
+    parent.is[Template]
+  }
+
+  def EnumTemplate(tree: Template, parent: Tree, destination: String): Boolean = {
+    val hasEnumCase = tree.stats.exists(st => st.is[Defn.EnumCase] || st.is[Defn.RepeatedEnumCase])
```
tgodzik, line 113: Added the two checks here, which together make sure that EnumCase is in Template and if template contains enum case then the parent can only be Enum. @olafurpg waht do you think?

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/EnumSuite.scala
```diff
@@ -272,67 +272,120 @@ class EnumSuite extends BaseDottySuite {
...
+            Name(""),
+            List(List(Term.Param(Nil, Term.Name("a"), Some(Type.Name("Int")), None)))
+          ),
+          List(Init(Type.Name("Color"), Name(""), List(List(Lit.Int(65280)))))
+        )
+      )
+    )
+  }
+
+  test("enum-case-toplevel") {
```
tgodzik, line 366: This test checks that we should fail to get an enum case in any other situation.

scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/ParentChecks.scala
tgodzik: We will need to go through all stats, so this might not be perfect :thinking:

scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/ParentChecks.scala
```diff
@@ -105,6 +105,15 @@ object ParentChecks {
     tree.tpe.is[Type.Singleton] ==> (parent.is[Ctor.Secondary] && destination == "init")
   }
 
+  def EnumCase(tree: Tree, parent: Tree, destination: String): Boolean = {
+    parent.is[Template]
+  }
+
+  def EnumTemplate(tree: Template, parent: Tree, destination: String): Boolean = {
```
olafurpg, line 112: Let's remove this

scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/ParentChecks.scala
```diff
@@ -105,6 +105,15 @@ object ParentChecks {
     tree.tpe.is[Type.Singleton] ==> (parent.is[Ctor.Secondary] && destination == "init")
   }
 
+  def EnumCase(tree: Tree, parent: Tree, destination: String): Boolean = {
+    parent.is[Template]
```
olafurpg, line 109: Add check if the grandparent is non-empty, then it must be an enum

olafurpg (approved):

scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/ParentChecks.scala
tgodzik: Removed!

scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/ParentChecks.scala
tgodzik: Added.

PR #2195, opened 2020-12-11
Fix product names in macros for 2.13

tgodzik:
`productElementName` was introduced in 2.13 and some libraries might use it to print names for field. When for example using pprint, previously we would get:

```scala
Source(
   = List(
    Defn.Def( = List(),  = Term.Name( = "b"),  = List(),  = List(),  = None,  = Lit.Int( = 1))
  )
)
```

now we get
```
Source(
  stats = List(
    Defn.Def(
      mods = List(),
      name = Term.Name(value = "b"),
      tparams = List(),
      paramss = List(),
      decltpe = None,
      body = Lit.Int(value = 1)
    )
  )
)
```


olafurpg (approved):
Awesome! This will greatly improve the debugging experience with pprint and experimenting with Scalameta in worksheets!

PR #2196, opened 2020-12-11
Add support for dependent function types

tgodzik:
Dependent function types are a new features in Scala 3, which is explained [here](https://dotty.epfl.ch/docs/reference/new-types/dependent-function-types.html). While working on this I noticed that MethodType is something that is no longer in any dialect and is not really used at all. Because of that and the fact that it caused some issues parsing dependent function types I decided to get rid of it and deprecate all dialect methods.


olafurpg (approved):
LGTM 👍 

It's OK to deprecate method types. They got introduced to unblock a Scalafix use-case when we were exploring how to implement semantic APIs, but it's is no longer relevant and it never got used by any application.

PR #2197, opened 2020-12-15
Fix position bugs and add position testing infrastructure.

olafurpg:
Previously, many Scala 3 AST nodes were missing positions, which caused
problems in Scalafmt. This commit adds positions for all Scala 3 AST
nodes and a new testing infrastructure to make it easier to visually
catch position bugs.

The `BaseTokenizerCoverageSuite` class is superseded by
`BasePositionSuite`. Position suites have several benefits over coverage
suites:

* output of failing test cases is copy-pasteable back into source code
* position suites test all children of a given code snippet instead of a
  single node.
* test cases can be written without copy-pasting unicode characters
* a single `check` method, instead of multiple `checkTerm`/`checkType`/`checkEnumerator`/...

Other kind of unrelated changes

* require `.origin` to be non-empty for all parsed trees. This should
  make parse tests for new AST nodes to fail if the positions is missing.
* deprecate "scala.meta.dialects.Dotty" in favor of
  "scala.meta.dialects.Scala3"


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3786,9 +3809,9 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     assert(token.text == "end")
     next()
     if (token.is[Ident]) {
-      EndMarker(termName())
+      EndMarker(autoPos(termName()))
```
tgodzik, line 3812: Should term name not have `autoPos` inside?  There is a number of invocations of `termName()` and most do not seem to be inside `autoPos` or `atPos`.

tests/shared/src/test/scala-2.13/scala/meta/tests/parsers/dotty/Scala3PositionSuite.scala
```diff
@@ -0,0 +1,117 @@
...
+import scala.meta.tests.parsers.BasePositionSuite
+
+class Scala3PositionSuite extends BasePositionSuite {
+  override def defaultDialect: Dialect = dialects.Scala3
+
+  check[Type]("A & B")
+  check[Type]("A | B")
+  check[Type](
+    "[X] =>> (X, X)",
+    """|Type.Bounds [X→←] =>> (X, X)
```
tgodzik, line 13:
```suggestion
    """|Type.Bounds [X@@] =>> (X, X)
```
Or something it's kind of hard to see in the git diff.

image

tests/shared/src/test/scala/scala/meta/tests/parsers/BasePositionSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+import scala.meta.parsers.Parse
+import scala.meta.internal.inputs._
+import munit.TestOptions
+import scala.meta.internal.trees.Origin
+import munit.Location
+import scala.meta.tests.parsers.MoreHelpers
+
+abstract class BasePositionSuite extends ParseSuite {
+  import scala.meta._
+  def defaultDialect: Dialect = dialects.Scala213
```
tgodzik, line 13:
```suggestion
  def defaultDialect: Dialect
```
I think it would be good to have it explicitely set in the implementing classes, so that it's clearer which dialect is each testing

tests/shared/src/test/scala/scala/meta/tests/parsers/BasePositionSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  def check[T <: Tree: Parse](code: TestOptions)(implicit loc: Location): Unit = {
+    check[T](code, "")
+  }
+  def check[T <: Tree: Parse](code: TestOptions, expected: String)(implicit loc: Location): Unit = {
```
tgodzik, line 18:
```suggestion
  def checkPositions[T <: Tree: Parse](code: TestOptions, expected: String)(implicit loc: Location): Unit = {
```
?

tgodzik: Also, could you a comment about the context of what we are testing here? so that in the future no one will have problem what is going on.

tgodzik (approved):
Thanks for the fix! We have missed a lot of positions for the new ASTs. I left a few comments, but nothing major, so LGTM

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
olafurpg: Good point, I removed `autoPos` and the tests still pass so it's not needed

tests/shared/src/test/scala-2.13/scala/meta/tests/parsers/dotty/Scala3PositionSuite.scala
olafurpg: Good idea, done.

tests/shared/src/test/scala/scala/meta/tests/parsers/BasePositionSuite.scala
olafurpg: Good point. I made the dialect a constructor argument.

tests/shared/src/test/scala/scala/meta/tests/parsers/BasePositionSuite.scala
olafurpg: Good idea. Done, I added a docstring explaining the motivation for these kinds of tests.

olafurpg (commented):
@tgodzik Thank you for the review! Great suggestions 🙏

PR #2198, opened 2020-12-21
Fix TypeCase positions to not include a previous character

tgodzik:
Previously, type case would include the opening `{` due to the code being copied from `caseClause` method. Now it only included the exact position.

I also added more tests to see if the positions are incorrect anywhere else.


kpbochenek (approved):
LGTM :+1:

PR #2200, opened 2020-12-21
Remove `as` soft keyword in patterns and givens

tgodzik:
Scala 3 no longer uses `as` in given declartions and instead uses a colon. Because of that and the limited need for
`as` it was also removed from patterns where it was a replacements for `@`

This still doesn't fix the givens with indentation that use the `with` keyword. The syntax is still fluctuating there, so we might need to work on it later.


kpbochenek (approved):
LGTM :+1:

PR #2201, opened 2020-12-23
Add inline match available in Scala 3

tgodzik:
Scala 3 introduces the concept of inline match that can be created within an inline method:
```scala
inline def g(x: Any): Any = inline x match {
  case x: String => (x, x) // Tuple2[String, String](x, x)
  case x: Double => x
}
```
This needs to be parse accordingly and also makes it imporrisble to use `inline` as identifiers in Apply trees.

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/InlineSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/InlineSuite.scala
@@ -170,9 +170,31 @@ class InlineSuite extends BaseDottySuite {
   }
 
   test("inline-soft-ident") {
-    // inline is ident declared as val inline = 4
+
     runTestAssert[Stat](
-      "object X { inline + 3 }"
+      "inline"
+    )(
+      Term.Name("inline")
+    )
+
+    runTestAssert[Stat](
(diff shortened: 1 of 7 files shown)


tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/InlineSuite.scala
```diff
@@ -187,12 +209,161 @@ class InlineSuite extends BaseDottySuite {
...
+        None,
+        Term.Match(
+          Term.Name("x"),
+          List(Case(Pat.Var(Term.Name("x")), None, Term.Name("x"))),
+          List(Mod.Inline())
+        )
+      )
+    )
+  }
+
```
kpbochenek, line 263: also `inline new X() match { ... }`

kpbochenek (approved):
Nice work! :+1:  (just one additional example to handle)

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/InlineSuite.scala
tgodzik: Good catch! Fixed!

PR #2208, opened 2021-01-04
Type aliases and function types should take into account indentation

tgodzik:
```
An indentation region can start after one of the following tokens:

=  =>  ?=>  <-  catch  do  else  finally  for
if  match  return  then  throw  try  while  yield
```

Previously, we forgot about types when checking this rule.


kpbochenek (approved):
good catch! :+1:

PR #2209, opened 2021-01-06
Allow extension to declare using and braces in another line

tgodzik:
This is done the same way when working with templates and method definitions.


kpbochenek (approved):
LGTM! :+1:

PR #2210, opened 2021-01-06
Fix givens to accept `with` keyword for refinements

tgodzik:
In the previous PR that touched the changes in givens I forgot to include `with`, which I am fixing in this PR. I also removed any test cases that were ignored, but don't have an easy way for the parser to show an error (also Dotty parser mostly works in those case also).

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1047,7 +1047,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       token.is[KwYield] || token.is[KwTry] || token.is[KwCatch] || token.is[KwFinally] ||
       token.is[KwMatch] || token.is[KwDo] || token.is[KwFor] || token.is[KwThen] ||
       token.is[KwElse] || token.is[Equals] || token.is[KwWhile] || isColonEol(token) ||
-      token.is[RightArrow]
+      token.is[RightArrow] || (token.is[KwWith] && token.next.isNot[Ident])
     }
   }
 
@@ -3908,9 +3908,11 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     givenDef(mods)
   }
 
-  // GivenDef          ::=  [GivenSig] [‘_’ ‘<:’] Type ‘=’ Expr
-  //                     |  [GivenSig] ConstrApps [TemplateBody]
-  // GivenSig          ::=  [id] [DefTypeParamClause] {UsingParamClause} ‘as’
+  /**
+   * GivenDef          ::=  [GivenSig] (Type [‘=’ Expr] | StructuralInstance)
+   * GivenSig          ::=  [id] [DefTypeParamClause] {UsingParamClause} ‘:’
+   * StructuralInstance ::=  ConstrApp {‘with’ ConstrApp} ‘with’ TemplateBody
+   */
   private def givenDef(mods: List[Mod]): Stat = autoPos {
     val anonymousName = autoPos(scala.meta.Name.Anonymous())
 
@@ -3933,15 +3935,40 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
     val decltpe = startModType()
 
+    def parents() = {
+      val parents = ListBuffer[Init](
+        atPos(decltpe.startTokenPos, decltpe.endTokenPos)(
+          Init(decltpe, autoPos(Name.Anonymous()), Nil)
+        )
+      )
+      while (token.is[KwWith] && token.next.is[Ident]) { next(); parents += init() }
+      parents.toList
+    }
+
     if (token.is[Equals]) {
       accept[Equals]
       Defn.GivenAlias(mods, sigName, sigTparams, sigUparamss, decltpe, exprMaybeIndented())
     } else {
-      val rhs = autoPos {
-        val (slf, stats) = templateBodyOpt(true)
-        Template(List.empty, List.empty, slf, stats)
+      val inits = parents()
+      val (slf, stats) = if (token.is[KwWith]) {
+        accept[KwWith]
+        newLineOptWhenFollowedBy[LeftBrace]
+        if (token.is[LeftBrace]) {
+          inBraces(templateStatSeq())
+        } else if (token.is[Indentation.Indent]) {
+          indented(templateStatSeq())
+        } else {
+          syntaxError("expected '{' or indentation", at = token.pos)
+        }
+      } else {
+        (autoPos(Self(autoPos(Name.Anonymous()), None)), Nil)
       }
-      Defn.Given(mods, sigName, sigTparams, sigUparamss, decltpe, rhs)
+      val rhs = if (slf.decltpe.nonEmpty) {
+        syntaxError("given cannot have a self type", at = slf.pos)
+      } else {
+        atPos(decltpe.startTokenPos, auto)(Template(List.empty, inits, slf, stats))
+      }
+      Defn.Given(mods, sigName, sigTparams, sigUparamss, rhs)
     }
   }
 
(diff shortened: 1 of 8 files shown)


tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/GivenUsingSuite.scala
```diff
@@ -236,11 +239,6 @@ class GivenUsingSuite extends BaseDottySuite {
     )
   }
 
-  test("given-subtype-error".ignore) {
-    // it is treaten as alias without '=' sign at the end and {...} is refinement part
-    runTestError("given intOrd: ? <: Ord[Int] { def f(): Int = 1 }", "missing = at the end")
```
tgodzik, line 241: This doesn't parse in Scala 3, so I removed the test case. We do parse it in both Scala 2 and 3, but I think form the parser perspective it is fine. It's also the check needed here would be a bit complicated.

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/GivenUsingSuite.scala
```diff
@@ -287,13 +285,6 @@ class GivenUsingSuite extends BaseDottySuite {
     )
   }
 
-  test("given-alias-override-block-error".ignore) {
```
tgodzik, line 290: This is actually separate from givens, so I removed it from here.

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/GivenUsingSuite.scala
```diff
@@ -339,44 +330,14 @@ class GivenUsingSuite extends BaseDottySuite {
   }
 
   test("given-alias-inline-subtype") {
-    runTestAssert[Stat]("inline given intOrd: ? <: Ord[Int] = ???")(
```
tgodzik, line 342: This doesn't parse in Scala 3 so I changed the test case.

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/GivenUsingSuite.scala
```diff
@@ -339,44 +330,14 @@ class GivenUsingSuite extends BaseDottySuite {
...
-        Type.Placeholder(Type.Bounds(None, Some(Type.Apply(pname("Ord"), List(pname("Int")))))),
-        tname("???")
-      )
-    )
-  }
-
-  test("given-alias-subtype-noinline-error".ignore) {
-    runTestError(
-      "given intOrd: ? <: Ord[Int] = ???",
-      "is only allowed for given with inline modifier"
```
tgodzik, line 357: This actually is not allowed ever currently.

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/GivenUsingSuite.scala
```diff
@@ -485,31 +446,6 @@ class GivenUsingSuite extends BaseDottySuite {
     )
   }
 
-  test("using-mix-named-anonymous-error".ignore) {
```
tgodzik, line 488: I removed the ignored test cases as it doesn't seem it will be easy to show sensible errors and we can focus on the happy path. Should be easy to return the test if they are needed

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/GivenUsingSuite.scala
kpbochenek: :+1: (change a test name to not include 'subtype')

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/GivenUsingSuite.scala
```diff
@@ -105,10 +108,10 @@ class GivenUsingSuite extends BaseDottySuite {
     )
   }
```
kpbochenek, line 110: (random place)
with can start in that case indentation block, could you add a test for that?
sth like:
```
given intOrd: Ord[Int] with
   def fa: Int = 1
   def fb: Int = 2
```

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3937,9 +3937,19 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       accept[Equals]
       Defn.GivenAlias(mods, sigName, sigTparams, sigUparamss, decltpe, exprMaybeIndented())
     } else {
```
kpbochenek, line 3939: comments section for this method:
` // GivenDef          ::=  [GivenSig] [‘_’ ‘<:’] Type ‘=’ Expr`
are no longer relevant, probably best to remove

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3937,9 +3937,19 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       accept[Equals]
```
kpbochenek, line 3937: There is a line above:
`val (sigName, sigTparams, sigUparamss) = if (token.is[Colon]) {`
according to spec colon must always be present.
At least one of sigName, sigTparams, sigUparamss must be present.
With this there is probably no need to fork and rollback how it was done before.

Someone noticed parsing was not straightforward and adjusted syntax? :)

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/GivenUsingSuite.scala
```diff
@@ -485,31 +446,6 @@ class GivenUsingSuite extends BaseDottySuite {
     )
```
kpbochenek, line 446: could you add test:
```
scala> given intM: Ord[Int] with Eq[Int] with {
     |   def fx = 3
     |}
// defined object intM
```
Didn't check if it is handled, but seems like a tricky case

kpbochenek (changes requested):
I slowly start to lose a track of what should be parseable :face_with_head_bandage:

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: You can still have:
`given Ord[Int] with { def compare(x: Int, y: Int): Int = 0 }` 

which does not contain `:`, since `[GivenSig]` is optional.

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/GivenUsingSuite.scala
tgodzik: That might have never actually worked :thinking:

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/GivenUsingSuite.scala
tgodzik: Fixed!

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/GivenUsingSuite.scala
tgodzik: Actually it was not supported, but it's not that hard to do. Fixed it and added a test for both indentation and legit `with`.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Updated!

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/GivenUsingSuite.scala
tgodzik: Ok, so that was much more difficult than aniticipated - I had to rework given defs, there is no need for separate `decltype` anymore, but we can just use the inits from the Template.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kpbochenek: ahh right 👍

kpbochenek (approved):
LGTM! :+1:

PR #2211, opened 2021-01-06
Trees: Support branch methods

kitbellew:
Currently, the code hardcodes exactly three methods (`value`, `name` and `tpe`) which can be declared in a `@branch` trait.

This is an attempt to support other methods as well, although it still is not flexible enough to support methods in a branch which is a parent of another branch.

Required for #2207 to work.

### DIFF
--- a/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/NamerMacros.scala
+++ b/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/NamerMacros.scala
@@ -5,6 +5,12 @@ package trees
 import scala.collection.mutable.ListBuffer
 import org.scalameta.internal.MacroHelpers
 
+object CommonNamerMacros {
+
+  private val quasiName = "Quasi"
+
+}
+
 trait CommonNamerMacros extends MacroHelpers {
   import c.universe._
 
@@ -34,6 +40,9 @@ trait CommonNamerMacros extends MacroHelpers {
     classifierBoilerplate
   }
 
+  private val quasiTypeName = TypeName(CommonNamerMacros.quasiName)
+  def isQuasiClass(cdef: ClassDef) = cdef.name.toString == CommonNamerMacros.quasiName
+
   def mkQuasi(
       name: TypeName,
       parents: List[Tree],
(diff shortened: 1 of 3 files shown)


kitbellew:
@tgodzik you were right about the rabbit hole, but this looks to work. if we merge this, i also have an adjusted version of your #2207 submission (although i am not clear how to share that with you).

as usual, it might be easier to review my code one commit at a time, rather than all at once. they are all individually packaged :)

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/NamerMacros.scala
```diff
@@ -34,46 +40,66 @@ trait CommonNamerMacros extends MacroHelpers {
...
+    val stub = {
       val unsupportedUnquotingPosition = "unsupported unquoting position"
       val unsupportedSplicingPosition = "unsupported splicing position"
       val message =
         q"if (this.rank == 0) $unsupportedUnquotingPosition else $unsupportedSplicingPosition"
       q"throw new $UnsupportedOperationException($message)"
     }
-    val qstubs = (paramss.flatten.map(_.name.toString) ++ extraStubs).distinct.map(TermName.apply)
-    qstubs.foreach(name => qstats += q"def $name: $NothingClass = ${stub()}")
+    val stubbedMembers = new mutable.HashSet[String]
```
tgodzik, line 75: Why is `stubbedMembers` needed? I see that it's never really read.

tgodzik (approved):
Looks good! Great job with that - I was a bit afraid to touch it :fearful: 

I think one value might not be needed, but aside from that everything seems perfectly clear. I will rebase my PR once this is ready. I might try also to see if there are any other ast nodes that can benefit from additional parent (for example Case and TypeCase). We can release once all that is merged.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/NamerMacros.scala
kitbellew: to make sure that, for instance, `value` is not defined twice, since `extraStubs` are hardcoded to contain it. 

in the original code, `distinct` was used to remove duplicates between class parameters and extra stubs. but we also need to apply the same to extra abstract definitions as well.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/NamerMacros.scala
```diff
@@ -34,46 +40,66 @@ trait CommonNamerMacros extends MacroHelpers {
...
+    def markStubbedMemberName(name: TermName): Boolean = stubbedMembers.add(name.toString)
```
kitbellew, line 76: this is where this structure is used: when the method returns true, it means the value hadn't been seen before and can be stubbed.

PR #2212, opened 2021-01-08
Update branches for scalameta trees

tgodzik:
- add common branch for import and export
- add fields for CaseTree

### DIFF
--- a/project/Mima.scala
+++ b/project/Mima.scala
@@ -76,6 +76,8 @@ object Mima {
     ProblemFilters.exclude[DirectMissingMethodProblem](
       "scala.meta.Defn#Given#DefnGivenImpl.copy$default$6"
     ),
-    ProblemFilters.exclude[DirectMissingMethodProblem]("scala.meta.Defn#Given#DefnGivenImpl.this")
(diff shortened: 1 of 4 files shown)


kitbellew (approved):

project/Mima.scala
```diff
@@ -76,6 +76,8 @@ object Mima {
     ProblemFilters.exclude[DirectMissingMethodProblem](
       "scala.meta.Defn#Given#DefnGivenImpl.copy$default$6"
     ),
-    ProblemFilters.exclude[DirectMissingMethodProblem]("scala.meta.Defn#Given#DefnGivenImpl.this")
+    ProblemFilters.exclude[DirectMissingMethodProblem]("scala.meta.Defn#Given#DefnGivenImpl.this"),
```
kitbellew, line 79: what was the problem with mima? adding a method causes this? or did we change the signature?

tgodzik: This happens when adding a methods for public traits. However since no one really can sensibly extend CaseTree, I think it's not a problem.

PR #2213, opened 2021-01-11
Add Decl.Given for abstract givens

tgodzik:
All other tree nodes like val, var or def have a separate Tree node when they are declared abstract. To keep in line with that I added Decl.Given to handle abstract given declarations

Abstract givens cannot be annonymous or have multiple types declared using `with`, so I added checks for that.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -3918,7 +3918,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
     val forked = in.fork
     val name: meta.Name =
-      if (token.is[Ident]) typeName() else anonymousName
+      if (token.is[Ident]) termName() else anonymousName
     val tparams = typeParamClauseOpt(
       ownerIsType = false,
       ctxBoundsAllowed = true,
@@ -3948,7 +3948,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     if (token.is[Equals]) {
       accept[Equals]
       Defn.GivenAlias(mods, sigName, sigTparams, sigUparamss, decltpe, exprMaybeIndented())
-    } else {
+    } else if (token.is[KwWith]) {
       val inits = parents()
       val (slf, stats) = if (token.is[KwWith]) {
         accept[KwWith]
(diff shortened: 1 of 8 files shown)


scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -313,6 +313,13 @@ object Decl {
       tparams: List[scala.meta.Type.Param],
       bounds: scala.meta.Type.Bounds
   ) extends Decl with Member.Type
+  @ast class Given(
+      mods: List[Mod],
+      name: scala.meta.Name,
+      tparams: List[scala.meta.Type.Param],
+      sparams: List[List[Term.Param]],
+      decltpe: scala.meta.Type
+  ) extends Defn
```
kpbochenek, line 322: extends Decl?

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -313,6 +313,13 @@ object Decl {
       tparams: List[scala.meta.Type.Param],
       bounds: scala.meta.Type.Bounds
   ) extends Decl with Member.Type
+  @ast class Given(
+      mods: List[Mod],
+      name: scala.meta.Name,
```
kpbochenek, line 318: use Term.Name if name cannot be Anonymous

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
kpbochenek, line 322: extends Decl

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
kpbochenek: or actually Type.Name here

kpbochenek (changes requested):

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
tgodzik: Good catch! Though I think givens should actually be `Term.Name` the same as defs, vals and vars. `Type.Name` is for more class like entities.

Changed here to TermName and fixed to use it always.

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
tgodzik: Fixed! That was some bad copy paste :sweat:

kpbochenek (approved):
Nice work! looking good :+1:

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3961,14 +3961,21 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           syntaxError("expected '{' or indentation", at = token.pos)
         }
       } else {
-        (autoPos(Self(autoPos(Name.Anonymous()), None)), Nil)
+        syntaxError("expected 'with' <body>", at = token.pos)
```
tgodzik, line 3964: Removed `abstract givens cannot be annonymous` here since it's actually misleading. Givens with `with` cannot be abstract at all.

PR #2215, opened 2021-01-13
Make Type.With show up always instead of Type.And where `with` keywor…

tgodzik:
…d is involved

Without this change it's not possible to see where `With` is still used and the parser should show the code as is.

This came up as an issue in scalafmt where `with` chains were translated as `Type.And` and it causes the formatter to behave differently. Having Type.With translated as Type.And despite initial ideas does not seem to be actually helpful


kpbochenek (approved):
LGTM! :+1:

PR #2216, opened 2021-01-14
Add support for 2.12.13 and fix warning reporting

tgodzik:
Major issue here was that warnings were not reported anymore for 2.12.13 and 2.13.2+ versions. This was most likely due to https://github.com/scala/scala/pull/9248, which delays warnings until the end.

I fixed via explicitely making the compiler report warnings in the `InteractiveSemaniticDB`, which unfortunately required some additional  Compat class and since that became a bit of a mess with the `2.13.1+` I reverted back to using separate directories for each version. It's not perfect, but I don't really see any other choice.

I also made InteractiveSuite run on all supported Scala versions.


semanticdb/scalac/library/src/main/scala-2.13.2/scala/meta/internal/semanticdb/scalac/VersionCompilerOps.scala
```diff
@@ -0,0 +1,11 @@
+package scala.meta.internal.semanticdb.scalac
+
+import scala.tools.nsc.interactive.Global
+
+trait VersionCompilerOps {
+
+  def forceWarnings(global: Global) = {
+    global.runReporting.reportSuspendedMessages()
```
tgodzik, line 8: This is the main thing that was needed to be added for some of the Scala versions.

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/InteractiveSuite.scala
```diff
@@ -0,0 +1,154 @@
+package scala.meta.tests
```
tgodzik, line 1: The suite was moved and reworked with a compat section.

olafurpg (approved):
LGTM 👍 Nice job tracking down the root cause of this regression.

PR #2217, opened 2021-01-15
Fix given exports to adhere to current spec

tgodzik:
Previously, given exports would be written like this:
```scala
export given A
```

now it got changed to be similar to imports:
```scala
export A.given
```


kpbochenek (approved):
good catch! 👍

PR #2220, opened 2021-01-20
[Scala 3] Fix givens to actually contain the given keyword 

tgodzik:
It seems that positions are sometimes different depending on whether we are in a top level tree or not. I adjusted the test for givens and also added one for exports.


kpbochenek (approved):
good idea :+1:

PR #2221, opened 2021-01-22
Make sure children contains binary compat fields

tgodzik:
Binary compat fields were added to allow for the older AST nodes to be forward binary compatible.

Previously, when invoking `children` method on trees that contained binaryCompat fields, those fields would not exist in the resulting list. Now, they are contained and can be used to traverse the tree.

It's important to note that traverse will still not working with those additional fields, but in most places (for example scalafmt) the `children` method is being used. We can try to fix it in the future if need arises.

This came up when working on `derives` in scalafmt. @kitbellew could you take a look? You are probably the only other person that recently dealt with thise macros :sweat_smile:

### DIFF
--- a/scalameta/common/shared/src/main/scala/org/scalameta/adt/Reflection.scala
+++ b/scalameta/common/shared/src/main/scala/org/scalameta/adt/Reflection.scala
@@ -117,6 +117,11 @@ trait Reflection {
   class Leaf(sym: Symbol) extends Adt(sym) {
     if (!sym.isLeaf) sys.error(s"$sym is not a leaf")
     def fields: List[Field] = sym.fields.map(_.asField)
+    def binaryCompatFields: List[Field] = {
+      sym.info.decls.collect {
+        case s if s.hasAnnotation[AstMetadata.binaryCompatField] => s.asField
+      }.toList
+    }
     def allFields: List[Field] = sym.allFields.map(_.asField)
     override def toString = s"leaf $prefix"
   }
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/ParseSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/ParseSuite.scala
@@ -51,12 +51,12 @@ class ParseSuite extends FunSuite with CommonTrees {
 }
 
 object MoreHelpers {
-  def requireNonEmptyOrigin(tree: Tree): tree.type = {
+  def requireNonEmptyOrigin(tree: Tree)(implicit dialect: Dialect): tree.type = {
     val missingOrigin = tree.collect {
       case t if t.origin == Origin.None => t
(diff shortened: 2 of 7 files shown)


tgodzik:
Och no, this actually showed some other errors we have :fearful:

scalameta/common/shared/src/main/scala/org/scalameta/adt/Reflection.scala
```diff
@@ -117,6 +117,12 @@ trait Reflection {
   class Leaf(sym: Symbol) extends Adt(sym) {
     if (!sym.isLeaf) sys.error(s"$sym is not a leaf")
     def fields: List[Field] = sym.fields.map(_.asField)
+    def binaryCompatFields: List[Field] = {
+      sym.info.decls
+        .filter(_.hasAnnotation[AstMetadata.binaryCompatField])
+        .map(_.asField)
```
kitbellew, line 123: optional: could do it via
```
   .collect {
     case x if x.hasAnnotation[...] => x.asField
   }
```

kitbellew (commented):
why are tests failing? doesn't look like a big change...

tests/shared/src/test/scala/scala/meta/tests/parsers/ParseSuite.scala
```diff
@@ -51,12 +51,12 @@ class ParseSuite extends FunSuite with CommonTrees {
 }
 
 object MoreHelpers {
-  def requireNonEmptyOrigin(tree: Tree): tree.type = {
+  def requireNonEmptyOrigin(tree: Tree)(implicit dialect: Dialect): tree.type = {
     val missingOrigin = tree.collect {
```
tgodzik, line 55: So this actually relies on the `children` method, which was not working for `binaryCompat` fields previosuly and now that it started working it turns out that origin is not set for a number of cases. They should be fixed now.

I also added dialect and `.map(_.structure)` otherwise an exception was thrown while trying to print the trees.

scalameta/common/shared/src/main/scala/org/scalameta/adt/Reflection.scala
tgodzik: Good idea! Changed.

tests/shared/src/test/scala/scala/meta/tests/parsers/ParseSuite.scala
tgodzik: This also explains why I was having problems getting a proper formatting for `inline if` :sweat_smile:

kitbellew (approved):

tgodzik:
Two more tests update, that I did not notice and added a `derives` test case for the `Scala3PositionsSuite`.

kitbellew (approved):

Issue #2228, opened 2021-01-24
\u005c before a closing quote or backtick causes an unclosed quote error


tjm1518:
Steps to reproduce:  
1. Have a `\u005c` escape code before a closing quote or backtick

Example:  
```scala
object Main extends App {
  println('\u005c' == '\\') // Evaluates to true, Metals detects an error
}
```

Scala Version: 2.13.4  
Metals Version: VSCode/0.9.10

PR #2229, opened 2021-01-24
Fix for \u005c causing an unclosed quote error

tjm1518:
I believe this fixes issue #2228 where `'\u005c'` would cause an unclosed character literal error.

### DIFF
--- a/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala
+++ b/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala
@@ -451,7 +451,7 @@ class LegacyScanner(input: Input, dialect: Dialect) {
             syntaxError("can't unquote into character literals", at = charOffset - 1)
           else if (isIdentifierStart(ch))
             charLitOr(getIdentRest _)
-          else if (isOperatorPart(ch) && (ch != '\\'))
+          else if (isOperatorPart(ch) && (ch != '\\' || isUnicodeEscape))
             charLitOr(getOperatorRest _)
           else {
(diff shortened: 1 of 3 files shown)


tjm1518:
The tests appear to be failing due to the same issue from what I understand?
The test that's failing (`val Lit("\u005c") = term("\"\\\"")`) gives the same failure message as I get from Metals (`= expected but } found`), however when I change it to `val Lit("\\") = term("\"\\\"")` Metals gives no errors?

tgodzik:
> The tests appear to be failing due to the same issue from what I understand?
> The test that's failing (`val Lit("\u005c") = term("\"\\\"")`) gives the same failure message as I get from Metals (`= expected but } found`), however when I change it to `val Lit("\\") = term("\"\\\"")` Metals gives no errors?

So it seems that it's a bug in Scala versions previous to `2.13.2`, which is why it doesn't compile. You could add a new suite something like:
```
class Lit213Suite extends ParseSuite {

}
```

that would reside inside `/tests/shared/src/test/scala-2.13.4/scala/meta/tests/parsers/LitSuite.scala` directory. However that would require changes in `build.sbt` to allow it for tests. 

I think for such small change I am willing to just merge it without the test, as it would require too much changes. I can just confirm it locally.

However, it seems that the test is actually failing on 2.13.4:
```
==> X scala.meta.tests.parsers.LitSuite.unicode_escaped_backslash  0.003s scala.meta.tokenizers.TokenizeException: <input>:1: error: unclosed string literal
"\"
^
```

tjm1518:
The tests were failing in 2.13.4 due to a missing backslash in the expected output, this was fixed and tested successfully in 2.13.2+ (see commit 2957cef).

tgodzik (approved):
LGTM!

PR #2231, opened 2021-01-28
Scala3 - parse multilne extension

dos65:


kpbochenek:
nice work!
Could you add a test for double safety:
```
val code = """|object X:
              |  extension (c: Circle)
              |    def cra: Int = 2
              |  def fn: Int = 1
```
And expectation would be that 'fn' belongs to X not as part of extension

dos65:
@kpbochenek 
Thx! Will add this test-case.

__Upd__, after an offline discussion with @tgodzik, we decided to give a try to rework how the parser handles significant indentation to make it easier and more close to the implementation in the compiler.

So, don't spend time reviewing this PR. I'm going to create another one that will cover multiline extensions too. I leave this one open for a while in case of the idea about refactoring will fail.

dos65:
Implemented in #2237

PR #2232, opened 2021-01-29
[Scala 3] Fix transparent traits

tgodzik:
Transparent traits are described here: https://dotty.epfl.ch/docs/reference/other-new-features/transparent-traits.html


kpbochenek (approved):
nice one :+1:

PR #2233, opened 2021-02-05
Add information about overrides to symbol information

tgodzik:
This is a follow up to https://github.com/scalameta/metals/issues/1645 

We should be easily able to include information about all symbols that a given symbol overrides, but I am not sure that it's the right way to go (not sure of a full impact on semanticdb). We could also try to just a Boolean flag that says whether or not a given symbols overrides anything. @olafurpg what do you think?


semanticdb/semanticdb/semanticdb.proto
```diff
@@ -280,6 +280,7 @@ message SymbolInformation {
   Signature signature = 17;
   repeated Annotation annotations = 13;
   Access access = 18;
+  repeated string overriden_symbols = 19;
```
olafurpg, line 283: Changes to semanticdb.proto need to be documented in three different places in the spec.

* Generic section on `SymbolInformation` https://scalameta.org/docs/semanticdb/specification.html#symbolinformation
* Scala section on `SymbolInformation`. We can link to the Scala Language Spec here https://www.scala-lang.org/files/archive/spec/2.12/05-classes-and-objects.html#overriding
* Java section on `SymbolInformation`. The relevant link to the Java Language spec https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.8

olafurpg (changes requested):
👍 I am in favor of adding this feature to SemanticDB. In fact, I was planning to submit a PR soon to add it so I'm glad you got ahead of me :p 

This feature actually existed before in SemanticDB, and it got removed in the commit [`42f110e` (#1506)](https://github.com/scalameta/scalameta/pull/1506/commits/42f110e4677287063f321c8e9d06f8f8a0a6b6c1).

I don't remember the exact motivation for removing `overrides`. I suspect this feature got removed because I assumed it was too expensive to persist overrides, but at the time we used to persist `SymbolInformation` for defined AND referenced symbols. My intuition is that this feature is cheap to support if we only do it for definitions.

There is one caveat that's worth keeping in mind. The `overrides` symbols can be different at compile-time and runtime. For example, a program can compile against `guava:1.0` and run with `guava:2.0`. If `guava:2.0` changes the ordering of methods, or introduces new method overloads, the `overrides` symbol that are generated at compile-time becomes outdates. This shouldn't be a big issue within the context of a single workspace.

One blocking comment about missing documentation in the spec.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolInformationOps.scala
```diff
@@ -226,13 +226,22 @@ trait SymbolInformationOps { self: SemanticdbOps =>
       }
     }
 
+    def overridenSymbols: List[String] = {
+      if (gsym.overrideChain.size > 1) {
```
olafurpg, line 230: This method can be replaced with `gsym.allOverriddenSymbols`, which returns `Nil` if this method doesn't override symbols.

Note that this is an expensive operation that's not cached, so it's good to avoid duplicate `overrideChain`.

tests/jvm/src/test/resources/metac-metacp.diff
```diff
@@ -160,6 +160,51 @@ annot/Alias.A#
 -              symbol: "scala/annotation/meta/param#"
 
 
+====================
+classes/C1#equals().
+====================
+--- metac
++++ metacp
+   public_access {
+-overriden_symbols: "scala/Any#equals()."
```
olafurpg, line 169: I'm torn on including overrides for these default methods. I don't think they will be useful for our applications, and they introduce noise in our tests and overhead in the SemanticDB payloads. 

My vote is to include these overrides in the payload because that's the **correct** thing.

olafurpg (changes requested):
Can we add a couple `TargetedSuite` tests to assert the behavior works as expected?

https://github.com/scalameta/scalameta/blob/b5bc9f1d9b4621d33aa4286087f9934114b0465f/tests/jvm/src/test/scala/scala/meta/tests/semanticdb/TargetedSuite.scala

Also, we should update metap to print `overrides` here https://github.com/scalameta/scalameta/blob/b5bc9f1d9b4621d33aa4286087f9934114b0465f/semanticdb/metap/src/main/scala/scala/meta/internal/metap/SymbolInformationPrinter.scala#L81

This should cause the expect tests to fail.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolInformationOps.scala
```diff
@@ -226,13 +226,22 @@ trait SymbolInformationOps { self: SemanticdbOps =>
       }
     }
 
+    def overridenSymbols: List[String] = {
```
olafurpg, line 229:
```suggestion
    def overriddenSymbols: List[String] = {
```

semanticdb/semanticdb/semanticdb.proto
tgodzik: Added documentation in all the needed places (I think)

tests/jvm/src/test/resources/metac-metacp.diff
tgodzik: Should we read the overriden symbols in `scalacp` ? Not sure how to get that information from scalap

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolInformationOps.scala
tgodzik: Changed it base don suggestions, thanks!

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolInformationOps.scala
tgodzik: Fixed

tgodzik:
@olafurpg finally got back to this, sorry it took so long. The biggest question for me here is whether we should try to get overrides information in scalacp?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolInformationOps.scala
```diff
@@ -226,13 +226,18 @@ trait SymbolInformationOps { self: SemanticdbOps =>
       }
     }
 
+    def overriddenSymbols: List[String] = {
+      gsym.allOverriddenSymbols.map(_.toSemantic)
```
olafurpg, line 230: Should we make this behavior configurable? It should be enabled by default, but it might be helpful to have an option to disable it in case there are bugs we're not aware of. Not a blocking issue, just an idea.

olafurpg (approved):
LGTM 👍 This looks great. I'm looking forward to implement this for Java 😄 

> Should we read the overriden symbols in scalacp ? Not sure how to get that information from scalap

I don't think that's necessary. That would require us to re-implement asSeenFrom like in Metals. Scalacp doesn't have access to the compiler. This means that `overrides` are not reliably available for Scalafix rules but I think it's OK.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolInformationOps.scala
tgodzik: Good idea, added. Let me know what you think @olafurpg

olafurpg: Looks great!

PR #2235, opened 2021-02-08
Fix syntax errors

tgodzik:
Fixes two issues:
- `case case` statement that cause InvariantException now shows a nice error
- `()` literal type is no longer supported in any dialect (checked with 2.13.4) and added a ncie message instead of unreachable.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1270,14 +1270,15 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   }
 
   def makeTupleType(body: List[Type]): Type = {
-    def emptyTupleResult = if (dialect.allowLiteralUnitType) Lit.Unit() else unreachable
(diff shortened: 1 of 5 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1270,7 +1270,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   }
 
   def makeTupleType(body: List[Type]): Type = {
-    def emptyTupleResult = if (dialect.allowLiteralUnitType) Lit.Unit() else unreachable
+    def emptyTupleResult = if (dialect.allowLiteralUnitType) Lit.Unit()
```
kpbochenek, line 1273: if no longer supported I would remove if and just inline syntaxError

kpbochenek (approved):
nice! 👍

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Good idea, thanks!

PR #2237, opened 2021-02-10
Parser: rework significant indentation processing.

dos65:
This PR is mostly about rework. In addition, it fixes `multiline-extension` parsing.

The logic from `loop` method that removes whitespaces and produces `Intent`/`Outdent` is moved on `TokenIterator` level.
It observes tokens on demand that adds an ability to put indentation on the parser level using `observeIndented` method.
Dotty parser uses a similar approach.

So, instead of trying to predict __if this `LF` that comes after `:` should add Indent or not__ now it's handled at the moment of parsing template. The same for extension methods.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -231,45 +231,6 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   // NOTE: Special case for Enum region is needed because parsing of 'case' statement is done differently
   case object RegionEnumArtificialMark extends SepRegion
 
-  class AllowColonEol {
-    private var unclosed = 0
-    def unclosedParensCount = unclosed
-    def openParen = unclosed += 1
-    def closeParen = unclosed -= 1
-  }
-
-  /* Heuristic to try to determine if colonEol can exist at a given point in code
-   * needed to deal with situations like:
-   * def a(b:
-   * Int)
-   */
-  def canBeFollowedByColonEol(
-      colonEolCanStartIndent: List[AllowColonEol],
-      curr: Token
-  ): List[AllowColonEol] = {
-    if (dialect.allowSignificantIndentation)
-      colonEolCanStartIndent match {
-        case list if curr.is[CanStartColonEol] =>
-          new AllowColonEol :: list
-        // This means we closed more parents than opened, which means we are outside of a region where colonEol is allowed
-        case allowColonEol :: tail if allowColonEol.unclosedParensCount < 0 =>
-          tail
-        // This means we found a token that should not exist before colonEol
-        case allowColonEol :: tail
-            if (curr.is[LeftBrace] || curr
-              .is[DclIntro]) && allowColonEol.unclosedParensCount == 0 =>
-          tail
-        case list @ allowColonEol :: tail if curr.is[LeftParen] =>
-          allowColonEol.openParen
-          list
-        case list @ allowColonEol :: tail if curr.is[RightParen] =>
-          allowColonEol.closeParen
-          list
-        case allow => allow
-      }
-    else Nil
-  }
-
   // NOTE: Scala's parser isn't ready to accept whitespace and comment tokens,
   // so we have to filter them out, because otherwise we'll get errors like `expected blah, got whitespace`
   // However, in certain tricky cases some whitespace tokens (namely, newlines) do have to be emitted.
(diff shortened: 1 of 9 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -279,24 +240,74 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+  }
+
+  case class TokenRef(
+      token: Token,
+      pos: Int,
+      nextPos: Int
+  )
 
-    @tailrec def loop(
+  class NewTokenIterator(
```
dos65, line 267: I have no idea how to name this Iterator 
The previous one was called `SimpleTokenIterator` but this name doesn't fit anymore :)

Is there a need to keep `TokenIterator` interface?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -279,24 +240,74 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+      scannerTokens: Tokens,
+      var sepRegions: List[SepRegion],
+      var curr: TokenRef,
+      var prevPos: Int
+  ) extends TokenIterator {
+
+    override def hasNext: Boolean = curr.token.is[EOF]
```
tgodzik, line 274:
```suggestion
    override def hasNext: Boolean = curr.token.isNot[EOF]
```
?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: `LazyTokenIterator` ?

dos65: That one is good but what do you think about removing `TokenInterator` interface? Then this name might be used for the current implementation.
Is it used anywhere outside of `ScalametaParser`?

dos65:
__Upd__
Also implemented indented `if` clauses (without `then`)
Now the following parsed correctly:
```scala
if (cond)
  f
  g
```

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -279,185 +240,226 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+          true
+        } else false
+      }
+    }
+
+    def observeIndented(): Boolean = {
+      observeIndented0 { (i, prev) =>
+        val undoRegionChange =
+          prev.headOption match {
+            case Some(RegionParen) if token.is[LeftParen] => prev.tail
```
tgodzik, line 311: We should have `RegionIndent` and the `RegionParen` next - instead of just dropping it.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -279,185 +240,226 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+            case Some(RegionEnumArtificialMark) if token.is[KwEnum] => prev.tail
+            case Some(_: RegionBrace) if token.is[LeftBrace] => prev.tail
```
tgodzik, line 313: Same case as with the above comment. We should still have RegionBrace, but most likely change the order.

`RegionBrace :: RegionIndent`

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -279,185 +240,226 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+        } else false
+      }
+    }
+
+    def observeIndented(): Boolean = {
+      observeIndented0 { (i, prev) =>
+        val undoRegionChange =
+          prev.headOption match {
+            case Some(RegionParen) if token.is[LeftParen] => prev.tail
+            case Some(RegionEnumArtificialMark) if token.is[KwEnum] => prev.tail
```
tgodzik, line 312: This should not pop up aside from enum, isn't thi shandled by `observeIndentedEnum`

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -279,185 +240,226 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
       val curr = scannerTokens(currPos)
+      val currentIndent = countIndent(currPos)
       val nextPos = {
         var i = currPos + 1
         while (i < scannerTokens.length && scannerTokens(i).is[Trivia]) i += 1
         if (i == scannerTokens.length) i = -1
         i
       }
       val next = if (nextPos != -1) scannerTokens(nextPos) else null
+      // println(s"--curr ${curr.token} ${curr.token.getClass()} $sepRegions ${hashCode()}")
```
tgodzik, line 346:
```suggestion
      
```

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -279,185 +240,226 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
-            // Now if we have token 'case' and top of stack is '$' we know it is Enum-case.
-            // In any other case it is 'match-case' or 'try-case'
+      def mkIndent: TokenRef =
+        TokenRef(
+          new Indentation.Indent(curr.input, curr.dialect, curr.start, curr.end),
+          prevPos,
+          currPos
+        )
+
+      def mkOutdent: TokenRef =
```
tgodzik, line 361:
```suggestion
      def prependOutdent: TokenRef =
```
?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2071,8 +2129,10 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
         in = forked
         cond = expr()
       }
+      cond
```
tgodzik, line 2132: this looks unused

tgodzik (changes requested):
Really great work @dos65 ! It also looks a bit more clear, which I did not think is possible :sweat_smile: 

Left a few comments, but nothing major.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: In my sense `prepend`  doesn't fit there. The new implementation can't prepend, it only returns the next token based on the current state.

If you are unsure about shortened `make` what if it will be `makeOutdent`/`createOutdent` or maybe just `outdent`?

tgodzik: Just a suggestion, you can ignore me really here.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: It will be applied one more time.
After `observerIndented` were called the next token after `Indent` will be `(` that will put `RegionParen` again.

For example, in the code below at the moment when `observeIndented` is called iterator already reached `(` token on the second line and updated `sepRegions` on `RegionParen :: Nil`.

```scala
if (cond)
  <(>f)
```

tests/shared/src/test/scala-2.13/scala/meta/tests/parsers/dotty/Scala3PositionSuite.scala
```diff
@@ -163,6 +163,8 @@ class Scala3PositionSuite extends BasePositionSuite(dialects.Scala3) {
     """|Enumerator.CaseGenerator case a: TP <- iter
        |Pat.Typed a: TP
        |Enumerator.Guard if cnd
+       |Term.Name do
```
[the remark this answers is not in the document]
dos65: Fixed

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: There is the same case as with `RegionParen`
Even if it might be strange but it possible to define `enum` inside `if`

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: The same as above

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/ControlSyntaxSuite.scala
```diff
@@ -198,6 +198,46 @@ class ControlSyntaxSuite extends BaseDottySuite {
...
+      Term.If(
+        Term.Name("cond"),
+        Term.Block(List(Term.Name("fx1"), Term.Name("fx2"))),
+        Lit.Unit()
+      )
+    )
+  }
+
+  test("new-if-else-indented") {
+    val code = """|if cond
```
tgodzik, line 221: I think this should not be accepted, but I will follow up with a PR later.

tgodzik (approved):
LGTM! Great work! Looks like we are almost on par with the Dotty parser!

PR #2242, opened 2021-02-16
[Scala 3] Adjust extension methods to 3.0.0-RC1

tgodzik:
Extension methods can now:
- have using parameter before the main parameter
- each def can now have a separate generic type parameter


kpbochenek (approved):
Nice work!

(should be okay but could you double check community-tests still pass? :P )

PR #2243, opened 2021-02-17
[Scala 3] Fix minor issues and make dotty community test run on CI

tgodzik:
This also fixes issues:
- `type T` being the only thing in an indented block
- given patterns work properly with functions
- typed pattern is not always allowed at the end, which is a change from Scala 2
- no longer allow for `if cond <indentation>`

### DIFF
--- a/community-test/src/test/scala/CommunityDottySuite.scala
+++ b/community-test/src/test/scala/CommunityDottySuite.scala
@@ -67,8 +67,8 @@ class CommunityDottySuite extends FunSuite {
     ),
     CommunityBuild(
       "https://github.com/scalameta/munit.git",
-      // latest commit from 27.11.2020
-      "5f384b548960ee7731649f5f900eb1d854e7827a",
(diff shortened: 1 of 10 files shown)


community-test/src/test/scala/CommunityDottySuite.scala
```diff
@@ -68,7 +68,7 @@ class CommunityDottySuite extends FunSuite {
     CommunityBuild(
       "https://github.com/scalameta/munit.git",
       // latest commit from 27.11.2020
```
kpbochenek, line 70: might be good to keep date synced with hash.
Or if you find it not useful maybe just removing this comment

kpbochenek (approved):
LGTM! So close to be fully compliant with compiler :D

(one minor question/suggestion)

community-test/src/test/scala/CommunityDottySuite.scala
tgodzik: Fixed!

PR #2244, opened 2021-02-17
[Scala 3] Allow to have some of the keywords on the same indentation level as the preceeding block

tgodzik:
This PR also fixes any related issues that were still not working in case dotty source code, which includes:
- indentation start in `{ a => ` block
- unicode escapes like `f"\u$oct%04x"`

### DIFF
--- a/community-test/src/test/scala/CommunityDottySuite.scala
+++ b/community-test/src/test/scala/CommunityDottySuite.scala
@@ -60,8 +60,8 @@ class CommunityDottySuite extends FunSuite {
   val communityBuilds = List(
     CommunityBuild(
       "https://github.com/lampepfl/dotty.git",
-      //commit hash from 16.12.2020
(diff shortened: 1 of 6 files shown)


community-test/src/test/scala/CommunityDottySuite.scala
```diff
@@ -156,11 +156,7 @@ class CommunityDottySuite extends FunSuite {
 
   final def dottyExclusionList = List(
     // [scalameta] erased modifier - for now used internally, will be available in 3.1
-    "library/src/scala/compiletime/package.scala",
-    // if then - else without outdentation before else.
-    // it's unclear what to do in this case
-    // https://github.com/lampepfl/dotty/issues/10372
```
kpbochenek, line 162: I would close this issue as we actually support this syntax anyway, WDYT?

tgodzik: Good idea, closed it despite it being a bit weird.

PR #2245, opened 2021-02-18
remove antiquated sbt launch script

SethTisue:
this one is realllllly old, and anyway it's bad practice (IMO) not to just use the standard launch script, hence I'm suggesting removing it rather than refreshing it


tgodzik (approved):
Thanks! I do agree it makes sense to remove it instead of maintaining a custom script.

Issue #2246, opened 2021-02-18
`testkit/test` fails, and is not tested in CI


SethTisue:
in a fresh checkout:

```
sbt:scalametaRoot> testkit/test
scala.meta.tests.prettyprinters.PrettyPrinterSuite:
  + val∙x∙=∙1 0.271s
  + (_:∙Throwable)∙⇒∙1 0.042s
  + 1∙join∙(()) 0.016s
  + foo∙match{∙case∙_∙=>∙_∙=>∙false} 0.047s
scala.meta.tests.contrib.ScaladocParserProps:
  + parser does not crash 1.843s
scala.meta.tests.transversers.TraverserOrder:
==> X scala.meta.tests.transversers.TraverserOrder.traversal order is preserved  0.879s munit.FailException: /Users/tisue/scalameta/scalameta/testkit/src/test/scala/scala/meta/tests/transversers/TraverserOrder.scala:27 assertion failed
26:    }
27:    assert(errors.isEmpty)
28:  }
    at munit.FunSuite.assert(FunSuite.scala:11)
    at scala.meta.tests.transversers.TraverserOrder.$anonfun$new$1(TraverserOrder.scala:27)
```

in `.github/workflows/ci.yml` I see `"++2.13.4 testsJVM/test"` but apparently that doesn't include the testkit tests, hence CI didn't catch this

(but the Scala community build did!)

tgodzik:
Thanks for reporting! I ignored the test for now in https://github.com/scalameta/scalameta/pull/2248

Traverser does not currently work with the new fields we added for Scala 3.

PR #2248, opened 2021-02-18
Add testkit/test to CI and ignore failing test

tgodzik:
Traversal doesn't work correctly for @binaryCompatFields currently. This will be an issue for anyone using it for Scala 3 sources, but should not be a problem otherwise. To fix it we need to make traversal work  with those fields or bump the version to 5.0.0 and turn all of those fields to real ones.

Fixes https://github.com/scalameta/scalameta/issues/2246

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,7 @@ jobs:
           - "++2.13.3 testsJVM/test"
           - "++2.13.4 testsJVM/test"
           - "++2.13.4 testsJS/test"
+          - "++2.13.4 testkit/test"
           - "++2.13.4 download-scala-library testsJVM/slow:test"
           - "communitytest/test"
           - "mima"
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -23,6 +23,7 @@ jobs:
           - "++2.13.3 testsJVM/test"
           - "++2.13.4 testsJVM/test"
           - "++2.13.4 testsJS/test"
+          - "++2.13.4 testkit/test"
```
olafurpg, line 26: Can we move these tests to the testsJVM project?

olafurpg (approved):
LGTM. I prefer having all tests in dedicated test projects. If the tests are slow to run then we we can tag them as such so they don’t run by default.

.github/workflows/ci.yml
tgodzik: Moved!

PR #2255, opened 2021-02-23
Update Scala to 2.13.5

sh0hei:


tgodzik (changes requested):
Thanks for the contribution! There is a bit more needed here - we also need to copy the semanticdb/scalac/library/src/main/scala-2.13.4 to scala-2.13.5 and update the `InteractiveSuite`

I can add that if you don't have time right now.

sh0hei:
Thanks for your comment @tgodzik !

> Thanks for the contribution! There is a bit more needed here - we also need to copy the semanticdb/scalac/library/src/main/scala-2.13.4 to scala-2.13.5 and update the InteractiveSuite

I'll challenge it  now 👍

.github/workflows/ci.yml
```diff
@@ -23,7 +23,9 @@ jobs:
           - "++2.13.3 testsJVM/test"
           - "++2.13.4 testsJVM/test"
           - "++2.13.4 testsJS/test"
```
tgodzik, line 25: We can remove this test - we run scala js only on the newest one.

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/InteractiveSuite.scala
```diff
@@ -110,7 +110,7 @@ class InteractiveSuite extends FunSuite {
     // typer hijacking is not working as expected with interactive.Global.
     expectedPrevious,
     compat = List(
-      "2.13.4" -> expectedLatest,
+      "2.13.5" -> expectedLatest,
```
tgodzik, line 113: I would leave `2.13.4` key still here. The test will fail otherwise.

tgodzik (changes requested):
We should also bump the scala-js plugin in `project/plugins.sbt` and maybe munit - like in the scala steward PR

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/InteractiveSuite.scala
```diff
@@ -93,7 +93,7 @@ class InteractiveSuite extends FunSuite {
   )
 
   val expectedPrevious213 = expected(
-    10,
+    11,
```
tgodzik, line 96: I think this should stay as is and just add the additional key below.

sh0hei:
@tgodzik I submitted the following PR as a preliminary preparation 🙏 
- https://github.com/scalameta/scalameta/pull/2256
- https://github.com/scalameta/scalameta/pull/2257

tgodzik:
> @tgodzik I submitted the following PR as a preliminary preparation 
> 
> * #2256
> * #2257

The updates together with the changes from the comment should make the CI green :+1:

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/InteractiveSuite.scala
tgodzik: We should also add `2.13.5` key, sorry if I wasn't clear about it. It should look like:
```
      "2.13.4" -> expectedLatest,
      "2.13.5" -> expectedLatest,
```

tgodzik: I added the small change myself to unblock.

tgodzik (approved):

PR #2261, opened 2021-03-03
Add support for Scala Native 0.4.0

tgodzik:
CC @ekrich


WojciechMazur (approved):
Looks good, `linkStubs(false)` and `gc(immix)` in nativeSettings can be ommited since they're using default values, but that just personal preference

tgodzik:
> Looks good, `linkStubs(false)` and `gc(immix)` in nativeSettings can be ommited since they're using default values, but that just personal preference

Removed it then and changed to releaseFast, since it seems that full release is not finishing on scalameta!

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -266,19 +266,22 @@ object TreeSyntax {
         }
         looksLikePatVar && thisLocationAlsoAcceptsPatVars
       }
-      // soft keywords might need to be written with backquotes in some places
+      /* Soft keywords might need to be written with backquotes in some places.
+       * Previously used match clause fails due to:
+       * https://github.com/scala-native/scala-native/issues/2187
```
olafurpg, line 271: Interesting! Nice job with finding a minimal repro!

olafurpg (approved):
LGTM 🚀 

I'm really excited to compare the performance of Scalafmt with Scala Native vs. native-image 😁

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
tgodzik: that was all @WojciechMazur :tada:

WojciechMazur:
> I'm really excited to compare the performance of Scalafmt with Scala Native vs. native-image
For performance comparison, we need to wait for SN 0.4.1, many performance regressions were fixed there. As an example, databricks/sjsonnet works around 2x faster with the current 0.4.1-SNAPSHOT in comparison to 0.4.0.

PR #2262, opened 2021-03-03
Add documentation field to SymbolInformation

Strum355:
This PR arose as a result of @olafurpg and mine's work on https://github.com/sourcegraph/lsif-java and it's underlying use of the semanticdb format.

In order to support displaying javadoc on the hovering of symbols, we need to store the javadoc string within the LSIF dump file, which is emitted by https://github.com/sourcegraph/lsif-semanticdb from the input semanticdb files. To accomplish this, we add the `documentation` string field to `SymbolInformation`. The expected string should not contain comment-denoting leading characters such as `/**`, leading `*` etc (should this be detailed in the spec?).

I've been made aware that in the current state of the Scala side of the semanticdb ecosystem, this field wouldn't be populated for Scala, so this is (for the time being) more oriented towards Java use of semanticdb. Hopefully this can change in the future :slightly_smiling_face:


docs/semanticdb/specification.md
```diff
@@ -810,6 +811,12 @@ signatures in supported languages.
 
 `access`. [Access](#access) modifier of the corresponding definition.
 
+`documentation`. The documentation string associated with a symbol.
```
olafurpg, line 814: I'm wondering if it makes sense to specify the format of the string in a separate field 🤔 

```proto
message SymbolInformation {
  // ...
  string documentation = 20;
  DocumentationFormat documentationFormat = 21;
}

enum DocumentationFormat {
  MARKDOWN = 0;
  HTML = 1;
  PLAINTEXT = 2;
  JAVADOC = 3;
  SCALADOC = 4;
}
```
What do you think?

docs/semanticdb/specification.md
```diff
@@ -810,6 +811,12 @@ signatures in supported languages.
...
+
+The availability of this information depends on the underlying method used to
```
olafurpg, line 816: This paragraph can be removed. All protobuf fields are optional and it's normal that different indexers emit different fields. It's the responsibility of each indexer to document what fields they emit.

docs/semanticdb/specification.md
```diff
@@ -1471,6 +1479,10 @@ message SymbolInformation {
     <td><code>access</code></td>
     <td>Explained below on per-definition basis.</td>
   </tr>
+  <tr>
+    <td><code>documentation</code></td>
+    <td>Functionality not exposed by the Scala compiler plugin API.</td>
```
olafurpg, line 1484:
```suggestion
    <td>Always empty. Not supported by semanticdb-scalac and metacp.</td>
```

olafurpg: FWIW, `metacp` generates `SymbolInformation` from a classpath alone (no sources)

semanticdb/semanticdb3/semanticdb3.md
```diff
@@ -632,6 +632,7 @@ message SymbolInformation {
   Signature signature = 17;
   repeated Annotation annotations = 13;
   Access access = 18;
+  string documentation = 20;
```
olafurpg, line 635: This can be removed. This file is the legacy version of the protocol. The file can honestly be removed, but no need to worry about that here.

tests/jvm/src/test/resources/metac-metacp.diff
```diff
@@ -494,6 +494,7 @@ example/Synthetic#x$2.
  access {
 -  public_access {
```
olafurpg, line 495: What caused this diff?

olafurpg (changes requested):
Thank you for this contribution! It's cumbersome to copy-paste the changes in so many places but it's the best approach we have for now, for better or worse :/

A few minor comments and one question about the format, otherwise this looks great!

tests/jvm/src/test/resources/metac-metacp.diff
Strum355: Thats not a diff caused by this PR, its just github rendering it with a red outline :sweat_smile:

olafurpg: lololol

Screenshot 2021-03-03 at 16 40 14


This diff file always trips me up during review.

docs/semanticdb/specification.md
Strum355: I was thinking about something like this, but only denoting the language that it comes from. Ultimately decided against it for the following reasons:

1. Looking at the different JVM language doc types (Javadoc, Scaladoc, KDoc, Groovydoc etc), they all seem to follow the baseline format and features. Hypothetically, if javadoc was the baseline, one could simply use a javadoc formatter (with *X*doc features copied verbatim)
2. Giving the responsibility of what format it should be to the consumer of a SemanticDB file rather than having them support all possible kinds. While technically you could convert from any format to the rest, consumers would probably prefer not to have to support them all, but be able to decide themselves based on it being _at least_ Javadoc. For additional features present in e.g. Scaladoc, one can switch based on the `langauge` field.

olafurpg: There's quite a big difference between Scaladoc and Javadoc. My motivation to add the format is that there can be different indexers for the same language that emit different formats. For example, you could implement a tool that takes SemanticDB as input and converts Javadoc into HTML/Markdown.

I would love to require this field to be markdown formatted but it's actually quite complicated to render docstrings correctly (subclasses can inherit from superclasses, Scaladoc allows interpolating variables that can be defined in subclasses, ...). A compiler plugin may not have all of the necessary information.


cc/ @tgodzik do you have any thoughts?

olafurpg: Alternative design to avoid the awkward `documentation` and `documentationFormat` sibling fields
```
message SymbolInformation {
  // ...
  Documentation documentation = 20;
}

message Documentation {
  enum Format {
    HTML = 0;
    MARKDOWN = 1;
    JAVADOC = 2;
    SCALADOC = 3;
    KDOC = 4;
  }
  string message = 0;
  Format format = 1;
}
```

tgodzik: I like the second option and I think it would be super useful to have DocumentationKind similar to `MarkupKind` that we have in LSP.

Strum355: Updated spec with @olafurpg's example. Let me know if the associated spec docs are as expected :slightly_smiling_face:

olafurpg (approved):
LGTM 👍

@tgodzik What do you think about the new format?

tgodzik (approved):
Looks good!

Strum355:
`sbt ++2.13.5 testsJVM/slow:test` is failing with the following, not sure how thats related to this PR
```
==> X scala.meta.tests.semanticdb.MetacScalaLibrary.compile scala-library  0.009s java.lang.AssertionError: assertion failed: /home/noah/Scala/scalameta/target/scala-library/scala-2.13.5/src/library is not a directory! Run `sbt download-scala-library`
    at scala.Predef$.assert(Predef.scala:280)
    at scala.meta.tests.semanticdb.MetacScalaLibrary$.process(MetacScalaLibrary.scala:37)
    at scala.meta.tests.semanticdb.MetacScalaLibrary.$anonfun$new$1(MetacScalaLibrary.scala:18)
```

olafurpg:
@Strum355 that failure can be ignored, it's unrelated to this PR. I'm not sure what's going on there 🤔

olafurpg:
@tgodzik should we ignore the CI failures and merge this?

tgodzik:
> @tgodzik should we ignore the CI failures and merge this?

I think so, the failures seem related to actions themselves.

PR #2264, opened 2021-03-19
[Scala 3] Allow return statement to contain indented block

tgodzik:
Previously, return could not start an indentation, but starting with https://github.com/lampepfl/dotty/pull/11752 it can. This fixes scalameta to parse accordingly.

Additionally, this change also created some issues with empty `return` statements in pattern matches. We would incorrectly think that the next statement is indented, because we never saved case block indentation. I added a new CaseRegion in this PR, so that we can remember the indentation. This is only added in `match` with braces as indented match clauses work correctly here.

### DIFF
--- a/community-test/src/test/scala/CommunityDottySuite.scala
+++ b/community-test/src/test/scala/CommunityDottySuite.scala
@@ -60,15 +60,15 @@ class CommunityDottySuite extends FunSuite {
   val communityBuilds = List(
     CommunityBuild(
       "https://github.com/lampepfl/dotty.git",
-      //commit hash from 17.02.2021
-      "8bbb0ba745453a0d2ffdb94c5966752dfe57bb5f",
+      //commit hash from 31.03.2021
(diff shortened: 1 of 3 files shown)


kpbochenek:
looks like community tests found something

tgodzik:
> looks like community tests found something

Yeah, trying to fix that, but seems a bit more complex. Not neccessarily related to returns.

community-test/src/test/scala/CommunityDottySuite.scala
```diff
@@ -154,10 +154,7 @@ class CommunityDottySuite extends FunSuite {
     build.excluded.exists(el => path.endsWith(el))
   }
 
-  final def dottyExclusionList = List(
-    // [scalameta] erased modifier - for now used internally, will be available in 3.1
```
tgodzik, line 158: This got removed from dotty for now.

dos65 (approved):
LGTM!

PR #2265, opened 2021-03-24
Fix issues with wrongly detected blocks

tgodzik:
The issues were found when using optional braces for scalafmt.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1017,7 +1017,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       token.is[KwThis] || token.is[KwThrow] || token.is[KwTry] || token.is[KwWhile] ||
       token.is[LeftParen] || token.is[LeftBrace] || token.is[Underscore] ||
       token.is[Unquote] || token.is[MacroSplice] || token.is[MacroQuote] ||
-      token.is[Indentation.Indent]
+      token.is[Indentation.Indent] || (token.is[LeftBracket] && dialect.allowPolymorphicFunctions)
     }
   }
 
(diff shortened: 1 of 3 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -581,7 +581,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
                   // if does not work with indentation in pattern matches
                   val shouldNotIndentIf =
                     prev.is[KwIf] && sepRegions.headOption.exists(r => r == RegionArrow)
-                  !shouldNotIndentIf && prev.is[CanStartIndent] && !next.is[RightBrace]
+                  !shouldNotIndentIf && prev.is[CanStartIndent] && !next.is[RightBrace] &&
+                  !next.is[LeftBracket]
```
dos65, line 585: I was curious why it works for `LeftBrace` without a similar clause and found that the actual reason is in `ExprIntro.unapply` 
It doesn't have `token.is[LeftBracket]` clause.

https://github.com/scalameta/scalameta/blob/a51e009563e69591d3f057e7b777ed4f6a224bc1/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L1012-L1021

tgodzik: Good catch! I added `!next.is[CantStartStat]` instead

dos65: Hmm. Sorry for being nonclear.
I think there should be an indent before `RightBrace`.

The sample you tried to fix isn't parsed correctly because of `ExprIntro`. It should have smth like:
 ```scala
  ... || (token.is[LeftBracket] && dialect.allowsPolymorphicFunctions)
```

Without indentation the following will be parsed incorrectly:
```scala
val a: Int = 
  [A] => (x: List[A]) => x.headOption
  42
```

tgodzik: Ach! Right! Changed, should be ok now... :sweat_smile:

dos65 (approved):

PR #2266, opened 2021-03-24
Use fastparse and geny from com.lihaoyi

lolgab:
com.lihaoyi libs are going to be maintained again for Scala 2.11 too


tgodzik:
I created a PR for geny itself here: https://github.com/scalameta/scalameta/pull/2274

Let's leave fastparse as is for now. I am not sure if we won't need to shade everything really.

PR #2275, opened 2021-04-06
ClassSignature specification had swapped self and declaration fields

VlachJosef:
Sync specification docs with actual [definition](https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb/semanticdb.proto#L215-L216).


olafurpg (approved):
Thank you! Good catch

PR #2276, opened 2021-04-08
Fix issues with end match

tgodzik:
There is some logic around optional braces that relies on the match keyword, however it should not take into account `end match`. This is now fixed.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -561,7 +561,9 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
             def needOutdent: Boolean = {
               sepRegions.headOption.exists {
                 case r if r.isIndented =>
-                  if (nextIndent < r.indent) prev.isNot[CanContinueOnNextLine]
+                  // need to check prev.prev in case of `end match`
+                  if (nextIndent < r.indent)
+                    prev.isNot[CanContinueOnNextLine] || prev.prev.is[soft.KwEnd]
                   else r.closeOnNonCase && next.isNot[KwCase] && nextIndent == r.indent
                 case _ => false
               }
@@ -602,7 +604,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
                 } else
                   // always add indent for indented `match` block
                   // check the previous token to avoid infinity loop
-                  prev.is[KwMatch] && next.is[KwCase] && token.isNot[Indentation.Indent]
+                  (!prev.prev.is[soft.KwEnd] && prev.is[KwMatch]) &&
+                  next.is[KwCase] && token.isNot[Indentation.Indent]
               } else false
             }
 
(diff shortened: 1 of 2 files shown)


kitbellew (approved):

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/EndMarkerSuite.scala
```diff
@@ -151,4 +151,78 @@ class EndMarkerSuite extends BaseDottySuite {
       )
     )
   }
+
+  test("if-then-with-end") {
```
dos65, line 155:
```suggestion
  test("match-with-end") {
```

dos65 (approved):

kitbellew:
@tgodzik this one doesn't work:
```
object a:
  val (foo, boo) =
    bar
    baz
  end val
  val (foo, boo) =
    bar
    baz
  end val
```

tgodzik:
I will take a look tomorrow!

dos65:
@tgodzik I almost fixed it

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -602,7 +604,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
                 } else
                   // always add indent for indented `match` block
                   // check the previous token to avoid infinity loop
-                  prev.is[KwMatch] && next.is[KwCase] && token.isNot[Indentation.Indent]
+                  (!prev.prev.is[soft.KwEnd] && prev.is[KwMatch]) &&
+                  next.is[KwCase] && token.isNot[Indentation.Indent]
```
kitbellew, line 608: sorry for being a pain but this is something that i learned the hard way in scalafmt. does prev refer to the previous token, skipping both whitespace and comments?

will this handle something like that:
```
    foo match // blah
    case bar => ...
    end match // blah
  case qux => ...
```

tgodzik: It skips them:
```scala
    def prev: Token = {
      val prev = scannerTokens.apply(Math.max(token.index - 1, 0))
      if (prev.is[Whitespace] || prev.is[Comment]) prev.prev
      else prev
    }
```

kitbellew: what about this:
```
val end = ...
end match
case theend => ...
end match
```

tgodzik: this doesn't compile in Dotty from what I see.

kitbellew: why not? `end` is not restricted https://scalacenter.github.io/scala-3-migration-guide/docs/incompatibilities/syntactic-changes.html#restricted-keywords

tgodzik: Might be a bug? Or really there is a number of scenarios, where soft keywords don't work, but it's not stated.

Same with:
```scala
def extension(a : Int) = a + 2
extension(2)
```
will not work.

You need to do:
```scala
def extension(a : Int) = a + 2
`extension`(2)
```

So soft keywords are kind of dodgy and when I asked Martin he said that this is fine.

tgodzik: https://github.com/lampepfl/dotty/issues/10076 for some context (I kind of argued with him there :sweat_smile: )

PR #2277, opened 2021-04-08
Parser: fix issues with `end` marker syntax for val

dos65:


kpbochenek (approved):
run scalafmt before merging, quick fix thanks :+1:

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1192,7 +1192,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       token.is[Interpolation.End] || token.is[Xml.End] ||
       token.is[KwReturn] || token.is[KwThis] || token.is[KwType] ||
       token.is[RightParen] || token.is[RightBracket] || token.is[RightBrace] ||
-      token.is[Underscore] || token.is[Ellipsis] || token.is[Unquote]
+      token.is[Underscore] || token.is[Ellipsis] || token.is[Unquote] ||
+      token.prev.is[EndMarkerIntro]
```
kitbellew, line 1196: does this change apply to all end markers? should we undo @tgodzik's special-case handling for match? #2276

tgodzik: Just checked and we still need that one.

dos65: These tho cases with `match` and `val` are different

Issue #2279, opened 2021-04-11
Body of an empty catch case includes right arrow


kitbellew:
In this code:
```
object a:
   def foo =
     try foo
     catch
       case a =>
     finally bar
```
the body of catch case is the following tree:
```
TYPE: Term$Block$TermBlockImpl
SOURCE: =>
STRUCTURE: Term.Block(Nil)
TOKENS: =>
```
this leads `scalafmt` to think `=>` starts a statement and format it as:
```
object a:
   def foo =
     try foo
     catch
       case a
         =>
     finally bar
```

tgodzik:
Fixed in https://github.com/scalameta/scalameta/pull/2282

PR #2280, opened 2021-04-12
Fix issues with wrong positions for empty self and block

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/2279

Otherwise, we would include tokens that do not belong to a tree.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -802,6 +802,12 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     def startTokenPos = in.tokenPos
     def endTokenPos = in.prevTokenPos
   }
+
+  case object EmptyPos extends Pos {
+    val startTokenPos = in.tokenPos
+    val endTokenPos = in.tokenPos - 1
+  }
+
   implicit def intToIndexPos(index: Int): IndexPos = IndexPos(index)
   implicit def tokenToTokenPos(token: Token): TokenPos = TokenPos(token)
   implicit def treeToTreePos(tree: Tree): TreePos = TreePos(tree)
@@ -815,6 +821,9 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   def atPos[T <: Tree](token: Token)(body: => T): T = {
     atPos(token.startTokenPos, token.endTokenPos)(body)
   }
+
+  def emptyPos[T <: Tree](body: => T) = atPos(EmptyPos, EmptyPos)(body)
+
   def atPos[T <: Tree](start: Pos, end: Pos)(body: => T): T = {
     val startTokenPos = start.startTokenPos
     val result = body
@@ -3038,6 +3047,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
             blockStatSeq() match {
               case List(q: Quasi) => q.become[Term.Quasi]
               case List(term: Term) => term
+              case Nil => emptyPos(Term.Block(Nil))
               case other => atPos(start, end)(Term.Block(other))
             }
           }
@@ -4860,7 +4870,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   }
 
   def templateStatSeq(enumCaseAllowed: Boolean = false): (Self, List[Stat]) = {
-    val emptySelf = autoPos(Self(autoPos(Name.Anonymous()), None))
+    val emptySelf = emptyPos(Self(emptyPos(Name.Anonymous()), None))
     var selfOpt: Option[Self] = None
     var firstOpt: Option[Stat] = None
     if (token.is[ExprIntro] && !token.is[KwExtension]) {
--- a/tests/shared/src/test/scala-2.13/scala/meta/tests/parsers/dotty/Scala3PositionSuite.scala
+++ b/tests/shared/src/test/scala-2.13/scala/meta/tests/parsers/dotty/Scala3PositionSuite.scala
@@ -301,13 +301,64 @@ class Scala3PositionSuite extends BasePositionSuite(dialects.Scala3) {
        |    42
        |  def b: String =
        |    "b"
-       |Self def
-       |Name.Anonymous def
+       |Self   @@def a: Int =
        |Defn.Def def a: Int =
        |    42
        |Defn.Def def b: String =
        |    "b"
        |Lit.String "b"
+       |""".stripMargin
+  )
+
+  checkPositions[Stat](
+    """|object a:
+       |   def foo =
+       |     try foo
+       |     catch
+       |       case a =>
+       |       case b =>
+       |         stmt1
+       |         stmt2
+       |       case c =>
+       |     finally bar
+       |""".stripMargin,
+    """|Template :
+       |   def foo =
+       |     try foo
+       |     catch
+       |       case a =>
+       |       case b =>
+       |         stmt1
+       |         stmt2
+       |       case c =>
+       |     finally bar
+       |Self    @@def foo =
+       |Defn.Def def foo =
+       |     try foo
+       |     catch
+       |       case a =>
+       |       case b =>
+       |         stmt1
+       |         stmt2
+       |       case c =>
+       |     finally bar
+       |Term.Try try foo
+       |     catch
+       |       case a =>
+       |       case b =>
+       |         stmt1
+       |         stmt2
+       |       case c =>
+       |     finally bar
+       |Case case a =>
+       |Term.Block    @@def foo =
+       |Case case b =>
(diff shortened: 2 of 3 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -279,7 +279,9 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
     override def next(): Token = {
       val (newSepRegions, newTokenRef) = nextToken(curr.pos, curr.nextPos, sepRegions)
-      prevPos = curr.pointPos
+      // don't update prevPos on artificially inserted tokens
+      if (curr.token.isNot[Indentation.Indent] && curr.token.isNot[Indentation.Outdent])
+        prevPos = curr.pointPos
```
dos65, line 284: I think we can also remove `TokenRef.pointPos`.
I added it only to fix smth similar to this issue. For any token exept `Indent` and `Outdent` `pointPos` equals to `TokenRef.pos`

tgodzik: When removing `pointPos` some tests seem to fail. we are still using prevPos set by Indent and Outdent, only on the next token we don't update it, so that we point to the previous pos before Indentation tokens.

dos65: I've tried to run a test on the snippet from the issue and it seems it's seems this fix doesn't help.
```scala
object a:
   def foo =
      try foo
        catch
          case a =>
        finally bar
```

output:
```
=> Obtained
    """|Template :
       |   def foo =
       |     try foo
       |     catch
       |       case a =>
       |     finally bar
       |Self    @@def foo =
       |Defn.Def def foo =
       |     try foo
       |     catch
       |       case a =>
       |     finally bar
       |Term.Try try foo
       |     catch
       |       case a =>
       |     finally bar
       |Case case a =>
       |Term.Block => // <--- `=>` is printed 
       |""".stripMargin

```

tests/shared/src/test/scala-2.13/scala/meta/tests/parsers/dotty/Scala3PositionSuite.scala
```diff
@@ -301,13 +301,58 @@ class Scala3PositionSuite extends BasePositionSuite(dialects.Scala3) {
...
+    """|Template :
+       |   def foo =
+       |     try foo
+       |     catch
+       |       case a =>
+       |       case b =>
+       |         st1
+       |         st2
+       |     finally bar
+       |Self    @@def foo =
```
kitbellew, line 333: why is self pointing to `def`?

tests/shared/src/test/scala-2.13/scala/meta/tests/parsers/dotty/Scala3PositionSuite.scala
```diff
@@ -301,13 +301,58 @@ class Scala3PositionSuite extends BasePositionSuite(dialects.Scala3) {
...
+       |     finally bar
+       |Term.Try try foo
+       |     catch
+       |       case a =>
+       |       case b =>
+       |         st1
+       |         st2
+       |     finally bar
+       |Case case a =>
+       |Term.Block        @@case b =>
```
kitbellew, line 350: shouldn't the block be empty?

kitbellew:
@tgodzik how did the test behave before this change?

tests/shared/src/test/scala-2.13/scala/meta/tests/parsers/dotty/Scala3PositionSuite.scala
tgodzik: It's just an empty Position, it needs to be somewhere so it's at the start of the template, it contains no tokens. It's the same as in the other tests.

tests/shared/src/test/scala-2.13/scala/meta/tests/parsers/dotty/Scala3PositionSuite.scala
tgodzik: It is, @@ just points to where the empty range is located at.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: ok it seems that the fix is not working everywhere :/ Need to dig in more.

tgodzik: Ok, another approach. Setting empty position on purpose.

tgodzik:
> @tgodzik how did the test behave before this change?

The test would include `def` for Self and `=>` for Block.

I changed the problematic places to use empty position on purpose, which means in Scalameta that the end is before the start. This worked in `autoPos` without optional braces, but seems quite complex to get right this way when we add the additional indentation tokens.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3038,6 +3047,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
             blockStatSeq() match {
               case List(q: Quasi) => q.become[Term.Quasi]
               case List(term: Term) => term
+              case Nil => emptyPos(Term.Block(Nil))
```
tgodzik, line 3050: This should not be an issue aside from case clauses, because in most other indented places we can't get an empty Block ended up by outdentation (which comes here from match).

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -4860,7 +4870,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   }
 
   def templateStatSeq(enumCaseAllowed: Boolean = false): (Self, List[Stat]) = {
-    val emptySelf = autoPos(Self(autoPos(Name.Anonymous()), None))
+    val emptySelf = emptyPos(Self(emptyPos(Name.Anonymous()), None))
```
tgodzik, line 4873: Self is the only one that can be empty at a start of Indentation as far as I can tell.

tests/shared/src/test/scala-2.13/scala/meta/tests/tokenizers/TokensPositionSuite.scala
```diff
@@ -642,7 +642,7 @@ class TokensPositionSuite extends BasePositionSuite(dialects.Scala213) {
     "new { val a = 1 } with A {}",
     """|Template { val a = 1 } with A {}
        |Defn.Val val a = 1
-       |Self new { val a = 1 } with A {@@}
+       |Self new { @@val a = 1 } with A {}
```
tgodzik, line 645: Not sure about this change :thinking:

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -802,6 +802,12 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     def startTokenPos = in.tokenPos
     def endTokenPos = in.prevTokenPos
   }
+
+  case object EmptyPos extends Pos {
+    val startTokenPos = in.tokenPos
+    val endTokenPos = in.tokenPos - 1
```
kitbellew, line 808: wouldn't the position be empty if `start == end` (since `end` is excluded)?

tests/shared/src/test/scala-2.13/scala/meta/tests/tokenizers/TokensPositionSuite.scala
kitbellew: probably wrong

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: I think it's only empty if the end is before the start. That happens with autoPos if ther is nothing to parse for the body.

dos65:
@tgodzik 
I found that there is an issue with `pointPos` in `Indent` and `Outdent`. It seems that at least they have to point on `LF` token in `scannerTokens` because they actually produced at the moment oh handling newline.

I have a [fix](https://github.com/dos65/scalameta/commit/8af19a9290d2fff6d92aa5903cdc3037be84bc35) for that.
However, it isn't fully correct as there are more `LF` that appears in token ranges.

tgodzik:
Thanks @dos65 ! I merged your fix!

PR #2281, opened 2021-04-12
Expand semanticdb spec to improve annotation modeling 

Strum355:
This PR expands the SemanticDB spec to allow for representing Annotation AST's (incl. parameters) by introducing two new types to the `Tree` family: `AnnotationTree` and `AssignTree`. 

As `AnnotationTree` is binary compatible with the `Annotation` type, this proposal deprecates `Annotation` in favour of `AnnotationTree`, replacing its usages in `SymbolInformation` and `AnnotatedType`. Alternatively, I would also be open to adding `Annotation` to the `Tree` oneof instead, if source-level backwards compatibility is also strongly desired.

The specification doc is updated to include some more examples, given the new capabilities to also model the parameters of annotations.

Working example of rendering annotations from the structures built via the changes in this PR can be found at https://github.com/sourcegraph/lsif-java/pull/148

### DIFF
--- a/docs/semanticdb/specification.md
+++ b/docs/semanticdb/specification.md
@@ -94,10 +94,10 @@ SemanticDB payloads must include the version of the SemanticDB model in the
   <tr>
     <td><b>Version</b></td>
     <td><b>Explanation</b></td>
-    <td><b>Data model<b></td>
+    <td><b>Data model</b></td>
   </tr>
   <tr>
-    <td><code>LEGACY<code></td>
+    <td><code>LEGACY</code></td>
     <td>Legacy SemanticDB payloads</td>
     <td><a href="https://github.com/scalameta/scalameta/blob/v3.0.0/semanticdb/semanticdb2/semanticdb2.proto">semanticdb2.proto</a></td>
   </tr>
@@ -471,13 +471,13 @@ and `declarations`. Declarations are modelled by a [Scope](#scope).
 ```protobuf
 message AnnotatedType {
   reserved 2;
-  repeated Annotation annotations = 3;
+  repeated AnnotationTree annotations = 3;
   Type tpe = 1;
 }
 ```
 
 `AnnotatedType` represents a type `tpe` annotated by one or more
-[Annotations](#annotation).
+[AnnotationTrees](#annotation-tree).
 
 ```protobuf
 message ExistentialType {
@@ -597,7 +597,7 @@ message SymbolInformation {
   int32 properties = 4;
   string display_name = 5;
   Signature signature = 17;
-  repeated Annotation annotations = 13;
+  repeated AnnotationTree annotations = 13;
   Access access = 18;
   repeated string overridden_symbols = 19;
   Documentation documentation = 20;
@@ -808,7 +808,7 @@ definitions have which display and symbol names in supported languages.
 See [Languages](#languages) for more information on which definitions have which
 signatures in supported languages.
 
-`annotation`. [Annotations](#annotation) of the corresponding definition.
+`annotation`. [AnnotationTree](#annotationtree) of the corresponding definition.
 
 `access`. [Access](#access) modifier of the corresponding definition.
 
@@ -832,19 +832,22 @@ message Documentation {
 }
 ```
 
-`Documentation` represents the documentation associated with a [Symbol](#symbol). The `format` fields
+`Documentation` represents the documentation associated with a [Symbol](#annotation-deprecated). The `format` fields
 specifies the format of the text stored in `message`.
 
-### Annotation
+### Annotation (deprecated)
 
 ```protobuf
 message Annotation {
+  option deprecated = true;
   Type tpe = 1;
 }
 ```
 
 `Annotation` represents annotations. See [Languages](#languages) for information
 on how annotations in supported languages map onto this data structure.
+
+Superseded by [AnnotationTree](#annotation-tree).
 
 ### Access
 
@@ -1091,6 +1094,29 @@ message TypeApplyTree {
 
 A `TypeApplyTree` represents the type application of a method, providing that
 method with type arguments.
+
+<a name="annotation-tree"></a>
+
+```protobuf
+message AnnotationTree {
+  Type tpe = 1;
+  repeated Tree parameters = 2;
+}
+```
+
+An `AnnotationTree` represents an annotation with parameters, if any. The
+`parameters` would normally be `AssignTree` in Java where the parameters
+generally are required to be assignment expressions, or else any other
+`Tree`.
+
+```protobuf
+message AssignTree {
+  Tree lhs = 1;
+  Tree rhs = 2;
+}
+```
+
+An `AssignTree` represents an assignment expression.
 
 ## Data Schemas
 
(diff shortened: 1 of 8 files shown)


semanticdb/semanticdb/semanticdb.proto
```diff
@@ -278,7 +278,7 @@ message SymbolInformation {
   int32 properties = 4;
   string display_name = 5;
   Signature signature = 17;
-  repeated Annotation annotations = 13;
+  repeated AnnotationTree annotations = 13;
```
olafurpg, line 281:
```suggestion
  repeated AnnotationTree annotations = 13; // NOTE: it's possible to manually change the type to `Annotation` for source compatibility with older SemanticDB versions.
```

semanticdb/semanticdb/semanticdb.proto
```diff
@@ -118,7 +118,7 @@ message StructuralType {
 
 message AnnotatedType {
   reserved 2;
-  repeated Annotation annotations = 3;
+  repeated AnnotationTree annotations = 3;
```
olafurpg, line 121:
```suggestion
  repeated AnnotationTree annotations = 3; // NOTE: it's possible to manually change the type to `Annotation` for source compatibility with older SemanticDB versions.
```

semanticdb/semanticdb/semanticdb.proto
```diff
@@ -413,3 +416,13 @@ message TypeApplyTree {
...
+
+message AnnotationTree {
+  Type tpe = 1;
+  repeated Tree parameters = 2;
+}
+
+message AssignTree {
+  Tree lhs = 1;
+  Tree rhs = 2;
+}
```
olafurpg, line 428: Let's add a newline at the end of file
```suggestion
}

```

docs/semanticdb/specification.md
```diff
@@ -3573,30 +3623,55 @@ modelled in [Scala symbols](#scala-symbol).
...
+  </tr>
+  <tr>
+    <td>
+      <code>
+        AnnotationTree(TypeRef(None, &lt;EnumConstArg#&gt;, List()),
+        List(AssignTree(IdTree(&lt;EnumConstArg#value().&gt;),
+          SelectTree(IdTree(&lt;X#&gt;), IdTree(&lt;X#CONST.&gt;))))
+      </code>
+    </td>
+    <td><code>@EnumConstArg(X.CONST)</code></td>
```
olafurpg, line 3674: Can we add one example for `AssignTree` and one example for nested `AnnotationTree`

semanticdb/semanticdb/semanticdb.proto
olafurpg: @tgodzik @gabro I'm curious to hear your thoughts about this Protobuf binary-compatible but Scala binary+source-incompatible change. It's possible to keep the old `Annotation` signature to preserve source compatibility but the new `AnnotationTree` is more consistently named with other `Tree` subclasses. 

A few sidenotes:

- The ScalaPB-generated classes have intentionally never been marked as public APIs exactly so that we can make Protobuf binary-compatible changes like this.
- SemanticDB annotations are not really used by any meaningful Scalameta project from what I can tell because they are incomplete for Scala (no parameters until this PR) and they were officially unsupported for Java.
- Scalafix will be able to incorporate these changes into its public API without breaking changes.

WDYT?

olafurpg (changes requested):
Thank you for this contribution. This is a fantastic milestone! Lack of comprehensive support for annotations has always been a weak spot in SemanticDB and it's really exciting to see that we now have a complete implementation for Java annotations.

I have a few minor comments, otherwise this looks good to me 👍

kitbellew:
still relevant?

Strum355:
> still relevant?

We've been using this in https://github.com/sourcegraph/scip-java for the past ~2 years, along with https://github.com/scalameta/scalameta/pull/2292 which also has not been merged here yet

tgodzik:
I think the issue was the `Scala binary+source-incompatible change` and we never really settled if it's ok to break it. I think the biggest issue would be with Scalafix, @bjaglin do you think the breaking change for semanticdb in this PR might cause issues for Scalafix? 

Maybe we could keep Annotation at least as an existing class and just AnnotationTree as a completely new thing? This might break some logic, but not break binary compat.

bjaglin:
That change _looks_ non-breaking from a Scalafix perspective as ScalaPB classes are not leaking to the public Scalafix API (they are all converted to case classes manually-defined in `scalafix.v1`).

When parameters are actually filled by semanticdb-scalac (since they don't seem in this PR?), we'd have to see if it is possible to expose `scalafix.v1.AnnotationTree` through ` scalafix.v1.AnnotatedType` in a smooth way, but that's another story.

kitbellew:
> > Not sure though how things are protobug compatible
> 
> `AnnotationTree`, at the protobuf byte-level, is essentially just an extension of `Annotation`: the latter had a field at index 1 of type `Type`, the former also has a field at index 1 of type `Type` and an additional (repeated) field at index 2 of type `Tree`. So long we're not changing the type of any existing fields of `Annotation` in `AnnotationTree` to something that is recursively of a different type (as type _names_ don't matter here, only their byte representation), then itll deserialize successfully. Hope that answers the question 🙂

does this mean we don't need to create a new type, and can stick with the old one? protobuf is backwards-compatible when it comes to _adding_ new fields, is our usage of it as well?

Strum355:
> > > Not sure though how things are protobug compatible
> > 
> > 
> > `AnnotationTree`, at the protobuf byte-level, is essentially just an extension of `Annotation`: the latter had a field at index 1 of type `Type`, the former also has a field at index 1 of type `Type` and an additional (repeated) field at index 2 of type `Tree`. So long we're not changing the type of any existing fields of `Annotation` in `AnnotationTree` to something that is recursively of a different type (as type _names_ don't matter here, only their byte representation), then itll deserialize successfully. Hope that answers the question 🙂
> 
> does this mean we don't need to create a new type, and can stick with the old one? protobuf is backwards-compatible when it comes to _adding_ new fields, is our usage of it as well?

I guess technically? I dont remember if there was a reason beyond having the naming more aligned with the rest of the `*Tree` types, its been a while since this was opened.

This change is _byte_ compatible, yes, but not _source_ compatible. Scala code would most likely complain as theyre different types in the generated code, but programs using either the old or the new schema can communicate with each other just fine

aicodeguard:
### Code Review – Key Findings

**Summary:**
The review identified a significant breaking change in the Protobuf definitions. The replacement of the `Annotation` message with `AnnotationTree` in core data structures will break source compatibility for all consumers of the generated code. While the wire format remains largely compatible due to field index preservation, the API surface change is substantial.

## Issues

1. **[Maintainability]** Breaking Protobuf API Change
   - **File:** `semanticdb/semanticdb/semanticdb.proto` (Lines 121)
   - **Link:** https://github.com/scalameta/scalameta/blob/ae56801a619aa3364e8f76d2a93a5128da65ac7f/semanticdb/semanticdb/semanticdb.proto#L121
   - **Details:** The `annotations` field in `AnnotatedType` (and `SymbolInformation`) was changed from `repeated Annotation` to `repeated AnnotationTree`. This changes the return type in generated Java/Scala classes, breaking the build for any project depending on this library.
   - **Fix:** Ensure this change is accompanied by a major version bump or appropriate migration guide. Alternatively, consider introducing a new field (e.g., `annotation_trees`) and deprecating the old one to allow a transition period, though this may complicate binary compatibility.

kitbellew:
closing as our "due diligence" has outlasted the author's interest in it :)

PR #2282, opened 2021-04-13
[Scala3] Fix issues with wrong `Indent`/`Outdent` positions.

dos65:
#2279  was caused by the wrong `Indent`/`Outdent` tokens positions.
They have to point at `LF` that triggered their creation.

To cleanup tree syntax from whitespaces now `atPos` performs a kind of `trim` operation.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -289,21 +289,22 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       if (!dialect.allowSignificantIndentation) false
       else {
         val existingIndent = sepRegions.find(_.isIndented).map(_.indent).getOrElse(0)
-        val hasLF = (prevPos to tokenPos).exists(i => scannerTokens(i).is[LF])
+        val lastLF = (prevPos to tokenPos).reverse.collectFirst({
+          case i if scannerTokens.tokens(i).is[LF] => i
+        })
         val expected = countIndent(tokenPos)
-        if (hasLF && expected > existingIndent) {
-          val prependIndentRegion = expected != existingIndent
-          if (prependIndentRegion) {
+        lastLF match {
+          case Some(pointPos) if expected > existingIndent =>
             sepRegions = f(expected, sepRegions)
             curr = TokenRef(
               new Indentation.Indent(token.input, token.dialect, token.start, token.end),
               curr.pos,
               curr.pos,
-              curr.pointPos
+              pointPos
             )
-          }
-          true
-        } else false
+            true
+          case _ => false
+        }
       }
     }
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM! Thanks for working on this, I was kind of lost with my own approach.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -289,21 +289,22 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       if (!dialect.allowSignificantIndentation) false
       else {
         val existingIndent = sepRegions.find(_.isIndented).map(_.indent).getOrElse(0)
-        val hasLF = (prevPos to tokenPos).exists(i => scannerTokens(i).is[LF])
+        val lastLF = (prevPos to tokenPos).reverse.collectFirst({
```
kitbellew, line 292: i'd short circuit, perhaps, and avoid reverse search if the condition is false
```
expected < existingIndent && .... collectFirst {
  case pointPos if ... =>
    curr = TokenRef ...
}.isDefined

File: scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala [hunk 2]
```diff
@@ -820,16 +821,31 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
     if (endTokenPos < startTokenPos) endTokenPos = startTokenPos - 1
-    val pos = TokenStreamPosition(startTokenPos, endTokenPos + 1)
+    val pos = trimmedPos(startTokenPos, endTokenPos)
     result.withOrigin(Origin.Parsed(input, dialect, pos)).asInstanceOf[T]
   }
 
+  def trimmedPos(startTokenPos: Int, endTokenPos: Int): TokenStreamPosition = {
+    def skipTrivia(range: Range): Int = {
+      range
+        .dropWhile(i => scannerTokens(i).is[Trivia])
```
kitbellew, line 831: is `find(!...)` available on `Range`?

kitbellew (commented):
thanks for fixing this! very minor ocd nitpicks follow...

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: `dropWhile` is specifically implemented for `Range` and should work slightly faster than generic `find`

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: thx! I'll look at it in the next PR

Issue #2283, opened 2021-04-14
contrib.ScaladocParser.parseScaladoc `@param` parse bug since 4.4.11


xuwei-k:
## 4.4.11 or later

```scala
import scala.meta.contrib.ScaladocParser
import scala.meta.inputs.Input
import scala.meta.Source
import scala.meta.tokens.Token.Comment

scala> ScaladocParser.parseScaladoc(Input.String("/** @param a b */").parse[Source].get.tokens.collectFirst{ case c: Comment => c }.get)
res0: Option[List[scala.meta.contrib.DocToken]] = Some(List(Description(@param a b)))
```

## 4.4.10

```scala
import scala.meta.contrib.ScaladocParser
import scala.meta.inputs.Input
import scala.meta.Source
import scala.meta.tokens.Token.Comment

scala> ScaladocParser.parseScaladoc(Input.String("/** @param a b */").parse[Source].get.tokens.collectFirst{ case c: Comment => c }.get)
res0: Option[List[scala.meta.contrib.DocToken]] = Some(List(Param(name=a, body=b)))
```

tgodzik:
Thanks for reporting! It does seem to be related to the changes I did with fastparse. I will take a look at it!

PR #2284, opened 2021-04-14
Fix parsing Param tags

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/2283

This was working in the other `internal.parser.ScaladocParser`. I guess we can drop this one if we ever do a 5.0 release.


kitbellew (approved):
some voodoo magic

tgodzik:
> some voodoo magic

I think the previous parser might have been lazy and now it tries to evaluate right away? Not sure, we defs seem to work everywhere else.

PR #2285, opened 2021-04-14
[Scala3] Fix `For`/`While` expr that might have indented body without `do` keyword

dos65:


tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/ControlSyntaxSuite.scala
```diff
@@ -740,6 +740,26 @@ class ControlSyntaxSuite extends BaseDottySuite {
     )
   }
 
+  test("new-for-indented-without-do") {
+    val code = """|for ( a <- x )
```
tgodzik, line 744: Could you also add a test for:
```
  val code = """|for { a <- x }
                  |  fx
                  |  fy
                  |""".stripMargin
```
?

tgodzik (changes requested):
Thanks a lot for the fix! I would just add the test for the last remaining case

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/ControlSyntaxSuite.scala
dos65: Added

tgodzik (approved):
LGTM!

PR #2290, opened 2021-04-16
Allow indented partial function values

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/2289

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2154,7 +2154,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
   private def exprMaybeIndented(): Term = {
     if (token.is[Indentation.Indent]) {
-      block()
+      blockExpr()
     } else {
       expr()
     }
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2154,7 +2154,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
   private def exprMaybeIndented(): Term = {
     if (token.is[Indentation.Indent]) {
-      block()
+      blockExpr()
```
tgodzik, line 2157: Funnily enough this also allows:
```
if
   case a =>
   case b =>
then
  stat
```

which is not entirely correct, but it's also accepted by the Dotty parser, only to be rejected at typer. I think more permissive parser is ok in this case.

dos65 (approved):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
olafurpg: I agree it's better to be permissive in this case. I think the Scalameta parser is probably too strict in some cases.

PR #2291, opened 2021-04-16
Remove most checks for Pat.Type

tgodzik:
Dotty allows type to almost always be defined

More info about the syntax is here https://dotty.epfl.ch/docs/internals/syntax.html#expressions

Connected to https://github.com/scalameta/scalafmt/issues/2424

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MinorDottySuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MinorDottySuite.scala
@@ -904,4 +904,69 @@ class MinorDottySuite extends BaseDottySuite {
       )
     )
   }
+
+  test("regex-pattern") {
+    runTestAssert[Stat](
+      """|s match
+         |  case re(v): String => v.toDouble
+         |  case other => o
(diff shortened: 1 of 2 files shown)


tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MinorDottySuite.scala
```diff
@@ -904,4 +904,37 @@ class MinorDottySuite extends BaseDottySuite {
       )
     )
   }
+  test("regex-pattern") {
+    runTestAssert[Stat](
+      """|s match
+         |  case re(v): String => v.toDouble
```
dos65, line 910: Probably, it's needed to remove all checks from `Typed.lhs`

I tried the following code and dotty compiles it:
```scala
s match {
  case (v: String): String =>  // <- scalameta tree `Pat.Typed(Pat.Typed(...), ...)`
}
```

tgodzik: Good catch, I removed all of them and also added another test.

dos65 (approved):

PR #2293, opened 2021-04-20
[Scala3] Fix selftype parsing for enum.

dos65:
In the following example: `enum A { self => ...`  right arrow should start `RegionIndentedEnum`  instead of `RegionIndented` to not to break `cases` parsing.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -251,6 +251,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     def observeIndented(): Boolean
     def observeOutdented(): Boolean
     def observeIndentedEnum(): Boolean
+    def undoIndent(): Unit
   }
   var in: TokenIterator = {
     new LazyTokenIterator(
@@ -300,6 +301,15 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           )
           true
         } else false
+      }
+    }
+
+    def undoIndent(): Unit = {
+      sepRegions = sepRegions match {
+        case region :: others if region.isIndented && curr.token.is[Indentation.Indent] =>
+          next()
+          others
+        case regions => regions
       }
     }
 
@@ -4892,6 +4902,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           in = beforeFirst
           selfOpt = Some(self())
           next()
+          in.undoIndent
         } catch {
           case ex: ParseException =>
             in = afterFirst
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/EnumSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/EnumSuite.scala
@@ -385,6 +385,108 @@ class EnumSuite extends BaseDottySuite {
     )
   }
 
+  test("enum-self-braces") {
+    val code =
+      """|enum A { self =>
+         |  case B(v: Int)
+         |  case C(v: String)
+         |}
+      """.stripMargin
+    runTestAssert[Stat](code)(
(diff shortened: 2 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -607,9 +611,14 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
             if (needOutdent)
               (sepRegions.tail, mkOutdent(lastNewlinePos))
-            else if (needIndent)
-              (RegionIndent(nextIndent, prev.is[KwMatch]) :: sepRegions, mkIndent(lastNewlinePos))
-            else if (canProduceLF) {
+            else if (needIndent) {
+              val nextRegion =
+                if (sepRegions.headOption.exists(_.isEnum))
+                  RegionIndentEnum(nextIndent)
```
tgodzik, line 617: Are we doubling the region in case of `=> <indent>` ? We should not be inserting indentation at all after => if there is already an enum region.

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/EnumSuite.scala
```diff
@@ -385,6 +385,105 @@ class EnumSuite extends BaseDottySuite {
...
+          ),
+          Nil
+        )
+      )
+    )
+  }
+
+  test("enum-self-indented") {
+    val code =
+      """|enum A:
```
tgodzik, line 437: This is basically the same as:
```scala
enum A:
   self =>
   case B(v: Int)
   case C(v: String)
```

Indentation after `self =>` doesn't seem to be playing any role. We should most likely add a test case there too.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -209,11 +209,15 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
   /* ------------- PARSER-SPECIFIC TOKENS -------------------------------------------- */
 
-  sealed trait SepRegion {
+  sealed trait SepRegion { self =>
     def indent = -1
     def closeOnNonCase = false
     def indentOnArrow = true
     def isIndented = false
+    def isEnum: Boolean = self match {
```
tgodzik, line 217:
```suggestion
    def isEnum: Boolean = this match {
```
should be enough, no?

tgodzik (changes requested):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: We put `Indent` for any template in this case. For `class A{ self => ..` and others.
It seems that it will be harder no to insert it after `SelfType =>` than preserving the enum region type in this case.

tgodzik: `class A{ self => ..` <- this one for sure
However, maybe it doesn't make sense here:
```scala
enum A:
  self => ..
```
?

dos65: Ha, you right!

I found the case that doesn't work. 
```scala
enum A:
  self =>
    case B, C
def foo: Int = ???
//^ error: outdent expected but def found
```

Moreover, Dotty parser has exactly the same hack as you propose: https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/parsing/Parsers.scala#L3836


Probably will port this trick.

dos65: Reworked this by adding `undoIndent` method to `TokenIterator`

tgodzik (approved):
LGTM!

Issue #2294, opened 2021-04-20
java.lang.NoSuchMethodError: scala.meta.Dialect.allowSignificantIndentation()Z


cacilhas:
I was trying the [Tree Guide](https://scalameta.org/docs/trees/guide.html), and in the first example I’m “welcomed” by the following message:

```scala
cacilhas-apps@ import scala.meta._
import scala.meta._

cacilhas-apps@ val program = """object Main extends App { print("Hello!") }"""
program: String = "object Main extends App { print(\"Hello!\") }"

cacilhas-apps@ val tree = program.parse[Source].get
java.lang.NoSuchMethodError: scala.meta.Dialect.allowSignificantIndentation()Z
  scala.meta.internal.parsers.ScalametaParser.isColonEol(ScalametaParser.scala:207)
  scala.meta.internal.parsers.ScalametaParser.countIndentInternal$1(ScalametaParser.scala:660)
  scala.meta.internal.parsers.ScalametaParser.scala$meta$internal$parsers$ScalametaParser$$countIndent(ScalametaParser.scala:666)
  scala.meta.internal.parsers.ScalametaParser$LazyTokenIterator.nextToken(ScalametaParser.scala:367)
  scala.meta.internal.parsers.ScalametaParser$LazyTokenIterator.next(ScalametaParser.scala:281)
  scala.meta.internal.parsers.ScalametaParser$LazyTokenIterator.next(ScalametaParser.scala:271)
  scala.meta.internal.parsers.ScalametaParser.next(ScalametaParser.scala:688)
  scala.meta.internal.parsers.ScalametaParser.accept(ScalametaParser.scala:860)
  scala.meta.internal.parsers.ScalametaParser.parseRule(ScalametaParser.scala:51)
  scala.meta.internal.parsers.ScalametaParser.parseSource(ScalametaParser.scala:142)
  scala.meta.parsers.Parse$.$anonfun$parseSource$1(Parse.scala:29)
  scala.meta.internal.parsers.ScalametaParser$$anon$257.apply(ScalametaParser.scala:5067)
  scala.meta.parsers.Api$XtensionParseDialectInput.parse(Api.scala:25)
  scala.meta.parsers.Api$XtensionParseInputLike.parse(Api.scala:14)
  ammonite.$sess.cmd3$.<clinit>(cmd3.sc:1)
```

I **strictly** followed the guide.

kpbochenek:
[...]

PR #2296, opened 2021-04-21
Update reference to Ammonite in guide.md

tgodzik:
Ammonite itself is using scalameta, which causes issue when users want to use it to experiment with the parser.

I propose we remove that section altogether as it seem we will never be able to make it work properly.

Fixes https://github.com/scalameta/scalameta/issues/2294

### DIFF
--- a/docs/trees/guide.md
+++ b/docs/trees/guide.md
@@ -40,11 +40,14 @@ import scala.meta._
 ### Ammonite REPL
 
 A great way to experiment with Scalameta is to use the
-[Ammonite REPL](http://ammonite.io/#Ammonite-REPL).
+[Ammonite REPL](http://ammonite.io/#Ammonite-REPL). To make it work with
+scalameta it needs to be run with `--thin` option.
 
-```scala
-// Ammonite REPL
-import $ivy.`org.scalameta::scalameta:@VERSION@`, scala.meta._
+```sh
+$ amm --thin
+Loading...
+Welcome to the Ammonite Repl 2.3.8 (Scala 2.13.5 Java 11.0.7)
+@ import $ivy.`org.scalameta::scalameta:@VERSION@`, scala.meta._
 ```
 
 ### ScalaFiddle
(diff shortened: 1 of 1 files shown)


dos65 (approved):
LGTM.

Btw, it looks similar to the problem in mdoc https://github.com/scalameta/mdoc/pull/473

kpbochenek:
Just thought that maybe we could change example to use worksheets from metals, but they suffer the same problem, right?

olafurpg:
I opened an Ammonite issue here https://github.com/com-lihaoyi/Ammonite/issues/1177

If we publish a separate `org.scalameta:semanticdb` module that only contains the auto-generated ScalaPB classes then we should be able to fix the root cause of this problem. Users will continue to want to use Ammonite with Scalameta, removing the mention is an improvement for beginners reading the docs but it's not a sustainable solution.

olafurpg (approved):

docs/trees/guide.md
```diff
@@ -37,16 +37,6 @@ All code examples assume you have the following import
...
 ```
 
-### Ammonite REPL
-
-A great way to experiment with Scalameta is to use the
-[Ammonite REPL](http://ammonite.io/#Ammonite-REPL).
-
-```scala
-// Ammonite REPL
-import $ivy.`org.scalameta::scalameta:@VERSION@`, scala.meta._
```
olafurpg, line 47: What do you think about recommending to use a hardcoded version of an older Scalameta release?

```
❯ amm
Loading...
Welcome to the Ammonite Repl 2.3.8-32-64308dc3 (Scala 2.13.4 Java 1.8.0_272)
@ q"a"
res1: Term.Name = Term.Name(value = "a")
```

We can add a note that newer versions don't work yet and link to the issue?

olafurpg: The import line is missing from the code above
```
❯ amm
Loading...
Welcome to the Ammonite Repl 2.3.8-32-64308dc3 (Scala 2.13.4 Java 1.8.0_272)
@ import $ivy.`org.scalameta::scalameta:4.4.6`, scala.meta._
qimport $ivy.$                               , scala.meta._

@ q"a"
res1: Term.Name = Term.Name(value = "a")
```

alexarchambault:
As [commented](https://github.com/com-lihaoyi/Ammonite/issues/1177#issuecomment-824017102) in https://github.com/com-lihaoyi/Ammonite/issues/1177, just pass the `--thin` option to Ammonite, and there shouldn't be any conflict.

docs/trees/guide.md
tgodzik: I added the `--thin` optino as suggested by @alexarchambault Thanks!

tgodzik:
> As [commented](https://github.com/com-lihaoyi/Ammonite/issues/1177#issuecomment-824017102) in [com-lihaoyi/Ammonite#1177](https://github.com/com-lihaoyi/Ammonite/issues/1177), just pass the `--thin` option to Ammonite, and there shouldn't be any conflict.

Good idea, thanks!

olafurpg (approved):
LGTM 👍

PR #2297, opened 2021-04-21
Fix issues with complex types in anonymous given

tgodzik:
The issue popped up when parsing https://github.com/kubukoz/slang/tree/165fc48c1a3bfff4a0349378855ca89c6d90873f

Connected to https://github.com/scalameta/scalafmt/issues/2216#issuecomment-823910262

We are first trying to parse non-anonymous given, which
 - requires `:` for type, if none is found it means it's anonymous
 - might fail in cases that anonymous type looks like type param
   `given Conversion[AppliedName, Expr.Apply[Id]] = ???`
   This will fail because type params cannot have `.`

### DIFF
--- a/community-test/src/test/scala/CommunityDottySuite.scala
+++ b/community-test/src/test/scala/CommunityDottySuite.scala
@@ -84,14 +84,17 @@ class CommunityDottySuite extends FunSuite {
     fetchCommunityBuild(build)
 
     val stats = checkFilesRecursive(communityDirectory.resolve(build.name))
-    val timePer1KLines = stats.timeTaken / (stats.linesParsed / 1000)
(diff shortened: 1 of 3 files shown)


kpbochenek (approved):
always a little scary to wrap things in try catch-all but it looks like best solution we can do here 👍

dos65 (approved):
I don't mind wrapping it in `try` but I have to notice that alternatively, it might be fixed by a look-ahead check if there are `:` before `=` . Feel free to ignore this )

tgodzik:
> I don't mind wrapping it in `try` but I have to notice that alternatively, it might be fixed by a look-ahead check if there are `:` before `=` . Feel free to ignore this )

I feel `try` is safer and easier in this case and serves the same role as a lookahead.

PR #2298, opened 2021-04-21
Allow Export to be a top level statement

tgodzik:
Found while testing https://gitlab.com/heksesang/scala3-derivation.git


kpbochenek (approved):
good catch :+1:

PR #2300, opened 2021-04-21
Allow indented block after `<-` and `?=>`

tgodzik:
We missed it previously and that seems to be the case according to https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html#optional-braces


dos65 (approved):
:+1:
It's only need to fix fmt

PR #2301, opened 2021-04-22
[Scala3] Fix Oudent position before `}`, `)`, `,`

dos65:
It addresses this problem - https://github.com/scalameta/scalafmt/pull/2447#issuecomment-824851176.
In previous similar fixes about position and indentation tokens, I forgot about these cases :disappointed: 

I double-checked all places where these artificial tokens are created and hope there will be no more such bugs.


kitbellew (approved):
based largely on the tests

tgodzik (approved):
LGTM! Thanks for the fix!

PR #2313, opened 2021-04-26
Drop procedure syntax for the Scala 3 dialect

tgodzik:
It was also needed to fix the macro used in tests, since it started generating method that was too large, but we haven't actually used most of that, so no test was harmed. This should also be covered by MiMa.


dos65 (commented):
LGTM!

PR #2319, opened 2021-04-29
Fix position for anonymous trait ctor

### LINKED ISSUE #2318: if `!allowTraitParameters`, the `extends` keyword is part of `Name.Anonymous` tree
if dialect used were `scala211`, the following code formats incorrectly:
```
trait SampleTrait extends A with B with C with D with E{
  self: LongNameMixin with B with C with D with EEEEEEEEE =>

  def foo: Boolean = true
}
```
like this:
```
trait SampleTrait extends A with B
with C with D with E {
  self: LongNameMixin
    with B with C with D
    with EEEEEEEEE =>

  def foo: Boolean = true
}
```

and the reason is that `extends` belongs not only to `Template` tree but also to a narrower `Name.Anonymous`. this problem does not surface if dialect is `scala211.withAllowTraitParameters(true)`.

dos65:
Fixes #2318

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -4534,7 +4534,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       val paramss = paramClauses(ownerIsType = true, owner == OwnedByCaseClass)
       Ctor.Primary(mods, name, paramss)
     } else if (owner.isTrait) {
-      Ctor.Primary(Nil, atPos(in.tokenPos, in.tokenPos)(Name.Anonymous()), Nil)
+      Ctor.Primary(Nil, atPos(in.tokenPos, in.tokenPos - 1)(Name.Anonymous()), Nil)
     } else {
       unreachable(debug(owner))
     }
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -4534,7 +4534,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       val paramss = paramClauses(ownerIsType = true, owner == OwnedByCaseClass)
       Ctor.Primary(mods, name, paramss)
     } else if (owner.isTrait) {
-      Ctor.Primary(Nil, atPos(in.tokenPos, in.tokenPos)(Name.Anonymous()), Nil)
+      Ctor.Primary(Nil, atPos(in.tokenPos, in.tokenPos - 1)(Name.Anonymous()), Nil)
```
kitbellew, line 4537: @dos65 quick question: below, on lines 4546 and 4557 (and maybe elsewhere), the same construct was used. should they be changed there as well? if Name.Anonymous should never have any tokens, should we define a shortcut:
```
  def atNameAnonymous(in: ...): ... = atPos(x, x - 1)(...)
```
(and what does `autoPos` on 4533 do?)

dos65: oops)

Good catch, thx! I will fix these cases too

> and what does autoPos on 4533 do?

:thinking:

tgodzik: Should do a `-1` range correctly in that case I think.

PR #2320, opened 2021-04-29
[Scala 3] Include inline in Term.Match tree position

tgodzik:
Connected to https://github.com/scalameta/scalafmt/issues/2478


dos65 (approved):
LGTM

PR #2324, opened 2021-04-30
Update master to main branch

tgodzik:
This requires also changing the primary branch in the repo.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 on:
   push:
     branches:
-      - master
+      - main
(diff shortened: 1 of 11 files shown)


dos65 (approved):
:+1:

ckipp01 (approved):

kitbellew:
there's no main branch yet, right? and why do you want to change this?

tgodzik:
We would update the branch just after merging, so that everything works as usual.

We are changing it as part of the diversity proposal at Scala Center. It's a minimal thing we can do and is a good gesture towards inclusivity.

Issue #2326, opened 2021-05-03
`extends` keyword is not part of a template


kitbellew:
In `class A extends B with C`, the template part is `B with C`. Is there any reason why it's not `extends B with C` (since both `B` and `C` are `inits` which are part of template)? Leads to some bugs and added complexity in `scalafmt`.

kitbellew:
@tgodzik @dos65 ^^^

PR #2327, opened 2021-05-04
Include `extends` keyword in Template

tgodzik:
We already include derive and it makes sense to keep it there.

Fixes https://github.com/scalameta/scalameta/issues/2326

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -3573,7 +3573,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   }
 
   def ctorModifiers(): Option[Mod] = token match {
-    case Unquote() => Some(unquote[Mod])
+    case Unquote() if ahead(token.is[LeftParen]) => Some(unquote[Mod])
     case Ellipsis(_) => Some(ellipsis(1, astInfo[Mod]))
     case KwPrivate() => Some(accessModifier())
     case KwProtected() => Some(accessModifier())
@@ -4743,7 +4743,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
   def quasiquoteTemplate(): Template = entrypointTemplate()
 
-  def entrypointTemplate(): Template = template(enumCaseAllowed = false)
+  def entrypointTemplate(): Template = autoPos(template(enumCaseAllowed = false))
 
   def ensureEarlyDef(tree: Stat): Stat = tree match {
     case q: Quasi => q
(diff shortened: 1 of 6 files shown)


tgodzik:
@kitbellew It seems the excluding `extend` was intended and it serves a purpose in quasiquotes. I am not sure how much they are currently used and not sure if we can change the behaviour currently. @olafurpg might know more.

We would also need to drop allowing quasiquotes for Mods in constructor params. So quite a few changes for this.

I wonder if we should at some point just extract the parser from scalameta and turn it all legacy. We could drop macros, allow only specific well tested features and cross publish for Scala 3 also. We could also experiment much more without fear of breaking changes. Any change here can have quite a large consequences.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3573,7 +3573,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   }
 
   def ctorModifiers(): Option[Mod] = token match {
-    case Unquote() => Some(unquote[Mod])
+    // case Unquote() => Some(unquote[Mod])
```
tgodzik, line 3576: We would need to drop this to make `class A $template` work.

tgodzik: Or actually, we can possibly keep this

tests/shared/src/test/scala/scala/meta/tests/quasiquotes/ErrorSuite.scala
```diff
@@ -720,19 +720,21 @@ class ErrorSuite extends FunSuite {
...
-      |<macro>:4: type mismatch when unquoting;
-      | found   : scala.meta.Type.Select
-      | required: scala.meta.Template
-      |Note: This shape of a quasiquote tells scala.meta that you're unquoting a template.
-      |If you'd like to unquote a parent reference, add {} immediately after the unquote.
-      |For more details, see https://github.com/scalameta/scalameta/issues/223.
-      |      q"class C extends $notReallyAParent"
-      |                        ^
-    """.trim.stripMargin)
+      q"class C $notReallyAParent"
```
kitbellew, line 727: this type of quasiquote is probably meaningless. as a result of your [welcome] change, there's no more confusion between matching a template and a parent class.

anyone who uses `$foo` without `extends` would know it matches a template. and using `extends $foo` would now require `init"..."` and nothing else.

so i would do this:
- in `ConversionMacros.scala:typeMismatchMessage`, replace the entire template-related block (and `MetaTemplate` field) and remove the `wholeMessage` variable as well.
- change this `#448` test into two:
  - `q"class C $foo` where `foo = t"..."` (this one); wants template, gets type
  - `q"class C extends $foo` where `foo = t"..."` (original, with `extends`); wants init, gets type

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3573,7 +3573,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   }
 
   def ctorModifiers(): Option[Mod] = token match {
-    case Unquote() => Some(unquote[Mod])
+    case Unquote() if ahead(token.is[LeftParen]) => Some(unquote[Mod])
```
kitbellew, line 3576: i am not sure i understand this. does this mean unquoting a mod requires `$(mod)`, with parens?

kitbellew (changes requested):
i think this is actually cleaner, no more confusion and need for `{}` (what do these braces do, anyway?).

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: It means as far as I can tell `$mod (` so essentially the constructor should be next.

kitbellew: ah, so this is not for `private class C(...)`, it's for `class C private(...)`? got it.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -4753,28 +4753,25 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
-            !token.is[LeftParen] && !token.is[LeftBracket] && !token.is[LeftBrace] &&
-            !token.is[KwWith]
-        )) {
-        unquote[Template]
-      } else {
-        template(afterExtend = true, enumCaseAllowed = owner.isEnum)
-      }
+      template(afterExtend = true, enumCaseAllowed = owner.isEnum)
     } else {
       val startPos = in.tokenPos
```
kitbellew, line 4767: this is not used, i think

tests/shared/src/test/scala/scala/meta/tests/quasiquotes/SuccessSuite.scala
```diff
@@ -2317,7 +2317,7 @@ class SuccessSuite extends FunSuite {
   test("#448") {
     val parent = init"_root_.scala.AnyVal"
     val template = template"$parent"
-    assert(q"class C extends $template".syntax == "class C extends _root_.scala.AnyVal")
+    assert(q"class C $template".syntax == "class C extends _root_.scala.AnyVal")
```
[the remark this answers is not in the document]
tgodzik: Added! It wasn't tested anywhere else though.

tests/shared/src/test/scala/scala/meta/tests/quasiquotes/ErrorSuite.scala
tgodzik: Did both, I think we should be fine now.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Removed

tests/shared/src/test/scala/scala/meta/tests/quasiquotes/SuccessSuite.scala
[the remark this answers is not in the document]
kitbellew: because it wasn't possible before your change

kitbellew (approved):

PR #2328, opened 2021-05-04
[Scala 3] Add support for indented unnamed anonymous classes

tgodzik:


dos65 (approved):
:+1:

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/SignificantIndentationSuite.scala
```diff
@@ -54,6 +54,54 @@ class SignificantIndentationSuite extends BaseDottySuite {
     )
   }
 
+  test("anonymous-class") {
```
kitbellew, line 57: you are testing anonymous classes. did it work, without this change, with regular
```
class A:
  ...
```
and does it work with early defs (that is, should line 4733 be changed as well):
```
class A extends { def ... }:
  ...
```

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -4736,7 +4736,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
         Template(Nil, Nil, self, body)
       }
     } else {
-      val parents = templateParents(afterExtend)
+      val parents = if (token.is[Colon]) Nil else templateParents(afterExtend)
```
kitbellew, line 4739: should this whole function be changed to:
```
    val edefsOrTemplate = if (token.is[LeftBrace]) {
      // @S: pre template body cannot stub like post body can!
      val (self, body) = templateBody(enumCaseAllowed)
      if (token.is[KwWith] && self.name.is[Name.Anonymous] && self.decltpe.isEmpty) {
        val edefs = body.map(ensureEarlyDef)
        next()
        Left(edefs)
      } else {
        Right(Template(Nil, Nil, self, body))
      }
    } else {
      Left(Nil)
    }
    edefsOrTemplate match {
      case Right(t) => t
      case Left(edefs) =>
        val parents = if (token.is[Colon]) Nil else templateParents(afterExtend)
        template(edefs, parents, enumCaseAllowed)
    }
```

kitbellew (changes requested):
see previous comments

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/SignificantIndentationSuite.scala
tgodzik: The first one worked previously, only `new:` was not handled.

As for `early defs` they are no longer allowed in Dotty, so this is safe.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: We should be fine as is, since early defs are not allowed in Dotty, so they will not exist with colons.

It would be probably good to remove early defs from Scala 3 dialect, but we could do that separately if needed.

kitbellew (approved):

PR #2329, opened 2021-05-04
[Scala 3] Include braces in extension method's block

tgodzik:


kitbellew (approved):
thank you! finally, we get more indicators.

there are probably still cases remaining where blocks don't include their braces (at least, with multi-stat blocks it was easy to tell they should have braces, but not with single-stat blocks).

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -4225,28 +4219,23 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+    val body: Stat =
       if (token.is[LeftBrace]) {
-        inBraces(templateStats())
+        autoPos(Term.Block(inBraces(templateStats())))
       } else {
         newLinesOpt()
         if (in.observeIndented()) {
-          indented(templateStats())
+          val block = autoPos(Term.Block(indented(templateStats())))
+          if (block.stats.size == 1) block.stats.head
```
kitbellew, line 4231: does this mean you will NOT keep braces for this single-stat block? i would prefer that you had. if block had braces, they need to be part of it, regardless of the number of stats.

kitbellew (changes requested):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: That's only for indentation, that doesn't have braces at all.

kitbellew: zrozumiał...

kitbellew (approved):

PR #2330, opened 2021-05-04
Use Term.Block in functions when {} are added explicitely.

tgodzik:
Previously, we would always remove the outer most {} and not report a Block even in:
```
List("").map{
s => {
    s + s
  }
}
```

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2255,10 +2255,9 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       case KwTry() =>
         next()
         val body: Term = token match {
-          case Indentation.Indent() => block()
           case _ if dialect.allowTryWithAnyExpr => expr()
-          case LeftBrace() => autoPos(inBracesOrUnit(block()))
           case LeftParen() => inParensOrUnit(expr())
+          case LeftBrace() | Indentation.Indent() => block()
           case _ => expr()
         }
         val catchopt =
@@ -2519,13 +2518,13 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
               }
               val params = convertToParams(t)
               val trm =
-                if (token.is[Indentation.Indent]) blockExpr()
-                else if (location != BlockStat) expr()
-                else block()
+                if (location != BlockStat) expr()
+                else blockExpr()
+
               if (contextFunction)
-                Term.ContextFunction(params, dropTrivialBlock(trm))
+                Term.ContextFunction(params, trm)
               else
-                Term.Function(params, dropTrivialBlock(trm))
+                Term.Function(params, trm)
             })
           } else {
             // do nothing, which will either allow self-type annotation parsing to kick in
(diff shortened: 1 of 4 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2521,11 +2521,22 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
                 if (token.is[Indentation.Indent]) blockExpr()
                 else if (location != BlockStat) expr()
-                else block()
+                else
+                  autoPos {
+                    if (token.is[LeftBrace]) {
+                      Term.Block(inBraces(blockStatSeq()))
+                    } else {
+                      blockStatSeq() match {
+                        case (head: Term) :: Nil => head
```
kitbellew, line 2530: should we keep it as block? who cares if it's just one statement. generally allows one to understand when blocks are expected in the code.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2521,11 +2521,22 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
               val trm =
                 if (token.is[Indentation.Indent]) blockExpr()
                 else if (location != BlockStat) expr()
-                else block()
```
kitbellew, line 2524: should we redefine
```
  def block(): Term = autoPos {
    val res = Term.Block(blockStatSeq())
    if (token.is[LeftBrace]) inBraces(res) else res
  }
```
and just use it everywhere? for instance, line 2257 (under `Try`) would become:
```
          case _ if dialect.allowTryWithAnyExpr => expr()
          case LeftParen() => inParensOrUnit(expr())
          case _ => block()
        }
```

tgodzik: Might be a better idea, I will take a look at it.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: I would not add a block if {} are not used I think. This was also done on purpose before.

kitbellew: i will leave it to you. in scalafmt, we can mostly adjust.

but if you asked me, i would use it whenever it's expected (in `if`, `case`, `def` etc.), regardless whether it's one statement, and whether it had braces. not using a block with a single statement and/or no braces is akin to not using a template if the body contains just one statement and no parents.

if you _had_ kept the block, solving the long-standing scalafmt request to add back the braces would have been a lot easier.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Reworked it a bit and managed to get rid of most repeated stuff.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: I will leave as is for now, I don't really want to change the base behaviour, since that can have unindented consequences,

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3032,26 +3031,24 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
-      indented(
-        dropTrivialBlock(Term.Block(blockStatSeq()))
-      )
-    else Term.Block(blockStatSeq())
+    if (token.is[LeftBrace]) {
+      Term.Block(inBraces(blockStatSeq()))
+    } else {
+      val possibleBlock =
+        if (token.is[Indentation.Indent])
+          indented(Term.Block(blockStatSeq()))
```
kitbellew, line 3048: above, you did inBraces and then Block. here, first Block and then indented. does the order matter?

if it doesn't, I'd first create Block over blockStatSeq as a variable, then overlay inBraces or indented.

kitbellew (approved):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: The order doesn't matter in this case, changed!

PR #2332, opened 2021-05-04
Add support for markdown code blocks

tdudzik:
Previously, only code blocks between symbols `{{{` and `}}}` were supported. This PR adds a support for markdown code blocks. We need it to implement formatting of code blocks in scaladoc using scalafmt.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
@@ -49,12 +49,19 @@ object ScaladocParser {
   private def codePrefix[_: P] = P("{{{")
   private def codeSuffix[_: P] = P(hspaces0 ~ "}}}")
 
+  def markdownCodePrefix[_: P] = P(leadHspaces0 ~ "```scala" ~ (!nl ~ AnyChar).rep ~ nl)
+  def markdownCodeSuffix[_: P] = P(hspaces0 ~ "```")
+
   private def linkPrefix[_: P] = P("[[" ~ hspaces0)
   private def linkSuffix[_: P] = P(hspaces0 ~ "]]")
 
   private def codeLineParser[_: P]: P[String] = {
     def codeLineEnd = P(nl | codeSuffix)
     P((!codeLineEnd ~ AnyChar).rep.!)
+  }
+
+  private def markdownCodeLineParser[_: P]: P[String] = {
+    P(!markdownCodeSuffix ~ (!nl ~ AnyChar).rep.!)
   }
 
   private def codeExprParser[_: P]: P[CodeExpr] = {
@@ -68,6 +75,12 @@ object ScaladocParser {
     def code = codeLineParser.rep(1, sep = nl)
     def pattern = leadHspaces0 ~ codePrefix ~ nl ~ code ~ codeSuffix
     P(pattern.map { x => CodeBlock(if (x.last.nonEmpty) x.toSeq else x.view.dropRight(1).toSeq) })
+  }
+
+  private def markdownCodeBlockParser[_: P]: P[CodeBlock] = {
+    def code = markdownCodeLineParser.rep(0, sep = nl)
+    def pattern = markdownCodePrefix ~ (code ~ nl).? ~ markdownCodeSuffix
+    P(pattern.map(x => CodeBlock(x.getOrElse(Seq()))))
   }
 
   private def headingParser[_: P]: P[Heading] = {
@@ -209,6 +222,7 @@ object ScaladocParser {
   private def termParser[_: P] =
     listBlockParser() |
       codeBlockParser |
+      markdownCodeBlockParser |
       headingParser |
       tagParser |
       tableParser |
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
@@ -206,6 +206,57 @@ class ScaladocParserSuite extends FunSuite {
     assertEquals(result, expectation)
   }
 
+  test("markdown code blocks") {
+    val codeBlock = // keep all newlines and leading spaces
+      """|  object Foo {
+         |    def bar(): String = "foobar"
+         |  }
+         |""".stripMargin.split("\n")
+    val codeBlockAsComment = codeBlock.mkString("\n *")
+
+    val result =
+      parseString(
+        s"""
+          /**
+            *  ```scala
+            *  ```
+            * 
+            *  ```scala
+            *$codeBlockAsComment  
+            *  ``` 
+            * 
+            *  ```scala
+            *$codeBlockAsComment ``` 
+            *  ```
+            * 
+            *  ```scala sc:compile
+            *$codeBlockAsComment 
+            *  ```
+            */
+       """.stripMargin
+      )
+
+    val expectation = Option(
+      Scaladoc(
+        Seq(
+          Paragraph(
+            Seq(CodeBlock(Seq()))
+          ),
+          Paragraph(
+            Seq(CodeBlock(codeBlock))
+          ),
+          Paragraph(
+            Seq(CodeBlock(codeBlock.dropRight(1) ++ Seq("  } ```")))
+          ),
+          Paragraph(
+            Seq(CodeBlock(codeBlock))
+          )
+        )
+      )
+    )
+    assertEquals(result, expectation)
+  }
+
   test("headings") {
     val level1HeadingBody = "Level 1"
     val level2HeadingBody = "Level 2"


kitbellew:
@tdudzik thanks for the submission. the change looks quite straightforward.

however, can you point out where scaladoc standard allows for this type of markdown (and why it needs to be used instead of `{{{`)? (https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html or anywhere else)

and if we add `` ``` `` as alternative to `{{{`, should we add `` ` `` as alternative to `{`?

tgodzik:
The new scaladoc for Scala 3 allows to use the markdown syntax in the comments. There are plans also to typecheck those snippets to make sure they are correct.

@tdudzik could you ask the scaladoc dotty team where it's all documented?

kitbellew:
> The new scaladoc for Scala 3 allows to use the markdown syntax in the comments. There are plans also to typecheck those snippets to make sure they are correct.

does it only support `` ```scala `` comments?

tdudzik:
Sure, I can ask them for more info if needed but there is documentation here: https://github.com/lampepfl/dotty/tree/master/scaladoc, and I think there are some test cases: https://github.com/lampepfl/dotty/blob/master/scaladoc-testcases/src/tests/tests.scala#L20 and there is a PR with type checking https://github.com/lampepfl/dotty/blob/2d4e88bc6f175ec4334365cf2c02c85beb312fcf/scaladoc-testcases/src/tests/snippetCompilerTests.scala#L5 .

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
```diff
@@ -206,6 +206,37 @@ class ScaladocParserSuite extends FunSuite {
     assertEquals(result, expectation)
   }
 
+  test("markdown code blocks".only) {
```
tgodzik, line 209:
```suggestion
  test("markdown code blocks") {
```

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
```diff
@@ -206,6 +206,37 @@ class ScaladocParserSuite extends FunSuite {
...
+         |    def bar(): String = "foobar"
+         |  }
+         |""".stripMargin.split("\n")
+    val codeBlockAsComment = codeBlock.mkString("\n *")
+
+    val result =
+      parseString(
+        s"""
+          /**
+            *  ```scala sc:compile
```
tgodzik, line 221: Could you also add "```scala\n" test case ?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -57,6 +60,10 @@ object ScaladocParser {
     P((!codeLineEnd ~ AnyChar).rep.!)
   }
 
+  private def markdownCodeLineParser[_: P]: P[String] = {
+    P(!markdownCodeSuffix ~ (!nl ~ AnyChar).rep.!)
+  }
+
   private def codeExprParser[_: P]: P[CodeExpr] = {
     def pattern = codePrefix ~ hspaces0 ~ codeLineParser ~ codeSuffix ~ punctParser.!
```
tgodzik, line 68: We should also add support for parsing expression witihin `` as suggested by @kitbellew

tgodzik (changes requested):

tgodzik:
Also, some back ground here. This would be the first step to support the new scaladoc code blocks. We were also thinking about maybe formatting those code blocks with scalafmt, but I am not sure if that is plausible. We can follow up with an issue in scalafmt and discuss it there @tdudzik

kitbellew:
> Also, some background here. This would be the first step to support the new scaladoc code blocks. We were also thinking about maybe formatting those code blocks with scalafmt, but I am not sure if that is plausible. We can follow up with an issue in scalafmt and discuss it there @tdudzik

problem is, `scalafmt` doesn't know how `CodeBlock` was obtained, and will output using `{{{`: https://github.com/scalameta/scalafmt/blob/master/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala#L608

either we need to create a different element, or add a pair of fields in `CodeBlock` to identify open and close delimiters. other elements, if they are alternatives for existing ones rather than completely new, would need to be rethought similarly.

tgodzik:
> > Also, some background here. This would be the first step to support the new scaladoc code blocks. We were also thinking about maybe formatting those code blocks with scalafmt, but I am not sure if that is plausible. We can follow up with an issue in scalafmt and discuss it there @tdudzik
> 
> problem is, `scalafmt` doesn't know how `CodeBlock` was obtained, and will output using `{{{`: https://github.com/scalameta/scalafmt/blob/master/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala#L608
> 
> either we need to create a different element, or add a pair of fields in `CodeBlock` to identify open and close delimiters. other elements, if they are alternatives for existing ones rather than completely new, would need to be rethought similarly.

Maybe we could do a MarkdownCodeBlock, it could contain the language and the additional markers?

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
tdudzik: Done

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
tdudzik: Fixed

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -49,6 +49,9 @@ object ScaladocParser {
   private def codePrefix[_: P] = P("{{{")
   private def codeSuffix[_: P] = P(hspaces0 ~ "}}}")
 
+  def markdownCodePrefix[_: P] = P(leadHspaces0 ~ "```scala" ~ (!nl ~ AnyChar).rep ~ nl)
```
kitbellew, line 52: is scala required? does dotty support only a subset of markdown, and a simple triple backquote without scala not supported?

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
```diff
@@ -206,6 +206,44 @@ class ScaladocParserSuite extends FunSuite {
...
```
kitbellew, line 221: both scala and sc:compile will be lost, is that ok?

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
```diff
@@ -206,6 +206,44 @@ class ScaladocParserSuite extends FunSuite {
...
+            *$codeBlockAsComment
+            *  ```
```
kitbellew, line 223: what happens if closing quotes are on the last line?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -70,6 +77,12 @@ object ScaladocParser {
     P(pattern.map { x => CodeBlock(if (x.last.nonEmpty) x.toSeq else x.view.dropRight(1).toSeq) })
   }
 
+  private def markdownCodeBlockParser[_: P]: P[CodeBlock] = {
+    def code = markdownCodeLineParser.rep(sep = nl)
+    def pattern = markdownCodePrefix ~ code ~ nl ~ markdownCodeSuffix
+    P(pattern.map { x => CodeBlock(if (x.last.nonEmpty) x.toSeq else x.view.dropRight(1).toSeq) })
```
kitbellew, line 83: what does this do? the check for empty, that is.

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
tdudzik: I've added a test case for it. The last line of the markdown code block should be just 3 backticks. They can't be on the same line as the code.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
tdudzik: As far as I can see, syntax highlighting in scaladoc is supported for both triple backquotes with and without scala. However, code block with scala is presented a little bit differently (it's surrounded by a frame), and a new snippet compilation feature supports only ```scala. Also, I think scalafmt should try to format only scala code.

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
tdudzik: I think it's a good idea to create a new `MarkdownCodeBlock` element (as it was already suggested) that will store information about language and options because of course, we need them.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
kitbellew: > As far as I can see, syntax highlighting in scaladoc is supported for both triple backquotes with and without scala. However, code block with scala is presented a little bit differently (it's surrounded by a frame), and a new snippet compilation feature supports only ```scala. Also, I think scalafmt should try to format only scala code.

scalafmt will format the scaladoc. if you don't call out `` ```java ... ``` `` code as a code block, this part will simply be turned into simple text and wrapped into a paragraph, losing structure.

if you additionally want to teach scalafmt to format the scala code within scaladoc, that would be a separate task -- and i am not sure i know off the top of my head how to do it.

tdudzik: > if you additionally want to teach scalafmt to format the scala code within scaladoc, that would be a separate task -- and i am not sure i know off the top of my head how to do it.

That was actually the goal.

tdudzik: @kitbellew I created an issue for it: https://github.com/scalameta/scalafmt/issues/2498

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
tdudzik: I think in that case we also need to distinguish between ` `` ` and ` {{{}}} `. Should I create a new element for markdown code expression or add the information about the symbol to the existing `CodeExpr` class?

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
kitbellew: @tdudzik any new scaladoc features we implement need to be clearly documented first. otherwise, people will keep filing issues about something that they expected to work, based on a random interpretation or some thread they read somewhere. for instance, you mentioned `sc:compile` but i have no idea what it does, and don't know where to find out.

so, documentation should be the first step, before this change. @tgodzik thoughts?

the link you shared before, https://github.com/lampepfl/dotty/tree/master/scaladoc, contains no specific information on what is supported (just says "markup"). all the `virtuslab` links it contains produce a "resource not found" error. i'd like to see all elements documented and explained, at least as well as in https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html which is what was used for the current scaladoc parser implementation. (keep in mind that there's another implementation in scalameta, under contrib, which doesn't fully follow that page and is therefore a bit less structured.)

tgodzik: I think we should fix the docs for sure, I will talk to the team responsible. I think scaladoc for Scala 3 follows the same guidelines as the one for Scala 3 plus Markdown code blocks, which are actually documented, so this wouldn't be a surprise to the users.

tgodzik: We can wait with with the merge until scaladoc is better documented and concurrently see if it's actually possible to format code blocks in scalafmt using the wiki syntax `{{{`

BarkingBad: Hi @kitbellew, here is [PR](https://github.com/lampepfl/dotty/pull/12083) adding docs for scaladoc.
To have a quick preview how these docs would look like after merge look at [rendered docs](https://scala3doc.virtuslab.com/pr-update-docs/scala3/docs/usage/scaladoc/index.html) or simple README under `scaladoc/` [directory](https://github.com/lampepfl/dotty/blob/0288354bf504130ca546313d3abc58ac0f759deb/scaladoc/README.md)
I'd love to hear your opinion and what sections do we should cover.

As it comes to `sc:compile` this part of work is not yet documented, as it's experimental feature developed by @pikinier20 and is subject to changes, but it's related to snippet compiler about which you can read more [here](https://contributors.scala-lang.org/t/snippet-validation-in-scaladoc-for-scala-3/4976)

Please let us know what we could improve

kitbellew: @barkingbad thank you, i made a small comment on that pr.

@tdudzik found some documentation for commonmark, seems that we need to support `` ``` `` with or without `scala` label. in fact, it's plausible that the code comes from another language as an example of what the code below the scaladoc had been based on.

so, why don't we indeed add a new `MDCodeBlock` class, with optional language, optional parameters a-la `sc:compile`, and some code lines (preserving indentation)?

tdudzik:
Implemented in https://github.com/scalameta/scalameta/pull/2338

PR #2337, opened 2021-05-10
Fix Name.Anonymous positions once again. 

dos65:
In case if there is an underlying token `this` or `_` it should point on it.
Also fixed `using` mod position in lambdas

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2151,7 +2151,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       case Ident(_) => termName()
       case LeftBrace() => dropTrivialBlock(expr(location = NoStat, allowRepeated = true))
       case KwThis() =>
-        val qual = autoPos(Name.Anonymous()); next();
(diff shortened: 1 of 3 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2470,7 +2470,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
                 case name: Term.Placeholder =>
                   Some(
                     atPos(tree, tree)(
-                      Term.Param(Nil, autoPos(Name.Anonymous()), None, None)
+                      Term.Param(Nil, atPos(name, name)(Name.Anonymous()), None, None)
```
kitbellew, line 2473: is this case tested? i thought this was the one that was failing for me originally, i don't see it in new tests.

kitbellew (approved):

kitbellew:
@dos65 still see some problems: printing trees for
```
val foo = new Bar {  _ =>
  println("a")
}
```
results in
```
SELF: TYPE: Self$SelfImpl
SOURCE: _
STRUCTURE: Self(Name(""), None)
TOKENS: _

NAME: TYPE: Name$Anonymous$NameAnonymousImpl
SOURCE: 
STRUCTURE: Name("")
TOKENS: 
```

PR #2341, opened 2021-05-13
Allow braceless package definitions

tgodzik:
It turns out we are still missing some syntax changes for optional braces. We can now also define packages with body using `:`, which is similar to previously available:
```
package a {
  object A
}
```

This commit fixes that.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -5023,7 +5023,16 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
   def packageDef(): Pkg = autoPos {
     accept[KwPackage]
-    Pkg(qualId(), inBracesOrNil(topStatSeq()))
+    def packageBody = {
+      if (isColonEol(token)) {
+        accept[Colon]
+        in.observeIndented()
+        indented(topStatSeq())
+      } else {
+        inBracesOrNil(topStatSeq())
+      }
+    }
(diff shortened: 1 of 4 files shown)


scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -439,7 +439,6 @@ object Defn {
 
 @ast class Pkg(ref: Term.Ref, stats: List[Stat]) extends Member.Term with Stat {
   checkFields(ref.isQualId)
-  checkFields(stats.forall(_.isTopLevelStat))
```
tgodzik, line 442: This is no longer true for packages in Scala 3 and we have no way of checking that here. However this is still checked while parsing.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -5055,7 +5064,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     def inBracelessPackage(): Boolean = token.is[KwPackage] && !ahead(token.is[KwObject]) && ahead {
       qualId()
       if (token.is[LF]) { next() }
-      token.isNot[LeftBrace]
+      token.isNot[LeftBrace] && (!dialect.allowSignificantIndentation || !isColonEol(token))
```
kitbellew, line 5067: completely optional but marginally easier (for me :) ):
```
&& !(dialect.allowSignificantIndentation && isColonEol(token))
```

kitbellew (approved):

dos65 (approved):
LGTM!

PR #2343, opened 2021-05-16
Allow QuotedMacroType to be used in pattern matches

tgodzik:
Connected to https://github.com/scalameta/scalafmt/issues/2512


kitbellew (approved):
optional: create a `QuotedMacroBody` interface and derive both from it, removing `checkFields`.

tgodzik:
> optional: create a `QuotedMacroBody` interface and derive both from it, removing `checkFields`.

I think that would currently break Mima, so I'd rather keep as is especially that would don't actually have common fields.

Issue #2323, opened 2021-04-30
Support for some Scala 3 syntactic constructs in Scala 2 with -Xsource:3


smarter:
Not sure if this is the correct place to discuss this, but this is a heads-up that Scala 2.13.6 and 2.12.14 will support a subset of the Scala 3 syntax under `-Xsource:3`:
- [Support ? as wildcard marker](https://github.com/scala/scala/pull/9560)
- [Support case in pattern bindings](https://github.com/scala/scala/pull/9558)
- [Support Scala 3 wildcard and renaming imports](https://github.com/scala/scala/pull/9582)
- [Allow soft keywords open and infix](https://github.com/scala/scala/pull/9580)
- [Support Scala 3 vararg splice syntax](https://github.com/scala/scala/pull/9584)
- [Support writing & instead of with in types](https://github.com/scala/scala/pull/9594)

These are all just parser changes intended to ease cross-compilation and it would be great if metals/scalameta/scalafmt/scalafix/... could understand these constructs in a Scala 2 codebase without the user having to do anything in particular. Note that while all these changes are currently under `-Xsource:3`, some of them might end up being enabled by default in future Scala 2 releases to ease migration.

tgodzik:
Thanks for the heads up! We have two choices here:
1. Add these features to Scala212 and Scala213 dialects, which would make it all easier to use and would not require any changes in other tools, but would mean that we would parse things with the assumption of `-Xsource:3`
2. Create separate dialects, which means we would have to detect `-Xsource:3` it in every tool that uses scalameta parser
3. Modification of 1., where we would also create LegacyScala212 and LegacyScala213 dialect, which can be used explicitely whenever needed.

I would honestly opt for `3` as it would mean most users will not really be affected and would not need to do anything. Scalafmt, Scalafix, Mdoc do not need errors from the parser and we should be able to handle it Metals thanks to recent changes by @dos65 

Opinions, anyone? @olafurpg @ckipp01 @kitbellew  @mlachkar ?

kitbellew:
@tgodzik scalafmt rewrite rules use these flags, so they had better be enabled in reality.

given the previous difficulty in supporting `try (x + y).foo` before 2.13.3, i'd say that scalameta needs to implement
```
case class ScalaVersion(major: Int, minor: Int, patch: Int)
object Dialect {
  ...
  def getByVersion(version: ScalaVersion): Dialect
}
```

[...]

PR #2344, opened 2021-05-17
Add support for Scala 2.13.6

tgodzik:
Also fixes https://github.com/scalameta/scalameta/issues/2323 by adding an additional dialect for use with source:3 option.

I don't think we can do much more on Scalameta side, since when parsing we don't have access to scalac options.

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
@@ -58,6 +58,30 @@ package object dialects {
     .withAllowNumericLiteralUnderscoreSeparators(true)
     .withAllowTryWithAnyExpr(true)
 
+  /**
+   *  Dialect starting with Scala 2.13.6 for `-Xsource:3` option
+   */
+  implicit val Scala213Source3 = Scala213
+    .withAllowQuestionMarkPlaceholder(true)
+    .withAllowAsForImportRename(true)
+    .withAllowStarWildcardImport(true)
+    .withAllowOpenClass(true)
+    .withAllowInfixMods(true)
+    .withAllowPostfixStarVarargSplices(true)
+    .withAllowAndTypes(true)
+
+  /**
+   *  Dialect starting with  Scala 2.12.14 for `-Xsource:3` option
+   */
+  implicit val Scala212Source3 = Scala212
+    .withAllowQuestionMarkPlaceholder(true)
+    .withAllowAsForImportRename(true)
+    .withAllowStarWildcardImport(true)
+    .withAllowOpenClass(true)
+    .withAllowInfixMods(true)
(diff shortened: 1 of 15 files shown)


scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -58,6 +58,30 @@ package object dialects {
     .withAllowNumericLiteralUnderscoreSeparators(true)
     .withAllowTryWithAnyExpr(true)
 
+  /**
+   *  Dialect for Scala 2.13.6 with `-Xsource:3` option
+   */
+  implicit val Scala213Source3 = Scala213
+    .withAllowQuestionMarkPlaceholder(true)
```
tgodzik, line 65: I haven't added `case` in for expression since we don't actually have a separate option for that, I can add it if needed later, but it's rather a minor problem - it will parse it always currently.

tgodzik:
At this point we are waiting for JS and Native releases which are on their way.

semanticdb/scalac/library/src/main/scala-2.13.6/scala/meta/internal/semanticdb/scalac/VersionCompilerOps.scala
```diff
@@ -0,0 +1,9 @@
+package scala.meta.internal.semanticdb.scalac
+
+import scala.tools.nsc.interactive.Global
+
+trait VersionCompilerOps {
+
+  def forceWarnings(global: Global) = {}
```
tgodzik, line 7: This got fixed in 2.13.6

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/LanguageOps.scala
```diff
@@ -2,20 +2,41 @@ package scala.meta.internal.semanticdb.scalac
...
+import scala.util.control.NonFatal
+import scala.tools.nsc.settings.ScalaVersion
 
 trait LanguageOps { self: SemanticdbOps =>
 
   lazy val language: String = {
     val version = Properties.versionNumberString
     if (version.startsWith("2.10")) "Scala210"
     else if (version.startsWith("2.11")) {
-      if (!version.endsWith("-bin-typelevel-4")) "Scala211"
```
tgodzik, line 12: Typelevel no longer has a separate compiler.

dos65 (approved):
:+1:

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/LanguageOps.scala
gabro: Good call, I've also removed these options from AST explorer

PR #2346, opened 2021-05-17
Enable xml and symbol literals in the Scala3 dialect

smarter:
- xml literals can be parsed by Scala 3 when scala-xml is on the
  classpath, just like in Scala 2.
- Symbol literals can still be used with
  `import scala.language.deprecated.symbolLiterals`


tgodzik (approved):
LGTM! Thanks for the contribution, I haven't realised that both are still supported.

Issue #2347, opened 2021-05-17
Parse underscore type lambdas (And `+_` and `-_` symbols)


neko-kai:
Recently added to Scala 2 in: https://github.com/scala/scala/pull/9605 (Under flag `-Xsource:3` + kind-projector flag `-P:kind-projector:underscore-placeholders` https://github.com/typelevel/kind-projector/pull/186)
And to Scala 3 in: https://github.com/lampepfl/dotty/pull/12378 (Under flag `-Ykind-projector:underscores`)

Note: Scala (with the above flags) now accepts tokens `+`/`-` followed by `_` as a type lambda placeholder. Some examples w/ edge cases below:
- https://github.com/scala/scala/pull/9605/files#diff-9cc41709c754ef3f0d8f37556e954ff08ce6a8531205c710a4fa84eb59db145a
- https://github.com/lampepfl/dotty/pull/12378/files#diff-95ae2d65c0a03a3dee556e5d610eaa65bfb4d614282904302d3baed217ba659c
- https://scalacenter.github.io/scala-3-migration-guide/docs/tutorials/kind-projector.html

dos65:
@neko-kai Will this change be included in 2.12.14  + `-Xsource:3`?

neko-kai:
@dos65 Yes

PR #2348, opened 2021-05-18
Allow `-_` and `+_` as type names for `-Xsource3` flag

dos65:
Addresses #2347

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
@@ -158,7 +158,16 @@ final class Dialect private (
     // Scala 3 no longer allows def hello(){} - `=` is always needed
     val allowProcedureSyntax: Boolean,
     // Scala 3 no longer allows `do {...} while(...)`
-    val allowDoWhile: Boolean
+    val allowDoWhile: Boolean,
+    /* Kind-project support
+     * works under -Xsource3 flag
+     * https://github.com/scala/scala/pull/9605
+     */
+    val allowPlusMinusUnderscoreAsIdent: Boolean,
+    /* The same as previous but for Scala3
+     * works under -Ykind-projector:underscores
+     */
+    val allowPlusMinusUnderscoreAsPlaceholder: Boolean
 ) extends Product with Serializable {
 
   // NOTE(olafur) checklist for adding a new dialect field in a binary compatible way:
@@ -253,7 +262,9 @@ final class Dialect private (
       allowAsForImportRename = false,
       allowStarWildcardImport = false,
       allowProcedureSyntax = true,
-      allowDoWhile = true
+      allowDoWhile = true,
+      allowPlusMinusUnderscoreAsIdent = false,
+      allowPlusMinusUnderscoreAsPlaceholder = false
       // NOTE(olafur): declare the default value for new fields above this comment.
     )
   }
@@ -455,6 +466,14 @@ final class Dialect private (
   def withAllowDoWhile(newValue: Boolean): Dialect = {
     privateCopy(allowDoWhile = newValue)
   }
+
+  def withAllowPlusMinusUnderscoreAsIdent(newValue: Boolean): Dialect = {
+    privateCopy(allowPlusMinusUnderscoreAsIdent = newValue)
+  }
+
+  def withAllowPlusMinusUnderscoreAsPlaceholder(newValue: Boolean): Dialect = {
+    privateCopy(allowPlusMinusUnderscoreAsPlaceholder = newValue)
+  }
   // NOTE(olafur): add the next `withX()` method above this comment. Please try
   // to use consistent formatting, use `newValue` as the parameter name and wrap
   // the body inside curly braces.
@@ -520,7 +539,9 @@ final class Dialect private (
       allowAsRenames: Boolean = this.allowAsForImportRename,
       allowStarWildcardImport: Boolean = this.allowStarWildcardImport,
       allowProcedureSyntax: Boolean = this.allowProcedureSyntax,
-      allowDoWhile: Boolean = this.allowDoWhile
+      allowDoWhile: Boolean = this.allowDoWhile,
+      allowPlusMinusUnderscoreAsIdent: Boolean = this.allowPlusMinusUnderscoreAsIdent,
+      allowPlusMinusUnderscoreAsPlaceholder: Boolean = this.allowPlusMinusUnderscoreAsPlaceholder
       // NOTE(olafur): add the next parameter above this comment.
   ): Dialect = {
     new Dialect(
@@ -583,7 +604,9 @@ final class Dialect private (
       allowAsRenames,
       allowStarWildcardImport,
       allowProcedureSyntax,
-      allowDoWhile
+      allowDoWhile,
+      allowPlusMinusUnderscoreAsIdent,
+      allowPlusMinusUnderscoreAsPlaceholder
       // NOTE(olafur): add the next argument above this comment.
     )
   }
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
@@ -69,6 +69,7 @@ package object dialects {
     .withAllowInfixMods(true)
     .withAllowPostfixStarVarargSplices(true)
     .withAllowAndTypes(true)
+    .withAllowPlusMinusUnderscoreAsIdent(true)
 
   /**
    *  Dialect starting with  Scala 2.12.14 for `-Xsource:3` option
@@ -81,6 +82,7 @@ package object dialects {
     .withAllowInfixMods(true)
     .withAllowPostfixStarVarargSplices(true)
     .withAllowAndTypes(true)
+    .withAllowPlusMinusUnderscoreAsIdent(true)
 
   implicit val Scala = Scala213 // alias for latest Scala dialect.
 
@@ -152,6 +154,7 @@ package object dialects {
     .withAllowStarWildcardImport(true)
     .withAllowProcedureSyntax(false)
     .withAllowDoWhile(false)
+    .withAllowPlusMinusUnderscoreAsPlaceholder(true)
 
   @deprecated("Use Scala3 instead", "4.4.2")
   implicit val Dotty = Scala3
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1719,6 +1719,9 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       else Type.Annotate(t, annots)
     }
 
+    private def allowPlusMinusUnderscore: Boolean =
+      dialect.allowPlusMinusUnderscoreAsIdent || dialect.allowPlusMinusUnderscoreAsPlaceholder
+
     def simpleType(): Type = {
       simpleTypeRest(autoPos(token match {
         case LeftParen() => autoPos(makeTupleType(inParens(types())))
@@ -1729,6 +1732,18 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           Type.Macro(macroSplice())
         case Ident("?") if dialect.allowQuestionMarkPlaceholder =>
           next(); atPos(in.prevTokenPos, auto)(Type.Placeholder(typeBounds()))
+        case ident: Ident
+            if (ident.value == "+" || ident.value == "-") &&
+              ahead(token.is[Underscore]) &&
+              allowPlusMinusUnderscore =>
+          autoPos {
+            accept[Ident]
+            accept[Underscore]
+            if (dialect.allowPlusMinusUnderscoreAsPlaceholder)
+              Type.Placeholder(typeBounds())
+            else
+              Type.Name(s"${ident.value}_")
+          }
         case Literal() =>
           if (dialect.allowLiteralTypes) literal()
           else syntaxError(s"$dialect doesn't support literal types", at = path())
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/TypeSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/TypeSuite.scala
@@ -159,4 +159,11 @@ class TypeSuite extends ParseSuite {
     assertNoDiff(exceptionScala2.shortMessage, "illegal literal type (), use Unit instead")
 
   }
+
+  test("plus-minus-then-underscore-source3") {
+    val Type.Function(List(Type.Name("+_")), Type.Name("Int")) =
+      tpe("+_ => Int")(dialects.Scala213Source3)
+    val Type.Apply(Type.Name("Option"), List(Type.Name("-_"))) =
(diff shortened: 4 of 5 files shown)


scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -152,6 +154,7 @@ package object dialects {
     .withAllowStarWildcardImport(true)
     .withAllowProcedureSyntax(false)
     .withAllowDoWhile(false)
+    .withAllowPlusMinusBeforeUnderscore(true)
```
dos65, line 157: In Scala3 it will work under `-Ykind-projector:underscores` but I think it's safe to enable it by default

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1729,12 +1729,18 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+            ) && dialect.allowPlusMinusBeforeUndescore =>
+          autoPos {
+            accept[Ident]
+            accept[Underscore]
+            Type.Name(s"${ident.value}_")
+          }
         case Literal() =>
           if (dialect.allowLiteralTypes) literal()
           else syntaxError(s"$dialect doesn't support literal types", at = path())
-        case Ident("-") if ahead { token.is[NumericLiteral] } && dialect.allowLiteralTypes =>
```
tgodzik, line 1735: Why remove this one? It's unrelated to the new syntax

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
tgodzik: I wonder if we should have Scala32 dialect for it?

tgodzik (changes requested):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: oops :facepalm:

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
dos65: Scala3 compiler doesn't accept such syntax without this flag.
It's a kind of `less-restrictive` setting by default.

neko-kai: @tgodzik Scala32 will accept underscores for type lambdas, but the `+_`/`-_` will forever be optional under a flag (they make no sense in Scala 3 since lambda parameter variance is inferred, they are parsed so that Scala 2 code can compile, but the marker itself is ignored)

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -158,7 +158,12 @@ final class Dialect private (
     // Scala 3 no longer allows def hello(){} - `=` is always needed
     val allowProcedureSyntax: Boolean,
     // Scala 3 no longer allows `do {...} while(...)`
-    val allowDoWhile: Boolean
+    val allowDoWhile: Boolean,
+    /* Kind-project support
+     * works under -Xsource3 flag
+     * https://github.com/scala/scala/pull/9605
+     */
+    val allowPlusMinusUnderscoreIdent: Boolean
```
kitbellew, line 166: I would rename this. Looking at the scala code and doc, I'd call this `allowUnderscoreAsTypeWildcard` and set to false in scala3 and scala2source3. the plus/minus thing is just a side effect because it needs co/contravariant types.

kitbellew (commented):
sorry for butting in with yet another naming suggestion... @tgodzik @dos65 see if it makes more semantic (rather than syntactic) sense to you, as it did for me.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1729,6 +1729,15 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
         case Ident("?") if dialect.allowQuestionMarkPlaceholder =>
           next(); atPos(in.prevTokenPos, auto)(Type.Placeholder(typeBounds()))
+        case ident: Ident
+            if (ident.value == "+" || ident.value == "-") &&
+              ahead(token.is[Underscore]) &&
+              dialect.allowPlusMinusUnderscoreIdent =>
+          autoPos {
+            accept[Ident]
+            accept[Underscore]
+            Type.Name(s"${ident.value}_")
```
kitbellew, line 1739: should Type.Placeholder be subclassed or supplemented with `Type.CovariantPlaceholder` and `Type.ContravariantPlaceholder` and used here? after all, `+_` is closer to a placeholder than an identifier, no?

tests/shared/src/test/scala/scala/meta/tests/parsers/TypeSuite.scala
```diff
@@ -159,4 +159,11 @@ class TypeSuite extends ParseSuite {
     assertNoDiff(exceptionScala2.shortMessage, "illegal literal type (), use Unit instead")
 
   }
+
+  test("plus-minus-then-undescore-source3") {
+    val Type.Function(List(Type.Name("+_")), Type.Name("Int")) =
+      tpe("+_ => Int")(dialects.Scala213Source3)
+    val Type.Apply(Type.Name("Option"), List(Type.Name("-_"))) =
+      tpe("Option[- _]")(dialects.Scala213Source3)
```
kitbellew, line 167: what about scala3, not just scala2source3?

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -152,6 +154,7 @@ package object dialects {
     .withAllowStarWildcardImport(true)
     .withAllowProcedureSyntax(false)
     .withAllowDoWhile(false)
+    .withAllowPlusMinusUnderscoreIdent(true)
```
kitbellew, line 157: should we start creating additional scala3 dialects? if we have lots of scala2.1x, why not scala3.1 and scala3.2?

tgodzik: We will most likely need to, but currently I am not sure what will be in 3.1 at this point.

tests/shared/src/test/scala/scala/meta/tests/parsers/TypeSuite.scala
tgodzik: We usually just test the one dialect that it's enabled in.

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
kitbellew: we know that this flag we are adding is only in 3.2, right? at least according to the docs.

tests/shared/src/test/scala/scala/meta/tests/parsers/TypeSuite.scala
kitbellew: but we are also enabling it in scala 3, which should be tested or removed

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: do we need to add a test for this, so that no accidental regressions?

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
neko-kai: @kitbellew The new syntax is already available under a flag `-Ykind-projector:underscores` in 3.
See
* https://docs.scala-lang.org/scala3/guides/migration/plugin-kind-projector.html
* https://dotty.epfl.ch/docs/reference/changed-features/wildcards.html

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: no, actually I was distracted by GitHub actions failure so I didn't look on the diff.

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
dos65: We already have `allowTypeParamUndesrcore` setting

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: I thought about that. However, it's unclear if just a `Name` or `Placeholder`. 
In Scala2 parser treat it as `Name` and then if kind-projector is enabled it might be transformed into `Placeholder`.

In Scala3 it will be `Placeholder`. Are there any cons of having `Placeholder` here?

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
kitbellew: so why not use it? this is how it is described:
```
    // Dotty rejects placeholder as Type parameter
    val allowTypeParamUnderscore: Boolean,
```
yet it is never set to false anywhere and never used in parsing.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: meaning, ignore the covariance (since, according to @neko-kai, it will be inferred) and just pretend it was a simple `_`? maybe.

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
dos65: I agree that this one fits better. 
But in this case non `Source3` scala2 dialects will accept this new syntax too.
Also, I forgot that we also need to enable `allowTypeParamUnderscore` for scala3

@kitbellew @tgodzik  are you ok with that?

dos65: Oh. `allowTypeParamUnderscore` is about different thing.
```scala
def f[_](a: Int): Int = ???
```

Ok, then the question is: are you ok to enable this syntax by default without any flag?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: Currently, scalameta parses kind-projector syntax (`+*`/`-*`/`+?`/`-?`) just as `Type.Name`.
An additional thing here that we don't have variance mods for `Type.Placeholder`.

I think the best option here would be to turn `+_`, `-_` into `Type.Placeholder` for Scala3 and keep them as `Type.Name` for `Scala2*`.

@kitbellew @tgodzik wdyt?

kitbellew: > I think the best option here would be to turn `+_`, `-_` into `Type.Placeholder` for Scala3 and keep them as `Type.Name` for `Scala2*`.

sounds good

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
kitbellew: > Oh. `allowTypeParamUnderscore` is about different thing.
> 
> ```scala
> def f[_](a: Int): Int = ???
> ```

Right, it's for a type parameter of a definition, not for a type definition. still, why can't i find any reads for `allowTypeParamUnderscore`, and never `false` for it?

> Ok, then the question is: are you ok to enable this syntax by default without any flag?

`allowUnderscoreAsTypeWildcard` (or whatever you decide to call it) shouldn't be part of `scala213`.

i am thinking that we should implement some sort of reflection to allow a user to specify config
```
dialect {
  base = scala213
  options {
    allowUnderscoreAsTypeWildcard = true
  }
}
```
that way, we don't need to add it as part of any explicit dialect, we don't need `scala213source3` either.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: :+1:

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
dos65: > Right, it's for a type parameter of a definition, not for a type definition. still, why can't i find any reads for allowTypeParamUnderscore, and never false for it?

It's used in several places. For example - [here](https://github.com/scalameta/scalameta/blob/55606a8a37c6be603c7eecc2639a75573af72ad7/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L4114)

What about setting rename. I decided to keep the existing one.
I looked on it one more time and as we currently always parse `_` as `Type.Placeholder` without any settings so the name you proposed doesn't fit for that.

dos65:
Summarizing all comments the final changeset is:
- new flag `allowPlusMinusUnderscoreAsIdent = true` for `Scala2*Source3`
- new flag `allowPlusMinusUnderscoreAsPlaceholder = true` for `Scala3`

tgodzik (approved):
LGTM!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1729,6 +1732,18 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+            if (ident.value == "+" || ident.value == "-") &&
+              ahead(token.is[Underscore]) &&
+              allowPlusMinusUnderscore =>
+          autoPos {
+            accept[Ident]
+            accept[Underscore]
+            if (dialect.allowPlusMinusUnderscoreAsPlaceholder)
+              Type.Placeholder(typeBounds())
+            else
+              Type.Name(s"${ident.value}_")
```
kitbellew, line 1745: does this version not allow type bounds?

dos65: yep, but it is in line with scala/scala#9605

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
kitbellew: > What about setting rename. I decided to keep the existing one.
> I looked on it one more time and as we currently always parse `_` as `Type.Placeholder` without any settings so the name you proposed doesn't fit for that.

Perhaps I should have been clearer. My proposal (with wildcards) would have had the opposite value; i.e., if you [still] allow wildcards, then it will not be parsed with +/-.

PR #2362, opened 2021-05-24
Allow next apply after newline and indentation

tgodzik:
Related to https://github.com/lampepfl/dotty/issues/12554 and https://github.com/scalameta/scalafmt/issues/2522

It is now possible to have the next apply in curried functions to start in the next line if there is an indentation:

```scala
def f(s: String)(t: String) = s"$t: $s"

def g =
  f("Foo")
    ("Bar")
```

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -247,6 +247,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   trait TokenIterator extends Iterator[Token] {
     def prevTokenPos: Int
     def tokenPos: Int
+    def currentIndentation: Int
     def token: Token
     def fork: TokenIterator
     def observeIndented(): Boolean
@@ -336,6 +337,14 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
         }
         RegionIndentEnum(i) :: nextPrev
       })
+    }
+
+    def currentIndentation: Int = {
+      val foundIndentation = countIndent(curr.pointPos)
+      if (foundIndentation < 0)
+        sepRegions.headOption.fold(-1)(_.indent)
+      else
+        foundIndentation
     }
 
     def observeOutdented(): Boolean = {
@@ -679,6 +688,10 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
   private def countIndent(tokenPosition: Int): Int =
     countIndentAndNewlineIndex(tokenPosition)._1
+
+  private def currentIndentation: Int = {
+    in.currentIndentation
+  }
 
   @tailrec
   private def isAheadNewLine(currentPosition: Int): Boolean = {
@@ -2110,6 +2123,15 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     if (token.is[LF] && ahead { token.is[T] }) newLineOpt()
   }
 
+  def newLineOptWhenFollowedBySignificantIndentationAnd(cond: Token => Boolean): Unit = {
+    def nextLineHasGreaterIndentation = {
+      val prev = currentIndentation
+      prev > 0 && ahead { cond(token) && currentIndentation > prev }
+    }
+    if (token.is[LF] && nextLineHasGreaterIndentation)
+      newLineOpt()
+  }
+
   def newLineOptWhenFollowing(p: Token => Boolean): Unit = {
     // note: next is defined here because current is token.LF
     if (token.is[LF] && ahead { p(token) }) newLineOpt()
@@ -2929,7 +2951,13 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   }
 
   def simpleExprRest(t: Term, canApply: Boolean): Term = atPos(t, auto) {
-    if (canApply) newLineOptWhenFollowedBy[LeftBrace]
+    if (canApply) {
+      if (dialect.allowSignificantIndentation) {
+        newLineOptWhenFollowedBySignificantIndentationAnd(x => x.is[LeftBrace] || x.is[LeftParen])
+      } else {
+        newLineOptWhenFollowedBy[LeftBrace]
+      }
+    }
     token match {
       case Dot() =>
         next()
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/SignificantIndentationSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/SignificantIndentationSuite.scala
@@ -1563,4 +1563,170 @@ class SignificantIndentationSuite extends BaseDottySuite {
       )
     )
   }
+
+  test("indented-apply") {
+    runTestAssert[Stat](
+      """|def method = 
+         |  fun(a,b,c)
+         |    (d, e)
+         |""".stripMargin,
+      assertLayout = Some(
+        "def method = fun(a, b, c)(d, e)"
+      )
+    )(
+      Defn.Def(
+        Nil,
+        Term.Name("method"),
+        Nil,
+        Nil,
+        None,
+        Term.Apply(
+          Term.Apply(Term.Name("fun"), List(Term.Name("a"), Term.Name("b"), Term.Name("c"))),
+          List(Term.Name("d"), Term.Name("e"))
+        )
+      )
+    )
+  }
+
+  test("indented-double-apply") {
+    runTestAssert[Stat](
+      """|def method = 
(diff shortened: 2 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2110,6 +2126,12 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     if (token.is[LF] && ahead { token.is[T] }) newLineOpt()
   }
 
+  def newLineOptWhenFollowedByIndentationAnd[T <: Token: TokenInfo]: Unit = {
+    val prev = currentIndentation
```
kitbellew, line 2130: this operation is pretty expensive. I'd check for LF and dialect flag first, then do this in the block.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2110,6 +2126,12 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+    if (token.is[LF] && prev > 0 && dialect.allowSignificantIndentation &&
+      ahead { token.is[T] && currentIndentation > prev }) newLineOpt()
```
kitbellew, line 2132: does the formatter allow this? i thought ahead was the body..

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -680,6 +692,10 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   private def countIndent(tokenPosition: Int): Int =
     countIndentAndNewlineIndex(tokenPosition)._1
 
+  private def currentIndentation: Int = {
```
kitbellew, line 695: unrelated question to method 10 lines above. why does it check for whitespace in two differently ways? isWhitespace vs is[Whitespace].

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -338,6 +339,17 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       })
     }
 
+    def currentIndentation: Int = {
+      val foundIndentation = countIndent(curr.pointPos)
+      if (foundIndentation < 0)
+        sepRegions.headOption match {
```
kitbellew, line 345: instead of match, you could .fold(-1)(_.indent).

kitbellew (changes requested):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: `isWhitespace` is only used to count spaces or tabs, so in reality it should be `isSpacesOrTabs`, but it's an inner function, so this should not be a big issue.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: sure, moved it to a separate def together with the ahead check.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Not sure I understand? This parses and formats properly. I moved it out to a separate def anyway.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2929,7 +2951,10 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   }
 
   def simpleExprRest(t: Term, canApply: Boolean): Term = atPos(t, auto) {
-    if (canApply) newLineOptWhenFollowedBy[LeftBrace]
+    if (canApply) {
+      newLineOptWhenFollowedBy[LeftBrace]
```
dos65, line 2955: I played with the original sample and it looks like Scala3 also has a similar rule for `LeftBrace`.

```scala
def fun(a: Int, b: Int, c: Int)(d: Int): String = ???
def method: String = 
  fun(1, 2, 3) 
    {4} // <- this  Term.Apply(Term.Apply("fun", ...), "d")
def method2: String = 
  fun(1, 2, 3)
  {4} // <- doesn't compile - `{4}` is not part of apply
      // but scalameta parses it the same way as above -  Term.Apply(Term.Apply("fun", ...), "d")
```

It seems that this part should be smth like:
```scala
if (canApply) {
  if (dialect.allowSignificantIndentation) {
     // combine these methods in one?
     newLineOptWhenFollowedByIndentationAnd[LeftParen]
     newLineOptWhenFollowedByIndentationAnd[LeftBrace]
  } else {
    newLineOptWhenFollowedBy[LeftBrace]
  }
}
```

tgodzik: Good catch! I was mislead a bit by Martin's comment, since he explicitely mentioned `)(`

Thanks! This should be fixed now.

dos65 (approved):
LGTM!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2929,7 +2951,14 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   }
 
   def simpleExprRest(t: Term, canApply: Boolean): Term = atPos(t, auto) {
-    if (canApply) newLineOptWhenFollowedBy[LeftBrace]
+    if (canApply) {
+      if (dialect.allowSignificantIndentation) {
+        newLineOptWhenFollowedByIndentationAnd[LeftBrace]
+        newLineOptWhenFollowedByIndentationAnd[LeftParen]
```
kitbellew, line 2957: you perform a ton of the same checks... i'd merge into one

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2110,6 +2123,15 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
 
+  def newLineOptWhenFollowedByIndentationAnd[T <: Token: TokenInfo]: Unit = {
+    def nextLineHasGreaterIndentation = {
+      val prev = currentIndentation
+      prev > 0 && ahead { token.is[T] && currentIndentation > prev }
+    }
+    if (token.is[LF] && dialect.allowSignificantIndentation && nextLineHasGreaterIndentation)
+      newLineOpt()
+  }
+
```
kitbellew, line 2134:
```suggestion
  def newLineOptWhenFollowedBySignificantIndentationAnd(cond: Token => Boolean): Unit = {
    def nextLineHasGreaterIndentation = {
      val prev = currentIndentation
      prev > 0 && ahead { cond(token) && currentIndentation > prev }
    }
    if (token.is[LF] && nextLineHasGreaterIndentation)
      newLineOpt()
  }

```

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew, line 2957:
```suggestion
        newLineOptWhenFollowedBySignificantIndentationAnd(x => x.is[LeftBrace] || x.is[LeftParen])
```

kitbellew (changes requested):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Good idea!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala

kitbellew (approved):

PR #2363, opened 2021-05-24
Dialect: add scala version, remove `allowPlusMinusUnderscoreAsIdent`

kitbellew:
Instead, use a combination of `allowPlusMinusUnderscoreAsPlaceholder` with `scalaMajor`.

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
@@ -33,8 +33,10 @@ package object dialects {
     allowXmlLiterals = true, // Not even deprecated yet, so we need to support xml literals
     toplevelSeparator = ""
   )
+    .withScalaVersion(2, 1, 0)
 
   implicit val Scala211 = Scala210
+    .withScalaVersion(2, 1, 1)
     .withAllowCaseClassWithoutParameterList(false)
     .withAllowSpliceUnderscores(true) // SI-7715, only fixed in 2.11.0-M5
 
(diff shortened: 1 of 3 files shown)


scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -33,8 +33,10 @@ package object dialects {
     allowXmlLiterals = true, // Not even deprecated yet, so we need to support xml literals
     toplevelSeparator = ""
   )
+    .withScalaVersion(2, 1, 0)
 
   implicit val Scala211 = Scala210
+    .withScalaVersion(2, 1, 1)
```
tgodzik, line 39:
```suggestion
    .withScalaVersion(2, 12, 0)
```
otherwise we risk people getting confused by this. Same in other places.

tgodzik (commented):
I wouldn't do the change unless we have more use cases than this single one. We are trading in one parameter for 3 new ones, which are not especially intuitive and exactly correct in regards to how semantic versioning normally works (I do understand the intent here, but it might be super confusing in the future). I am also not sure how this helps us in the long run.

Having separate dialect options for each feature proved to be super useful especially with the `-Xsource:3` flag. With `withScalaVersion` we would embed the dialect specific logic inside the code, which will be much less clear than the current situation. Now, we are able to see exactly what are the different features of the parser inside of the dialect, but with this change we would need to go through the code to see why exactly some of the code is parsed as is. 

TLDR; I think we don't need this change and it is against the current practice of putting everything into the Dialect.

kitbellew:
> I wouldn't do the change unless we have more use cases than this single one. We are trading in one parameter for 3 new ones, which are not especially intuitive and exactly correct in regards to how semantic versioning normally works (I do understand the intent here, but it might be super confusing in the future). I am also not sure how this helps us in the long run.

ok, fair enough.

PR #2365, opened 2021-05-25
Trees: add an option to get original input syntax

kitbellew:
Currently, the syntax is requested with an implicit dialect, and most of the time the dialect isn't set explicitly but uses whatever happened to be provided by default.

For this type of case, let's provide a way to obtain the original syntax as it was used when parsing the code.


kitbellew:
turns out the simple `.toString` already does it.

Issue #2364, opened 2021-05-24
extends token is part of the template when parsing, but not when pretty-printing


bjaglin:
https://github.com/scalameta/scalameta/issues/2326 introduced a breaking change for Scalafix rule authors. Rules would have to detect the runtime scalameta version to be able to work with older versions of Scalafix (<0.9.28) and new ones (>=0.9.28).

Before, rewriting rules replacing a `Template` tree with a pretty-printed representation of another one didn't have to include the `extends` token (unless the class didn't have any super class/trait before the rewrite).

Now, the `extends` token being captured by the `template` (in terms of position), and the `extends` token being added only when `Template` is pretty-printed within a `Defn.class`, an explicit `extends` prefix must be added to avoid ending up with invalid code `class Foo NewSuperClass`.

bjaglin:
@kitbellew I might have misqualified the problem, but in the context of [Scalafix's `replaceTree`](https://scalacenter.github.io/scalafix/docs/developers/patch.html#replacetree), one can expect (and this was the behavior before https://github.com/scalameta/scalameta/issues/2326) that replacing (in terms of position range) a `Template` in a parsed tree with another pretty-printed `Template` should not generate invalid code.

I imagine it might be uncommon to pretty-print a `Template` alone, but to minimize the impact on source code, it's a recommendation for Scalafix rule authors to replace as little as possible. We might revisit this guideline if there are exceptions?
> It's best to target as precise tree nodes as possible when using replaceTree() to avoid conflicts with other patches.

https://scalacenter.github.io/scalafix/docs/developers/patch.html#replacetree

bjaglin:
Thanks @tgodzik, I could confirm that this fixes the regression, for which I added tests in https://github.com/scalacenter/scalafix/pull/1415

PR #2366, opened 2021-05-26
Print extends when printing template

tgodzik:
Previously, we introduced the change to positions that made templates contain the `extends` keyword. We haven't had the same for printing tempaltes, which created an unintended assymetry. This fixes that behaviour.

Fixes https://github.com/scalameta/scalameta/issues/2364


scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -149,7 +149,6 @@ object TreeSyntax {
         templ.self.name.is[Name.Anonymous] &&
         templ.self.decltpe.isEmpty &&
         templ.stats.isEmpty) s()
-      else if (templ.inits.nonEmpty || templ.early.nonEmpty) s(" extends ", templ)
```
kitbellew, line 152: i don't understand how this is helping. this method (`templ`) is called on line 925 for `Defn.Class`, 936 for `Defn.Trait`, 971 for `Defn.Enum`, 993 for `Defn.Object` and 1016 for `Pkg.Object`. so for those cases it should work.

below, you are covering 4 out of these 5 cases (no `Pkg.Object`). perhaps the original submitter meant some other cases which we aren't covering yet. such as `Defn.Given`? should we ask for examples?

also, I would add plenty of tests, to make sure something didn't work before but is working now.

kitbellew (changes requested):
i don't understand how this improves the outcome.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
kitbellew: understood. template needs to print extends, not the outer tree.

tgodzik: Added some tests that should additionally highlight the issue

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
```diff
@@ -505,6 +505,36 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
     assert(templStat("class C extends (() => Int)").syntax == "class C extends (() => Int)")
   }
 
+  test("package object e extends D") {
```
kitbellew, line 508: could you please tests with `given` and `new` anonymous? and whatever else can potentially use a template.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -1083,7 +1082,14 @@ object TreeSyntax {
         if (isTemplateEmpty) s()
         else {
           val pearly = if (!t.early.isEmpty) s("{ ", r(t.early, "; "), " } with ") else s()
-          val pparents = w(r(t.inits, " with "), " ", !t.inits.isEmpty && !isBodyEmpty)
+          val suffix = if (t.inits.nonEmpty && !isBodyEmpty) " " else ""
+          val canHaveExtend = t.parent.exists(p =>
```
kitbellew, line 1086: what about `t.parent.exists(!_.is[Term.NewAnonymous])` (or `(_.is[Defn])`)?

kitbellew (changes requested):

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
kitbellew: > `Defn.Given` cannot have `extends` and I forgot about Pkg.Object.

didn't see this, sorry. still, would be good to have tests for all constructs which use template.

kitbellew: so, should this method be removed completely? it doesn't do anything anymore.

dos65 (approved):
Looks good!

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
tgodzik: Changed to `t.parent.exists(p => p.isNot[Defn.Given] && p.isNot[Term.NewAnonymous])`

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
tgodzik: Added tests in a separate suite for Scala 3 (I had some issues with the implicit dialect, so decided that it's safer)

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/Scala3SyntaxSuite.scala
```diff
@@ -0,0 +1,30 @@
...
+      )
+    assertEquals(template.syntax, "Ord[Int] with \n{ def f(): Int = 1 }")
+  }
+
+  test("enum C extends A with B { case D }") {
+    assertEquals(
+      templStat("enum C extends A with B { case D }").syntax,
+      "enum C extends A with B { case D }"
+    )
+    val q"enum $name $template" = q"enum C extends A with B { case D }"
```
tgodzik, line 27: @bjaglin quasiquotes with enums seems to work. But I had some issues with givens, however those might caused by the fact that givens can take a lot of forms. We might need to work on that depending on how much quasiquotes are sued here.

kitbellew (approved):

PR #2369, opened 2021-05-27
Show: change `wrap()` to use `Sequence` or `Wrap`, change `repeat` to use `Wrap`

kitbellew:


tgodzik (commented):
Looks good, but could you add more background on why this change is needed? I am a bit confused by the change :sweat_smile:

kitbellew:
> Looks good, but could you add more background on why this change is needed? I am a bit confused by the change 😅

I wanted to get rid of the `def templ` that you modified earlier today, because it's no longer doing anything, and it was a bit hard with all these checks for what's set and what's not.

this approach skips empty parts automatically, should be easier and less error-prone.

the change to templates is forthcoming. i didn't want to bundle with this more general pr.

tgodzik (approved):
Ok thanks for explaining!

PR #2371, opened 2021-05-28
Add support for Scala 2.12.14

tgodzik:


tests/jvm/src/test/scala/scala/meta/tests/semanticdb/OccurrenceSuite.scala
```diff
@@ -21,7 +21,16 @@ class OccurrenceSuite extends FunSuite {
     OccurrenceSuite.testCases().foreach { t =>
       test(t.name) {
         val body = t.body()
-        assertNoDiff(body.obtained, body.expected)
+        val expectedCompat = if (ScalaVersion.atLeast212_14) {
```
tgodzik, line 24: We could use some better way to work with compatibility in tests, but I don't want to reinvent it all here at this time.

dos65 (approved):
:+1:

Issue #2372, opened 2021-05-31
Scala 3 new keywords are not escaped correctly when used in names


vigoo:
Example:

```scala
  import scala.meta.dialects.Scala3

  println(Term.Name("trait").syntax)
  println(Term.Name("enum").syntax)
```

prints

```
`trait`
enum
```

This makes it hard to generate scala 3 compatible source code from arbitrary external sources where these new keywords may appear as names.

tgodzik:
Thanks for finding this! I added the new keywords to the list https://github.com/scalameta/scalameta/pull/2373

vigoo:
Wow that was fast, thank you :)

PR #2373, opened 2021-05-31
Make sure the new Scala 3 keywords are backticked

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/2372

### DIFF
--- a/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/package.scala
+++ b/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/package.scala
@@ -5,7 +5,7 @@ package object tokenizers {
   type Offset = Int
   type LegacyToken = Int
 
-  val keywords = Set(
+  private val baseKeywords = Set(
     "abstract",
     "case",
     "do",
@@ -46,7 +46,6 @@ package object tokenizers {
     "protected",
     "throw",
     "val",
-    "given",
     "_",
     ":",
     "=",
@@ -61,4 +60,18 @@ package object tokenizers {
     "\u21D2",
     "\u2190"
   )
+
+  def keywords(dialect: Dialect) = {
+    val dialectKeywords = Set.newBuilder[String]
+
+    if (dialect.allowEnums) dialectKeywords += "enum"
+    if (dialect.allowExportClause) dialectKeywords += "export"
+    if (dialect.allowGivenUsing) {
+      dialectKeywords += "given"
+      dialectKeywords += "?=>"
+    }
+    if (dialect.allowSignificantIndentation) dialectKeywords += "then"
+
+    baseKeywords ++ dialectKeywords.result()
+  }
 }
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
@@ -15,7 +15,6 @@ import Show.{
 }
 import scala.meta.internal.trees.{root => _, branch => _, _}
 import scala.meta.internal.tokenizers.Chars._
-import scala.meta.internal.tokenizers.keywords
 import org.scalameta.adt._
 import org.scalameta.invariants._
 import org.scalameta.unreachable
@@ -80,6 +79,7 @@ object TreeSyntax {
     }
   }
   private final class SyntaxInstances(dialect: Dialect) {
+    val keywords = tokenizers.keywords(dialect)
     import SyntaxInstances.SyntacticGroup
     import SyntacticGroup.Type._, SyntacticGroup.Term._, SyntacticGroup.Pat._,
     SyntacticGroup.Literal, SyntacticGroup.Path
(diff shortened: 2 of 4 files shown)


scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/package.scala
```diff
@@ -10,6 +10,8 @@ package object tokenizers {
     "case",
     "do",
     "else",
+    "enum",
+    "export",
```
kitbellew, line 14: what about `using`, `extension`, `derives` and all the other scala3 innovations?

vigoo: Also there are these: https://dotty.epfl.ch/docs/reference/soft-modifier.html
But I think I will test these separately and provide examples in a separate issue if needed

kitbellew: @vigoo @tgodzik so all hard keywords are covered, then? if so, lgtm.

vigoo: Based on this:
https://dotty.epfl.ch/docs/internals/syntax.html#regular-keywords

i think these two are missing:
```
then
?=>
```

tgodzik: Those are soft keywords and I think we covered most of the problematic cases in tests. Inline and extension being the the hardest. Let me check if it will not break the Scala 2 code actually.

kitbellew: > i think these two are missing:
> 
> ```
> ?=>
> ```

wouldn't this be covered as a non-alphanum identifier?

vigoo: > wouldn't this be covered as a non-alphanum identifier?

Well 

```scala
val ?=> = 0 // invalid
val `?=>` = 0 // invalid
val ?==> = 0 // valid
```

i think it behaves exactly as the other keywords in this regard

kitbellew: interesting :) let's add it, then.

separately, I'd sort these keywords, hard to find whether something is already there.

tgodzik: Added both and also changed it to make sure we only backtick those keywords with a proper dialect option on.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -80,6 +79,7 @@ object TreeSyntax {
     }
   }
   private final class SyntaxInstances(dialect: Dialect) {
+    val keywords = scala.meta.internal.tokenizers.keywords(dialect)
```
kitbellew, line 82: optional: since we already within `internal`, wouldn't `tokenizers.keywords(...)` work?

kitbellew (approved):

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/package.scala
```diff
@@ -61,4 +60,13 @@ package object tokenizers {
...
     "\u2190"
   )
+
+  def keywords(implicit dialect: Dialect) = {
+    val enumKeyword = if (dialect.allowEnums) Set("enum") else Set.empty[String]
+    val exportKeyword = if (dialect.allowExportClause) Set("export") else Set.empty[String]
+    val givenUsingKeywords = if (dialect.allowGivenUsing) Set("given", "?=>") else Set.empty[String]
+    val thenKeyword = if (dialect.allowSignificantIndentation) Set("then") else Set.empty[String]
+
+    baseKeywords ++ enumKeyword ++ givenUsingKeywords ++ exportKeyword ++ thenKeyword
```
kitbellew, line 70: another option:
```
val dialectKeywords - Set.newBuilder[String]
if (dialect.allows...) dialectKeywords += ...
...
baseKeywords ++ dialectKeywords.result()
```

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/package.scala
```diff
@@ -61,4 +60,13 @@ package object tokenizers {
     "\u21D2",
     "\u2190"
   )
+
+  def keywords(implicit dialect: Dialect) = {
```
kitbellew, line 64: is it to be used implicitly?

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
tgodzik: It works! Changed

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/package.scala
tgodzik: Good idea, changed it.

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/package.scala
tgodzik: Better to make it explicit. Changed

kitbellew (approved):

PR #2375, opened 2021-06-02
Fix issue with braces after function arrow

tgodzik:
Previosly, we would not parser correctly when braces would appear after function arrow and were not actually the function body. Now, we make sure to treat that situation differently so that the braces are not accepted greedily.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2557,7 +2557,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
               val trm =
                 if (location != BlockStat) expr()
                 else {
-                  blockExpr() match {
+                  blockExpr(isBlockOptional = true) match {
                     case partial: Term.PartialFunction if token.is[Colon] =>
                       accept[Colon]
                       atPos(partial, auto)(Term.Ascribe(partial, typeOrInfixType(location)))
@@ -2591,7 +2591,10 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     )
     accept[RightArrow]
     atPos(implicitPos, auto)(
-      Term.Function(List(param), if (location != BlockStat) expr() else blockExpr())
+      Term.Function(
+        List(param),
+        if (location != BlockStat) expr() else blockExpr(isBlockOptional = true)
+      )
     )
   }
 
@@ -3080,18 +3083,18 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     trees
   }
 
-  def blockExpr(): Term = autoPos {
+  def blockExpr(isBlockOptional: Boolean = false): Term = autoPos {
     if (ahead(token.is[CaseIntro] || (token.is[Ellipsis] && ahead(token.is[KwCase])))) {
       if (token.is[LeftBrace])
         Term.PartialFunction(inBraces(caseClauses()))
       else
         Term.PartialFunction(indented(caseClauses()))
-    } else block()
+    } else block(isBlockOptional)
   }
 
-  def block(): Term = autoPos {
+  def block(isBlockOptional: Boolean = false): Term = autoPos {
     def blockWithStats = Term.Block(blockStatSeq())
-    if (token.is[LeftBrace]) {
+    if (!isBlockOptional && token.is[LeftBrace]) {
       inBraces(blockWithStats)
     } else {
       val possibleBlock =
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/DefnSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/DefnSuite.scala
@@ -2,7 +2,7 @@ package scala.meta.tests
 package parsers
 
 import scala.meta._
-import scala.meta.dialects.Scala211
+import scala.meta.dialects.Scala213
 
 class DefnSuite extends ParseSuite {
   test("val x = 2") {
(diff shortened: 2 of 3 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3080,18 +3083,18 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
         Term.PartialFunction(indented(caseClauses()))
-    } else block()
+    } else block(isAfterArrow)
   }
 
-  def block(): Term = autoPos {
+  def block(isAfterArrow: Boolean = false): Term = autoPos {
     def blockWithStats = Term.Block(blockStatSeq())
-    if (token.is[LeftBrace]) {
+    if (!isAfterArrow && token.is[LeftBrace]) {
```
kitbellew, line 3097: should we rename this new parameter based on what we're doing with it here, not where it came from? what if another case could lead to this as well?

tests/shared/src/test/scala/scala/meta/tests/parsers/DefnSuite.scala
```diff
@@ -172,6 +172,61 @@ class DefnSuite extends ParseSuite {
     ) = templStat("def f(x: Int): Int = macro impl")
   }
 
+  test("braces-in-functions") {
```
kitbellew, line 175: how's this treated:
```
if (...) { ... }.select else {...} + {...}
try {...}.select
```
what is `location == BlockStat`?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: This is just the situation where {} is optional for blocks, which is only after `=>`. I think the name makes sense here.

tests/shared/src/test/scala/scala/meta/tests/parsers/DefnSuite.scala
tgodzik: I think `BlockStat` means we are in a block, which can be either indented, `=>` or {}. At least that's what I think.
I will check the other example

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: so why not rename `isBlockOptional`?

tgodzik: changed!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2294,7 +2294,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
         val body: Term = token match {
           case _ if dialect.allowTryWithAnyExpr => expr()
           case LeftParen() => inParensOrUnit(expr())
-          case LeftBrace() | Indentation.Indent() => block()
```
tgodzik, line 2297: Turns out `try {...}.select` was also problematic. `expr()` is enough here, since it will also catch a block. I checked anything invoking `block()` or `blockExpr()` if doesn't have the same issue.

kitbellew: `try {...}.select` is only allowed when `allowTryWithAnyExpr` is set (see line 2295). so it looks like you don't need to change this line.

kitbellew (changes requested):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Och, it seems we are using Scala211 dialect as default in a lot places :thinking: 

We could try setting the dialect based on the current version of Scala, which would test much more things. I will follow up with another PR.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3080,18 +3083,18 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     trees
   }
 
-  def blockExpr(): Term = autoPos {
+  def blockExpr(isBlockOptional: Boolean = false): Term = autoPos {
```
dos65, line 3086: I found that for parsing cases we use `expr(location = BlockStat, ...)`: https://github.com/scalameta/scalameta/blob/c058f6dfd2d220a8cef2e44669f3adb393ff6d6d/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L3124

I'm wondering if the same approach would work for `function` and `implicitClosure`

dos65: I tried that and seems like `expr` usage instead of `blockExpr` fixes that problem - [commit](https://github.com/dos65/scalameta/commit/37f1fd38a7931d944d01b36dcf2098f7dabff4b3).
I'm not fully sure if it fixes any additional cases but at least it looks more generic.
@tgodzik wdyt?

tgodzik: That's a good idea! Thanks! And also changed!

tgodzik: Actually, it doesn't work after checking again. Reverted to the last version.

dos65 (approved):
Looks good!
Sorry for the wrong suggestion

tgodzik:
> Looks good!
> Sorry for the wrong suggestion

No worries, it seemed to make sense and would be a nice simplification.

PR #2378, opened 2021-06-08
Remove not needed libraries

tgodzik:
I started looking into this in order to start running Scalafmt tests on Native, but it seems I can instead just copy the small class that is needed for the tests. However, I think it's an improvement to remove two libraries here.

### DIFF
--- a/tests/jvm/src/test/scala-2.12/scala/meta/tests/semanticdb/ExpectSuite.scala
+++ b/tests/jvm/src/test/scala-2.12/scala/meta/tests/semanticdb/ExpectSuite.scala
@@ -17,6 +17,7 @@ import scala.meta.io.Classpath
 import scala.meta.io.AbsolutePath
 import scala.meta.tests.cli._
 import munit.FunSuite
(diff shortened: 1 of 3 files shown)


tgodzik:
`diffutils` are needed, I will add it back then.

olafurpg (approved):
Nice!

tests/jvm/src/test/scala-2.12/scala/meta/tests/semanticdb/ExpectSuite.scala
```diff
@@ -17,6 +17,7 @@ import scala.meta.io.Classpath
 import scala.meta.io.AbsolutePath
 import scala.meta.tests.cli._
 import munit.FunSuite
+import munit.internal.difflib.DiffUtils
```
tgodzik, line 20: `Diffutils` from munit work and I don't think we plan to change it at any point in the future. What do you think @olafurpg ?

kitbellew (approved):
i am not @olafurpg but...

PR #2381, opened 2021-06-15
[Scala3] While condition - accept multiply exprs in parens

### LINKED ISSUE #2380: scalameta parse error for valid Scala 3 code
**Describe the bug**

**To Reproduce**

```scala
object Test:

  var x = 1
  var y = 2
  while (x > 0) && (y > 0) do
    x -= 1
    y -= 1
```
This program gives a parse error on the `do`: 
```
    `;` expected but `do` found. (scalameta)
```

**Expected behavior**

It should parse. Interestingly, it does parse if `while` is replaced by `if` and `do` is replaced by `then`.

dos65:
Fixes #2380

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2253,12 +2253,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       accept[KwThen]
       (cond, exprMaybeIndented())
     } else {
-      val forked = in.fork
-      var cond = condExpr()
-      if ((token.is[Ident] && isLeadingInfixOperator(token)) || token.is[Dot]) {
-        in = forked
-        cond = expr()
-      }
+      var cond = condExprInParens[KwThen]
       newLinesOpt()
       if (!tryAcceptWithOptLF[KwThen])
         in.observeIndented()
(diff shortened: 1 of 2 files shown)


tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/ControlSyntaxSuite.scala
```diff
@@ -1103,6 +1103,26 @@ class ControlSyntaxSuite extends BaseDottySuite {
     )
   }
 
+  test("while-expr-do") {
+    val code =
+      """|  while (x > 0) && (y > 0) do
```
kitbellew, line 1108: should this also work:
```
  while (x > 0) && (y > 0) x += 1
```
also, is `do` allowed only in scala3?

dos65: It shouldn't. It seems that the chained infix application works only if there is `do` keyword in the dotty parser.

Thanks for pointing at that. I will slightly rework that pr.

dos65: Added checks on having `do`/`then` kw after complex expressions as it's works in scala3.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2274,6 +2269,21 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     }
   }
 
+  def condExprInParens[T <: Token: TokenInfo]: Term = {
+    if (dialect.allowSignificantIndentation) {
```
kitbellew, line 2273: i think this branch is not tested. you don't use scala 2 in any test cases?

kitbellew (approved):
implementation is good but one comment on test coverage

tgodzik:
It seems to fail now on compiler/src/dotty/tools/dotc/core/Decorators.scala:171 in the dotty repo, you might need to add that test case too.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: For scala2 this part branch isn't involved because the condition has to be inside braces.
There are tests for scala2 `if`/`while` in [TermSuite](https://github.com/scalameta/scalameta/blob/main/tests/shared/src/test/scala/scala/meta/tests/parsers/TermSuite.scala).

tgodzik (approved):
LGTM!

PR #2382, opened 2021-06-16
[Scala3] Allow context bounds in trait type parameters

dos65:
Originally was reported here - https://github.com/scalameta/metals/issues/2874


tgodzik (approved):
LGTM! Let's wait until the build is green, but this looks good! Thanks!

Issue #2379, opened 2021-06-08
[Scala 3] Scalameta fails to parse valid ‘if … then’ statement that has a newline immediately following the if


djneades:
These two (admittedly rather strangely formatted) code examples compile with Scala 3.0.0:

```scala
val x = Some(10)

def f =
   if
         x.exists(x => x == 10) then
      println("Yes")
   else
      println("No")
```

```scala
val x = Some(10)

def f =
   if
      x.exists
         (x => x == 10) then
      println("Yes")
   else
      println("No")
```

However, for both of them, Scalameta chokes on the `then`. Running Scalafmt 3.0.0-RC5 on the first example above gives:

```
sbt:test> scalafmt
[info] Formatting 1 Scala sources...
[error] C:\Git\BugReport\src\main\scala\Main.scala:5: error: ; expected but then found
[error]          x.exists(x => x == 10) then
[error]                                 ^: C:\Git\BugReport\src\main\scala\Main.scala
[error] (Compile / scalafmt) C:\Git\BugReport\src\main\scala\Main.scala:5: error: ; expected but then found
[error]          x.exists(x => x == 10) then
[error]                                 ^: C:\Git\BugReport\src\main\scala\Main.scala
[error] Total time: 0 s, completed 09-Jun-2021 00:42:34
```

Interestingly, Scalameta is fine with this valid Scala 3 code:

```scala
val x = Some(10)

def f =
   if
      x.exists
         (x => x == 10)
   then
      println("Yes")
   else
      println("No")
```

I noticed this issue while trying to characterize [this Scala 3 issue](https://github.com/lampepfl/dotty/issues/12757).

tgodzik:
Thanks for reporting! This is due to the fact that we treat the statement after `if` as an indented block, which cannot contain a standalone `then`. Not sure how to fix that currently, since I thought that is the expected behaviour :thinking:

djneades:
@tgodzik Ah :-)

This issue is not a problem for me, since I don’t actually have any code that looks like this. Perhaps the Scala 3 folks will take a look at my [issue](https://github.com/lampepfl/dotty/issues/12757) and tweak the compiler to be more in line with your expectations …

kitbellew:
> Thanks for reporting! This is due to the fact that we treat the statement after `if` as an indented block, which cannot contain a standalone `then`. Not sure how to fix that currently, since I thought that is the expected behaviour 🤔

[...]

PR #2384, opened 2021-06-16
[Scala 3]Keywords can also end indentation region if on the same line

tgodzik:
This follows the additional rules in https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html

```
The following two additional rules support parsing of legacy code with ad-hoc layout. They might be withdrawn in future language versions:

An <outdent> is also inserted if the next token following a statement sequence starting with an <indent> closes an indentation region, i.e. is one of then, else, do, catch, finally, yield, }, ), ] or case.
```

Fixes https://github.com/scalameta/scalameta/issues/2379


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -358,7 +358,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           token.is[KwDo] || token.is[KwCatch] || token.is[KwFinally] || token.is[KwYield] ||
           token.is[KwMatch]
 
-        if (existingIndent == expected && canEndIndentation(curr.token)) {
+        // we also insert outdent if correct keyword is on the same line
+        if ((existingIndent == expected || expected == -1) && canEndIndentation(curr.token)) {
```
dos65, line 362: With this additional rule, it seems that calling `countIndentAndNewline` isn't necessary at all.
It would be enough just to check that we are inside the indented region and keyword can close it.
Smth like:
```scala

sepRegions match {
  case region :: tail if region.isIndented && canEndIndetation(curr.token) =>
     // procude outdent
     true
   case _ => false
}
```

tgodzik: Ach, you are right. This actually works in more situations that it didn't previously work like:
```
    val a = if false then
      1
      2 
       else
      3
```

dos65 (approved):
:rocket:

Issue #938, opened 2017-06-12
Publish testkit for scala.js


LMnet:
At the moment there is no testkit for scala.js. As a temporary workaround I just copypasted classes from testkit project to my scala.js project, but it would be convenient to have testkit cross-published.

olafurpg:
Thank you for reporting! This should be doable. Testkit uses some JVM specific code to download and unzip. However, it should be possible to move that to the `jvm/` directory. Would you be interested in contributing this fix @LMnet ?

LMnet:
I can try, but I'm playing with scalameta only with my pet project in free time, so I can't promise anything.
About implementation: as far as I understand, `AnyDiff`, `Observation`, `Phase`, `StructurallyEqual` and `SyntaxAnalysis` could be cross-published. `Corpus`, `CorpusFile`, `FileOps` should be moved to the jvm-specific folder. Am I right?

olafurpg:
That sounds about right @LMnet .

xeno-by:
@olafurpg @LMnet Guys, is this ticket still relevant?

LMnet:
@xeno-by I'm not sure I will have time for this issue in the near future.

About relevance - in my project, I still have a copy-pasted testkit. I'm not sure how many scala.js projects uses scalameta, but overall it looks like this task is pretty useful.

olafurpg:
This is a several minute effort to separate out the JVM-only parts in to the `jvm/` directory so I propose we keep this issue and I will complete it once I have some time to kill

PR #2385, opened 2021-06-16
Cross publish testkit for JS and Native

tgodzik:
This will be need to properly test Scalafmt on Native.

Fixes https://github.com/scalameta/scalameta/issues/938


dos65 (approved):
Looks good!

tgodzik:
I wonder if we should try to run:
``` scala   
"tar -xf $tarball -C $destination"
```

for Native and JS which would make it possible to use Corpus from those platforms :thinking:

olafurpg (approved):

PR #2386, opened 2021-06-17
Don't publish testkit JS and Native for 2.11

tgodzik:
This fixes the release, since scalameta is not released for Native or JS on 2.11.


tgodzik:
I will go ahead and merge this as it should fix the release, which I broke :O

PR #2392, opened 2021-06-23
[Scala 3] Allow inline matches on literals

tgodzik:
Found while working on https://github.com/lampepfl/dotty/pull/12902


dos65 (approved):
Looks good!

Do you have an idea why `EqualProps` fails on CI?
I can't reproduce it locally

tgodzik:
> Looks good!
> 
> Do you have an idea why `EqualProps` fails on CI?
> I can't reproduce it locally

No idea, tried to reproduce it, but somehow it just fails on CI :/

PR #2393, opened 2021-06-24
[Scala 3] Add the rest of possible inline matched expressions

tgodzik:
In the previous PR, I haven't added all possible cases, which inline can be followed by with a match. I went through all ExprIntro tokens and checked each one, so we should be good now. We could use `ExprIntro` classifier in theory, but it seems around half of the tokens defined there would be invalid after inline.

This issue came up when I was looking through the Dotty test cases and it seems we should try to parse as many as possible there, however it seems we do have some bugs still and I don't want to fix it all in one PR.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1097,9 +1097,13 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   @classifier
   trait InlineMatchMod {
     def unapply(token: Token): Boolean = {
-      token.is[soft.KwInline] &&
-      (token.next.is[LeftParen] || token.next.is[LeftBrace] || token.next.is[KwNew] ||
-        token.next.is[Ident] || token.next.is[Literal])
+      token.is[soft.KwInline] && {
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1099,7 +1099,9 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     def unapply(token: Token): Boolean = {
       token.is[soft.KwInline] &&
       (token.next.is[LeftParen] || token.next.is[LeftBrace] || token.next.is[KwNew] ||
-        token.next.is[Ident] || token.next.is[Literal])
+        token.next.is[Ident] || token.next.is[Literal] || token.next.is[Interpolation.Id] ||
```
kitbellew, line 1102: since `token.next` is actually a somewhat non-trivial operation, should we save in a variable or just use
```
  token.is[soft.KwInline] && (token.next match {
    case _: LeftParen | _: LeftBrace | _: KwNew | ... => true
    case _ => false
  })
```

tgodzik: Fixed!

kitbellew (approved):

PR #2396, opened 2021-06-24
[Scala3] Allow empty template with `:` + end marker for object/class/trait

### LINKED ISSUE #2387: Scala 3 empty object with `end` token error
In Scala 3 dialect (under Scalafmt) we get parse error when an object has no body, but has end token.
```scala
object Foo:
end Foo
```

dos65:
Fixes #2387

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -907,6 +907,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   def acceptStatSep(): Unit = token match {
     case LF() | LFLF() => next()
     case _ if in.observeOutdented() =>
+    case t if t.is[EndMarkerIntro] =>
     case _ => accept[Semicolon]
   }
   def acceptStatSepOpt() =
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -907,6 +907,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   def acceptStatSep(): Unit = token match {
     case LF() | LFLF() => next()
     case _ if in.observeOutdented() =>
+    case t if t.is[EndMarkerIntro] =>
```
tgodzik, line 910: Is this needed? Outdent should be always before end marker and from what I see we don't need it below.

tgodzik (changes requested):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: For empty template body, there is no indent/outdent
```scala
object Foo: // <- here at colon  we call `observeIndented` while the next line doesn't have any indentation
end Foo
```

tgodzik: Generally, it should end on the LF after `:`, but this doesn't seem to happen and we are skipping the LF. That would be the underlying issue here. However, not sure how to fix it properly, so it's ok for now.

tgodzik (approved):
LGTM

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: and would this work:
```
object Foo:
// comment
end Foo
```

dos65: @kitbellew yep. Parser ignores comments in the same way as whitespaces

PR #2400, opened 2021-06-28
[Scala3] Trimmed pos - handle case with single Trivia token

### LINKED ISSUE #2397: Omitted `else` in optional-braces is not empty
in this expression:
```
if cond then
        sb.append(doc.text.substring(offset, end))
        sb.append(doc.text.substring(offset, end))
```
the `else` subtree of `Term.If` is a `Lit.Unit` consisting of a single `Token.LF` (rather than empty, as I would expect).

dos65:
Fixes #2397

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -861,7 +861,10 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       if (rangeFromStart.isEmpty) (startTokenPos, endTokenPos)
       else (skipTrivia(rangeFromStart), skipTrivia(rangeFromStart.reverse))
 
-    TokenStreamPosition(start, end + 1)
+    if (start == end && scannerTokens(start).is[Trivia])
+      TokenStreamPosition(start, end)
+    else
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -861,7 +861,10 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       if (rangeFromStart.isEmpty) (startTokenPos, endTokenPos)
       else (skipTrivia(rangeFromStart), skipTrivia(rangeFromStart.reverse))
 
-    TokenStreamPosition(start, end + 1)
+    if (start == end && scannerTokens(start).is[Trivia])
+      TokenStreamPosition(start, end)
```
kitbellew, line 865: what if start != end? does it ever make sense to start a tree with Whitespace?

dos65: if start != end than trivia is already removed fully.

kitbellew: got it... must be because it's not the start that has it but end.

kitbellew (approved):

tgodzik (approved):
LGTM!

PR #2402, opened 2021-07-01
SemanticdbReporter should set its own Xmaxwarns

mlachkar:
This pull request goal is make semanticbreporter store all diagnostics, without depending of -xmaxWarn set by the user in his build. Moreover, this pull request fix a bug, where in previous versions between 2.12 and 2.13.0, only 100 diagnotics could be stored.

- in previous version in 2.12 up to 2.13.0, we set `new settings()` for semanticdbRepoter
  which have a Xmaxwarns to default value (100 max warnings), which implies that the reporter wont report more than 100 diagnostics
- in 2.13.1+, by overloading `def filter` we keep the same the same setting Xmaxwarns set by the user.
  This setting is supposed to modify the number of warning printed, but should not modify the number of
  warning stored in semanticdb file.


mlachkar:
Running the test against `main` in 2.12.14 shows that the number of diagnostic is only 97 (I was expecting 100, but still it's what I wanted to show)
```
==> X scala.meta.tests.semanticdb.ExpectSuite.metac.expect  0.668s munit.ComparisonFailException: /Users/meriamlachkar/scalacenter/scalameta/tests/jvm/src/test/scala-2.12/scala/meta/tests/semanticdb/ExpectSuite.scala:51
50:      }
51:      this.assertNoDiff(loadObtained, expectedCompat)
52:    }
diff assertion failed
=> Diff (- obtained, + expected)
 Language => Scala
-Diagnostics => 97 entries
+Diagnostics => 101 entries
```

tgodzik (approved):
LGTM!

PR #2412, opened 2021-07-09
[Scala 3] Allow type param after ApplyUsing

tgodzik:


dos65 (approved):
:smile: I spent some time wondering if this should compile or not. 
Smth like: `Learning Scala 3 by parser issues`

tgodzik:
>  I spent some time wondering if this should compile or not.
> Smth like: `Learning Scala 3 by parser issues`

That's exactly how I am learning Scala 3 :D

tgodzik:
Also, from the parser perspective it doesn't matter. We don't know what is returned after each apply.

PR #2413, opened 2021-07-09
[Scala 3] Better detection of conditions without parens

tgodzik:


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -989,6 +989,27 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+      var braces = 0
+      if (braces == 0 && token.is[T]) true
+      else if (braces == 0 && (token.is[StopScanToken] || endBlock)) false
+      else if (token.is[EOF]) false
+      else {
+        if (startBlock) braces += 1
+        if (endBlock) braces -= 1
+        next()
+        lookForToken()
+      }
```
dos65, line 1006: `braces` should be a parameter in `lookForToken`.

tgodzik: Good idea! Changed!

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/ControlSyntaxSuite.scala
```diff
@@ -1814,4 +1814,31 @@ class ControlSyntaxSuite extends BaseDottySuite {
       )
     )
   }
+
+  test("if-infix") {
+    runTestAssert[Stat](
+      """|if (1) max 10 gt 0
+         |
```
kitbellew, line 1821: is this really valid in scala 3?!

tgodzik: Yep, it's taken straight from the tests. I think this is the most hardcore edge case they could find.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -989,6 +989,28 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     case _ => false
   }
 
+  def followedByToken[T <: Token: TokenInfo]: Boolean = {
+    def startBlock = token.is[LeftBrace] || token.is[Indentation.Indent]
+    def endBlock = token.is[RightBrace] || token.is[Indentation.Outdent]
```
dos65, line 994: should it skip parens too?

tgodzik: It doesn't in dotty, so I would rather keep it the same as the Scala 3 implementation.

dos65 (approved):

PR #2418, opened 2021-07-12
[Scala 3] Emit outdent before commas in arg position

tgodzik:
This follows:

```
An <outdent> is finally inserted in front of a comma that follows a statement sequence starting with an <indent> if the indented region is itself enclosed in parentheses
```

from https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -411,14 +411,20 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
       def currRef: TokenRef = TokenRef(curr, currPos, currPos + 1, currPos)
 
+      def indentationWithinParenRegion: Boolean = sepRegions.headOption.exists(_ != RegionParen) &&
+        sepRegions
+          .collectFirst {
+            case RegionParen => true
+            case other if !other.isIndented => false
+          }
+          .getOrElse(false)
+
       if (isTrailingComma) {
         nextToken(currPos, currPos + 1, sepRegions)
       } else if (curr.isNot[Trivia]) {
         if (curr.is[LeftParen]) (RegionParen :: sepRegions, currRef)
         else if (curr.is[LeftBracket]) (RegionBracket :: sepRegions, currRef)
-        else if (curr.is[Comma] &&
-          sepRegions.headOption.exists(_.isInstanceOf[RegionIndent]) &&
-          sepRegions.tail.headOption.contains(RegionParen)) {
+        else if (curr.is[Comma] && indentationWithinParenRegion) {
           (sepRegions.tail, mkOutdent(prevPos))
         } else if (curr.is[LeftBrace]) {
           val indentInBrace = if (isAheadNewLine(currPos)) countIndent(nextPos) else -1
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -356,16 +356,29 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
-            sepRegions = sepRegions.tail
-            curr = TokenRef(
-              new Indentation.Outdent(token.input, token.dialect, token.start, token.end),
-              curr.pos,
-              curr.pos,
-              prevPos
-            )
-            true
+            outdent()
+          case _ if existParenRegion && token.is[Comma] =>
```
dos65, line 380: :thinking: There is a similar handler in `LazyTokenIterator` https://github.com/scalameta/scalameta/blob/ea3c552a3a056a22586a7129d0d9c246c5fa672e/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L419-L422

It looks like a better place for fix. The only problem I don't get why it doesn't work.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -356,16 +356,29 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+          )
+          true
+        }
+
+        def existParenRegion: Boolean = sepRegions
+          .collectFirst {
+            case RegionParen => true
+            case other if !other.isIndented => false
+          }
+          .getOrElse(false)
```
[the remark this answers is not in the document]
tgodzik: This doesn;t work, there might be multiple indented regions as is the case here.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: `headOption` is the reason, there might be mutliple indented regions - I will fix it there

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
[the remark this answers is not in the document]
tgodzik: Moved it to LazyTokenIterator

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -411,14 +411,20 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
       def currRef: TokenRef = TokenRef(curr, currPos, currPos + 1, currPos)
 
+      def indentationWithinParenRegion: Boolean = sepRegions.headOption.exists(_ != RegionParen) &&
+        sepRegions
+          .collectFirst {
+            case RegionParen => true
```
tgodzik, line 417: So this checks if all regions up to parens are indented.

dos65 (approved):

PR #2425, opened 2021-07-19
[Scala 3] Don't print braces for single given import

tgodzik:


dos65 (approved):
LGTM

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -1244,6 +1244,7 @@ object TreeSyntax {
       case (t: Importee.Name) :: Nil => s(t)
       case (t: Importee.Wildcard) :: Nil => s(t)
       case (t: Importee.GivenAll) :: Nil => s(t)
+      case (t: Importee.Given) :: Nil => s(t)
       case (t: Importee.Rename) :: Nil if dialect.allowAsForImportRename => s(t)
```
kitbellew, line 1248: why not just have a single case with one importee and check it for unimport or rename?

tgodzik: This would most likely make it shorter, but it seem quite readable this way.

PR #2426, opened 2021-07-19
[Scala3] Significant indentation - handle comments properly

### LINKED ISSUE #2424: [Scala 3] Spurious `expected template body` when line begins with comment
Using Scala 3 optional braces style, I commented out the `inline` keyword to debug something and got an `expected template body` error (Metals 0.10.5 in VS Code):

```scala
object Foo:
  /*inline*/ def bar: String = ???
```
expected_template_body

dos65:
Indentation calculation rules:
- for single-line comment, spaces between newline and the start of comment
  ```scala
  object Foo:
     /* comment */ def foo: Int = ??? // <- indent = 3
- for multiline, spaces between newline and firts non-whitespace character
  ```scala
  object Foo: /* multi
   line
     comment */ def foo: Int = ??? // <- indent = 3
  ```
  
Fixes #2424

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -186,12 +186,6 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
     def nextSafe: Token = scannerTokens.apply(Math.min(token.index + 1, scannerTokens.length - 1))
 
-    def nextEnd: Token = {
-      val next = nextSafe
-      if ((next.is[Whitespace] && !next.is[LineEnd]) || next.is[Comment]) next.nextEnd
-      else next
-    }
-
     def next: Token = {
       val next = nextSafe
       if (next.is[Whitespace] || next.is[Comment]) next.next
@@ -204,8 +198,18 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     }
   }
 
+  def isMultilinePos(p: Position): Boolean = p.endLine > p.startLine
+
   def isColonEol(token: Token) = {
-    dialect.allowSignificantIndentation && token.is[Colon] && token.nextEnd.is[LineEnd]
+
+    def isNextEOL(t: Token): Boolean = {
+      val next = t.nextSafe
+      if (next.is[LineEnd] || next.is[MultilineComment]) true
+      else if ((next.is[Whitespace] && !next.is[LineEnd]) || next.is[Comment]) isNextEOL(next)
+      else false
+    }
+
+    dialect.allowSignificantIndentation && token.is[Colon] && isNextEOL(token)
   }
 
   /* ------------- PARSER-SPECIFIC TOKENS -------------------------------------------- */
@@ -532,7 +536,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
         var newlineStreak = false
         var newlines = false
         while (i < nextPos) {
-          if (scannerTokens(i).is[LF] || scannerTokens(i).is[FF]) {
+          val token = scannerTokens(i)
+          if (token.is[LF] || token.is[FF] || token.is[MultilineComment]) {
             lastNewlinePos = i
             if (newlineStreak) newlines = true
             newlineStreak = true
@@ -666,6 +671,21 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
   }
 
+  private def multineCommentIndent(t: Comment): Int = {
+    @tailrec
+    def loop(idx: Int, indent: Int, isAfterNewline: Boolean): Int = {
+      if (idx == t.value.length) indent
+      else {
+        t.value.charAt(idx) match {
+          case '\n' => loop(idx + 1, 0, isAfterNewline = true)
+          case ' ' | '\t' if isAfterNewline => loop(idx + 1, indent + 1, isAfterNewline)
+          case _ => loop(idx + 1, indent, isAfterNewline = false)
+        }
+      }
+    }
+    loop(0, 0, false)
+  }
+
   /**
    * When token on `tokenPosition` is not a whitespace and is
    * a first non-whitespace character in a current line then a result is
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -681,9 +699,17 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
     def countIndentInternal(pos: Int, acc: Int = 0): (Int, Int) = {
-      if (pos < 0 || scannerTokens(pos).is[LF]) (acc, pos)
-      else if (isWhitespace(scannerTokens(pos))) countIndentInternal(pos - 1, acc + 1)
-      else (-1, -1)
+      val token = scannerTokens(pos)
+      if (pos < 0) (acc, pos)
+      else
+        token match {
+          case _: LF => (acc, pos)
+          case c: Comment if token.pos.endLine > token.pos.startLine =>
```
tgodzik, line 707: I think `MultilineComment` is unused, so maybe:
```suggestion
          case MultilineComment() =>
```

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/SignificantIndentationSuite.scala
```diff
@@ -2038,4 +2038,148 @@ class SignificantIndentationSuite extends BaseDottySuite {
...
+            Defn.Def(Nil, Term.Name("foo"), Nil, Nil, Some(Type.Name("Int")), Term.Name("???")),
+            Defn.Def(Nil, Term.Name("bar"), Nil, Nil, Some(Type.Name("Int")), Term.Name("???"))
+          ),
+          Nil
+        )
+      )
+    )
+  }
+
+  Source(
```
tgodzik, line 2113: I think this is not needed?

tgodzik (changes requested):
Looks good! Just some smaller comments.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: I would use it but here it's needed to upcast `Token` to `Comment` that's impossible with classifier macro.
Changed it a little - extracted this check into function

tgodzik (approved):
LGTM!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -681,9 +701,17 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
     @tailrec
     def countIndentInternal(pos: Int, acc: Int = 0): (Int, Int) = {
-      if (pos < 0 || scannerTokens(pos).is[LF]) (acc, pos)
-      else if (isWhitespace(scannerTokens(pos))) countIndentInternal(pos - 1, acc + 1)
-      else (-1, -1)
+      val token = scannerTokens(pos)
```
kitbellew, line 704: should you access token before checking `pos < 0`?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -666,6 +671,21 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
   }
 
+  private def multineCommentIndent(t: Comment): Int = {
+    @tailrec
+    def loop(idx: Int, indent: Int, isAfterNewline: Boolean): Int = {
+      if (idx == t.value.length) indent
+      else {
+        t.value.charAt(idx) match {
+          case '\n' => loop(idx + 1, 0, isAfterNewline = true)
+          case ' ' | '\t' if isAfterNewline => loop(idx + 1, indent + 1, isAfterNewline)
```
kitbellew, line 681: a tab is a single indent?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: @dos65 a bug?

dos65: yep. thx!

kitbellew:
@dos65 can't upgrade `scalafmt` with new version of `scalameta`. see examples below.

kitbellew:
@dos65 here's one example:
```
object a:
  val a =
    if a then
      if aa then
        aaa
        // c1
    else
      if bb then
        bbb
        // c2
```
the `aa` tree is:
```
TYPE: Term$If$TermIfImpl
SOURCE: if aa then
        aaa
STRUCTURE: Term.If(Term.Name("aa"), Term.Name("aaa"), Lit.Unit(()), Nil)
TOKENS: if,∙,aa,∙,then,¶,∙,∙,∙,∙,∙,∙,∙,∙,aaa
```
while a very similar `bb` tree is:
```
TYPE: Term$If$TermIfImpl
SOURCE: if bb then
        bbb
        // c2
STRUCTURE: Term.If(Term.Name("bb"), Term.Name("bbb"), Lit.Unit(()), Nil)
TOKENS: if,∙,bb,∙,then,¶,∙,∙,∙,∙,∙,∙,∙,∙,bbb,¶,∙,∙,∙,∙,∙,∙,∙,∙,//∙c2,
```
specifically, the `then` part is
```
TYPE: Term$Name$TermNameImpl
SOURCE: bbb
        // c2
STRUCTURE: Term.Name("bbb")
TOKENS: bbb,¶,∙,∙,∙,∙,∙,∙,∙,∙,//∙c2,
```
(also, there's something after `// c2` since there's a trailing comma).

kitbellew:
@dos65 here's another example: this code is not parsable anymore
```
object a {
  @foo
  /**
   * scaladoc
   */
  class QueueSequentialBenchmark {}
}
```
nor is this:
```
object a {
  @foo(bar)
  /**
   * scaladoc
   */
  class QueueSequentialBenchmark {}
}
```

dos65:
@kitbellew the second one with comment and annotation should be fixed in #2437


what about the first one and positions. I can't reproduce it. 
Tried the following:
```scala
object a:
  val a =
    if a then
      if aa then
        aaa
        // c1
    else
      if bb then
        bbb
        // c2
        fx
```

Positions test output:
```
Template :
  val a =
    if a then
      if aa then
        aaa
        // c1
    else
      if bb then
        bbb
        // c2
        fx
Self   @@val a =
Defn.Val val a =
    if a then
      if aa then
        aaa
        // c1
    else
      if bb then
        bbb
        // c2
        fx
Term.If if a then
      if aa then
        aaa
        // c1
    else
      if bb then
        bbb
        // c2
        fx
Term.If if aa then
        aaa
Lit.Unit         // c1@@  // <- if body includes comment
Term.If if bb then
        bbb
        // c2
        fx
Term.Block bbb
        // c2
        fx
Lit.Unit         fx@@
```

Could you add more details/check it one more time?

dos65:
@kitbellew upd - I found this broken test case in scalafmt. Will look at it

PR #2428, opened 2021-07-20
[Scala 3] Recognize implicit closures only if next token is not a SoftModifier

tgodzik:
Previously, we would only check for Ident, which is not enough for Scala 3.

Causes https://github.com/scalameta/scalafmt/issues/2640


dos65 (approved):
:+1:

PR #2439, opened 2021-08-02
Support new Scala3 modifiers

tanishiking:
This PR adds the following new modifiers for Scala3 to SemanticDB specification.

- Given
  - https://docs.scala-lang.org/scala3/reference/contextual/givens.html
  - for `given x: Int`, `given Int`, `using ctx: Context` and `using Context`
- Inline
  - https://docs.scala-lang.org/scala3/reference/metaprogramming/inline.html
- Open
  - https://docs.scala-lang.org/scala3/reference/other-new-features/open-classes.html
- Transparent
  - https://docs.scala-lang.org/scala3/reference/other-new-features/transparent-traits.html
- Infix
  - https://docs.scala-lang.org/scala3/reference/changed-features/operators.html
- Opaque
  - https://docs.scala-lang.org/scala3/reference/other-new-features/opaques.html

FYI here's the generated code from this new SemanticDB schema https://github.com/tanishiking/semanticdb-for-scala3/pull/2

---

Maybe we should also support `extension` modifier in the future, but how to encode `extension` seems a bit complicated. Therefore let's start with simple modifiers first.

- https://github.com/lampepfl/dotty/issues/11690
- https://discord.com/channels/632642981228314653/632642981228314657/867097982860394526 

---

After this PR merged, we have to

- [ ] Support new modifiers in `metap`
- [ ] Extend `ExtracctSemanticDB` of Scala3

Or should we implement the Scala3 side first, before updating the schema?

### DIFF
--- a/docs/semanticdb/specification.md
+++ b/docs/semanticdb/specification.md
@@ -794,6 +794,37 @@ languages map onto these properties.
     <td><code>DEFAULT</code></td>
     <td>Is a default parameter or a default method?</td>
   </tr>
+  <tr>
+    <td><code>0x10000</code></td>
+    <td><code>GIVEN</code></td>
+    <td>Is a given instance?</td>
+    <td>Is a <code>given instance</code> (e.g. <code>given x: Int = ...</code> or <code>using x: Int</code>)?</td>
+  </tr>
+  <tr>
+    <td><code>0x20000</code></td>
+    <td><code>INLINE</code></td>
+    <td>Has an <code>inline</code> modifier?</td>
+  </tr>
+  <tr>
+    <td><code>0x40000</code></td>
+    <td><code>OPEN</code></td>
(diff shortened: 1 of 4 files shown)


tgodzik (approved):
> Support new modifiers in metap

Could we add it in this PR or doesn't it make sense since we wouldn't have any way to tests it here?

tanishiking:
Though we can't test it until Scala3 extract that information, they're so simple and I will include `metap` implementation to this PR :)

tanishiking:
I opened a PR to dotty to extract those new information from Scala3 https://github.com/lampepfl/dotty/pull/13239
The metap implementation is basically the same between this PR and Scala3, and it looks good (seeing the metap.expect in the https://github.com/lampepfl/dotty/pull/13239 )

tgodzik (approved):
LGTM!

PR #2441, opened 2021-08-04
Sbt - migrate to slash syntax

dos65:


build.sbt
```diff
@@ -712,11 +706,10 @@ lazy val publishableSettings = Def.settings(
 )
 
 lazy val nonPublishableSettings = Seq(
-  skip in publish := true,
+  publish / skip := true,
   mimaPreviousArtifacts := Set.empty,
-  publishArtifact in (Compile, packageDoc) := false,
-  publishArtifact in packageDoc := false,
-  sources in (Compile, doc) := Seq.empty,
+  Compile / packageDoc / publishArtifact := false,
```
tgodzik, line 711: Is `packageDoc / publishArtifact := false` not needed?

dos65: Yep. Even before migration there was a warning from sbt-linter that it isn't used

tgodzik (approved):
LGTM! Thanks for migrating the syntax, I never managed to get up to do it.

PR #2448, opened 2021-08-10
Fix UnreachableError when printing a pattern match bind to a case object

### LINKED ISSUE #2447: UnreachableError when parsing a pattern bind to a case object
The following code throws an `UnreachableError`

```scala
import scala.meta._

q"{ case x @ Y => x }".syntax
```

[Scastie here](https://scastie.scala-lang.org/xKpv6hjQTkSH7AvF2SfjkA)

Lower-casing the pattern bind doesn't cause an exception, and also parses to a different tree:

`{ case x @ Y => x }` becomes:

```
Term.PartialFunction(List(Case(Pat.Bind(Pat.Var(Term.Name("x")), Term.Name("Y")), None, Term.Name("x"))))
```

While `{ case x @ y => x }` becomes:

```
Term.PartialFunction(List(Case(Pat.Bind(Pat.Var(Term.Name("x")), Pat.Var(Term.Name("y"))), None, Term.Name("x"))))
```

Full stacktrace:

```
org.scalameta.UnreachableError: this code path should've been unreachable
	at org.scalameta.UnreachableError$.raise(package.scala:40)
	at scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances.isAmbiguousWithPatVarTerm$1(TreeSyntax.scala:252)
	at scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances.isAmbiguousInParent$1(TreeSyntax.scala:288)
	at scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances.$anonfun$guessIsBackquoted$7(TreeSyntax.scala:295)
	at scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances.$anonfun$guessIsBackquoted$7$adapted(TreeSyntax.scala:295)
	at scala.Option.exists(Option.scala:376)
	at scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances.guessIsBackquoted(TreeSyntax.scala:295)
	at scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances.$anonfun$syntaxTree$1(TreeSyntax.scala:351)
	at scala.meta.prettyprinters.Syntax$$anon$1.apply(Syntax.scala:9)
	at scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances.p(TreeSyntax.scala:126)
	at scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances.$anonfun$syntaxTree$1(TreeSyntax.scala:787)
	at scala.meta.prettyprinters.Syntax$$anon$1.apply(Syntax.scala:9)
	at scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances.p(TreeSyntax.scala:126)
	at scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances.$anonfun$syntaxTree$1(TreeSyntax.scala:1171)
	at scala.meta.prettyprinters.Syntax$$anon$1.apply(Syntax.scala:9)
	at scala.meta.prettyprinters.Show$.indent(Show.scala:72)
	at scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances.$anonfun$syntaxTree$34(TreeSyntax.scala:621)
	at scala.collection.immutable.List.map(List.scala:293)
	at scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances.$anonfun$syntaxTree$1(TreeSyntax.scala:621)
	at scala.meta.prettyprinters.Syntax$$anon$1.apply(Syntax.scala:9)
	at scala.meta.internal.prettyprinters.TreeSyntax$.$anonfun$apply$1(TreeSyntax.scala:1284)
	at scala.meta.prettyprinters.Syntax$$anon$1.apply(Syntax.scala:9)
	at scala.meta.prettyprinters.Api$XtensionSyntax.syntax(Api.scala:11)
```

hugo-vrijswijk:
Fixes #2447

Please take a good look at if this is not breaking anything 🙏.

`{ case x => Y => () }` parses differently than `{ case x @ y => () }`. I'm assuming this is on purpose and for a very good reason.

This PR changes the `thisLocationAlsoAcceptsPatVars` to return true for a `Pat.Bind`. I'm _assuming_ this is correct, as in a pattern match a `Pat.Var` can also take the place of a `Pat.Bind` (like in `{ case x => () }`).

EDIT: I just noticed that the [documentation](https://scalameta.org/docs/trees/quasiquotes.html#patterns-metapat-and-cases-metacase) mentions a `Pat.Bind` always consisting of two `Pat`'s. Is this correct, and does it mean that `x @ Y` should be parsed differently? (now it is `Pat.Bind(Pat.Var(Term.Name("x")), Term.Name("Y")))`)

### DIFF
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
@@ -239,7 +239,7 @@ object TreeSyntax {
           case p: Pat.Wildcard => unreachable
           case p: Pat.Var => false
           case p: Pat.Repeated => false
-          case p: Pat.Bind => unreachable
+          case p: Pat.Bind => true
           case p: Pat.Alternative => true
           case p: Pat.Tuple => true
           case p: Pat.Extract => p.args.exists(_ eq t)
--- a/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
@@ -1087,6 +1087,14 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
     checkTree(Lit.String("ラーメン"), "\"ラーメン\"")
   }
 
+  test("#2447 Pat.Bind on Term.Name") {
+    checkTree(q"{ case x @ Y => x }", "{\n  case x @ Y => x\n}")
+  }
+
+  test("#2447 Pat.Bind on Term.Name backticks") {
+    checkTree(q"{ case x @ `y` => x }", "{\n  case x @ `y` => x\n}")
+  }
+
   def checkTree(original: Tree, expected: String)(implicit loc: munit.Location): Unit = {
     assertNoDiff(original.syntax, expected)
     assertNoDiff(original.structure, expected.parse[Stat].get.structure)


tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
```diff
@@ -1087,6 +1087,10 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
     checkTree(Lit.String("ラーメン"), "\"ラーメン\"")
   }
 
+  test("#2447 Pat.Bind on Term.Name") {
+    checkTree(q"{ case x @ Y => x }", "{\n  case x @ Y => x\n}")
+  }
```
dos65, line 1092: Could you also add a test with object name in backtics?
```scala
checkTree(q"{ case x @ `y` => x }", "{\n case x @ `y` => x\n}")
```

dos65 (changes requested):
Thanks for the fix!
LGTM! 

I have only one minor notice.

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
hugo-vrijswijk: Added!

dos65:
@hugo-vrijswijk And the last thing before merge. Could you squash these commits into a single one?

hugo-vrijswijk:
Yeah, of course, I've squashed all commits into one

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -239,7 +239,7 @@ object TreeSyntax {
           case p: Pat.Wildcard => unreachable
```
kitbellew, line 239: wouldn't this and Typed below also need to be fixed? @dos65

dos65: I think there is no need. `Pat.Wildcard` can't be a parent of Term.Name.
What about `Pat.Typed` - I haven't come up with any case when it might be reachable here.

PR #2452, opened 2021-08-19
Always backtick `this`

stephenjudkins:


tests/jvm/src/test/scala-2.13/scala/meta/tests/testkit/prettyprinters/PrettyPrinterSuite.scala
```diff
@@ -44,4 +44,6 @@ class PrettyPrinterSuite extends FunSuite {
   checkOk("""1 join (())""")
 
   checkOk("""foo match{ case _ => _ => false}""")
+
+  checkOk("class Foo(`this`: Int)")
```
kitbellew, line 48: what about `class Foo(foo: String) { def this(foo: Int) = this(foo.toString) }`?

kitbellew (changes requested):

tests/jvm/src/test/scala-2.13/scala/meta/tests/testkit/prettyprinters/PrettyPrinterSuite.scala
stephenjudkins: Well, it appears to work fine. Added a test case. It seems to parse it as `Term.This`, and it generates the output correctly.

PR #2470, opened 2021-09-07
Remove invalid "default" JSExportTopLevel

gabro:
This causes linking errors in a Scala 3 / Scala.js 1 project

> Invalid top level export for name 'default' in class scala.meta.parsers.JSFacade$ when emitting a Script (NoModule) because it is not a valid JavaScript identifier (did you want to emit a module instead?)

The default export was also never used in practice, and it's better to access the single exported elements instead.

### DIFF
--- a/scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
+++ b/scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
@@ -108,13 +108,11 @@ object JSFacade {
         }
     }
 
-  @JSExportTopLevel("default")
   @JSExportTopLevel("parseSource")
   def parseSource(
       code: String
   ): js.Dictionary[Any] = parse[Source](code, defaultSettings)
 
-  @JSExportTopLevel("default")
   @JSExportTopLevel("parseSource")
   def parseSource(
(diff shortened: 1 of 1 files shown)


scalameta/parsers/js/src/main/scala/scala/meta/parsers/JSFacade.scala
```diff
@@ -108,13 +108,11 @@ object JSFacade {
         }
     }
 
-  @JSExportTopLevel("default")
```
kitbellew, line 111: should it be done via a JSExportTopLevel on the containing object, along with a JSExport the methods?

https://www.scala-js.org/doc/interoperability/export-to-javascript.html

gabro: I think we want "top level" exports, i.e. it should result in:

```js
module.exports = {
  parseSource,
  parseStat
}
```

gabro: specifically, we fall into the "Exporting top-level methods" case from the docs

kitbellew (approved):

PR #2468, opened 2021-09-06
Fix .scalafmt.conf and apply new version changes

dos65:

PR #2471, opened 2021-09-07
ScaladocParser - minor fix for linkParser

dos65:
Link might have some symbols at the end and we need to consume in as punc.
Otherwise it parses as separate `Word` and scalafmt writes is with an
additional space.

Example:
```scala
// before
"[Link]'s" -> Seq(Link("Link"), Word("'s"))
//after
"[Link]'s" -> Seq("Link", Seq.empty, "'s")
```

Should fix [this sample](https://github.com/scalameta/scalameta/pull/2468#discussion_r703208311)

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
@@ -107,7 +107,7 @@ object ScaladocParser {
   private def linkParser[_: P]: P[Link] = P {
     def end = space | linkSuffix
     def anchor = P((!end ~ AnyChar).rep(1).!.rep(1, sep = spaces1))
-    def pattern = linkPrefix ~ (anchor ~ linkSuffix ~ punctParser.!)
+    def pattern = linkPrefix ~ (anchor ~ linkSuffix ~ labelParser.?.!)
     pattern.map { case (x, y) => new Link(x, y) }
   }
 
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
@@ -133,7 +133,7 @@ class ScaladocParserSuite extends FunSuite {
     val words = descriptionBody.split("\\s+").toSeq.map(Word.apply)
     val refNone = Seq(Link("Description", Seq("Body"), ""))
     val refDots = Seq(Link("Description", Seq("Body"), "..."))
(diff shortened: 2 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -107,7 +107,7 @@ object ScaladocParser {
   private def linkParser[_: P]: P[Link] = P {
     def end = space | linkSuffix
     def anchor = P((!end ~ AnyChar).rep(1).!.rep(1, sep = spaces1))
-    def pattern = linkPrefix ~ (anchor ~ linkSuffix ~ punctParser.!)
+    def pattern = linkPrefix ~ (anchor ~ linkSuffix ~ wordParser.?.!)
```
kitbellew, line 110: do you need `.!` after the `wordParser` (since wordparser is already capturing)? also, `wordParser` returns a `Word`; if you need a string, should it be `labelParser` instead? or if it can be empty (labelParser is not), should it be `(!space ~ AnyChar).rep`?

ended up with lots of questions. last one was probably most relevant.

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
```diff
@@ -144,10 +144,10 @@ class ScaladocParserSuite extends FunSuite {
           * $descriptionBody [[ $descriptionBody ]]
           * $descriptionBody
           *
-          * [[ $descriptionBody ]]!$descriptionBody
-          *
           * $descriptionBody
           * [[ $descriptionBody ]] $descriptionBody
+          * 
+          * [[ $descriptionBody ]]'s
```
kitbellew, line 150: why reorder and not put this on original line 147? would have fewer changes

kitbellew (changes requested):

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
dos65: Yep, but the original input from 147 was a little bit confusing: 
  `[[ Description  Body ]]!Description Body`  -> `Link("Description", Seq("Body"), "!Description")`

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
dos65: thx! I missed that there is a `labelParser`

kitbellew (approved):

PR #2484, opened 2021-09-27
Remove option that breaks Metals&SBT on JDK 17

keynmol:
- It breaks metals quietly on startup
- Breaks SBT loudly


tgodzik (approved):
LGTM!

PR #2489, opened 2021-10-01
Parser - fix Scaladoc list item body parsing

dos65:
Add support for markdown blocks and tables.

Addresses [this issue](https://github.com/scalameta/scalameta/pull/2468#discussion_r703209012)

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
@@ -73,15 +73,20 @@ object ScaladocParser {
    * https://spec.commonmark.org/0.29/#fenced-code-blocks
    */
 
-  private def mdCodeBlockIndent[_: P] = hspace.rep(max = 3)
+  private def mdCodeBlockIndent[_: P](indent: Int) = hspace.rep(min = indent, max = indent + 3)
   private def mdCodeBlockFence[_: P] = "`".rep(3) | "~".rep(3)
 
-  private def mdCodeBlockParser[_: P]: P[MdCodeBlock] = P {
-    (startOrNl ~ mdCodeBlockIndent.! ~ mdCodeBlockFence.!).flatMap { case (indent, fence) =>
+  private def mdCodeBlockParser[_: P]: P[MdCodeBlock] = mdCodeBlockParser(0)
+
+  private def mdCodeBlockParser[_: P](indent: Int): P[MdCodeBlock] = P {
+    def indentParser =
+      mdCodeBlockIndent(indent)
+
+    (startOrNl ~ indentParser.! ~ mdCodeBlockFence.!).flatMap { case (indent, fence) =>
       def lineEnd = hspaces0 ~ nl
       def info = hspaces0 ~ labelParser.!.rep(0, sep = hspaces0) ~ lineEnd
       info.flatMap { infoSeq =>
-        def codeEnd = mdCodeBlockIndent ~ fence.substring(0, 1).rep(fence.length)
+        def codeEnd = indentParser ~ fence.substring(0, 1).rep(fence.length)
         def line = hspace.rep(max = indent.length) ~ (!lineEnd ~ AnyChar).rep.!
         def lines = !codeEnd ~ line.rep(1, sep = lineEnd ~ !codeEnd)
         def block = (lines ~ lineEnd).? ~ codeEnd ~ hspaces0 ~ &(nl | End)
@@ -111,14 +116,18 @@ object ScaladocParser {
     pattern.map { case (x, y) => new Link(x, y) }
   }
 
-  private def nextPartParser[_: P]: P[Unit] = P {
-    def mdCodeBlockPrefix = mdCodeBlockIndent ~ mdCodeBlockFence
+  private def nextPartParser[_: P]: P[Unit] = nextPartParser(0)
+
+  private def nextPartParser[_: P](indent: Int): P[Unit] = P {
+    def mdCodeBlockPrefix = mdCodeBlockIndent(indent) ~ mdCodeBlockFence
     def anotherBeg = P(CharIn("@=") | (codePrefix ~ nl) | listPrefix | tableSep | "+-" | nl)
     nl ~/ (nl | mdCodeBlockPrefix | hspaces0 ~/ anotherBeg)
   }
 
-  private def textParser[_: P]: P[Text] = P {
-    def end = P(End | nextPartParser)
+  private def textParser[_: P]: P[Text] = textParser(0)
+
+  private def textParser[_: P](indent: Int): P[Text] = P {
+    def end = P(End | nextPartParser(indent))
     def part: P[TextPart] = P(!paraEnd ~ (codeExprParser | linkParser | wordParser))
     def sep = P(!end ~ nlHspaces0)
     def text = hspaces0 ~ part.rep(1, sep = sep)
@@ -148,21 +157,36 @@ object ScaladocParser {
     }
   }
 
-  private def listBlockParser[_: P](minIndent: Int = 1): P[ListBlock] = P {
+  private def listItemParser[_: P](indent: Int): P[ListItem] = {
+    def innerBody =
+      (listBlockParser(indent) |
+        mdCodeBlockParser(indent) |
+        tableParser(indent) |
+        textParser(indent)).rep(0)
+
+    (textParser(indent) ~ innerBody)
+      .map { case (beginning, other) => ListItem(beginning +: other) }
+  }
+
+  private def listBlockParser[_: P]: P[ListBlock] = listBlockParser(0)
+
+  private def listBlockParser[_: P](minIndent: Int): P[ListBlock] = P {
+
     def listParser = (hspacesMinWithLen(minIndent) ~ listPrefix.! ~ hspaces1).flatMap {
       case (indent, prefix) =>
         def sep = (nl ~ hspacesMinWithLen(indent) ~ prefix ~ hspaces1).flatMap { x =>
           if (x != indent) Fail else Pass
         }
-        (textParser ~ listBlockParser(indent + 1).?)
-          .map { case (desc, list) => ListItem(desc, list) }
+        listItemParser(indent + 1)
           .rep(1, sep = sep)
           .map(x => ListBlock(prefix, x.toSeq))
     }
     startOrNl ~ listParser
   }
 
-  private def tableParser[_: P]: P[Table] = P {
+  private def tableParser[_: P]: P[Table] = tableParser(0)
+
+  private def tableParser[_: P](indent: Int): P[Table] = P {
     def toRow(x: Iterable[String]): Table.Row = Table.Row(x.toSeq)
     def toAlign(x: String): Option[Table.Align] = {
       def isEnd(y: Char) = y match {
@@ -186,7 +210,7 @@ object ScaladocParser {
     def cellChar = escape ~ AnyChar | !(nl | escape | tableSep) ~ AnyChar
     def row = (cellChar.rep.! ~ tableSpaceSep).rep(1)
     // non-standard but frequently used delimiter line, e.g.: +-----+-------+
-    def delimLine = hspaces0 ~ plusMinus
+    def delimLine = hspacesMin(indent) ~ plusMinus
     def sep = nl ~ (delimLine ~ nl).rep ~ tableSpaceSep
     // according to spec, the table must contain at least two rows
     def table = row.rep(2, sep = sep).map { x =>
@@ -226,7 +250,7 @@ object ScaladocParser {
   }
 
   private def termParser[_: P] =
-    listBlockParser() |
+    listBlockParser |
       mdCodeBlockParser |
       codeBlockParser |
       headingParser |
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/Scaladoc.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/Scaladoc.scala
@@ -97,10 +97,10 @@ object Scaladoc {
   /* List blocks */
 
   /** Represents a list item */
-  final case class ListItem(text: Text, nested: Option[ListBlock] = None)
+  final case class ListItem(terms: Seq[Term])
 
   /** Represents a list block */
-  final case class ListBlock(prefix: String, item: Seq[ListItem]) extends Term
+  final case class ListBlock(prefix: String, items: Seq[ListItem]) extends Term
 
   /* Tags */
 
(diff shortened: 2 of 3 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -148,21 +152,37 @@ object ScaladocParser {
...
+        mdCodeBlockParser(indent) |
+        tableParser(indent) |
+        textParser(indent)
+      ).rep(0, sep = paraSep)
+    
+    (textParser() ~ innerBody)
+     .map{case (beginning, other) => ListItem(beginning +: other)}
+  }
+
+  private def listBlockParser[_: P](indent: Int = 1): P[ListBlock] = P {
```
kitbellew, line 171: this `indent` conflicts with the one on line 153(left)/173(right). i used `minIndent` to avoid this clash.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -76,12 +76,16 @@ object ScaladocParser {
   private def mdCodeBlockIndent[_: P] = hspace.rep(max = 3)
   private def mdCodeBlockFence[_: P] = "`".rep(3) | "~".rep(3)
 
-  private def mdCodeBlockParser[_: P]: P[MdCodeBlock] = P {
-    (startOrNl ~ mdCodeBlockIndent.! ~ mdCodeBlockFence.!).flatMap { case (indent, fence) =>
+
+  private def mdCodeBlockParser[_: P](indent: Int = 1): P[MdCodeBlock] = P {
+    def indentParser = 
+      hspace.rep(min = indent, max = indent + 3)
```
kitbellew, line 82: so now default is from 1 to 4. before (line 76) was from 0 to 3. is that correct? do we have the right boundary tests?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -117,20 +121,20 @@ object ScaladocParser {
     nl ~/ (nl | mdCodeBlockPrefix | hspaces0 ~/ anotherBeg)
   }
 
-  private def textParser[_: P]: P[Text] = P {
+  private def textParser[_: P](indent: Int = 0): P[Text] = P {
```
kitbellew, line 124: optional: should we not have the default value? call with explicit?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -148,21 +152,37 @@ object ScaladocParser {
...
+    def innerBody = 
+      (!stop(indent) ~ 
+        listBlockParser(indent) |
+        mdCodeBlockParser(indent) |
+        tableParser(indent) |
+        textParser(indent)
+      ).rep(0, sep = paraSep)
+    
+    (textParser() ~ innerBody)
+     .map{case (beginning, other) => ListItem(beginning +: other)}
```
kitbellew, line 168: formatting: `map {`.

scalameta/trees/shared/src/main/scala/scala/meta/internal/Scaladoc.scala
```diff
@@ -97,10 +97,10 @@ object Scaladoc {
   /* List blocks */
 
   /** Represents a list item */
-  final case class ListItem(text: Text, nested: Option[ListBlock] = None)
+  final case class ListItem(term: Seq[Term])
 
   /** Represents a list block */
-  final case class ListBlock(prefix: String, item: Seq[ListItem]) extends Term
+  final case class ListBlock(prefix: String, items: Seq[ListItem]) extends Term
```
kitbellew, line 103: if you renamed `item` to `items`, then why not use `terms` on line 100?

kitbellew (commented):
good job figuring out this parser... i always struggle. a few comments below.

scalameta/trees/shared/src/main/scala/scala/meta/internal/Scaladoc.scala
dos65: Changed to `terms`
The reason is weird, but there are several places where parameter was named improperly. Like: `Paragraph.term`, `Text.part` ...

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
dos65: reverted it back to `minIndent`

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
dos65: Fixed.

Initially I noticed that `listParser` requires at least one space at the beginning and decided the same rule should be applied for mdCode but I wasn't right. Also changed it for `listParser` and added some tests - scaladoc parser allows list/mdcode/tables without leading space.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -148,21 +155,40 @@ object ScaladocParser {
...
+        tableParser(indent) |
+        textParser(indent)).rep(0, sep = paraSep)
+
+    (textParser ~ innerBody)
+      .map { case (beginning, other) => ListItem(beginning +: other) }
+  }
+
+  private def listBlockParser[_: P]: P[ListBlock] = listBlockParser(0)
+
+  private def listBlockParser[_: P](minIndent: Int): P[ListBlock] = P {
```
kitbellew, line 175: should we extract listParser below as, say, listParserImpl, and define listBlockParser as `startOrNl ~ listParserImpl(0)` and name the other one as `nestedListBlockParser` and define as `nl ~ listBlockParserImpl(...)`?

scalameta/trees/shared/src/main/scala/scala/meta/internal/Scaladoc.scala
kitbellew: yes, i think i started with `item: Type*` and don't like using plural for those.

kitbellew (changes requested):
could you add a lot more tests which show more complex nested list items, and where a list ends and another element (text, table, code, list with another labeling, tag annotation, etc), outside of the list, begins.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
dos65: I don't think that there will be any difference. Anyway it will be used only here so I think it's ok to leave it here.

dos65 (commented):
@kitbellew I did some changes. Could you re-review it?

I've also prepared [changes required for scalafmt](https://github.com/dos65/scalafmt/commit/5e8cfa3cf5010132e95a7b18e9f20c8748e04778). I've tried them locally on scalameta repo - looks like everything works well.

kitbellew:
@dos65 vadim, sorry for the delay, i will take a look soon, hopefully this week. thank you for checking thoroughly.

p.s. happy new year!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -148,21 +157,36 @@ object ScaladocParser {
...
+      (listBlockParser(indent) |
+        mdCodeBlockParser(indent) |
+        tableParser(indent) |
+        textParser(indent)).rep(0)
+
+    (textParser(indent) ~ innerBody)
+      .map { case (beginning, other) => ListItem(beginning +: other) }
+  }
+
+  private def listBlockParser[_: P]: P[ListBlock] = listBlockParser(0)
```
kitbellew, line 171: turns out indent can't be 0; the outermost list has to have at least one (hence `minIndent: Int = 1` above).

scalameta/trees/shared/src/main/scala/scala/meta/internal/Scaladoc.scala
kitbellew: i have extracted this change into a separate pr, #2532.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -148,21 +157,36 @@ object ScaladocParser {
     }
   }
 
-  private def listBlockParser[_: P](minIndent: Int = 1): P[ListBlock] = P {
+  private def listItemParser[_: P](indent: Int): P[ListItem] = {
+    def innerBody =
+      (listBlockParser(indent) |
+        mdCodeBlockParser(indent) |
```
kitbellew, line 163: here's what i found:
- while nested list can be indented by just 1 space, a nested markdown block has to be indented by the width of the list marker *and* the space that follows it, so we need two parameters here
- we can also include a nested `{{{` code block, it just doesn't need to have any special indentation.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -186,7 +210,7 @@ object ScaladocParser {
     def cellChar = escape ~ AnyChar | !(nl | escape | tableSep) ~ AnyChar
     def row = (cellChar.rep.! ~ tableSpaceSep).rep(1)
     // non-standard but frequently used delimiter line, e.g.: +-----+-------+
-    def delimLine = hspaces0 ~ plusMinus
+    def delimLine = hspacesMin(indent) ~ plusMinus
```
kitbellew, line 213: this is also not needed, a table doesn't have to be indented within a list, it looks like.

scalameta/trees/shared/src/main/scala/scala/meta/internal/Scaladoc.scala
```diff
@@ -97,10 +97,10 @@ object Scaladoc {
   /* List blocks */
 
   /** Represents a list item */
-  final case class ListItem(text: Text, nested: Option[ListBlock] = None)
+  final case class ListItem(terms: Seq[Term])
```
kitbellew, line 100: since every list item starts with text first, i'd keep `text` and replace `nested` with `terms: Seq[Term] = Nil`

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
```diff
@@ -452,25 +480,33 @@ class ScaladocParserSuite extends FunSuite {
...
-                    Some(ListBlock("1.", Seq(ListItem(Text(Seq(Word(list21)))))))
-                  ),
-                  ListItem(
-                    Text(Seq(Word(list12))),
-                    Some(
-                      ListBlock(
-                        "a.",
-                        Seq(ListItem(Text(Seq(Word(list21)))), ListItem(Text(Seq(Word(list22)))))
+
+    val expected =
```
kitbellew, line 484: most of this diff contains formatting changes (newline before `Option`) and type `Seq -> List`. i'd keep just the actual changed part.

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
```diff
@@ -516,14 +551,16 @@ class ScaladocParserSuite extends FunSuite {
...
-                    Text(Seq(Word(list11))),
-                    Some(
+                    Seq(
+                      Text(Seq(Word(list11))),
                       ListBlock(
                         "i.",
                         Seq(
                           ListItem(
-                            Text(Seq(Word(list21))),
-                            Some(ListBlock("I.", Seq(ListItem(Text(Seq(Word(list31)))))))
```
kitbellew, line 526: if we go with my suggestion for `ListItem`, this diff becomes very small: `Some -> Seq`.

tests/shared/src/test/scala/scala/meta/tests/parsers/ScaladocParserSuite.scala
```diff
@@ -660,7 +701,153 @@ class ScaladocParserSuite extends FunSuite {
...
+    val list11 = "List11"
+    val list12 = "List12"
+    val list21 = "List21"
+
+    val result =
+      parseString(
+        s"""
+        /**
+          * Some text:
+          *- $list11
```
kitbellew, line 833: good test but this is not a valid list, it has to be indented.

kitbellew (commented):
@dos65 since i have way too many comments or suggestions, i will simply check out your code and send a modified pr for your review.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
dos65: Oh, yep - found it by attached link to scaladoc spec. On the other hand, it's allowed in scala3.

I think in general it would be better to allow it. I think in general people don't know how comment is rendered in scaladoc and can't notice that they missed space before list item. I assume there will be more unexpected rewrites rather that invalid list detection.

kitbellew: current approach is requiring one space (not "one more space than the rest of text", as documentation describes), but it doesn't allow having absolutely no spaces at all. problem with that is, the list marker can really be a continuation of the previous line ("the default value of this parameter is <nl> 1." or "it can be 2 <nl> - or 3 - in some cases"). so, I wouldn't relax.

dos65: hmm, I've just realized that without this rule might become unpredictable. There is the case when scalfmt might move `-` to the beginning of next line and turn it into a list. I agree, let's keep this rule ignoring scala3 behavior

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
dos65: good points!

Do you have a fix for that or should I do that after we merge your PRs?

kitbellew: i do have a fix, will share in a couple of hours. expanded your solution, added embedded `{{{` code blocks, also option for tags (someone complained that `@example` tag needs to have embedded code).

kitbellew: shared part 1 (of 3): #2537. there are two more commits. let me know if you prefer to see everything at once, rather than one piece of functionality at a time, and i will upload more commits into same pr. otherwise, will submit more after review and merge.

Issue #2499, opened 2021-10-29
Signature error after 0.10.8


cquiroz:
**Describe the bug**

In one particular project after upgrading to metals 0.10.8 I started seeing these errors in metals.log

```
Oct 29, 2021 12:54:55 PM scala.meta.internal.pc.CompilerAccess handleError
SEVERE: unsupported signature: class scala.meta.internal.semanticdb.Signature$Empty$ Empty
java.lang.RuntimeException: unsupported signature: class scala.meta.internal.semanticdb.Signature$Empty$ Empty
	at scala.sys.package$.error(package.scala:27)
	at scala.meta.internal.semanticdb.scalac.SymbolInformationOps$XtensionGSymbolMSymbolInformation.sig(SymbolInformationOps.scala:153)
	at scala.meta.internal.semanticdb.scalac.SymbolInformationOps$XtensionGSymbolMSymbolInformation.toSymbolInformation(SymbolInformationOps.scala:243)
	at scala.meta.internal.semanticdb.scalac.TextDocumentOps$XtensionCompilationUnitDocument$traverser$3$.saveSymbol$1(TextDocumentOps.scala:217)
	at scala.meta.internal.semanticdb.scalac.TextDocumentOps$XtensionCompilationUnitDocument$traverser$3$.trySymbolDefinition(TextDocumentOps.scala:221)
	at scala.meta.internal.semanticdb.scalac.TextDocumentOps$XtensionCompilationUnitDocument$traverser$3$.tryFindMtree(TextDocumentOps.scala:328)
	at scala.meta.internal.semanticdb.scalac.TextDocumentOps$XtensionCompilationUnitDocument$traverser$3$.traverse(TextDocumentOps.scala:689)
	at scala.meta.internal.semanticdb.scalac.TextDocumentOps$XtensionCompilationUnitDocument$traverser$3$.traverse(TextDocumentOps.scala:205)
	at scala.reflect.api.Trees$Traverser.$anonfun$traverseStats$2(Trees.scala:2520)
	at scala.reflect.api.Trees$Traverser.atOwner(Trees.scala:2529)
	at scala.reflect.api.Trees$Traverser.$anonfun$traverseStats$1(Trees.scala:2520)
	at scala.reflect.api.Trees$Traverser.traverseStats(Trees.scala:2519)
	at scala.reflect.internal.Trees$Template.traverse(Trees.scala:548)
	at scala.reflect.internal.Trees.itraverse(Trees.scala:1604)
[...]

PR #2500, opened 2021-10-29
Don't fail on empty signature

tgodzik:
In some cases empty signature might show up when semanticdb is generated in the presentation compiler. In that case it's normal that it can happen.

Fixes https://github.com/scalameta/scalameta/issues/2499


dos65 (approved):

dos65:
@tgodzik it's needed to run fmt

PR #2504, opened 2021-11-08
Remove old mima filters

tgodzik:
These should no longer be needed and the API must now be stable.


dos65 (approved):
@tgodzik Sorry, I missed this review request.

tgodzik:
> @tgodzik Sorry, I missed this review request.

Totally forgot about it also :sweat_smile:

PR #2510, opened 2021-11-16
Testkit - ignore KafkaConfig file as it can't be transformed in EqualProps test

### LINKED ISSUE #2509: Tree transformer - StackOverflow on chained Apply/Select tree
I noticed that sometimes `EqualProps` test behaves flaky.

The reason is in [this source from Kafka](https://github.com/apache/kafka/blob/1.1/core/src/main/scala/kafka/server/KafkaConfig.scala#L748-L958).

It fails on transforming the selected block that has a highly-nested structure.

dos65:
I think it would be a challenge to fix #2509  so it's easier to ignore this file in test.


tgodzik (approved):
Och, so that's the issue here! Thanks for finding this one!

PR #2527, opened 2021-12-20
Revert "Update scalafmt-core to 3.2.1"

tgodzik:
Reverts scalameta/scalameta#2525

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,4 @@
-version = 3.2.1
+version = 2.7.5
 align.preset = none
 align.stripMargin = true
 assumeStandardLibraryStripMargin = true


kitbellew:
i have no objections... but since no files changed formatting, what exactly is the block with the linked pr?

tgodzik:
> i have no objections... but since no files changed formatting, what exactly is the block with the linked pr?

I think it got forgotten :sweat_smile:  @dos65 could you take a look at https://github.com/scalameta/scalameta/pull/2468 ?

dos65:
hmm... actually fmt in #2525 was failed because of invalid config. I remember that I did a fix to return non zero exit code in this case but it seems that it doesn't work.

what about #2468 - mostly issues with markdown parsing that would be nice to fix. I have an incomplete PR with fix but couldn't find time for it. Issue with comments always looks less prioritized in compare to new things that appears :smile: 

Let's do the following: I'll try to make one more try until the year. If I won't succeed, I'll turn them all into issues and adjust existing comments with current formatter state so we will be able to upgrade scalafmt.

tgodzik:
Ok cool @dos65 I am ok with turning into issues also, would be good to update scalafmt.

PR #2530, opened 2021-12-29
Show a clear message when using secondary constructors in a wrong place

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/2529

Ideally, we would make the invariant messages have an additional message parameter so that this never happens again, but this is more involved.


dos65 (approved):
LGTM!

PR #2534, opened 2021-12-31
ScaladocParser: take the rightmost code suffix

kitbellew:


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -47,7 +47,7 @@ object ScaladocParser {
   private def tableSpaceSep[_: P] = P(hspaces0 ~ tableSep)
 
   private def codePrefix[_: P] = P("{{{")
-  private def codeSuffix[_: P] = P(hspaces0 ~ "}}}")
+  private def codeSuffix[_: P] = P(hspaces0 ~ "}}}" ~~ !"}")
```
dos65, line 50: Maybe use repeat?
```suggestion
  private def codeSuffix[_: P] = P(hspaces0 ~ "}".rep(min = 3, max = "3"))
```

dos65 (approved):
LGTM!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
dos65: hm - wrong suggestion, ignore it

PR #2537, opened 2022-01-01
ScaladocParser: allow other list-embedded elements

kitbellew:
For now, that does not include markdown as that requires extra indent. Minor modification based on #2489, @dos65 did the bulk of the word.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
@@ -71,10 +71,18 @@ object ScaladocParser {
    * https://spec.commonmark.org/0.29/#fenced-code-blocks
    */
 
-  private def mdCodeBlockIndent[_: P] = hspace.rep(max = 3)
+  /* while spec mentions that a fence marker can be indented between 0 and 3 spaces,
+   * examples under "list item" show that it's *relative* to the containing element;
+   * also, lists are included in "container blocks" which have that property.
+   * however, for non-indented case (mdOffset = 0), since we don't capture the actual
+   * offset, let's allow 0 (next to asterisk) and 1 (one space from asterisk). */
+  private def getMdOffsetMax(mdOffset: Int) = math.max(1, mdOffset) + 3
+
   private def mdCodeBlockFence[_: P] = "`".rep(3) | "~".rep(3)
 
-  private def mdCodeBlockParser[_: P]: P[MdCodeBlock] = P {
+  private def mdCodeBlockParser[_: P](mdOffset: Int = 0): P[MdCodeBlock] = P {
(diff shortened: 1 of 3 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
```diff
@@ -235,6 +236,13 @@ object ScaladocParser {
       tableParser |
       leadTextParser // keep at the end, this is the fallback
 
+  private def embeddedTermsParser[_: P](indent: Int): P[Seq[Term]] = P {
+    (listBlockParser(indent) |
+      codeBlockParser |
```
dos65, line 241: `codeBlockParser` should take indent too.

I haven't covered all cases in tests but the additional ones are:
~~~
// case 1 - no indent
- item  
```
code
```
// result
Paragraph(
   List(
      ListBlock(...),
      MdCodeBlock(..)
   )
)
------
// case 2 - larger than list indent + 3
- item  
      ```
      code
      ```
// result
Paragraph(
   List(
      ListBlock(
        ListItem(List(Text(Word("item"), Word("```"), ...)))
      )
   )
)
~~~

kitbellew: > `codeBlockParser` should take indent too.

what makes you think that? and what should be indented, the delimiters `{{{` or the entire block?

what about table? should that be indented, too (and according to what spec)?

dos65: Oh, I confused it with mdcodeblock. I think in this case it would be easier to review to whole changeset to list item parsing than splitting it onto several PRs.

What about tables - probably they should be indented. 
There is no spec on it but I think it should work in a similar way as fenced codeblocks/quotes in markdown (like this one:  [example 254](https://spec.commonmark.org/0.30/#example-254))

kitbellew: my read is that they should not.

if scala3 implemented the full markdown, we have to write a completely new parser; this one doesn't support blockquotes, md lists (which use different different markers and increment), headings with other markers etc.

so, I'm considering this just an extension of scala 2 with fenced code blocks. and in scala 2, they are not documented to be indented, and i don't know what the actual scaladoc tool does (maybe that is what we should find out, as documentation is not very comprehensive).

kitbellew: and I will update this pr with additional commits (but do read them one at a time, should be easier).

however, i was hoping you'd review the other pre first, so i can rebase and adjust.

kitbellew: rebased, added the other two commits

dos65 (approved):
LGTM!

Thx for competing that! :rocket:

PR #2541, opened 2022-01-04
ScaladocParser: handle `@usecase` as special case

kitbellew:
Let's handle cases where label is expected but description is not, by reading in the entire tag line as label. This adds required handling to @usecase which must be formatted all on one line.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
@@ -135,13 +135,17 @@ object ScaladocParser {
 
   private def tagParser[_: P]: P[Tag] = P {
     def label = P((nl ~ !nextPartParser()).? ~ hspaces0 ~ wordParser)
+    // special case: @usecase takes a single code line, on the same line
+    def labelInline = P(hspaces0 ~ (!nl ~ AnyChar).rep(1).!.map(Word))
     def desc = P {
       (textParser().? ~ embeddedTermsParser())
         .map { case (x, terms) => x.fold(terms)(_ +: terms) }
     }
     hspaces0 ~ ("@" ~ labelParser).!.flatMap { tag =>
       val tagType = TagType.getTag(tag)
-      def labelOpt = if (tagType.hasLabel) label.map(x => Some(x)) else Pass(None)
+      def labelOpt =
+        if (!tagType.hasLabel) Pass(None)
+        else (if (tagType.optDesc) label else labelInline).map(x => Some(x))
       def descOpt = if (tagType.optDesc) desc else Pass(Nil)
       (labelOpt ~ descOpt).map { case (label, desc) => Tag(tagType, label, desc) }
     }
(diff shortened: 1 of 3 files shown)


dos65 (approved):
LGTM!

dos65:
@kitbellew I think we also need to add `optDesc= false` to [all group tags](https://github.com/scalameta/scalameta/blob/c39d246e12bd470602a388b88ca9a982223e8031/scalameta/trees/shared/src/main/scala/scala/meta/internal/Scaladoc.scala#L175-L181). According [to this doc (Member grouping tags)](https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html) they are all onliners

kitbellew:
> @kitbellew I think we also need to add `optDesc= false` to [all group tags](https://github.com/scalameta/scalameta/blob/c39d246e12bd470602a388b88ca9a982223e8031/scalameta/trees/shared/src/main/scala/scala/meta/internal/Scaladoc.scala#L175-L181). According [to this doc (Member grouping tags)](https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html) they are all onliners

didn't find where it says they must be on one line. other than group, all can have description.

dos65:
@kitbellew 
> Members that do not have a `@group` tag will be listed as “Ungrouped” in the resulting documentation.
>
>   `@group` <group> - mark the entity as a member of the <group> group.
>    `@groupname` <group> <name> - provide an optional name for the group. <name> is displayed as the group header before the group description.
>    `@groupdesc` <group> <description> - add optional descriptive text to display under the group name. Supports multiline formatted text.
>    `@groupprio` <group> <priority> - control the order of the group on the page. Defaults to 0. Ungrouped elements have an implicit priority of 1000. Use a value between 0 and 999 to set a relative position to other groups. Low values will appear before high values.

`@groupname` and `@groupprio` are one line tags

kitbellew:
> @kitbellew
> 
> > Members that do not have a `@group` tag will be listed as “Ungrouped” in the resulting documentation.
> > `@group`  - mark the entity as a member of the  group.
> > `@groupname`   - provide an optional name for the group.  is displayed as the group header before the group description.
> > `@groupdesc`   - add optional descriptive text to display under the group name. Supports multiline formatted text.
> > `@groupprio`   - control the order of the group on the page. Defaults to 0. Ungrouped elements have an implicit priority of 1000. Use a value between 0 and 999 to set a relative position to other groups. Low values will appear before high values.
> 
> `@groupname` and `@groupprio` are one line tags

see you saying that i can't do this:
```
/**
  * @groupname somegroup
  *    this is the name of somegroup
  */
```

group priority looks short, could be.

dos65:
@kitbellew 
> see you saying that i can't do this:
Yes.

Somehow I missed a part with tag formats.
Upd:
> Members that do not have a `@group` tag will be listed as “Ungrouped” in the resulting documentation.
>    `@group <group>` - mark the entity as a member of the <group> group.
>    `@groupname <group> <name>` - provide an optional name for the group. <name> is displayed as the group header before the group description.
>    `@groupdesc <group> <description>` - add optional descriptive text to display under the group name. Supports multiline formatted text.
>    `@groupprio <group> <priority>` - control the order of the group on the page. Defaults to 0. Ungrouped elements have an implicit priority of 1000. Use a value between 0 and 999 to set a relative position to other groups. Low values will appear before high values.

Only in`groupdesc` it's mentioned that `Supports multiline formatted text`.
Also, I did a grep in scala-library sources - all `@group` tags are one liners

kitbellew:
cool, let's change that. alas, documentation is really very sparse, and no examples. perhaps the assumption was that only `scaladoc` will implement the logic, and people will empirically figure out how to use it.

kitbellew:
@dos65 added in #2543.

PR #2544, opened 2022-01-04
ScaladocParser: fix check for table delim line

kitbellew:
The existing one was getting confused by a "-" list following the table.


dos65 (approved):
:+1:

Issue #2546, opened 2022-01-06
Scala 3 dialect rejects legal `for` comprehension


SethTisue:
- **Version**: 3.2.1
- **Configuration**:
```
version = 3.2.1
runner.dialect = scala3
```

## Steps

Given:
```scala
class C:
  def foo =
    (for a <- List(1)
         b <- List(2)
     yield a + b).toSet.size
```

When I run scalafmt like this:

```bash
scalafmt .
```

## Problem

Scalafmt bails with an error:
```
/Users/tisue/tmp/20220105/./87.scala:4: error: [dialect scala3] illegal start of simple expression
         b <- List(2)
           ^
```

## Workaround

inserting a line break after `for` avoids the problem

tgodzik:
This looks a bit weird and I am not sure how to treat it indentation wise. We assumed previously that either for can have `{` or indentation. This is none of that and yet it seems to compile.

I can't find any explanation for this in https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html

Should we report it as a bug?

dos65:
@tgodzik it seems that it's a valid case according to https://dotty.epfl.ch/docs/internals/syntax.html. 
```
ForExpr           ::=  ‘for’ ‘(’ Enumerators0 ‘)’ {nl} [‘do‘ | ‘yield’] Expr     ForYield(enums, expr) / ForDo(enums, expr)
                    |  ‘for’ ‘{’ Enumerators0 ‘}’ {nl} [‘do‘ | ‘yield’] Expr
                    |  ‘for’     Enumerators0          (‘do‘ | ‘yield’) Expr
Enumerators0      ::=  {nl} Enumerators [semi]
```

`Nl` before `Enumerators` is optional

tgodzik:
That looks like it should be on one line though:
```
‘for’     Enumerators0          (‘do‘ | ‘yield’) Expr
```

PR #2547, opened 2022-01-06
Allow first statement inside for to be on the same line as the keyword

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/2546

I don't think there is a way to make it simpler yet :thinking:

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -545,10 +545,10 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
             } else sepRegions
           (nextRegions, currRef)
         } else if (dialect.allowSignificantIndentation && curr.is[KwFor]) {
-          val updatedSepRegions =
-            if (sepRegions.nonEmpty && sepRegions.head.isInstanceOf[RegionParen])
-              RegionParen(true) :: sepRegions.tail
-            else sepRegions
+          val updatedSepRegions = sepRegions match {
+            case RegionParen(_) :: tail => RegionParen(true) :: tail
+            case _ => sepRegions
+          }
           (updatedSepRegions, currRef)
         } else {
           (sepRegions, currRef)
@@ -581,17 +581,12 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
         def canProduceLF: Boolean = {
           lastNewlinePos != -1 &&
           prev != null && (prev.is[CanEndStat] || token.is[Indentation.Outdent]) &&
-          next != null && next.isNot[CantStartStat] &&
-          (sepRegions.isEmpty ||
-            sepRegions.head.isInstanceOf[RegionBrace] ||
-            sepRegions.head.isInstanceOf[RegionCase] ||
-            sepRegions.head.isInstanceOf[RegionEnum] ||
-            sepRegions.head.isInstanceOf[RegionIndent] ||
-            sepRegions.head.isInstanceOf[RegionIndentEnum] ||
-            (
-              sepRegions.head.isInstanceOf[RegionParen] &&
-                sepRegions.head.asInstanceOf[RegionParen].canProduceLF
-            ))
+          next != null && next.isNot[CantStartStat] && sepRegions.headOption.forall {
+            case _: RegionBrace | _: RegionCase | _: RegionEnum => true
+            case _: RegionIndent | _: RegionIndentEnum => true
+            case x: RegionParen => x.canProduceLF
+            case _ => false
+          }
         }
 
         if (dialect.allowSignificantIndentation) {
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -543,7 +544,15 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
                 newRegions
             } else sepRegions
           (nextRegions, currRef)
-        } else (sepRegions, currRef)
+        } else if (dialect.allowSignificantIndentation && curr.is[KwFor]) {
```
kitbellew, line 547: is `KwFor` the only such case?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -543,7 +544,15 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+          val updatedSepRegions =
+            if (sepRegions.nonEmpty && sepRegions.head.isInstanceOf[RegionParen])
```
kitbellew, line 549: optionally, instead: `if (sepRegions.headOption.exists(_.isInstanceOf...)`, like line 430.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -578,7 +587,11 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
             sepRegions.head.isInstanceOf[RegionCase] ||
             sepRegions.head.isInstanceOf[RegionEnum] ||
             sepRegions.head.isInstanceOf[RegionIndent] ||
-            sepRegions.head.isInstanceOf[RegionIndentEnum])
+            sepRegions.head.isInstanceOf[RegionIndentEnum] ||
```
kitbellew, line 590: i think it might be time to change this entire parenthesized expression `(sepRegions.isEmpty || ...)` to
```
sepRegions.headOption.forall {
  case _: RegionBrace | _: RegionCase | _: RegionEnum => true
  case _: RegionIndent | _: RegionIndentEnum => true
  case x: RegionParen => x.canProduceLF
  case _ => false
}
```

dos65 (commented):
It seems that in dotty parser there is a trick with setting a special region to the scanner: https://github.com/lampepfl/dotty/blob/0b4c6e7b8d74394e8bad4babab86fdc6d9982971/compiler/src/dotty/tools/dotc/parsing/Parsers.scala#L2561

Here `enumerators` are parsed like they wold be in inside `{}`.

I can't say that it's a much better approach but it seems it would require less changes. Also, we need to check other usages `inSepRegion` - maybe there are similar issues to that.

tgodzik:
> I can't say that it's a much better approach but it seems it would require less changes. Also, we need to check other usages `inSepRegion` - maybe there are similar issues to that.

I think this need more changes as we need to add a lot of special places where we need to close the BraceRegion without the braces.

kitbellew:
strange question, perhaps. if scala already has a parser, wouldn't it be easier to refactor dotty to expose just the parser and then write a converter from their representation into ours? the only question is how to support multiple dialects, of course...

tgodzik:
> strange question, perhaps. if scala already has a parser, wouldn't it be easier to refactor dotty to expose just the parser and then write a converter from their representation into ours? the only question is how to support multiple dialects, of course...

There was a talk of doing that, but it would require more than just translating representation. Dotty has much less nodes in the AST and it doesn't always fit what is actually being displayed. I think, currently it's much more work than it's worth. Especially that we are only fixing some smaller issues currently.

kitbellew:
@tgodzik please let me know if you are planning to complete this soon. i was going to release a new version of scalameta and use that in scalafmt (scaladoc changes), so that @dos65 can finish #2468.

tgodzik:
> @tgodzik please let me know if you are planning to complete this soon. i was going to release a new version of scalameta and use that in scalafmt (scaladoc changes), so that @dos65 can finish #2468.

I hoped I would finish it today, but I will take a look tomorrow. If you could hold of on releasing until then it would be awesome.

kitbellew:
@tgodzik thank you! yes, i will. there are still a pair of prs in scalafmt that need your review anyway.

dos65 (approved):
I've tried my suggestion with hack on replacing `sepRegions` as it's done in dotty but it doesn't work.

Let's merge this solution.

I think that should be the only case for Bracket + multiline statements that isn't a single expression.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: sepRegions is used to simplify handling of whitespaces, is some of them they are automatically omitted so that the parser doesn't have to handle it.

I think KW is the only case like that, can't think of another really.

SethTisue:
thank you!

Issue #2516, opened 2021-12-11
ScalametaParser: comment before outdent not part of indented expression


kitbellew:
In this example:
```
val (xTranslation, yTranslation) = (isHMirror, isVMirror) match
      case (true, true)  => ((rotationWidth / 2).toInt, (rotationHeight / 2).toInt)  // Top left
      case (true, false) => ((rotationHeight / 2).toInt, -(rotationWidth / 2).toInt) // Bottom left
      case (false, true) => (-(rotationHeight / 2).toInt, (rotationWidth / 2).toInt) // Top right
      case (false, false) => (-(rotationWidth / 2).toInt, -(rotationHeight / 2).toInt) // Bottom right
```
the `// Bottom right` comment *should* be part of the `Term.Match` but it only belongs to `Defn.Val` instead.

PR #2549, opened 2022-01-15
ScalametaParser: various refactoring to prepare for indented comments

kitbellew:
Prepares for #2516.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -875,38 +875,30 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   }
   def atPos[T <: Tree](start: Pos, end: Pos)(body: => T): T = {
     val startTokenPos = start.startTokenPos
-    val result = body
-    var endTokenPos = end.endTokenPos
-    if (endTokenPos < startTokenPos) endTokenPos = startTokenPos - 1
-    val pos = trimmedPos(startTokenPos, endTokenPos)
-    result.withOrigin(Origin.Parsed(input, dialect, pos)).asInstanceOf[T]
+    atPosWithBody(startTokenPos, body, end)
   }
 
-  def trimmedPos(startTokenPos: Int, endTokenPos: Int): TokenStreamPosition = {
-    def skipTrivia(range: Range): Int = {
-      range
-        .dropWhile(i => scannerTokens(i).is[Trivia])
-        .headOption
-        .getOrElse(range.start)
-    }
+  def atPosWithBody[T <: Tree](startTokenPos: Int, body: T, endPos: Pos): T = {
+    var endTokenPos = endPos.endTokenPos
+    if (endTokenPos < startTokenPos) endTokenPos = startTokenPos - 1
+
+    def skipBy(range: Range, f: Token => Boolean): Int =
+      range.dropWhile(i => f(scannerTokens(i))).headOption.getOrElse(range.start)
+    @inline def skipTrivia(range: Range): Int = skipBy(range, _.is[Trivia])
+
     val rangeFromStart = startTokenPos to endTokenPos
     val (start, end) =
       if (rangeFromStart.isEmpty) (startTokenPos, endTokenPos)
       else (skipTrivia(rangeFromStart), skipTrivia(rangeFromStart.reverse))
 
-    if (start == end && scannerTokens(start).is[Trivia])
-      TokenStreamPosition(start, end)
-    else
-      TokenStreamPosition(start, end + 1)
+    val endExcl = if (start == end && scannerTokens(start).is[Trivia]) end else end + 1
+    val pos = TokenStreamPosition(start, endExcl)
+    body.withOrigin(Origin.Parsed(input, dialect, pos))
   }
 
   def atPosTry[T <: Tree](start: Pos, end: Pos)(body: => Try[T]): Try[T] = {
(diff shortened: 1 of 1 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -904,38 +870,30 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
-        .headOption
-        .getOrElse(range.start)
-    }
+  def atPosWithBody[T <: Tree](startTokenPos: Int, body: T, endPos: Pos): T = {
+    var endTokenPos = endPos.endTokenPos
+    if (endTokenPos < startTokenPos) endTokenPos = startTokenPos - 1
+
+    def skipBy(range: Range, f: Token => Boolean): Int =
+      range.dropWhile(i => f(scannerTokens(i))).headOption.getOrElse(range.start)
+    @inline def skipTrivia(range: Range): Int = skipBy(range, _.is[Trivia])
```
dos65, line 882: `skipBy` is called only from `skipTrivia`. Maybe merge them into a single func?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2389,15 +2349,15 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   }
 
   def condExprInParens[T <: Token: TokenInfo]: Term = {
-    def isFollowedBy[T <: Token: TokenInfo] = {
+    def isFollowedByT = {
```
dos65, line 2352: I would leave implicit in signature here as compiler might (or might not) optimize this function if it doesn't depend on parent function.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -671,13 +638,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
         } else {
           if (canProduceLF) {
-            if (isLeadingInfixOperator(next)) {
-              nextToken(prevPos, nextPos, sepRegions)
-            } else {
-              var token = scannerTokens(lastNewlinePos)
-              if (newlines) token = LFLF(token.input, token.dialect, token.start, token.end)
-              (sepRegions, TokenRef(token, lastNewlinePos, lastNewlinePos + 1, lastNewlinePos))
-            }
```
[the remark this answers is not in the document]
kitbellew: 1. `isLeadingInfixOperator` is defined as `dialect.allowSignificantIndentation && ...` (https://github.com/scalameta/scalameta/blob/main/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L764); the `else` two lines above closes `if (dialect.allowSignificantIndentation)` (https://github.com/scalameta/scalameta/blob/main/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L592). therefore, this `if` never fires.
2. the next three lines are defined in `lastWhitespaceToken` (https://github.com/scalameta/scalameta/blob/main/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L574).

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: i intentionally did this. in the next pr i will define skipping by whitespace only, keeping comments.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: this change is because
1. `isFollowedBy` uses the same symbol for the generic type (`T`) as the outer `condExprInParens`
2. however, instead of renaming it `U`, observe that `T` is exactly the same when used; and `condExprInParens` has all the implicits already.

kitbellew: i can also rename to `U` or move this method out. let me know what you prefer.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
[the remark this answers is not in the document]
dos65: Nice catch! Thx for explanation.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: yep I've seen that.
As I remember in case of  inner `def` with a full signature compiler moves this `def` into synthetic one in class body by itself. Otherwise if it references to smth in outer def it will be a lambda.

I don't think that it there is a significant difference in performance but there is a point of doing that event if types and implicits are same.

> let me know what you prefer.

any option

kitbellew: done. by the way, the method it calls (`followedByToken`) is not used anywhere else; perhaps this newline check should simply have been part of that method instead.

PR #2550, opened 2022-01-15
ScalametaParser: outdent after indented comments

### LINKED ISSUE #2516: ScalametaParser: comment before outdent not part of indented expression

kitbellew:
Previously, we would outdent immediately after a non-trivial token while ignoring any trailing comments which followed.
Fixes #2516.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -389,11 +389,39 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           case region :: tail if region.isIndented && canEndIndentation(curr.token) =>
             sepRegions = tail
             val outdent = mkOutdentToken(curr.token)
-            curr = TokenRef(outdent, curr.pos, curr.pos, prevPos)
+            val outdentPos = findOutdentPos(prevPos, curr.pos, region)
+            curr = TokenRef(outdent, curr.pos, curr.pos, outdentPos)
             true
           case _ => false
         }
       }
+    }
+
+    private def findOutdentPos(prevPos: Int, currPos: Int, region: SepRegion): Int = {
+      val outdent = region.indent
+      @tailrec
+      def iter(i: Int, pos: Int, indent: Int): Int = {
+        if (i >= currPos) {
+          if (pos < currPos) pos else currPos - 1
+        } else {
+          scannerTokens(i) match {
+            case _: LF | _: FF =>
+              iter(i + 1, if (pos == prevPos) i else pos, 0)
+            case _: Space | _: Tab if indent >= 0 =>
+              iter(i + 1, pos, indent + 1)
+            case Whitespace() =>
+              iter(i + 1, pos, indent)
+            case _: Comment if indent < 0 || outdent <= indent =>
+              iter(i + 1, i + 1, -1)
+            case _ => pos
+          }
+        }
+      }
+
+      val iterPos = 1 + prevPos
+      if (iterPos < currPos) iter(iterPos, prevPos, -1)
+      else if (scannerTokens(currPos).is[EOF]) currPos
+      else prevPos
     }
 
     @tailrec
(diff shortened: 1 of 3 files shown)


tests/shared/src/test/scala-2.13/scala/meta/tests/parsers/dotty/Scala3PositionSuite.scala
```diff
@@ -661,4 +866,68 @@ class Scala3PositionSuite extends BasePositionSuite(dialects.Scala3) {
...
+       |  case (false, true) => (baz, qux) // c2""".stripMargin,
+    """|Pat.Tuple (x, y)
+       |Term.Match (foo, bar) match
+       |  case (true, false) => (baz, qux) // c1
+       |  case (false, true) => (baz, qux) // c2
+       |Term.Tuple (foo, bar)
+       |Case case (true, false) => (baz, qux) // c1
+       |Pat.Tuple (true, false)
+       |Term.Tuple (baz, qux)
+       |Case case (false, true) => (baz, qux)
```
dos65, line 927: Shouldn't the second `case` include `// c2`?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -813,9 +849,29 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+    }
+    term.stats match {
+      case (b: Term.Block) :: Nil => keepInnerBlock(b)
+      case (s: Term) :: Nil if commentPosOpt(in.prevTokenPos).forall { commentPos =>
+            s.origin match {
+              case Origin.Parsed(_, _, pos) => pos.end > commentPos
+              case _ => false
+            }
+          } =>
+        s
```
dos65, line 866: Could you explain why it's needed?
It took me a bit but as I got the logic is the following: `keep Block in case if consist from single term in case if there is a trailing comment`. Is it for scalafmt?

Also, I've tried the following case:
```scala
object y {
 val a =
   if cond then
     fx // c1
   else
      gx
  /* c2 */}
```
Having this logic `fx //c1` is parsed as a `Term.Block` while it should be a `Term.Name`.

dos65 (commented):
Great work! Overall looks good.

I left few suggestions and not sure about one thing.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -378,13 +378,40 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+            case _: Comment if indent < 0 || outdent <= indent =>
+              iter(i + 1, i + 1, -1)
+            case _ => pos
+          }
+        }
+      }
+
+      val iterPos = 1 + prevPos
+      if (iterPos < currPos) iter(iterPos, prevPos, -1)
+      else { if (scannerTokens(currPos).is[EOF]) currPos else prevPos }
```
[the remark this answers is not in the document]
kitbellew: will do, thank you

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -624,8 +659,9 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
               } else false
             }
 
-            if (needOutdent)
-              (sepRegions.tail, mkOutdent(lastNewlinePos))
+            val needOutdent = getOutdentIfNeeded()
+            if (needOutdent ne null)
+              (sepRegions.tail, mkOutdentTo(needOutdent, nextPos))
```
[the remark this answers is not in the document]
kitbellew: i was thinking about this, too...

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: would you only keep the outer block if the comment was standalone? generally, i would prefer not to drop block at all, even if for a single statement, but for maximum backwards compatibility...

the original issue, #2516, had a problem with a trailing comment, not standalone.

tests/shared/src/test/scala-2.13/scala/meta/tests/parsers/dotty/Scala3PositionSuite.scala
kitbellew: good point. how do i do it? how do i find the code which sets the end position for case? perhaps it didn't use autopos there.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: if we don't put block around it, then Name will include the comment.

and more on why this shouldn't be just Name. `then` expects a block or at least an expression, not a specific identifier or constant; to me, it's similar to the difference between "number" (expression) and "digit" (basic component of a number).

kitbellew:
@bjaglin clarification. comments are already in intermediate tree positions, this change also allows them in trailing positions. only initial comments are [still] not included.

tests/shared/src/test/scala-2.13/scala/meta/tests/parsers/dotty/Scala3PositionSuite.scala
kitbellew: i think i figured it out. will add a separate commit fixing this. but first, please review and approve #2553.

kitbellew (commented):
made changes locally, will upload soon. need to merge #2553 first, though.

kitbellew:
submitted the changes, rebased. please take another look, @dos65.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
dos65: I'm still not sure about this change.
The comment says ` if block ends in a comment and single term doesn't contain it, keep block` but it's incorrect.

If you revert this part and run `Scala3PositionSuite` you will notice that single terms includes comment positions.

For example:
```scala
object y {
   val a =
      fx
      // c1
}
// positions with this change
    """|Template {
       |  val a =
       |    fx
       |    // c1
       |}
       |Self   @@val a =
       |Defn.Val val a =
       |    fx
       |    // c1
       |Term.Block fx
       |    // c1
       |""".stripMargin


// positions without this change
// c1 is included in fx-term

    """|Template {
       |  val a =
       |    fx
       |    // c1
       |}
       |Self   @@val a =
       |Defn.Val val a =
       |    fx
       |    // c1
       |Term.Name fx
       |    // c1
       |""".stripMargin

```
So, the question is if there is a need to do that if non-block term pos anyway contains comments?

kitbellew: that's exactly it. if you don't apply this change, then the comment will be part of `Term.Name` -- but it shouldn't be. and i am not sure if that's desirable, really.

the comment will not be lost, with or without the last commit, but i am simply suggesting that the comment should be part of the block, or that we should remove the block that was there in the first place.

how's this: can we merge the 5 commits which add comment handling, and then have a separate discussion on whether to keep or remove the so-called trivial blocks?

kitbellew:
@dos65 for now, removed the controversial keep-trivial-block commit. let's discuss that separately.

dos65:
@kitbellew let me do last check - will try this version on scalafix tests to be sure that  it won't affect it

dos65 (approved):

dos65:
> let me do last check - will try this version on scalafix tests to be sure that it won't affect it

checked that - tests pass

kitbellew:
thank you for double checking, @dos65

PR #2551, opened 2022-01-17
Properly parse patterns in splices inside quoted patterns

adampauls:
Discussion [here](https://contributors.scala-lang.org/t/bind-in-splices-in-quoted-patterns/5533) and Dotty PR [here](https://github.com/lampepfl/dotty/pull/14277). Previously, splices inside quoted patterns (like the `...` in `case '{foo(${...})`) were parsed as blocks instead of patterns, even though the typer treats such splices as patterns. This means that arbitrary blocks were syntactically permitted, even though they would never compile, and some patterns (like those containing bindings like `${y@...}` were not syntactically permitted, even though the typer would accept them. 

This requires a change in the API to add a Term than can wrap a spliced pattern. Let me know if there's anything better I can do here.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -45,6 +45,10 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     def isInside() = nested > 0
   }
 
+  private object QuotedPatternContext {
+    var inPattern: Boolean = false
+  }
+
   def parseRule[T <: Tree](rule: this.type => T): T = {
     // NOTE: can't require in.tokenPos to be at -1, because TokIterator auto-rewinds when created
     // require(in.tokenPos == -1 && debug(in.tokenPos))
@@ -3029,7 +3033,12 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   def macroSplice(): Term = autoPos {
     accept[MacroSplice]
     QuoteSpliceContext.enter()
-    val splice = Term.SplicedMacroExpr(autoPos(Term.Block(inBraces(blockStatSeq()))))
+    val splice =
+      if (QuotedPatternContext.inPattern)
+        Term.SplicedMacroPat(autoPos(inBraces(pattern())))
+      else
+        Term.SplicedMacroExpr(autoPos(Term.Block(inBraces(blockStatSeq()))))
+
     QuoteSpliceContext.exit()
     splice
   }
@@ -3567,7 +3576,11 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           val patterns = inParens(if (token.is[RightParen]) Nil else noSeq.patterns())
           makeTuple[Pat](patterns, () => Lit.Unit(), Pat.Tuple(_))
         case MacroQuote() =>
-          Pat.Macro(macroQuote())
+          val oldInPattern = QuotedPatternContext.inPattern
+          QuotedPatternContext.inPattern = true
+          val res = Pat.Macro(macroQuote())
+          QuotedPatternContext.inPattern = oldInPattern
+          res
         case MacroQuotedIdent() =>
           Pat.Macro(macroQuotedIdent())
         case KwGiven() =>
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -38,16 +38,23 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
   /* ------------- PARSER ENTRY POINTS -------------------------------------------- */
 
-  private object QuoteSpliceContext {
+  /** Utility for tracking a context, like whether we are inside a pattern or a quote. */
+  private trait NestedContextBase {
     private var nested = 0
-    def enter() = nested += 1
-    def exit() = nested -= 1
+    def within[T](body: => T): T = {
+      nested += 1
+      try {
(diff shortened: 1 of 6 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -45,6 +45,10 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     def isInside() = nested > 0
   }
 
+  private object QuotedPatternContext {
+    var inPattern: Boolean = false
```
kitbellew, line 49: i'd just copy the contents of splice context above, as-is.

or even:
```
private abstract class QuoteContextBase {
  private var nested = 0
  def enter() = nested += 1
  def exit() = nested -= 1
  def isInside() = nested > 0
}
private object QuoteSpliceContext extends QuoteContextBase
private object QuotedPatternContext extends QuoteContextBase
```

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -604,6 +605,8 @@ object TreeSyntax {
           case head :: Nil => s("${ ", head, " }")
           case other => s("${", r(other.map(i(_)), ""), n("}"))
         }
+      case Term.SplicedMacroPat(pat) =>
+        s("${ ", pat, " }")
```
kitbellew, line 609: do we need spaces around `pat`? why not `"${"` and `"}"`?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3567,7 +3576,11 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           val patterns = inParens(if (token.is[RightParen]) Nil else noSeq.patterns())
           makeTuple[Pat](patterns, () => Lit.Unit(), Pat.Tuple(_))
         case MacroQuote() =>
-          Pat.Macro(macroQuote())
+          val oldInPattern = QuotedPatternContext.inPattern
+          QuotedPatternContext.inPattern = true
+          val res = Pat.Macro(macroQuote())
+          QuotedPatternContext.inPattern = oldInPattern
+          res
```
kitbellew, line 3583: perhaps:
```
try {
  QuotedPatternContext.enter()
  Pat.Macro(macroQuote())
} finally {
  QuotedPatternContext.exit()
}
```

kitbellew (changes requested):

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
adampauls: I copied line 605. I think if I change one I should change the other?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
adampauls: Should I do this to the uses of enter/exit elsewhere too?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
adampauls: Done.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: ooh, tricky question :) and thank you for suggesting. my take is this:

if others (@tgodzik and/or @dos65) agree with my preference for this approach (poor man's "destructor") over something like
```
  enter()
  val res = ...
  exit()
  res
```
then it would be nice to make that change. i would, however, do it as a separate commit, to keep it clean.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
kitbellew: sorry, didn't notice this. then let's keep it consistent with line 605 (as well as 595 and probably others). please ignore the original comment :)

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3028,24 +3039,26 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
   def macroSplice(): Term = autoPos {
     accept[MacroSplice]
-    QuoteSpliceContext.enter()
-    val splice = Term.SplicedMacroExpr(autoPos(Term.Block(inBraces(blockStatSeq()))))
-    QuoteSpliceContext.exit()
-    splice
+    QuoteSpliceContext.within {
+      if (QuotedPatternContext.isInside())
```
adampauls, line 3043: The functional change.

kitbellew: lovely!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
adampauls: I will rebase, one sec.

adampauls: Okay, done.

kitbellew (approved):

PR #2563, opened 2022-01-19
ScalametaParser: use outer block ending in comment

kitbellew:
When we have a block containing a single statement, the existing code removes the so-called trivial block and keeps just the inner statement.

However, this might result in extending the end position of the statement itself and, after #2550, include a comment token. So, to continue the discussion started in #2550, resubmitting this commit.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -862,9 +862,30 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       case _ => term
     }
 
-  private def dropOuterBlock(term: Term.Block): Term =
+  private def dropOuterBlock(term: Term.Block): Term = {
+    def commentPosOpt(pos: Int) = scannerTokens(pos) match {
+      case _: Comment => Some(pos)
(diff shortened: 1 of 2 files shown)


dos65:
@kitbellew sorry for delay.

Technically, I agree that it's better to keep it as block in case of trailing comment.
On the other hand, it feels like this change isn't required by scalafmt and more about having a more clear result. Am I right?

I'm wondering maybe it would be easier just to remove `dropOuterBlock` and keep tree as a block if it's a single expr in braces or indented?

kitbellew:
first, yes, scalafmt is fine either way.

second, about your suggestion regarding keeping outer block. the removal of those outer blocks led to some kludgy code in scalafmt, although it mainly applied to cases when the block was removed even though there were explicit braces around it. so, essentially, my view is that if there's anything in the block that doesn't belong to the only statement it contains (comment or brace), I'd keep the block.

dos65:
@kitbellew 

What about the following cases:
 - single line comment - not clear why block is introduced 
   ```scala
   val a =
      fx // c1
   ```
 - single stm block that starts with comment - no block in this case
   ```scala
   val a = 
      /* c1 */
      fx
   ```
> first, yes, scalafmt is fine either way.

If it isn't required by any feature wouldn't be it better not to add this logic?
Having the cases I mentioned, I wouldn't say that existing logic makes it easier to reason about in which tree code with comment will be parsed.

PR #2564, opened 2022-01-21
Allow _IDENT in string interpolation

tgodzik:
This doesn't work in the Scala 2 parser, but it feels like a bug, so should be safe to fix in the Scalameta parser


dos65 (approved):
:+1:

Issue #1940, opened 2019-12-13
Add dialect to parse toplevel statements and packages


olafurpg:
Currently, when using the default dialect it's not possible to include toplevel statements
```scala
// Foo.scala
def foo() = 42 // error: expected class or package
```
With the `Sbt1` dialect, it's possible to parse toplevel statements but it's not valid to declare packages
```scala
// build.sbt
package bar // error:
```

It would be nice to have a third dialect that accepts both package declarations and toplevel statements. This dialect would be useful for example for Scalafmt, where we just want to format as much code as possible.

adampauls:
I would like to resurrect this issue. The Scala3 dialect has allowTopLevelTerms, but it forbids package statements ( see [here](https://github.com/scalameta/scalameta/blob/main/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L5088). I'm assuming this is because `allowToplevelTerms` is used for the REPL, where package statements don't make sense, but it still seems curious not to allow the user some flexibility here.

tgodzik:
It should be possible to change split this, since there is only one place this is being used:

https://github.com/scalameta/scalameta/blob/dc7320df79bb9ee8626281154fc36e8cb9cbc5cd/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L5088

I think it would be enough to add an additional dialect option and change to it in the code above. I don't have time to look at it currently, but I think we should be ok to merge this change.

PR #2565, opened 2022-01-25
ScalametaParser: Make `allowToplevelTerms` accept package declarations

adampauls:
Addresses https://github.com/scalameta/scalameta/issues/1940. A little more complicated than I thought. Comments welcome. I'm not sure if I put the test in the right place.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -5030,7 +5030,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   private val consumeStat: PartialFunction[Token, Stat] = {
     case KwImport() => importStmt()
     case KwExport() => exportStmt()
-    case KwPackage() if !dialect.allowToplevelTerms => packageOrPackageObjectDef()
+    case KwPackage() =>
+      packageOrPackageObjectDef(if (dialect.allowToplevelTerms) consumeStat else topStat)
     case DefIntro() => nonLocalDefOrDcl(secondaryConstructorAllowed = true)
     case EndMarkerIntro() => endMarker()
     case ExprIntro() => stat(expr(location = NoStat, allowRepeated = true))
@@ -5099,14 +5100,13 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     stats.toList
   }
 
-  def topStatSeq(): List[Stat] = statSeq(topStat, errorMsg = "expected class or object definition")
   def topStat: PartialFunction[Token, Stat] = {
     case Ellipsis(_) =>
       ellipsis(1, astInfo[Stat])
     case Unquote() =>
       unquote[Stat]
     case KwPackage() =>
-      packageOrPackageObjectDef()
+      packageOrPackageObjectDef(topStat)
     case KwImport() =>
       importStmt()
     case KwExport() =>
@@ -5253,21 +5253,21 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     stats.toList
   }
 
-  def packageOrPackageObjectDef(): Stat = autoPos {
+  def packageOrPackageObjectDef(statpf: PartialFunction[Token, Stat]): Stat = autoPos {
     require(token.is[KwPackage] && debug(token))
     if (ahead(token.is[KwObject])) packageObjectDef()
-    else packageDef()
+    else packageDef(statpf)
   }
 
-  def packageDef(): Pkg = autoPos {
+  def packageDef(statpf: PartialFunction[Token, Stat]): Pkg = autoPos {
     accept[KwPackage]
     def packageBody = {
       if (isColonEol(token)) {
         accept[Colon]
         in.observeIndented()
-        indented(topStatSeq())
+        indented(statSeq(statpf))
       } else {
-        inBracesOrNil(topStatSeq())
+        inBracesOrNil(statSeq(statpf))
       }
     }
     Pkg(qualId(), packageBody)
@@ -5280,25 +5280,14 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   }
 
   def source(): Source = autoPos {
-    if (dialect.allowToplevelTerms) scriptSource()
-    else batchSource()
+    batchSource(if (dialect.allowToplevelTerms) consumeStat else topStat)
   }
 
   def quasiquoteSource(): Source = entrypointSource()
 
   def entrypointSource(): Source = source()
 
-  def scriptSource(): Source = autoPos {
-    // WONTFIX: https://github.com/scalameta/scalameta/issues/368
-    if (dialect.toplevelSeparator == "") {
-      Source(parser.statSeq(consumeStat))
-    } else {
-      require(dialect.toplevelSeparator == EOL)
-      Source(parser.statSeq(consumeStat))
-    }
-  }
-
-  def batchSource(): Source = autoPos {
+  def batchSource(statpf: PartialFunction[Token, Stat] = topStat): Source = autoPos {
     def inBracelessPackage(): Boolean = token.is[KwPackage] && !ahead(token.is[KwObject]) && ahead {
       qualId()
       if (token.is[LF]) { next() }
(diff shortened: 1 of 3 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -5099,14 +5101,13 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     stats.toList
   }
 
-  def topStatSeq(): List[Stat] = statSeq(topStat, errorMsg = "expected class or object definition")
```
kitbellew, line 5102: why not just redefine it as
```
statSeq(
   if (dialect.allowToplevelTerms) consumeStat else topStat,
   "some appropriate error message"
)
```
and remove most of the method parameter changes?

tests/shared/src/test/scala/scala/meta/tests/parsers/TopLevelTermSuite.scala
```diff
@@ -0,0 +1,82 @@
...
+
+import munit._
+
+class TopLevelTermSuite extends FunSuite {
+  test("allowToplevelTerms simple") {
+    implicit val dialect: Dialect = dialects.Scala3.withAllowToplevelTerms(true)
+
+    val sourceString = """def foo(x: Int): Int = x
+                          |foo(x)
+""".trim.stripMargin
```
kitbellew, line 14: this is a weird string constant. indent all lines, including last, use margin char `|`

tests/shared/src/test/scala/scala/meta/tests/parsers/TopLevelTermSuite.scala
```diff
@@ -0,0 +1,82 @@
...
+
+    val tree = sourceString.parse[Source].get
+
+    assert(tree.syntax == sourceString)
+    assert(
+      tree.structure ==
+        """Source(List(
+          |Defn.Def(Nil, Term.Name("foo"), Nil, List(List(Term.Param(Nil, Term.Name("x"), Some(Type.Name("Int")), None))), Some(Type.Name("Int")), Term.Name("x")), 
+          |Term.Apply(Term.Name("foo"), List(Term.Name("x")))))
+          |""".trim.stripMargin.split('\n').mkString("")
```
kitbellew, line 24: is it the same as `replace("\n", "")`?

kitbellew (changes requested):
@tgodzik needs to review

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -5030,7 +5030,9 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   private val consumeStat: PartialFunction[Token, Stat] = {
     case KwImport() => importStmt()
     case KwExport() => exportStmt()
-    case KwPackage() if !dialect.allowToplevelTerms => packageOrPackageObjectDef()
```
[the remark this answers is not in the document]
adampauls: I would happily do that. I assumed I should keep existing behavior, but happy to change it. Anyone we should ask?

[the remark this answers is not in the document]
kitbellew: we should ask @tgodzik and @dos65

tests/shared/src/test/scala/scala/meta/tests/parsers/TopLevelTermSuite.scala
```diff
@@ -0,0 +1,82 @@
...
+
+    val tree = sourceString.parse[Source].get
+
+    assert(tree.syntax == sourceString)
+    val expected =
+      """Source(List(Pkg(Term.Name("bar"), List(
+                     |Defn.Def(Nil, Term.Name("foo"), Nil, List(List(Term.Param(Nil, Term.Name("x"), Some(Type.Name("Int")), None))), Some(Type.Name("Int")), Term.Name("x")), 
+                     |Term.Apply(Term.Name("foo"), List(Term.Name("x"))))
+                     |)))""".trim.stripMargin.split('\n').mkString("")
+    assert(tree.structure == expected)
```
tgodzik, line 59: Let's use assertEquals, which  should give us a better diff in case of issues.

tests/shared/src/test/scala/scala/meta/tests/parsers/TopLevelTermSuite.scala
```diff
@@ -0,0 +1,82 @@
...
+""".trim.stripMargin
+
+    val tree = sourceString.parse[Source].get
+
+    assert(tree.syntax == sourceString)
+    val expected =
+      """Source(List(Pkg(Term.Name("bar"), List(
+                     |Defn.Def(Nil, Term.Name("foo"), Nil, List(List(Term.Param(Nil, Term.Name("x"), Some(Type.Name("Int")), None))), Some(Type.Name("Int")), Term.Name("x")), 
+                     |Term.Apply(Term.Name("foo"), List(Term.Name("x"))))
+                     |)))""".trim.stripMargin.split('\n').mkString("")
```
tgodzik, line 58:
```suggestion
    val expected =
      Source(List(Pkg(Term.Name("bar"), List(
                     Defn.Def(Nil, Term.Name("foo"), Nil, List(List(Term.Param(Nil, Term.Name("x"), Some(Type.Name("Int")), None))), Some(Type.Name("Int")), Term.Name("x")), 
                     Term.Apply(Term.Name("foo"), List(Term.Name("x"))))
                     ))).structure
```

tgodzik (changes requested):
Thanks for the contribution! I think we could simplify it a bit and just allow package, this will not break any existing behaviour and we can make the parser more permissive.

tests/shared/src/test/scala/scala/meta/tests/parsers/TopLevelTermSuite.scala
adampauls: Done.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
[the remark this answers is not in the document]
adampauls: Fantastic.

tests/shared/src/test/scala/scala/meta/tests/parsers/TopLevelTermSuite.scala
adampauls: Probably. I stole this from another [suite](https://github.com/scalameta/scalameta/blob/main/tests/shared/src/test/scala/scala/meta/tests/parsers/SbtSuite.scala#L83). Happy to replace.

tests/shared/src/test/scala/scala/meta/tests/parsers/TopLevelTermSuite.scala
adampauls: Done.

tests/shared/src/test/scala/scala/meta/tests/parsers/TopLevelTermSuite.scala
adampauls: Much better, thanks.

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -717,6 +733,7 @@ object Dialect extends InternalDialect {
       allowSpliceUnderscores: Boolean,
       allowTermUnquotes: Boolean,
       allowToplevelTerms: Boolean,
+      allowPackageStatementsWithToplevelTerms: Boolean,
```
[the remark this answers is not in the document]
adampauls: Happily.

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -687,6 +702,7 @@ final class Dialect private (
       allowSpliceUnderscores,
       allowTermUnquotes,
       allowToplevelTerms,
+      allowPackageStatementsWithToplevelTerms,
```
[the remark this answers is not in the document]
adampauls: Done

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -661,6 +674,8 @@ final class Dialect private (
       allowSpliceUnderscores: Boolean = this.allowSpliceUnderscores,
       allowTermUnquotes: Boolean = this.allowTermUnquotes,
       allowToplevelTerms: Boolean = this.allowToplevelTerms,
+      allowPackageStatementsWithToplevelTerms: Boolean =
```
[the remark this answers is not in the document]
adampauls: Done

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
adampauls: I didn't like that because then the name is misleading, since it might permit things that are not top-level statements (as defined by `topStat` and the Scala 3 spec). I deleted it precisely so I would have to audit every position to see whether it really wanted a topStat or the dialect-specific notion of one. 
I would suggest leaving it as is I have it, or renaming `topStat` to be `scala3TopStat`. What do you prefer?

kitbellew: i would recommend considering this rename idea as part of some other pull request
 this one is about package, so let's stick to that.

adampauls: Okay sure. I'll just leave a comment warning about the confusion of topStatSeq() not always meaning a Seq of topStat.

tgodzik (approved):
Thank you a lot for the contribution! LGTM :rocket:

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: oh well...

kitbellew: @adampauls would you consider revising this like i suggested? it may have been a mistake to say "Okay sure" and yet leave this unchanged.

here's an example of the adverse effect: https://github.com/scalacenter/scalafix/pull/1552#pullrequestreview-882591159

PR #2566, opened 2022-01-27
ScalametaParser: Avoid stack overflow when annotation comes after modifier

adampauls:
Before this PR, parsing ```implicit @foo def foo(): Int``` would stack overflow. This is because `tmplDef` is happy to accept modifiers, even though in all cases it's been used, modifiers have already been parsed. This seems to only be a problem for dotty, but I added a test for both anyways.

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/ModSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/ModSuite.scala
@@ -479,4 +479,8 @@ class ModSuite extends ParseSuite {
     ) =
       templStat("protected[foo] private def foo = ???")
   }
+
+  test("Annotation after modifier") {
+    interceptParseErrors("implicit @foo def foo(a: Int): Int")
+  }
 }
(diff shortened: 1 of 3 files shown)


tests/shared/src/test/scala/scala/meta/tests/parsers/ModSuite.scala
```diff
@@ -479,4 +479,8 @@ class ModSuite extends ParseSuite {
     ) =
       templStat("protected[foo] private def foo = ???")
   }
+
+  test("Annotation after modifier") {
+    interceptParseErrors("implicit @foo def foo(a: Int): Int")
```
adampauls, line 484: This already passed.

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MinorDottySuite.scala
```diff
@@ -442,6 +442,18 @@ class MinorDottySuite extends BaseDottySuite {
     )
   }
 
+  test("annotation after modifier") {
+    runTestError[Stat](
+      "implicit @foo def foo(): Int",
```
adampauls, line 447: This would stack overflow.

tgodzik (approved):
LGTM! I broke my workspace a bit testing this out :smile:

Issue #2567, opened 2022-01-28
`ArrayIndexOutOfBoundsException` when parsing a wrong Scala 3 lambda parameter


KacperFKorban:
The error occurred in Scastie.
Link: https://scastie.scala-lang.org/KacperFKorban/xQWBNvfBRp6doXtHN2Rbtw/22
(Note: sometimes the full stack trace doesn't show)

Code snippet:

```scala
List(1,2,3).map { using i: Int => i }
```

Output

<details>

```
java.lang.ArrayIndexOutOfBoundsException: Index 105 out of bounds for length 56
	at scala.meta.tokens.Tokens.apply(Tokens.scala:34)
	at scala.meta.internal.parsers.ScalametaParser.$anonfun$trimmedPos$1(ScalametaParser.scala:898)
	at scala.collection.immutable.Range.dropWhile(Range.scala:294)
	at scala.meta.internal.parsers.ScalametaParser.skipTrivia$1(ScalametaParser.scala:898)
	at scala.meta.internal.parsers.ScalametaParser.trimmedPos(ScalametaParser.scala:905)
	at scala.meta.internal.parsers.ScalametaParser.atPos(ScalametaParser.scala:891)
	at scala.meta.internal.parsers.ScalametaParser.atPos(ScalametaParser.scala:881)
	at scala.meta.internal.parsers.ScalametaParser.$anonfun$expr$32(ScalametaParser.scala:2645)
	at scala.meta.internal.parsers.ScalametaParser.atPos(ScalametaParser.scala:888)
	at scala.meta.internal.parsers.ScalametaParser.convertToParam$1(ScalametaParser.scala:2645)
	at scala.meta.internal.parsers.ScalametaParser.convertToParams$1(ScalametaParser.scala:2666)
	at scala.meta.internal.parsers.ScalametaParser.$anonfun$expr$22(ScalametaParser.scala:2668)
	at scala.meta.internal.parsers.ScalametaParser.atPos(ScalametaParser.scala:888)
	at scala.meta.internal.parsers.ScalametaParser.$anonfun$expr$2(ScalametaParser.scala:2613)
	at scala.meta.internal.parsers.ScalametaParser.atPos(ScalametaParser.scala:888)
	at scala.meta.internal.parsers.ScalametaParser.autoPos(ScalametaParser.scala:922)
	at scala.meta.internal.parsers.ScalametaParser.expr(ScalametaParser.scala:2397)
[...]

PR #2568, opened 2022-01-28
Add test case for #2567

tgodzik:
@KacperFKorban  This seems to work correctly in the current Scalameta version, maybe it's just using an older version? Or this is actually a scastie issue.

closes https://github.com/scalameta/scalameta/issues/2567


kitbellew:
@tgodzik is it important that in the original example, there are no parens around the parameter?

tgodzik:
> @tgodzik is it important that in the original example, there are no parens around the parameter?

That's my bad, I tested that originally, but then forgot about `assertLayout` which we used for comparing printed tree :sweat_smile:

kitbellew (approved):

PR #2569, opened 2022-02-02
Update Algolia credentials

tgodzik:


kitbellew:
you have something which adds trailing commas...

tgodzik:
> you have something which adds trailing commas...

That would be prettifier and I think that's the default style in javascript?

kitbellew:
> > you have something which adds trailing commas...
> 
> That would be prettier and I think that's the default style in javascript?

i don't disagree :)

kitbellew (approved):
this looks good to me, and tests pass, even though i don't fully understand the significance of the apiKey change (or appId) introduction.

tgodzik:
> this looks good to me, and tests pass, even though i don't fully understand the significance of the apiKey change (or appId) introduction.

apparently that's a change needed for the search bar to work, but I am not sure exactly how that works :sweat_smile: 

I'll wait for @olafurpg to take a look.

olafurpg (approved):
The API key and app ID looks correct. However, I can't find the "scalameta" application in my algolia account so I don't know if the index has been enabled yet. I sent an email to Algolia DocSearch requesting to add you @tgodzik to all of the Scalameta-related Algolia apps. We may want to wait with merging this PR until we hear back from them

tgodzik:
> > The API key and app ID looks correct. However, I can't find the "scalameta" application in my algolia account so I don't know if the index has been enabled yet. I sent an email to Algolia DocSearch requesting to add you @tgodzik to all of the Scalameta-related Algolia apps. We may want to wait with merging this PR until we hear back from them
> 
> @tgodzik any news on this?

No update, I think we need to redo the indexes with a new account for algolia.

PR #2570, opened 2022-02-03
Fix issues with using atPos with tree position

tgodzik:
`Tree.pos` gives us the position in the file, while `atPos` needs to figure out the token index, so we can't just use start and end from Position. Instead we can jsut copy the origin of the original Tree.

The implicit conversions from Int are a bit at fault here, but when I tried to touch them everything blew up, so I'd rather escape with my life.

Fixes https://github.com/scalameta/scalameta/issues/2567


kitbellew (approved):
clever catch!

PR #2571, opened 2022-02-07
Upgrade sclafmt to 3.4.2

dos65:


tgodzik (approved):
LGTM!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -615,16 +615,15 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
-             *    and previous token can't continue expr on the next line
-             *  - At the end of `match` block even if indentation level is not changed
-             *    Example:
-             *    ```
-             *    x match
-             *    case 1 =>
-             *    case 2 => // <- produce outdent
-             *    foo()
-             *    ```
+             *   - If indentation on next line is less than current and previous token can't
```
kitbellew, line 618: that one looks weird... why did it get shifted another space?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -639,18 +638,17 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
              * Indent is needed in the following cases:
-             * - Indetation on new line is greater and previous token can start indentation
-             *   and token can start indentation
-             * - Indentation on the new line is the same and the next token is the first `case` clause in match
-             *   Example:
-             *   ```
-             *   x match // <- mk indent
-             *   case 1 =>
-             *   ```
+             *   - Indetation on new line is greater and previous token can start indentation and
```
kitbellew, line 641: i was expecting one space but it has two... same question as above

tests/jvm/src/test/scala-2.13/scala/meta/tests/api/SurfaceSuite.scala
```diff
@@ -8,10 +8,13 @@ import scala.reflect.runtime.universe._
 import scala.reflect.runtime.{universe => ru}
 
 class SurfaceSuite extends FunSuite {
-  object CoreReflection extends {
-    val u: ru.type = ru
-    val mirror: u.Mirror = u.runtimeMirror(classOf[scala.meta.Tree].getClassLoader)
-  } with scala.meta.internal.trees.Reflection with scala.meta.internal.tokens.Reflection
+  object CoreReflection
+      extends {
```
kitbellew, line 12: so you didn't use parent constructors = always? i thought that was the purpose of the change we merged

kitbellew (commented):
i think there's a bug with lists...

tests/jvm/src/test/scala-2.13/scala/meta/tests/api/SurfaceSuite.scala
dos65: I thought that this split before `extends` shouldn't depend on `binPack.parentConstructors` setting.
But then I've checked PR one more time and I found that it appears in case of `binPack.parentConstructors = Oneline` that we are currently using :disappointed: 

I decided just to go ahead as we already spent o lot of time on waiting fixes.

PR #2581, opened 2022-02-09
ScalametaParser: implement ammonite script parser

### LINKED ISSUE #2155: @ line in Ammonite scripts mistaken as an error
**Describe the bug**
In Ammonite scripts, such as the Mill tool's build.sc file, it is occasionally necessary to have a line containing only `@` to ensure a certain execution order in the interpreter.

For example, when adding nonstandard repositories for ivy imports, one would do the following:
```scala
interp.repositories() ++= Seq( ... )
@
import $ivy.`...::...:...`
```
This forces the first statement to be executed before the last one; otherwise `import` statements are executed first.

Currently, the scalameta interpreter does not understand this `@` line, giving (in this case) the following error:
```
identifier expected but import found
```

**To reproduce**
 Steps to reproduce the behaviour:
1. Create an Ammonite script (ie. `test.sc`) with the following contents:
```scala
interp.repositories() ++= Seq(coursierapi.MavenRepository.of("https://jitpack.io"))
@
import $ivy.`com.github.davoclavo:mill-ensime:0.0.3`
```
2. The error should appear on line 3. You can execute the file (ie. `amm test.sc`) to see that it is valid syntax.

**Expected behaviour**
There should be no errors in this file.

**Installation**
Operating system: macOS Catalina (10.15.7)
Editor: VSCode
Metals version: 1.9.4

**Search terms**
ammonite, ivy import, at sign, interp, mill

kitbellew:
Adds a new MultiSource tree, and a new Input.Ammonite wrapper which is used as an indication that multi-source tokenization is needed.

Fixes #2155.

### DIFF
--- a/scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Api.scala
+++ b/scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Api.scala
@@ -21,6 +21,9 @@ private[meta] trait Aliases {
 
     type Slice = scala.meta.inputs.Input.Slice
     val Slice = scala.meta.inputs.Input.Slice
+
+    type Ammonite = scala.meta.inputs.Input.Ammonite
+    val Ammonite = scala.meta.inputs.Input.Ammonite
   }
 
   type Position = scala.meta.inputs.Position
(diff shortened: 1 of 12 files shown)


dos65 (approved):
:+1: :+1: :+1:

kitbellew:
@dos65 thank you. wanted to get confirmation that this approach will work with metals (the original question). cc @tgodzik 

i have obviously already checked that it works with the formatter.

dos65:
@kitbellew it should work well. the only we need is to check places where create `Input` and turn it into ammonite one

PR #2582, opened 2022-02-12
Dialect: deprecate toplevelSeparator as unused

kitbellew:


tgodzik (approved):

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -228,7 +227,7 @@ final class Dialect private (
       allowViewBounds = allowViewBounds,
       allowWithTypes = allowWithTypes,
       allowXmlLiterals = allowXmlLiterals,
-      toplevelSeparator = toplevelSeparator,
+      toplevelSeparator = null, // unused
```
dos65, line 230: Wouldn't it be enough just to mark it as deprecated?
There is a probability that somebody might get NPE. I understand that it's low but who knows.

kitbellew: i can use "" if you prefer.

dos65: maybe keep it as it is?
we don't do anything similar with other deprecated settings here. for example: `allowWithTypes`

kitbellew: ok, will do

PR #2585, opened 2022-02-12
ScalametaParser: bug with currying in func types

kitbellew:
Since secondOpenParenPos was initialized to -1 but changed only on 0, it was never really changed, and the syntax error about multiple parameter lists in function types likely resulted in IndexOutOfBoundsException.


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -697,9 +696,11 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
         val rawtss = new ListBuffer[List[Tree]]
         while (!hasTypes && !hasImplicits && token.is[LeftParen]) {
-          if (openParenPos != in.tokenPos && secondOpenParenPos == 0)
-            secondOpenParenPos = in.tokenPos
-          accept[LeftParen]
+          if (rawtss.nonEmpty) {
```
tgodzik, line 699: Do we need to do this in a while then? Since in every step we add to `rawtss` the loop will effectively only run once, no?

Should we replace it with an `if` ?

tgodzik (changes requested):

dos65 (commented):
It seems that it would be nice to cover this part by a test

kitbellew:
> It seems that it would be nice to cover this part by a test

is there no test? 😂 I'll take a look

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: @tgodzik done. @dos65 added tests.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -648,87 +648,71 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
-          accept[Colon]
-          Type.TypedParam(name, typ())
-        }
-        @tailrec
-        def paramOrType(): Tree = token match {
-          case Ellipsis(rank) =>
-            ellipsis(rank, astInfo[Tree])
-          case Unquote() =>
-            unquote[Tree]
-          case KwImplicit() if !hasImplicits =>
```
dos65, line 674: it's not directly related to this pr but I got curious
@kitbellew @tgodzik does anybody know in which case there might be `implicit`? :thinking: 

Currently, we allow the following: `type X = (implicit A) => Int` but it's invalid.

tgodzik: Not really, this might be accepted by the parser because it was when we forked off from the original parser, but might no longer be relevant.

kitbellew:
> LGTM!

ah, missed it. thank you!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: I think it's good as is, no need to remove implicit here.

PR #2586, opened 2022-02-12
ScalametaParser.simpleType: don't set pos twice

kitbellew:
There is an outer autoPos which would override anyway.


tgodzik (approved):
LGTM!

PR #2591, opened 2022-02-13
Dialect: remove fields deprecated over a year ago

kitbellew:


tgodzik (changes requested):
We need to update mima and the next version we release I would bump at least the minor to acknowledge the changes.

kitbellew:
> We need to update mima and the next version we release I would bump at least the minor to acknowledge the changes.

Mima done. Will use version 4.5.0 next.

tgodzik (approved):

PR #2595, opened 2022-02-14
ScalametaParser: fix bug with Type.Refine position

kitbellew:
Because of parameter shadowing, it was setting position using an inner Option[Type] rather than the outer type as all the other cases (and the original Type.Compound code which preceded this implementation). Also, use copyPos for entire patternTyp.loop.


dos65 (approved):
LGTM!

PR #2597, opened 2022-02-15
ScalametaParser: no multiple checks of newlinesOpt

kitbellew:


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3785,11 +3777,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       name: Name,
       paramss: List[List[Term.Param]]
   ): Ctor.Secondary = {
-    newLineOptWhenFollowedBy[LeftBrace]
-    val (init, stats) = token match {
-      case LeftBrace() => constrBlock()
-      case _ => accept[Equals]; constrExpr()
-    }
+    if (!isAfterOptNewLine[LeftBrace]) accept[Equals]
+    val (init, stats) = constrExpr()
```
tgodzik, line 3781: It was `constrBlock()` for `{` before. Wouldn't this change the logic?

tgodzik (changes requested):
Just one question.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: thank you! tired eyes, i guess, thought i saw the same. tests are not 100% coverage... although in my other job, they are >95%.

tgodzik: most likely there is no big difference, since I think `constrExpr` can also invoke `constrBlock` (not 100% sure)

kitbellew: you were right. also, all these `constrXxx` methods are only used once, so i simply inlined, to reduce checks.

PR #2599, opened 2022-02-15
ScalametaParser: simplify argumentExprOrPrefixExpr

kitbellew:
No need to use recursion, no need to rebuild the list of arguments as they are simply checked for existence of Term.Repeated.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2185,26 +2185,21 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   def argumentExprsOrPrefixExpr(): List[Term] = {
     if (token.isNot[LeftBrace] && token.isNot[LeftParen]) prefixExpr(allowRepeated = false) :: Nil
     else {
-      def argsToTerm(args: List[Term], openParenPos: Int, closeParenPos: Int): Term = {
-        def badRep(rep: Term.Repeated) = syntaxError("repeated argument not allowed here", at = rep)
-        def loop(args: List[Term]): List[Term] = args match {
-          case Nil => Nil
-          case Term.Assign(_, rep: Term.Repeated) :: _ => badRep(rep)
-          case (rep: Term.Repeated) :: _ => badRep(rep)
-          case (t: Term) :: rest => t :: loop(rest)
-          case _ => unreachable(debug(args))
-        }
-        atPos(openParenPos, closeParenPos)(makeTupleTerm(loop(args)))
+      def findRep(args: List[Term]): Option[Term.Repeated] = args.collectFirst {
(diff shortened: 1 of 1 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2185,26 +2185,23 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
-          case (rep: Term.Repeated) :: _ => badRep(rep)
-          case (t: Term) :: rest => t :: loop(rest)
-          case _ => unreachable(debug(args))
-        }
-        atPos(openParenPos, closeParenPos)(makeTupleTerm(loop(args)))
+      def badRep(rep: Term.Repeated) = syntaxError("repeated argument not allowed here", at = rep)
+      def checkArgs(args: List[Term]): Unit = args.foreach {
+        case Term.Assign(_, rep: Term.Repeated) => badRep(rep)
+        case rep: Term.Repeated => badRep(rep)
+        case _ =>
```
tgodzik, line 2192:
```suggestion
      def checkArgs(args: List[Term]): Option[Term] = args.find {
        case Term.Assign(_, rep: Term.Repeated) => true
        case rep: Term.Repeated => true
        case _ => false
```

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2185,26 +2185,23 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
       }
       val openParenPos = in.tokenPos
       val args = argumentExprs()
       val closeParenPos = in.prevTokenPos
+      @inline def addPos(body: Term): Term = atPosWithBody(openParenPos, body, closeParenPos)
       token match {
         case Dot() | LeftBracket() | LeftParen() | LeftBrace() | Underscore() =>
-          simpleExprRest(argsToTerm(args, openParenPos, closeParenPos), canApply = true) :: Nil
+          checkArgs(args)
+          simpleExprRest(addPos(makeTupleTerm(args)), canApply = true) :: Nil
```
tgodzik, line 2201: and then
```suggestion
        checkArgs(args) match {
           case Some(t) => syntaxError("repeated argument not allowed here", at = t)
           case _ => simpleExprRest(addPos(makeTupleTerm(args)), canApply = true) :: Nil
        } 
```
?

tgodzik (changes requested):
Just a small comment.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: done, using collectFirst

tgodzik (approved):
LGTM!

PR #2601, opened 2022-02-16
ScalametaParser: wrap in AnonymousFunction if need

### LINKED ISSUE #1843: Information lost on some Term.ApplyInfix use cases
```scala
val tree1 = q"list foo (_ fun (_.bar))"
val tree2 = q"list foo (_ fun _.bar)"

tree1.structure == tree2.structure // true

// Both have this structure:
Term.ApplyInfix(
  Term.Name("list"),
  Term.Name("foo"),
  Nil,
  List(Term.ApplyInfix(
    Term.Placeholder(),
    Term.Name("fun"),
    Nil,
    List(Term.Select(Term.Placeholder(), Term.Name("bar")))
  ))
)
```

These are different trees, so the structure should have all the information to construct the original syntax.

kitbellew:
Fixes #1843.


dos65 (approved):
:+1:

tgodzik (approved):
:rocket:

PR #2602, opened 2022-02-16
Tranversers: split main apply match into several

kitbellew:
The current size of the Transformer.apply method is nearing the maximum allowed by JVM (64kb); in fact, adding one more simple Tree leads to build failures.

Therefore, create separate methods for Term, Type and Defn trees (these contain the largest -- and growing -- collections of trees), and split the primary apply().


kitbellew:
see #2601 build failures to explain the reason for this commit.

Issue #2561, opened 2022-01-19
Corruption of SemanticDB files


wchau:
It occurs around 1% of the time against our Scala codebase and is rather flaky. Rerunning it seems to fix it, so it's not 100% reproducible either.

SemanticDB version: 4.4.30
Scala versions: 2.11, 2.12, 2.13

Found while running lsif-semanticdb:

Generating LSIF... [######## ] 89% 39,861 files processedcom.google.protobuf.InvalidProtocolBufferException$InvalidWireTypeException: Protocol message tag had invalid wire type.
processedcom.google.protobuf.InvalidProtocolBufferException: Protocol message contained an invalid tag (zero).

tgodzik:
Thanks for reporting! Could this be an issue with LSIF for Scala? I have never seen this happen in any of the workspaces I worked with.

Maybe it's an issue when we're trying to read the file before it's finished writing?

@olafurpg any idea what might be going on?

olafurpg:
@wchau thank you for reporting! Can you upload the payload that’s failing to parse? This is the first time somebody reports this issue and SemanticDB sees quite heavy usage.

wchau:
@olafurpg I'll see if I can reproduce it in an OSS environment so that I can share the payload.

olafurpg:
Would you be OK with privately sharing the payload with me? I’m happy to troubleshoot this issue with you in the sourcegraph slack (I can join the channel on Monday). It might be tricky to reproduce in an OSS environment if the issue is rooted in your bespoke setup.

tgodzik:
Is this still an issue?

PR #2605, opened 2022-02-17
Use `java.nio.file.Files.write` to write SemanticDB files

olafurpg:
Related to https://github.com/scalameta/scalameta/issues/2561

Previously, the SemanticDB compiler plugin streamed bytes into the file
increasing the risk that the plugin would be interrupted mid-way. Now,
the plugin writes the payload into a byte array and then writes that
byte array to disk.

There's still a small risk that `Files.write` gets interrupted. We could
minimize that risk even further by writing to a temporary file first and
then moving that file. We can consider using that approach if the
problem persists, this change along should already significantly
minimize the risk of half-complete SemanticDB files.


tgodzik (approved):
LGTM!

This would explain the issue I've seen where the semanticdb was create without any content at first.

olafurpg:
@tgodzik thank you! Do you have any plans to cut a new release soon?

kitbellew:
released 4.5.0.

PR #2628, opened 2022-02-21
Token macros: create common abstract base class

kitbellew:
Also:
- simplify the root macro, matching only the needed fields
- define the branch macro, by copying root, removing custom parents and
  replacing root annotations with branch annotations.


tgodzik (approved):
Looks good! As far as I can see this mostly just move the code a bit?

kitbellew:
> Looks good! As far as I can see this mostly just move the code a bit?

yes, pretty much. and copy.

PR #2629, opened 2022-02-21
ScalametaParser: tokenSeparated no extra checks

kitbellew:
Currently, the token is checked multiple times in the loop. For example, if the next token is Ellipsis, 2 checks in the while condition (one for Sep and another for Ellipsis), then 2 more checks in the body, 4 total. If the next token is Sep, it will result in 3 checks.

Change to a tail-recursive implementation with 1 check for Ellipsis and 2 or 3 for Sep.


tgodzik (approved):
LGTM

PR #2630, opened 2022-02-21
ScalametaParser: remove broken branch in statSeq

kitbellew:
If we are calling a partial function on an argument, it will throw an exception unless it is defined for that argument.

Therefore, there's no need for the Ellipsis check as it will result in a failure anyway; let's issue a proper syntax error instead.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -4147,10 +4147,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     val isIndented = acceptOpt[Indentation.Indent]
 
     while (!token.is[StatSeqEnd]) {
-      def isDefinedInEllipsis = {
-        if (token.is[LeftParen] || token.is[LeftBrace]) next(); statpf.isDefinedAt(token)
-      }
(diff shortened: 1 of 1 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -4147,10 +4147,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     val isIndented = acceptOpt[Indentation.Indent]
 
     while (!token.is[StatSeqEnd]) {
-      def isDefinedInEllipsis = {
-        if (token.is[LeftParen] || token.is[LeftBrace]) next(); statpf.isDefinedAt(token)
-      }
-      if (statpf.isDefinedAt(token) || (token.is[Ellipsis] && ahead(isDefinedInEllipsis)))
```
tgodzik, line 4153: Do we have a test case for it? When does the Ellipsis token is actually used?

kitbellew: this code is from 5-6 years ago. it's possible that this execution path was never executed. also, it's possible that whatever `statpf` that is being passed here also takes care of ellipses.

kitbellew: tracked down all partial functions passed here:
- `consumeStat`
- `templateStat`
- `topStat`

They all handle Ellipsis.

tgodzik (approved):
Thanks for the explanation!

PR #2634, opened 2022-02-23
Token: add Keyword/Literal/Trivia etc branches

kitbellew:


scalameta/tokens/shared/src/main/scala/scala/meta/tokens/Token.scala
```diff
@@ -132,12 +154,12 @@ object Token {
...
   // Trivia
-  @fixed(" ") class Space extends Token
-  @fixed("\t") class Tab extends Token
-  @fixed("\r") class CR extends Token
-  @fixed("\n") class LF extends Token
-  @fixed("\f") class FF extends Token
-  @freeform("comment") class Comment(value: String) extends Token
+  @fixed(" ") class Space extends HSpace
+  @fixed("\t") class Tab extends HSpace
+  @fixed("\r") class CR extends AtEOL
```
tgodzik, line 159: Why `AtEOL` ? Not sure what does this category mean?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/TokenClassifiers.scala
```diff
@@ -5,64 +5,6 @@ import scala.meta.internal.classifiers.classifier
...
-@classifier
-trait Whitespace {
-  def unapply(token: Token): Boolean = {
-    token.is[Space] || token.is[Tab] ||
-    token.is[LineEnd] || token.is[FF]
-  }
-}
-
-@classifier
-trait LineEnd {
```
kitbellew, line 53: this is what I called `AtEOL`, it's used in a couple of places around the codebase.

scalameta/tokens/shared/src/main/scala/scala/meta/tokens/Token.scala
kitbellew: see comment above. `CR` and `LFLF`, which are part of it, are not _really_ EOL themselves, but they are indicators that an end of line is nearby :)

PR #2636, opened 2022-02-23
ScalametaParser: tailrec recursion for batchSource

kitbellew:


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -4361,37 +4369,46 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+        f(buf.toList)
+      case _ =>
+        statSeqBuf(buf, statpf)
+        f(buf.toList)
     }
     if (inBracelessPackage()) {
       val startPos = in.tokenPos
       next()
-      Source(List(autoEndPos(startPos)(Pkg(qualId(), bracelessPackageStats()))))
+      Source(List(autoEndPos(startPos)(Pkg(qualId(), bracelessPackageStats(identity)))))
```
tgodzik, line 4408: Why do we need the f parameter if it's always indentity? (As far as I can tell)

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -4361,37 +4369,46 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+          case _: LeftBrace =>
+            inPackage(inBraces(statSeq(statpf)))
+            bracelessPackageStats(f)
           case t if isColonEol(t) =>
             next()
             in.observeIndented()
             inPackage(indented(statSeq(statpf)))
-          case _ => List(autoEndPos(startPos)(Pkg(qid, bracelessPackageStats())))
+            bracelessPackageStats(f)
+          case _ => bracelessPackageStats(x => f(List(autoEndPos(startPos)(Pkg(qid, x)))))
```
kitbellew, line 4396: it's not identity here

tgodzik: I haven't noticed this one. Wouldn't it be simpler to just  do:
```
val stats = bracelessPackageStats()
List(autoEndPos(startPos)(Pkg(qid, stats))
```
to avoid passing the f function?

kitbellew: then you can't do `@tailrec` :)

tgodzik: ach, that's because of the tail recursion.

PR #2637, opened 2022-02-23
ScannerTokens: use case matching when possible

kitbellew:


dos65 (commented):
@kitbellew 
How much does it improve performance?
I'm wondering if we need to look if we can do any optimizations for `def is` and `Classifier` if they add a significant overhead.

kitbellew:
> @kitbellew How much does it improve performance? I'm wondering if we need to look if we can do any optimizations for `def is` and `Classifier` if they add a significant overhead.

i don't know how to measure this, to be honest. my guess would be not too much, even with repetitive accesses to `token` which call `in.token` which calls something else.

mostly, i find match a bit more readable and structured (personal preference), and i like consistency :)

PR #2661, opened 2022-03-02
ScalametaParser: don't need to strip backquotes

kitbellew:
LegacyScanner does not include backquotes in the `value` field of an Ident token, hence we don't need to do that.


tgodzik (approved):
:rocket:

PR #2662, opened 2022-03-02
LegacyScanner: don't use BACKQUOTED_IDENT

kitbellew:
Its handling is identical to IDENTIFIER; use a simple boolean in the one place where there's a difference.

### DIFF
--- a/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala
+++ b/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala
@@ -118,8 +118,8 @@ class LegacyScanner(input: Input, dialect: Dialect) {
   protected def emitIdentifierDeprecationWarnings = true
 
   /** Clear buffer and set name and token */
-  private def finishNamed(idtoken: LegacyToken = IDENTIFIER): Unit = {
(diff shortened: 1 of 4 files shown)


scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala
```diff
@@ -642,7 +642,7 @@ class LegacyScanner(input: Input, dialect: Dialect) {
         putChar(ch)
         nextRawChar()
       } while (ch != SU && Character.isUnicodeIdentifierPart(ch))
-      next.setIdentifier(cbuf, IDENTIFIER, dialect) { x =>
+      next.setIdentifier(cbuf, true, dialect) { x =>
```
tgodzik, line 645:
```suggestion
      next.setIdentifier(cbuf, isBackquoted = true, dialect) { x =>
```

tgodzik (approved):
LGTM! Just a minor comment

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala
kitbellew: thank you, used the suggestion.

PR #2663, opened 2022-03-02
LegacyScanner: refactor getIdentRest

kitbellew:
Also inline into it the no longer needed getIdentOrOperatorRest.

### DIFF
--- a/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala
+++ b/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala
@@ -516,30 +516,23 @@ class LegacyScanner(input: Input, dialect: Dialect) {
   }
 
   @tailrec
-  private def getIdentRest(): Unit = (ch: @switch) match {
-    case 'A' | 'B' | 'C' | 'D' | 'E' | 'F' | 'G' | 'H' | 'I' | 'J' | 'K' | 'L' | 'M' | 'N' | 'O' |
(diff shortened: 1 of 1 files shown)


scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala
```diff
@@ -536,30 +536,20 @@ class LegacyScanner(input: Input, dialect: Dialect) {
...
       putChar(ch)
       nextChar()
-      getIdentRest()
-    case '_' =>
+      if (isIdentifierPart(ch))
+        getIdentRest()
+      else
+        getOperatorRest()
+    } else if ( // strangely enough, Character.isUnicodeIdentifierPart(SU) returns true!
+      ch != SU && (if (ch == '$') getDollar() else Character.isUnicodeIdentifierPart(ch))) {
```
tgodzik, line 548: Could you extract the condition after `&&` to a def? It seems a bit hard to understand currently.

kitbellew: would it be easier to read if i change it to a match? SU vs $ vs rest?

tgodzik: Might be, but why not change it to a def ?

kitbellew: mostly because i don't know what to call this `def` :) any suggestions?

tgodzik (approved):
Looks better now, thanks!

PR #2669, opened 2022-03-05
Dialect: centralize creating unquote dialect

kitbellew:


scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -610,6 +610,12 @@ final class Dialect private (
       toplevelSeparator
     )
   }
+
+  def unquote(): Dialect = privateCopy(
```
tgodzik, line 614:
```suggestion
  def unquoteVariant(): Dialect = privateCopy(
```
?

tgodzik (approved):
I think we could rename it a bit, but otherwise LGTM

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
kitbellew: done

Issue #2315, opened 2021-04-27
[Scala 3] Enable infix type precedence


tgodzik:
Based on input from @smarter :
>To parse Scala 3 correctly there's another subtle difference that needs to be handled: infix type applications are parsed following precedence rules, like infix term applications, for example:
```scala
class ^[A, B]
trait Test {
  def foo: String Map Int ^ Double
}
```
>In Scala 2, the type of foo is (String Map Int) ^ Double aka ^[Map[String, Int], Double] but in Scala 3 it's String Map (Int ^ Double) aka Map[String, ^[Int, Double]], there's an unmerged PR implementing this for Scala 2: scala/scala#6147

This means we would need to apply similar mechanism that we use for terms to types. Something akin to:

https://github.com/scalameta/scalameta/blob/6c0a370cc278d473588f973c436d8378a691d2b1/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L2695

It should be possibly easier than normal terms, but will not trivial at all.

PR #2671, opened 2022-03-08
Trees: turn Type.{And, Or} into infix-like trees

kitbellew:
Otherwise we can't apply the complex infix formatting to them as the `|` and `&` tokens are not represented as separate children within the tree.

Needed by scalameta/scalafmt#3153.

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
@@ -9,6 +9,7 @@ import scala.collection.immutable.TreeMap
  */
 final class Dialect private (
     // Are `&` intersection types supported by this dialect?
+    @deprecated("allowAndTypes unneeded, infix types are supported", "4.5.1")
     val allowAndTypes: Boolean,
     // Are extractor varargs specified using ats, i.e. is `case Extractor(xs @ _*)` legal or not?
     val allowAtForExtractorVarargs: Boolean,
@@ -33,6 +34,7 @@ final class Dialect private (
     // Some quasiquotes only support single-line snippets.
     val allowMultilinePrograms: Boolean,
     // Are `|` (union types) supported by this dialect?
+    @deprecated("allowOrTypes unneeded, infix types are supported", "4.5.1")
     val allowOrTypes: Boolean,
     // Are unquotes ($x) and splices (..$xs, ...$xss) allowed?
     // If yes, they will be parsed as patterns.
@@ -242,6 +244,7 @@ final class Dialect private (
   // Are unquotes ($x) and splices (..$xs, ...$xss) allowed?
   def allowUnquotes: Boolean = allowTermUnquotes || allowPatUnquotes
 
+  @deprecated("allowAndTypes unneeded, infix types are supported", "4.5.1")
   def withAllowAndTypes(newValue: Boolean): Dialect = {
     privateCopy(allowAndTypes = newValue)
   }
@@ -274,6 +277,7 @@ final class Dialect private (
   def withAllowMultilinePrograms(newValue: Boolean): Dialect = {
     privateCopy(allowMultilinePrograms = newValue)
   }
+  @deprecated("allowOrTypes unneeded, infix types are supported", "4.5.1")
   def withAllowOrTypes(newValue: Boolean): Dialect = {
     privateCopy(allowOrTypes = newValue)
   }
@@ -620,6 +624,7 @@ final class Dialect private (
 
 object Dialect extends InternalDialect {
   def apply(
+      @deprecated("allowAndTypes unneeded, infix types are supported", "4.5.1")
       allowAndTypes: Boolean,
       allowAtForExtractorVarargs: Boolean,
       allowCaseClassWithoutParameterList: Boolean,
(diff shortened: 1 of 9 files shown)


tgodzik (commented):
This kind of looks like duplicated information. We already would know that And uses `&` operator.

This also break compatibility again and since this proved to cause issues in scalafix I would rather avoid it.

kitbellew:
a complete different question: Term.ApplyInfix contains a list arguments for the right-hand side; what does it look like when this list contains multiple items? I am curious why Type infix contains Type on the right, while Term or Pat infix contain a list (instead of also a single, via, say, a Tuple).

tgodzik:
That seems to have been done a while ago: https://github.com/scalameta/scalameta/pull/506 and https://github.com/scalameta/scalameta/pull/504

Maybe we could change all occurrences of And/Or to ApplyInfix ? We could then deprecate And and Or in that case.

What do you think?

tgodzik:
> a complete different question: Term.ApplyInfix contains a list arguments for the right-hand side; what does it look like when this list contains multiple items? I am curious why Type infix contains Type on the right, while Term or Pat infix contain a list (instead of also a single, via, say, a Tuple).

Huh, it seems that list always contains a single param. My guess is that it was added at a different time. We could remove it for `5.0.0`, where we would rework a lot of other things.

kitbellew:
> That seems to have been done a while ago: #506 and #504
> 
> Maybe we could change all occurrences of And/Or to ApplyInfix ? We could then deprecate And and Or in that case.
> 
> What do you think?

I am _very_ good with this approach! but then the question: what is the purpose of the `allowAndTypes` dialect flag?

tgodzik:
> > That seems to have been done a while ago: #506 and #504
> > Maybe we could change all occurrences of And/Or to ApplyInfix ? We could then deprecate And and Or in that case.
> > What do you think?
> 
> I am _very_ good with this approach! but then the question: what is the purpose of the `allowAndTypes` dialect flag?

Wait, in this case if we don't allowAndTypes aren't they just treated as infix? Then they actually might not have a purpose :O

tgodzik:
Just checked, yep they do not serve any purpose.

tgodzik:
One possibility might be precedence: ` A | B & C | D` - related to https://github.com/scalameta/scalameta/issues/2315

tgodzik:
On the other hand, do we support precendence currently? Maybe if we solve 2315 then it also solves this situation.

kitbellew:
> On the other hand, do we support precendence currently? Maybe if we solve 2315 then it also solves this situation.

looks like the question about list and type on the right is once again relevant :)

type infixes are parsed here, and likely simplistically:
https://github.com/scalameta/scalameta/blob/main/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L838

whereas term and pat infixes are parsed here (and they do handle precedence):
https://github.com/scalameta/scalameta/blob/main/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L1859

tgodzik:
> looks like the question about list and type on the right is once again relevant :)

Even with precendence why would we need a list here though? There no example in tests that does that for Term.ApplyInfix

kitbellew:
@tgodzik removed the trees, deprecated the flags, build passes. PTAL.

project/Mima.scala
```diff
@@ -27,33 +27,35 @@ object Mima {
...
-    ProblemFilters.exclude[DirectMissingMethodProblem]("scala.meta.Dialect.allowLiteralUnitType"),
-    ProblemFilters
-      .exclude[DirectMissingMethodProblem]("scala.meta.Dialect.withAllowLiteralUnitType"),
-    ProblemFilters.exclude[DirectMissingMethodProblem]("scala.meta.Dialect.allowSuperTrait"),
-    ProblemFilters.exclude[DirectMissingMethodProblem]("scala.meta.Dialect.withAllowSuperTrait"),
-    ProblemFilters.exclude[DirectMissingMethodProblem]("scala.meta.Dialect.allowAsPatternBinding"),
-    ProblemFilters
-      .exclude[DirectMissingMethodProblem]("scala.meta.Dialect.withAllowAsPatternBinding"),
-    ProblemFilters.exclude[DirectMissingMethodProblem]("scala.meta.Dialect.this")
+    ProblemFilters.exclude[MissingClassProblem]("scala.meta.Type$And"),
```
tgodzik, line 30: Could we leave the trees for now? Just add the deprecated message next to them. We can remove them later.

kitbellew: done

olafurpg (approved):
Are and/or expressions truly infix? Is there no workaround on the Scalafmt side to avoid this change? I’m ok merging this.

tgodzik:
> Are and/or expressions truly infix? Is there no workaround on the Scalafmt side to avoid this change? I’m ok merging this.

I think there is actually no difference between `|` / `&` and normal infix types. It looks like it's a syntactic sugar in essence.

There are not represented differently in the compile
`Foo | Bar[Files]` is shown as:

```scala
tpt = AppliedTypeTree(
              tpt = Ident(name = |),
              args = List(
                Ident(name = Foo),
                AppliedTypeTree(tpt = Ident(name = Bar), args = List(Ident(name = Files)))
              )
            )
```

olafurpg:
> There are not represented differently in the compiler

This change sounds good to me in that case 👍

PR #2676, opened 2022-03-14
Dialect: add useInfixTypePrecedence flag in scala3

### LINKED ISSUE #2315: [Scala 3] Enable infix type precedence

kitbellew:
Fixes #2315.

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
@@ -148,7 +148,9 @@ final class Dialect private (
      */
     val allowPlusMinusUnderscoreAsPlaceholder: Boolean,
     // import a.b.c.{ given, _} used for -X:source3
-    val allowGivenImports: Boolean
+    val allowGivenImports: Boolean,
+    // Scala 3 uses proper precedence rules for infix types, unlike Scala 2
+    val useInfixTypePrecedence: Boolean
 ) extends Product with Serializable {
 
   // NOTE(olafur) checklist for adding a new dialect field in a binary compatible way:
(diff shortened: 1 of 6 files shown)


tgodzik (approved):
LGTM! Great work! I was a bit afraid to delve into this one :sweat_smile:

kitbellew:
> LGTM! Great work! I was a bit afraid to delve into this one 😅

turned out to be quite simple. largely a copy of the method just above, but it did let me look into his pattern infix was implemented.

(and i found a minor bug with pattern while looking, will send you soon.)

p.s. all quiet on the western front? are people feeling safe in polska?

tgodzik:
> p.s. all quiet on the western front? are people feeling safe in polska?

I think a few days were scary for people in Poland, but now it's mostly just hectic trying to make sure that people are being helped.

kitbellew:
> > p.s. all quiet on the western front? are people feeling safe in polska?
> 
> I think a few days were scary for people in Poland, but now it's mostly just hectic trying to make sure that people are being helped.

are you talking about ukrainian refugees? are they finding their way into poland?

tgodzik:
> > > p.s. all quiet on the western front? are people feeling safe in polska?
> > 
> > 
> > I think a few days were scary for people in Poland, but now it's mostly just hectic trying to make sure that people are being helped.
> 
> are you talking about ukrainian refugees? are they finding their way into poland?

yes, there is almost 2 milion people who came through the polish boarder alone and some are staying while other are moving to other european countries. All of them are quite exhausted and need help.

PR #2680, opened 2022-03-16
ScalametaParser: recognize Pat.Tuple as infix rhs

kitbellew:


tgodzik (approved):
Looks good, just a single question from my end.

PR #2684, opened 2022-03-18
ScalametaParser: use tuple for interim infix rhs

kitbellew:
Uses for `Term.ApplyInfix` the approach we took for `Pat.ExtractInfix`. Simplifies the code and sets positions for the RHS correctly.


tgodzik (approved):
LGTM

PR #2681, opened 2022-03-17
Support MatchType for Scala3

tanishiking:
Extract match_type part from https://github.com/scalameta/scalameta/pull/2414

Now it's time to merge at least for `MatchType` as https://github.com/lampepfl/dotty/pull/14608 is ready to merge :)

Let me work on metap implementation for `MatchType` in separate PR.

PR #2690, opened 2022-03-21
Support MatchType in metacp

tanishiking:
Follow-up for https://github.com/scalameta/scalameta/pull/2681

This commit changes anything, as there are no MatchType included test files, and we don't invoke compilation using scala3 in the scalameta test at this moment :)


tgodzik (approved):
LGTM!

PR #2325, opened 2021-04-30
Always add () for constructor invocations when printing

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/2317

PR #2694, opened 2022-03-27
TreeSyntax: handle Init argss explicitly

### LINKED ISSUE #1917: Pretty-printer bug for constructor lambda param
Given the following class:
```scala
class Foo(f: Int => Int)
```

For this tree:
```scala
val tree = q"new Foo((a: Int) => a + 1)"
```

This is the `.syntax`:
```scala
tree.syntax
// res2: String = "new Foo { (a: Int) => a + 1 }"
```

However, the result does not compile:
```scala
// Compiles
new Foo((a: Int) => a + 1)

// Does not compile
new Foo { (a: Int) => a + 1 }
// not enough arguments for constructor Foo: (f: Int => Int)
// Unspecified value parameter f
```

### LINKED ISSUE #2029: Transform adds additional parentheses
Hello,

```
val program = """object A {
                 |  val b = Some(2).map("key" -> _.toString)
                 |  val c = 0
                 |}
                 |""".stripMargin
val res = program.parse[Source].get.transform({ case q"0" => q"3" })
```

**produces**

```
object A {
  val b = Some(2).map("key" -> (_.toString))
  val c = 3
}
```

There are added `()` around `_.toString`. **This new syntax doesn't compile.**

---
In the case the transform didn't match on the non-related part, there are no added parentheses:

```
val program = """object A {
                 |  val b = Some(2).map("key" -> _.toString)
                 |  val c = 0
                 |}
                 |""".stripMargin
val res = program.parse[Source].get.transform({ case q"1" => q"3" })
```

**produces**

```
object A {
  val b = Some(2).map("key" -> _.toString)
  val c = 0
}
```

which is the expected result.

Thank you !

kitbellew:
Fixes #1917. Follow on to #2325.

### DIFF
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
@@ -1018,7 +1018,11 @@ object TreeSyntax {
           )
 
       // Init
-      case t: Init => s(if (t.tpe.is[Type.Singleton]) kw("this") else p(RefineTyp, t.tpe), t.argss)
+      case t: Init =>
+        s(
+          if (t.tpe.is[Type.Singleton]) kw("this") else p(RefineTyp, t.tpe),
+          r(t.argss.map(x => s("(", r(x, ", "), ")")))
+        )
 
       // Self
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM!

PR #2697, opened 2022-03-28
ScalametaParser: try parse self in template first

### LINKED ISSUE #1776: Parser fails when inline keyword starts template
Copy-pasted from #1038 (since Sam doesn't want to have pending open issues).

The parser fails if a `Template` starts with `inline`, for example:

```scala
object a {
  inline def foo = 2
}
```

It parses fine, however, if `inline` is not at the beginning of the template body

```scala
object a {
  val dummy = 2
  inline def foo = 2
}
```

kitbellew:
Instead of trying to get a statement and then pivoting to a self, do the opposite. That way, parsing for first and later statements will be consistent. Fixes #1776 and similar.


dos65 (approved):
LGTM!

PR #2698, opened 2022-03-29
ScalametaParser: allow repeated in `apply { ... }`

### LINKED ISSUE #1063: "repeated argument not allowed here" false negative parse error
```
/home/geirsson/frontend/common/app/views/support/package.scala:87: error: repeated argument not allowed here
def withJsoup(html: Html)(cleaners: HtmlCleaner*): Html = withJsoup(html.body) { cleaners: _* }
                                                                                             ^
```

kitbellew:
Fixes #1063.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2216,7 +2216,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
   def argumentExprsWithUsing(location: Location = NoStat): (List[Term], Boolean) = token match {
     case LeftBrace() =>
-      (List(blockExpr()), false)
+      (List(blockExpr(allowRepeated = true)), false)
     case LeftParen() =>
       inParensOnOpen(token match {
         case RightParen() =>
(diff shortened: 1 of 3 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -4202,7 +4202,17 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
 
-  def blockStatSeq(): List[Stat] = listBy[Stat] { stats =>
+  def blockStatSeq(allowRepeated: Boolean = false): List[Stat] = listBy[Stat] { buf =>
+    object stats {
+      final def +=(elem: Stat): Unit = {
+        if (allowRepeated)
+          buf.lastOption match { // only last one can be repeated
+            case Some(t: Term.Repeated) => syntaxError("repeated argument not allowed here", at = t)
+            case _ =>
+          }
```
dos65, line 4212: It seems that in case of repeated in braces compiler allow only a single statement.
The following won't compile:
```
scala> def bar(a: Int*): Unit = println(s"ooo ${a}")
def bar(a: Int*): Unit

scala> bar {
     |   val a = List(1,2,3)
     |   a: _*
     | }
         a: _*
          ^
On line 4: error: no `: _*` annotation allowed here
```

kitbellew: done, PTAL

dos65 (approved):
:+1:

PR #2702, opened 2022-04-01
TreeSyntax: refactor printing Term.Param

kitbellew:
Similar logic is used for regular params and for the special case of constructor params. Let's extract a method and reuse.


tanishiking (approved):
LGTM:+1:

dos65 (approved):
:+1:

PR #2709, opened 2022-04-03
ScalametaParser: keep `(())` as `Lit.Unit`

### LINKED ISSUE #2708: Infix parsing with parens/Unit
Previously (as of version 4.5.0) ```scala
res == ()
``` 
parsed as `a.==()`, i.e with no arguments. That is not how Scastie 3.1.1 parses it, since this returns true:
```scala
val res = ()
res == ()
```

However, as of 4.5.1 (and 4.5.2), even `res == (())` parses as `res.==()`, which I think is definitely a bug.

kitbellew:
Fixes #2708.


dos65 (approved):
LGTM!

Issue #1384, opened 2018-03-01
Lit.String and Lit.Char format is not preserved


MasseGuillaume:
Obtained:

```scala
import scala.meta._
"\"\\n\"".parse[Term].get.asInstanceOf[Lit.String].value
res31: String =
"
"
"\'\\n\'".parse[Term].get.asInstanceOf[Lit.Char].value
res33: Char =


```

Expected:

```scala
import scala.meta._
"\"\\n\"".parse[Term].get.asInstanceOf[Lit.String].value == "\\n"
"\'\\n\'".parse[Term].get.asInstanceOf[Lit.Char].value == "\\n"
```

olafurpg:
I am unable to reproduce when reading from a file
```
$ cat string.txt
"\\n"
$ cat char.txt
'\n'
$ scala
> Input.File(new java.io.File("string.txt")).parse[Term].get.asInstanceOf[Lit.String].value)
res14: String = "\\n"
@ Input.File(new java.io.File("char.txt")).parse[Term].get.asInstanceOf[Lit.Char].value
res3: Char = '\n'
```

The repl is not always the best place to do these experiments because of how scalac processes literals.

MasseGuillaume:
Failing test case:
https://github.com/MasseGuillaume/scalameta/commit/d8686a6f6448c6ebe7b1457ddf4dedb6cc4e3ecc

MasseGuillaume:
@olafurpg you are right the information when you create the input. I think to avoid this, we should provide smarter constructors for Input.

For example, Input.String and Input.VirtualFile can be a macro and extract the string literal.

MasseGuillaume:
This one is incorrect:

```scala
"""\u005Cu0041"""
```

```scala
Input.File(new java.io.File("string.txt")).parse[Term].get.asInstanceOf[Lit.String].value
// result: \u0041
```

PR #2711, opened 2022-04-03
TreeSyntax: handle single trailing NL with dquotes

### LINKED ISSUE #2695: newline syntax is not optimal
Hi all

We have a piece of code that does some scala code generation. Now, at some point, I had a string that is just `"\n"` and I wanted to generate a code that looks consistent across all values of this string. 

As an example, let's say the string is supposed to be a delimiter and you're supposed to generate a splitter function. Something as simple as below:

```scala
object Test extends App {
  import scala.meta._
  import scala.meta.inputs._
  import scala.meta.inputs.Input._
  import scala.meta.internal.trees.Origin
  val delimiters = List("\n", "\t")
  for (delimiter <- delimiters) {
    val code =
      q"""
        def split(s: String) = s.split(${Lit.String(delimiter)})
       """
    println(s"\n----\n$code\n---\n")
  }
}
```

Now, the code that gets printed for `\t` looks fine

```scala
def split(s: String) = s.split("\t")
```

But for `\n`, ScalaMeta has a special behavior

```scala
def split(s: String) = s.split("""
""")
```

What I want, instead is this:
```scala
def split(s: String) = s.split("\n")
```

I spent some time going through the scala meta codebase and found that the reason is [this line](https://github.com/scalameta/scalameta/blob/main/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala#L846). The behavior seems to be hard-coded in scala-meta. 

I believe there might have been a good reason to do this, but this ends up generating dirty code, especially in the case where the string just contains a single character of newline. This behavior could have been kept restricted to when there is more than one occurrence of the newline character. 

## Potential Solutions

One potential solution I see is to define my own syntax function (`syntax2`), the implicit classes/methods around it, and call that. There, I am a little blocked because the [class is declared as final](https://github.com/scalameta/scalameta/blob/main/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala#L81). I could of-course copy the whole class, but I'd rather want to avoid that, and go that route only if I don't have any other options. 

Another potential solution would be if there were some flags in ScalaMeta pretty-printing where I can control this behavior. There doesn't seem to be such a flag available. 

Another potential solution could be to enhance this behavior, let it go in the next release, and then depend on the latest version of ScalaMeta.

kitbellew:
Fixes #2695.

### DIFF
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
@@ -822,10 +822,9 @@ object TreeSyntax {
       case Lit.Char(value) => m(Literal, s(enquote(value.toString, SingleQuotes)))
       // Strings should be triple-quoted regardless of what newline style is used.
       case Lit.String(value) =>
-        m(
-          Literal,
-          s(enquote(value.toString, if (value.contains("\n")) TripleQuotes else DoubleQuotes))
-        )
+        val firstNL = value.indexOf('\n') + 1
+        val style = if (firstNL > 0 && firstNL < value.length) TripleQuotes else DoubleQuotes
+        m(Literal, s(enquote(value, style)))
       case Lit.Symbol(value) => m(Literal, s("'", value.name))
       case Lit.Null() => m(Literal, s(kw("null")))
       case Lit.Unit() => m(Literal, s("()"))
--- a/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
@@ -1320,16 +1320,14 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
   test("#2695 1") {
     checkTree(
       q"""s.split("\n")""",
-      s"""s.split(${"\"" * 3}
-         |${"\"" * 3})""".stripMargin
+      """s.split("\n")""".stripMargin
     )
   }
 
   test("#2695 2") {
     checkTree(
       q"""s.split("foo\n")""",
-      s"""s.split(${"\"" * 3}foo
-         |${"\"" * 3})""".stripMargin
+      """s.split("foo\n")""".stripMargin
     )
   }
 
--- a/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
@@ -1317,4 +1317,28 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
     )
   }
 
+  test("#2695 1") {
+    checkTree(
+      q"""s.split("\n")""",
+      s"""s.split(${"\"" * 3}
+         |${"\"" * 3})""".stripMargin
+    )
+  }
+
+  test("#2695 2") {
+    checkTree(
+      q"""s.split("foo\n")""",
+      s"""s.split(${"\"" * 3}foo
+         |${"\"" * 3})""".stripMargin
+    )
+  }
+
+  test("#2695 3") {
+    checkTree(
+      q"""s.split("\nfoo")""",
+      s"""s.split(${"\"" * 3}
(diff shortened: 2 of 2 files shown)


dos65 (commented):
@kitbellew  :thinking: I'm wondering if we can add a field to `Lit.String` to mark if it's multiline or not. I'm not sure if handling of trailing newline changes a lot

albertikm:
> @kitbellew 🤔 I'm wondering if we can add a field to `Lit.String` to mark if it's multiline or not.

How would that help?

> I'm not sure if handling of trailing newline changes a lot

it's a simplification in case of a string which is not really multi-line but does end in a newline.

kitbellew:
> > @kitbellew 🤔 I'm wondering if we can add a field to `Lit.String` to mark if it's multiline or not.
> 
> How would that help?
> 
> > I'm not sure if handling of trailing newline changes a lot
> 
> it's a simplification in case of a string which is not really multi-line but does end in a newline.

hmm, looks like i replied from my other account :)

dos65:
@kitbellew yea I got it, but I wouldn't say that it fixes this problem in general except of the case in reported issue.
It doesn't work for `s.split("Single\nLine\n\String")`.

kitbellew:
> @kitbellew yea I got it, but I wouldn't say that it fixes this problem in general except of the case in reported issue. It doesn't work for `s.split("Single\nLine\n\String")`.

> @kitbellew yea I got it, but I wouldn't say that it fixes this problem in general except of the case in reported issue. It doesn't work for `s.split("Single\nLine\n\String")`.

would you like it to? generally, i am not sure how much fanciness we should introduce.

also, this seems a bit similar to #1384.

tgodzik:
The proper solution would probably need having Literal.MultilineString, otherwise we are relying on heuristics.

However, since this is such a niche issue, maybe we could check origin and see if it was multiline string originally? In case of no origin (which would be in the issue) we could just do a normal string?

dos65:
@tgodzik I tried adding the `isSingleline` flag as `binaryCompatField` to `Lit.String`. However I couldn't make it work in tests yet. Should this approach work in theory?

kitbellew:
> The proper solution would probably need having Literal.MultilineString, otherwise we are relying on heuristics.
> 
> However, since this is such a niche issue, maybe we could check origin and see if it was multiline string originally? In case of no origin (which would be in the issue) we could just do a normal string?

I like this input-based idea. I will try on the weekend.

tgodzik:
> @tgodzik I tried adding the `isSingleline` flag as `binaryCompatField` to `Lit.String`. However I couldn't make it work in tests yet. Should this approach work in theory?

It could, but we don't usually add a boolean field in that case but a separate class. This would have to be `Lit.MultilineString` which would be a more wholesome solution.

kitbellew:
> > @tgodzik I tried adding the `isSingleline` flag as `binaryCompatField` to `Lit.String`. However I couldn't make it work in tests yet. Should this approach work in theory?
> 
> It could, but we don't usually add a boolean field in that case but a separate class. This would have to be `Lit.MultilineString` which would be a more wholesome solution.

How would this be used? A code parser would create different-class instances for `" ... "` and `""" ... """` strings? Would we need to be (at least, source) backwards-compatible and turn `Lit.String` into a trait with an `unapply` method so that current traversal/transformation code still works?

kitbellew:
> > The proper solution would probably need having Literal.MultilineString, otherwise we are relying on heuristics.
> > However, since this is such a niche issue, maybe we could check origin and see if it was multiline string originally? In case of no origin (which would be in the issue) we could just do a normal string?
> 
> I like this input-based idea. I will try on the weekend.

#2723 uses input for syntax.

tgodzik:
> > > @tgodzik I tried adding the `isSingleline` flag as `binaryCompatField` to `Lit.String`. However I couldn't make it work in tests yet. Should this approach work in theory?
> > 
> > 
> > It could, but we don't usually add a boolean field in that case but a separate class. This would have to be `Lit.MultilineString` which would be a more wholesome solution.
> 
> How would this be used? A code parser would create different-class instances for `" ... "` and `""" ... """` strings? Would we need to be (at least, source) backwards-compatible and turn `Lit.String` into a trait with an `unapply` method so that current traversal/transformation code still works?

IT would be tricky to make it compatible, so probably best to avoid it for now if possible.

kitbellew:
avoiding it.

PR #2718, opened 2022-04-07
PlaceholderChecks: include ApplyUnary

### LINKED ISSUE #2717: `RedundantBraces` in pattern match clause with parenthesis creates invalid Scala code
```
version = 3.5.1
runner.dialect = scala213
maxColumn = 120
docstrings.style = Asterisk
assumeStandardLibraryStripMargin = true
rewrite.rules = [RedundantBraces, RedundantParens, SortImports, SortModifiers]
indentOperator.exemptScope = all
align.tokens = [
  {code = "%", owner = "Term.ApplyInfix"},
  {code = "%%", owner = "Term.ApplyInfix"},
  {code = ":=", owner = "Term.ApplyInfix"},
  "="
]
```

NB: before submitting, please confirm that the problem is observed in the
*latest published* version of the formatter! We don't publish hotfixes for older
versions, and the problem you have observed in an older version may have already
been fixed.

When I run scalafmt via CLI like this: scalafmt-native:3.5.1

See https://github.com/lightbend/paradox/runs/5872565759?check_suite_focus=true for more info

## Steps

Given code like this:
```scala
xs span { !separates(_) } match {
  case (h, b) => Frontin(loadProperties(Some(h.mkString(""))), if (b.isEmpty) "" else b.tail.mkString(""))
}
```

## Problem

Scalafmt formats code like this:
```scala
xs span !separates(_) match {
  case (h, b) => Frontin(loadProperties(Some(h.mkString(""))), if (b.isEmpty) "" else b.tail.mkString(""))
}
```

i.e. removes the `{` and `}` around `!separates(_)`

Which causes the following Scala compiler error

```
[error] /home/runner/work/paradox/paradox/core/src/main/scala/com/lightbend/paradox/markdown/Frontin.scala:46:28: missing parameter type for expanded function
[error] The argument types of an anonymous function must be fully known. (SLS 8.5)
[error] Expected type was: com.lightbend.paradox.markdown.Frontin
[error]         xs span !separates(_) match {
```

## Expectation

I would like the formatted output to look like this:
```scala
xs span { !separates(_) } match {
  case (h, b) => Frontin(loadProperties(Some(h.mkString(""))), if (b.isEmpty) "" else b.tail.mkString(""))
}
```

## Workaround

Removing `RedundantBraces` from `rewrite.rule`'s solves the issue. I suspect `//format: off` would also work.

## Notes

The definition of `seperates` is as follows

```scala
def separates(str: String): Boolean =
  (str.trim == separator) && (str startsWith separator)
```

and the type of `xs` is a `List[String]`

Project issue where I encountered this bug can be found at https://github.com/lightbend/paradox/pull/517/files#r845364890

Scala compiler error can be found at https://github.com/lightbend/paradox/runs/5872565753?check_suite_focus=true

kitbellew:
Fixes #2717.

### DIFF
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/PlaceholderChecks.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/PlaceholderChecks.scala
@@ -30,6 +30,7 @@ object PlaceholderChecks {
           isBlockPlaceholder(t.args) || t.args.exists(isPlaceholder) || { queue += t.fun; iter }
         case t: Term.ApplyInfix =>
           isBlockPlaceholder(t.args) || { queue += t.lhs; queue ++= t.args; iter }
+        case t: Term.ApplyUnary => queue += t.arg; iter
         case t: Term.Repeated => queue += t.expr; iter
         case _: Term.AnonymousFunction => queue.nonEmpty && iter
         case t => t.children.exists(isPlaceholder) || queue.nonEmpty && iter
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM!

Issue #2727, opened 2022-04-12
Failure on previously formatted macro code


valencik:
Pulled out from https://github.com/scalameta/munit/pull/518#issuecomment-1096576523

## Configuration (required) ##
```
version = "3.5.1"
runner.dialect = scala3
```

## Command-line parameters (required) ##

When I run scalafmt via CLI like this: `-c .scalafmt.conf test.scala`

## Steps

Given code like this:
```scala
package munit.internal

import munit.Clue
import scala.quoted._
import scala.language.experimental.macros

object MacroCompat {

  trait ClueMacro {
    inline implicit def generate[T](value: T): Clue[T] = ${ clueImpl('value) }
    implicit def generate[T](value: T): Clue[T] = macro MacroCompatScala2.clueImpl
  }

}
```

## Problem

Scalafmt formats code like this:
```scala
package munit.internal

import munit.Clue
import scala.quoted._
import scala.language.experimental.macros

object MacroCompat {

  trait ClueMacro {
    inline implicit def generate[T](value: T): Clue[T] = ${ clueImpl('value) }
    implicit def generate[T](value: T): Clue[T] =
      macro MacroCompatScala2.clueImpl
  }

}
```
Note the `macro MacroCompatScala2.clueImpl` on a new line.

Once formatted, rerunning scalafmt yields the following error:
```
org.scalafmt.dynamic.exceptions.PositionExceptionImpl:
  /home/andrew/src/scalafmttest/test.scala:12: error: [dialect scala3] illegal start of statement
      macro MacroCompatScala2.clueImpl
      ^: /home/andrew/src/scalafmttest/test.scala
```

## Expectation

The output format should be able to parsed/formatted again

tgodzik:
Looks like an issue with macros themselves:
```
val rhs = {
        accept[Equals]
        isMacro = acceptOpt[KwMacro]
        exprMaybeIndented()
      }
```
Notice that we do not accept newline there and it seems like it was always like this.  Can a macro keyword be after a newline? If it can't then it's a bug in scalafmt, otherwise we should just accept newline here.

We should probably also not accept the macro as keyword for Scala 3 dialect, since it's no longer a keyword, though this isn't causing any large issues currently.

tgodzik:
Looks like we can have a newline before macro, so just `acceptOpt[LF]` should help.

PR #2728, opened 2022-04-14
Don't allow indentation for macro definitions

tgodzik:
From what I understand macros must exist as the only statement in a definition, which is way we can't have a block there and indentation should not be tried.

Fixes https://github.com/scalameta/scalameta/issues/2727


kitbellew (approved):
so it's not about allowing LF before macro, then?

tgodzik:
> so it's not about allowing LF before macro, then?

It's not, I forgot that we remove irrelevant LFs

Issue #2719, opened 2022-04-09
Backticks in for comprehensions trigger a failure


Martynas-P:
The following code fails to generate SemanticDB (the same code compiles fine without backticks.)

```scala
object App {
  def main(args: Array[String]): Unit = {
    for {
      i <- Seq(1, 2, 3)
      `j`: Seq[Int] = Seq(4, 5, 6, 7)
    } yield `j` :+ i
  }
}
```

Error

<details>
<summary>
scalac: failed to generate semanticdb for App.scala: App.scala:5: error: <- expected but : found
</summary>

```
scalac: failed to generate semanticdb for /home/martynas/projects/scalameta-bug-reproduce/src/main/scala/App.scala:
/home/martynas/projects/scalameta-bug-reproduce/src/main/scala/App.scala:5: error: <- expected but : found
      `j`: Seq[Int] = Seq(4, 5, 6, 7)
         ^
	at scala.meta.internal.parsers.Reporter.syntaxError(Reporter.scala:16)
	at scala.meta.internal.parsers.Reporter.syntaxError$(Reporter.scala:16)
	at scala.meta.internal.parsers.Reporter$$anon$1.syntaxError(Reporter.scala:22)
	at scala.meta.internal.parsers.Reporter.syntaxError(Reporter.scala:17)
	at scala.meta.internal.parsers.Reporter.syntaxError$(Reporter.scala:17)
	at scala.meta.internal.parsers.Reporter$$anon$1.syntaxError(Reporter.scala:22)
	at scala.meta.internal.parsers.ScalametaParser.syntaxErrorExpected(ScalametaParser.scala:360)
	at scala.meta.internal.parsers.ScalametaParser.accept(ScalametaParser.scala:364)
	at scala.meta.internal.parsers.ScalametaParser.generatorBuf(ScalametaParser.scala:2385)
	at scala.meta.internal.parsers.ScalametaParser.enumeratorBuf(ScalametaParser.scala:2355)
	at scala.meta.internal.parsers.ScalametaParser.$anonfun$enumerators$1(ScalametaParser.scala:2342)
	at scala.meta.internal.parsers.ScalametaParser.$anonfun$enumerators$1$adapted(ScalametaParser.scala:2339)
	at scala.meta.internal.parsers.ScalametaParser.scala$meta$internal$parsers$ScalametaParser$$listBy(ScalametaParser.scala:516)
[...]

PR #2729, opened 2022-04-14
Allow using backquoted pattern vars in generators

tgodzik:
This is not allowed in pattern matches but in generators backquoted identifiers are always treated as pattern variables, even with a `case` keyword

Fixes https://github.com/scalameta/scalameta/issues/2719


kitbellew (approved):
one question: are backticks "only allowed in for-comprehension" or "only prohibited in case-match"?

dos65 (approved):
:+1:

Issue #2733, opened 2022-04-20
? will be allowed as a wildcard by default in Scala 2.13.9 and 2.12.16


smarter:
Previously, this was only allowed under -Xsource:3, but starting from 2.13.9 (cf https://github.com/scala/scala/pull/9990) and 2.12.16 (cf https://github.com/scala/scala/pull/10005) this syntax will be enabled by default, so scalameta should support it without requiring the use of `Scala213Source3`/`Scala212Source3`:
https://github.com/scalameta/scalameta/blob/d21ee313757d8b94c82de2b48d14f54342b8e6e5/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala#L55-L59

PR #2734, opened 2022-04-20
Allow `?` as placeholder as default in Scala213 and Scala212 dialects

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/2733


tgodzik:
> @tgodzik change is simple but... why? scala213 does not really support it.

It's due to https://github.com/scala/scala/pull/9990

We could add a separate dialect, but it would be too much hassle and the change only influences [?] which will not break anything that wasn't allowed before.

kitbellew (approved):

Issue #1223, opened 2018-01-19
Parenthesis affecting infix operator parsing


blorente:
Using `scalameta`'s `semanticdb`, I stumbled upon the following case:

```scala
object inAppl {
  val a = Seq(1, 2, 3)
  val b = Seq("a", "b", "c")
  (a zip b/*1*/)/*2*/
}
```
Here is a strange corner case in the matching of synthetic parameter lists to applications. In the following example, we have that the relevant synthetic `[74..74): *(scala.collection.Seq.canBuildFrom[Tuple2[Int, String]])` is inserted at position /* 1 \*/, but the application actually goes from 67 to 76 (/* 2 \*/).

We know that it is the synthetic that must be inserted at /* 2 */, because the signatures match and there is no other synthetic around that position.

In fact, if we remove he parentheses it works as expected:

```scala
object inAppl {
  val a = Seq(1, 2, 3)
  val b = Seq("a", "b", "c")
  a zip b/*1 and 2*/
}
```

We're currently doing analysis of large corpora with scalameta, so this will prevent us from taking these cases into account.

PR #2738, opened 2022-04-25
test: add test for #1223

tanishiking:
close https://github.com/scalameta/scalameta/issues/1223

Looks like this is no longer the case, please let me know if I'm missing something :) @blorente

### DIFF
--- a/tests/jvm/src/test/resources/metac.expect
+++ b/tests/jvm/src/test/resources/metac.expect
@@ -1942,6 +1942,58 @@ Synthetics:
   apply => scala/collection/immutable/List.apply().
   Any => scala/Any#
 
+semanticdb/integration/src/main/scala/example/Issue1223.scala
+-------------------------------------------------------------
+
+Summary:
+Schema => SemanticDB v4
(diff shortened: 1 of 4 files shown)


tests/jvm/src/test/resources/metac.expect
```diff
@@ -1942,6 +1942,58 @@ Synthetics:
...
+  Int => scala/Int#
+[4:10..4:13): Seq => *.apply[String]
+  apply => scala/collection/generic/GenericCompanion#apply().
+  String => java/lang/String#
+[5:3..5:8): a zip => *[Int, String, Seq[Tuple2[Int, String]]]
+  Int => scala/Int#
+  String => java/lang/String#
+  Seq => scala/collection/Seq#
+  Tuple2 => scala/Tuple2#
+[5:3..5:10): a zip b => *(Seq.canBuildFrom[Tuple2[Int, String]])
```
tanishiking, line 1990: Looks like the position is ok

tgodzik: Shouldn't it be `[5:3..5:11]` according to the issue?  So after the closing parenthesis?

PR #2744, opened 2022-05-01
exclude parens tokens in Lit Term.ApplyInfix.args

bjaglin:
Follows https://github.com/scalameta/scalameta/pull/2736

This is not related to https://github.com/scalacenter/scalafix/issues/1594 as there is no reason to lookup literals in semanticdb, but for consistency, I thought it could be good to exclude parens as well for literals.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2200,10 +2200,13 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       val rpPos = auto.endTokenPos
       @inline def addPos(body: Term): Term = {
         body match {
-          // For `Term.Ref`, avoid using autoPos to exclude parens tokens in Term.ApplyInfix.args.
+          // For `Term.Ref` & `Lit` other than Unit, avoid using autoPos
+          // to exclude parens tokens in Term.ApplyInfix.args.
           // For `a f (b)`, autoPos include `(b)` for ApplyInfix.args position.
           // https://github.com/scalacenter/scalafix/issues/1594
           case _: Term.Ref =>
+            body
+          case l: Lit if l.value != () =>
             body
           case _ =>
             atPosWithBody(lpPos, body, rpPos)
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/BasicPositionSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/BasicPositionSuite.scala
@@ -18,6 +18,14 @@ class BasicPositionSuite extends BasePositionSuite(dialects.Scala213) {
   )
 
   checkPositions[Term](
+    "a f (123)"
+  )
+
+  checkPositions[Term](
+    "a f ()"
+  )
+
+  checkPositions[Term](
     "(1 + 2).foo",
     """|Term.ApplyInfix (1 + 2)
        |""".stripMargin


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2200,11 +2200,14 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
       @inline def addPos(body: Term): Term = {
         body match {
-          // For `Term.Ref`, avoid using autoPos to exclude parens tokens in Term.ApplyInfix.args.
+          // For `Term.Ref` & `Lit` other than Unit, avoid using autoPos
+          // to exclude parens tokens in Term.ApplyInfix.args.
           // For `a f (b)`, autoPos include `(b)` for ApplyInfix.args position.
           // https://github.com/scalacenter/scalafix/issues/1594
           case _: Term.Ref =>
             body
+          case l: Lit if l.value != () =>
```
bjaglin, line 2209: without the `if`, getting

```
[error] /home/runner/work/scalameta/scalameta/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala:1507:15: exception during macro expansion: 
[error] java.lang.RuntimeException: internal error: unpositioned prototype (): Lit.Unit(())
```

https://github.com/scalameta/scalameta/blob/60083a2d2eb4c23609694725b54d80bbe98dffe6/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala#L1469-L1474

tests/shared/src/test/scala/scala/meta/tests/parsers/BasicPositionSuite.scala
```diff
@@ -17,6 +17,14 @@ class BasicPositionSuite extends BasePositionSuite(dialects.Scala213) {
     "a f (b)"
   )
 
+  checkPositions[Term](
+    "a f (123)"
+  )
+
+  checkPositions[Term](
+    "a f ()"
+  )
```
bjaglin, line 26: this was not failing before, but I thought I could add it preemptively since there is clearly a special case for Unit

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2200,11 +2200,14 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+            body
           case _ =>
             atPosWithBody(lpPos, body, rpPos)
```
bjaglin, line 2212: I am wondering if this should exist at all https://github.com/scalameta/scalameta/pull/2736/files#r862543397

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: let's not look at the value but instead directly as Lit.Unit? In fact, instead of Ref and this, should you simply add parens for Lit.Unit and Term.Tuple only?

kitbellew (changes requested):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: another, possibly cleaner option, is to undo the recent Ref change here and instead remove addPos on line 2220 below, in the second branch of that match. elsewhere, the output of this method is unfolded, by mapping Lit.Unit back to Nil, and Term.Tuple to its arguments. however, the first branch of the match below returns a different expression than what addPos wraps there.

dos65:
Superseded by #2744.

PR #2749, opened 2022-05-04
ScalametaParser: fix paren positions in infix expr

kitbellew:
- do not include parens in rhs args
- do include them in full infix when parens are part of lhs.

Supersedes #2744.


tgodzik (approved):
Thanks, looks good!

Issue #2750, opened 2022-05-09
Existential type triggers "unsupported type ?"


Martynas-P:
Hey, I have the following piece of code

```scala
import com.wix.accord.dsl._

object App {
  def main(args: Array[String]): Unit = {
    validator[MyClass] { myClass =>
      myClass.s is notEmpty // this triggers the exception
    }
  }
}

case class MyClass(s: Seq[_])
```

Which triggers the following error in the SemanticDB scalac plugin

```
scalac: failed to generate semanticdb for /home/martynas/projects/scalameta/src/main/scala/App.scala:
java.lang.RuntimeException: unsupported type ?: BoundedWildcardType(TypeBounds(TypeRef(ThisType(scala), scala.Nothing, List()), TypeRef(ThisType(scala), scala.Any, List())))
```

The library used here is Accord [^0]. It can be added to project with `libraryDependencies += "com.wix" %% "accord-core" % "0.7.3"`. It also requires Scala 2.12.

The `validator` method here uses a macro [^1]. 

I am really sorry, but I could not reproduce the bug without the Accord library :(

[^0]: https://github.com/wix/accord
[^1]: https://github.com/wix/accord/blob/master/core/src/main/scala/com/wix/accord/dsl/package.scala#L191

tanishiking:
Thank you for reporting! It might be a bug of semanticdb plugin, I'll take a look later.
Anyway, thank you very much to try to reproduce the issue without the dependency!

tanishiking:
Indeed `BoundedWildcardType` isn't supported by scalac-semanticdb, at least we should compile even if we hit a non-supported type.

tanishiking:
:eyes: https://github.com/scala/scala/blob/a3e918a10d2ed1fad78457377244106e22685277/src/reflect/scala/reflect/api/Types.scala#L996-L1014

PR #2752, opened 2022-05-12
fix: Do not crash when scalameta hit BoundedWildcardType

tanishiking:
fix https://github.com/scalameta/scalameta/issues/2750

Confirmed it fixes the issue in https://github.com/tanishiking/scalameta-2750

As commented by https://github.com/scala/scala/blob/a3e918a10d2ed1fad78457377244106e22685277/src/reflect/scala/reflect/api/Types.scala#L996-L1014 those types shouldn't be accessible after type inference, and it might be a bug of scalac.


dos65 (approved):
:+1: 
LGTM! Thx for the fix

PR #2756, opened 2022-05-22
Revert "Allow `?` as placeholder as default in Scala213 and Scala212 dialects"

### LINKED ISSUE #2751: Support for selectively overriding individual fields in Dialect
Seems possibly related to #2750
Definitely related to #2734

Now that `Foo[?]` syntax is default, I expected to be able to

```scala
implicit val d = scala.meta.dialects.Scala212.withAllowQuestionMarkPlaceholder(false)
```

... in order to restore the previous functionality, but I'm greeted with 

```
<console>:14: error: dialect of type scala.meta.Dialect is not supported by quasiquotes (to fix this, import something from scala.meta.dialects, e.g. scala.meta.dialects.Scala212)
```

Digging into the code, I found [this message](https://github.com/scalameta/scalameta/blob/3c9b813c542a7fcc165f9be3e238c3085420fab0/scalameta/quasiquotes/shared/src/main/scala/scala/meta/internal/quasiquotes/ReificationMacros.scala#L131-L134):

>     // For now I'll just prohibit quasiquotes for situations when `dialectTree` doesn't point to one of the predefined dialects.
>    // A natural extension to this would be to allow any static value, not just predefined dialects.
>    // Furthermore, we could further relax this restriction by doing parsing for a superset of all dialects and then
>    // delaying validation of resulting ASTs until runtime.

which doesn't inspire much hope. How can this be resolved?

I guess this is only a transitional issue, while everybody upgrades to the latest versions of Scala212 and 213, and I can stay on the old version of scalameta, but I've not seen any discussion around this so I thought I'd ask.

Thank you for your excellent library!

---

A full reproduction case:

```scala
[info] Starting scala interpreter...
Welcome to Scala 2.12.15 (OpenJDK 64-Bit Server VM, Java 11.0.11).
Type in expressions for evaluation. Or try :help.

scala> import scala.meta._
import scala.meta._

scala> q"Foo[_]".syntax
res0: String = Foo[?]  // My consumers can't compile this yet

// An attempt to supply an instance of Dialect is greeted with this message
scala> def foo(implicit dialect: Dialect): String = q"Foo[_]".syntax
<console>:14: error: dialect of type scala.meta.Dialect is not supported by quasiquotes (to fix this, import something from scala.meta.dialects, e.g. scala.meta.dialects.Scala212)
       def foo(implicit dialect: Dialect): String = q"Foo[_]".syntax
                                                    ^
```

blast-hardcheese:
Fixes #2751 

This reverts commit 817b016b656a683378187d8937bea7c1526e53f5.

This is the first of two PRs, first to temporarily revert this breaking change to 2.12 and 2.13 dialects, the second will be to re-establish the new defaults, intended to be merged once 2.12.16 and 2.13.9 are released.

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
@@ -45,19 +45,18 @@ package object dialects {
 
   implicit val Scala212 = Scala211
     .withAllowTrailingCommas(true)
-    .withAllowQuestionMarkPlaceholder(true)
 
   implicit val Scala213 = Scala212
     .withAllowImplicitByNameParameters(true)
     .withAllowLiteralTypes(true)
     .withAllowNumericLiteralUnderscoreSeparators(true)
     .withAllowTryWithAnyExpr(true)
-    .withAllowQuestionMarkPlaceholder(true)
 
   /**
    * Dialect starting with Scala 2.13.6 for `-Xsource:3` option
    */
   implicit val Scala213Source3 = Scala213
+    .withAllowQuestionMarkPlaceholder(true)
     .withAllowAsForImportRename(true)
     .withAllowStarWildcardImport(true)
     .withAllowOpenClass(true)
@@ -71,6 +70,7 @@ package object dialects {
    * Dialect starting with Scala 2.12.14 for `-Xsource:3` option
    */
   implicit val Scala212Source3 = Scala212
+    .withAllowQuestionMarkPlaceholder(true)
     .withAllowAsForImportRename(true)
     .withAllowStarWildcardImport(true)
     .withAllowOpenClass(true)


smarter:
It looks like the problem isn't that the syntax is allowed, but that it is used by default by the pretty-printer when it is allowed, maybe the pretty-printer could be changed to not do that on Scala 2 (or even better, use token-level information to know whether a ? or an _ should be pretty-printed?)

blast-hardcheese:
@smarter Right -- This change wasn't incorrect, it was just merged and released before the requisite versions of scalac were released.

Attempting to represent whether `?` or `_` was in the initial encoding would be another way to work around this, though I still think giving control to the `Dialect` responsible for choosing how it's rendered is still the more robust solution. By having Dialect supplied at the end of the world, users wouldn't need to special-case every time a token is constructed.

blast-hardcheese:
(see #2758)

kitbellew (changes requested):
honestly, i don't really understand why this is needed. you can provide a different dialect to `.syntax` if you wish. and i already commented on how to avoid this problem with a custom dialect (forgoing quasiquotes).

tgodzik:
Thanks for the contribution! What would you say about having an alternative approach here: https://github.com/scalameta/scalameta/pull/2759

I think this would be the safest we can do :thinking:

blast-hardcheese:
Closing in favor of the merged PR, I'll see if I still have difficulties after that gets released. Thank you for your assistance!

PR #2758, opened 2022-05-22
Adding a Scala212Pre16 dialect

### LINKED ISSUE #2751: Support for selectively overriding individual fields in Dialect

blast-hardcheese:
Resolves #2751 

In preparation for #2757, to aide consumers who desire a seamless transition during the shakeout, introduce a new dialect that disallows the new `?` placeholder syntax for both 2.12 and 2.13:

```scala
Scala212Pre16
Scala213Pre9
```

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
@@ -55,6 +55,12 @@ package object dialects {
     .withAllowQuestionMarkPlaceholder(true)
 
   /**
+   * Dialect ending with Scala 2.13.9 for legacy F[_] syntax instead of F[?]
+   */
+  implicit val Scala213Pre9 = Scala213
+    .withAllowQuestionMarkPlaceholder(false)
+
+  /**
    * Dialect starting with Scala 2.13.6 for `-Xsource:3` option
    */
   implicit val Scala213Source3 = Scala213
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
@@ -79,6 +79,12 @@ package object dialects {
     .withAllowPlusMinusUnderscoreAsIdent(true)
     .withAllowGivenImports(true)
 
+  /**
+   * Dialect ending with Scala 2.12.15 for legacy F[_] syntax instead of F[?]
+   */
+  implicit val Scala212Pre16 = Scala212
+    .withAllowQuestionMarkPlaceholder(false)
+
   implicit val Scala = Scala213 // alias for latest Scala dialect.
 
   implicit val Sbt0136 = Scala210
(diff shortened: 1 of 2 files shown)


scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -54,6 +54,12 @@ package object dialects {
     .withAllowTryWithAnyExpr(true)
     .withAllowQuestionMarkPlaceholder(true)
 
+  /**
+   * Dialect ending with Scala 2.13.9 for legacy F[_] syntax instead of F[?]
+   */
+  implicit val Scala213Pre9 = Scala213
```
kitbellew, line 60: why not `Scala_2_13_8` instead?

kitbellew (changes requested):

tgodzik:
Thanks for the contribution! After thinking it over a bit, I think having a new dialect for older version would introduce too much work for the downstream projects such as yours. I went with an alternative approach in https://github.com/scalameta/scalameta/pull/2759 

This should mean that no changes are needed for you downstream.

blast-hardcheese:
Makes sense, thanks for your consideration here

Issue #2777, opened 2022-06-29
InvariantFailedException in output


szymon-rd:
This error appears in output:
```
org.scalameta.invariants.InvariantFailedException: invariant failed:
when verifying value.!=(null).&&(scala.Predef.augmentString(value).nonEmpty)
found that scala.Predef.augmentString(value).nonEmpty is false
where value = 
	at org.scalameta.invariants.InvariantFailedException$.raise(Exceptions.scala:19)
	at scala.meta.Term$Name$.internal$39(Trees.scala:76)
	at scala.meta.Term$Name$.apply(Trees.scala:76)
	at scala.meta.internal.parsers.ScalametaParser.$anonfun$macroSplicedIdent$2(ScalametaParser.scala:2116)
	at scala.meta.internal.parsers.ScalametaParser.atPos(ScalametaParser.scala:304)
	at scala.meta.internal.parsers.ScalametaParser.atCurPos(ScalametaParser.scala:307)
	at scala.meta.internal.parsers.ScalametaParser.atCurPosNext(ScalametaParser.scala:310)
	at scala.meta.internal.parsers.ScalametaParser.$anonfun$macroSplicedIdent$1(ScalametaParser.scala:2116)
	at scala.meta.internal.parsers.ScalametaParser.atPos(ScalametaParser.scala:296)
	at scala.meta.internal.parsers.ScalametaParser.autoPos(ScalametaParser.scala:337)
	at scala.meta.internal.parsers.ScalametaParser.macroSplicedIdent(ScalametaParser.scala:2114)
	at scala.meta.internal.parsers.ScalametaParser.$anonfun$simpleExpr0$1(ScalametaParser.scala:2030)
	at scala.meta.internal.parsers.ScalametaParser.atPosTry(ScalametaParser.scala:334)
	at scala.meta.internal.parsers.ScalametaParser.autoPosTry(ScalametaParser.scala:344)
	at scala.meta.internal.parsers.ScalametaParser.simpleExpr0(ScalametaParser.scala:2019)
	at scala.meta.internal.parsers.ScalametaParser.simpleExpr(ScalametaParser.scala:2017)
	at scala.meta.internal.parsers.ScalametaParser.prefixExpr(ScalametaParser.scala:2001)
	at scala.meta.internal.parsers.ScalametaParser.postfixExpr(ScalametaParser.scala:1983)
	at scala.meta.internal.parsers.ScalametaParser.$anonfun$expr$2(ScalametaParser.scala:1556)
[...]

PR #2779, opened 2022-07-06
bugfix: Make sure that the macro ident is not just $

tgodzik:
Previously, if user started to write the macro spliced identifier the parser would throw an invariant exception. Now we check that we can construct proper name out of the Ident.

Fixes https://github.com/scalameta/scalameta/issues/2777

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -490,7 +490,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   @classifier
   trait MacroSplicedIdent {
     def unapply(token: Token): Boolean = {
-      dialect.allowSpliceAndQuote && QuotedSpliceContext.isInside() && isIdentAnd(_.head == '$')
+      dialect.allowSpliceAndQuote && QuotedSpliceContext.isInside() &&
+      isIdentAnd(name => name.size > 1 && name.head == '$')
     }
   }
 
@@ -2113,9 +2114,9 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     }
   }
 
-  def macroSplicedIdent(): Term = autoPos {
+  private def macroSplicedIdent(): Term = autoPos {
     token match {
-      case Ident(value) =>
+      case Ident(value) if value.length() > 1 && value.head == '$' =>
         val name = atCurPosNext(Term.Name(value.stripPrefix("$")))
         Term.SplicedMacroExpr(name)
       case _ =>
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2115,7 +2115,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
   def macroSplicedIdent(): Term = autoPos {
     token match {
-      case Ident(value) =>
+      case Ident(value) if value.size > 1 =>
```
kitbellew, line 2118: perhaps length instead of size?

also, what if the value doesn't start with `$` that we're trying to strip?

kitbellew: > > perhaps length instead of size?
> 
> The definition of size is `@`inline` def size: Int = s.length` so it looks to result in the same behavior.

what is the benefit of using size instead of the more explicit length? if i remember correctly, size carries the meaning that it might potentially need to iterate over the entire sequence while length is more likely to be O(1).

> 
> > also, what if the value doesn't start with $ that we're trying to strip?
> 
> the current behaviour is that `macroSplicedIdent` is only invoked if there is `$`. So currently there is no way that it doesn't start with $

I understand that. but what if the behavior changes, and the method is invoked without this guarantee? especially since this is a public method.

tgodzik: Ok, changed to `length`, I just thought that there is no real difference.

> I understand that. but what if the behavior changes, and the method is invoked without this guarantee? especially since this is a public method.

Honestly, I don't know what should happen if there is no `$`, that shouldn't happen unless someone breaks the parser code.  Added another check just in case.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -490,7 +490,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   @classifier
   trait MacroSplicedIdent {
     def unapply(token: Token): Boolean = {
-      dialect.allowSpliceAndQuote && QuotedSpliceContext.isInside() && isIdentAnd(_.head == '$')
+      dialect.allowSpliceAndQuote && QuotedSpliceContext.isInside() && 
+        isIdentAnd(name => name.size > 1 && name.head == '$')
```
tgodzik, line 494: I realized that we should check this before, since `$` can actually be a normal identifier.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2113,9 +2114,9 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     }
   }
 
-  def macroSplicedIdent(): Term = autoPos {
+  private def macroSplicedIdent(): Term = autoPos {
     token match {
-      case Ident(value) =>
+      case Ident(value) if value.length() > 1 && value.head == '$' =>
```
tgodzik, line 2119: Added the conditions just in case and made private.

kitbellew (approved):
nice, thank you!

PR #2780, opened 2022-07-10
ScalametaParser: refactor macro ident matching

kitbellew:
Remove syntax error because it cannot currently occur. Follow up to #2779.

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MacroSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MacroSuite.scala
@@ -421,11 +421,21 @@ class MacroSuite extends BaseDottySuite {
       )
     )
   }
+
   test("no-name") {
     runTestAssert[Stat](
       "'{ val x: Int = $ }"
     )(
       Term.QuotedMacroExpr(
+        Term.Block(
+          List(Defn.Val(Nil, List(Pat.Var(Term.Name("x"))), Some(Type.Name("Int")), Term.Name("$")))
+        )
+      )
+    )
+    runTestAssert[Stat](
+      "${ val x: Int = $ }"
+    )(
+      Term.SplicedMacroExpr(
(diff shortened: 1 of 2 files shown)


tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MacroSuite.scala
```diff
@@ -421,15 +421,9 @@ class MacroSuite extends BaseDottySuite {
...
-    runTestAssert[Stat](
-      "'{ val x: Int = $ }"
-    )(
-      Term.QuotedMacroExpr(
-        Term.Block(
-          List(Defn.Val(Nil, List(Pat.Var(Term.Name("x"))), Some(Type.Name("Int")), Term.Name("$")))
-        )
-      )
-    )
+    runTestError[Stat]("'{ val x: Int = $ }", "error: Expected macro spliced ident")
```
tgodzik, line 426: This can actually be correct single `$` can be a previously declared ident.

tgodzik (changes requested):

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MacroSuite.scala
kitbellew: should it be just `$`, or in reality `$$`?

how should this be encoded: `$+`, `$ + 1`?

tgodzik: I don't think you can include + as part of the identifier.

It should be $, you can already do `val $ = 123` and as I checked with Scala 3 macros it accepts a single $

kitbellew: done

tgodzik (approved):
LGTM!

PR #2784, opened 2022-07-16
release command

eed3si9n:
Problem
-------
sbt 1.7.0 made ++ command more strict, making it not possible
to use +publishSigned as is.
This is because compiler plugin has much finer-grained
crossScalaVersions compared to the core libraries.

Solution
--------
Make all subprojects as fine-grained as needed by the compiler plugins.
Adjust publishing by defining a custom command called release.


tgodzik:
@eed3si9n I am trying to run the release, but it seems that each time it hangs https://github.com/scalameta/scalameta/runs/7433154721?check_suite_focus=true

Looks like memory issues, which we didn't encounter before. Each time the version changes it seems to take on more and more memory, probably due to some caching? Anyway to force sbt to free some memory?

eed3si9n:
I suggest you give it more memory. For example, this is what we use for sbt - https://github.com/sbt/sbt/blob/1.7.x/.github/workflows/ci.yml:

```yaml
    env:
      JAVA_OPTS: -Xms800M -Xmx2G -Xss6M -XX:ReservedCodeCacheSize=128M -server -Dsbt.io.virtual=false -Dfile.encoding=UTF-8
      JVM_OPTS: -Xms800M -Xmx2G -Xss6M -XX:ReservedCodeCacheSize=128M -server -Dsbt.io.virtual=false -Dfile.encoding=UTF-8
```

tgodzik:
> I suggest you give it more memory. For example, this is what we use for sbt - https://github.com/sbt/sbt/blob/1.7.x/.github/workflows/ci.yml:
> 
> ```yaml
>     env:
>       JAVA_OPTS: -Xms800M -Xmx2G -Xss6M -XX:ReservedCodeCacheSize=128M -server -Dsbt.io.virtual=false -Dfile.encoding=UTF-8
>       JVM_OPTS: -Xms800M -Xmx2G -Xss6M -XX:ReservedCodeCacheSize=128M -server -Dsbt.io.virtual=false -Dfile.encoding=UTF-8
> ```

We have even higher memory set and it's still running out of memoery. I thought that it might have been the amount of classes being loaded up, so I increased `-XX:ReservedCodeCacheSize` but that didn't help.

Getting a lot of:
```
[warn] In the last 16 seconds, 8.898 (55.9%) were spent in GC. [Heap: 2.94GB free of 5.21GB, max 5.33GB] Consider increasing the JVM heap using `-Xmx` or try a different collector, e.g. `-XX:+UseG1GC`, for better performance.
[warn] one warning found
```
which is weird since it seems to have a lot of free memory. Why would it GC so heavily?

PR #2789, opened 2022-07-20
bugfix: Only set latest 2.13 and 2.12 versions for Native/JS projects

tgodzik:


tgodzik:
I'll merge it to see if this helped.

PR #2791, opened 2022-07-21
ScalametaParser: no double autopos in modifier

kitbellew:
Follow up to #2787.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2696,29 +2696,29 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   }
 
   private def modifier(isLocal: Boolean): Mod = {
-    val mod = autoPos(token match {
+    val mod = token match {
       case t: Unquote => unquote[Mod](t)
       case t: Ellipsis => ellipsis[Mod](t, 1)
-      case _: KwAbstract => next(); Mod.Abstract()
-      case _: KwFinal => next(); Mod.Final()
-      case _: KwSealed => next(); Mod.Sealed()
-      case _: KwImplicit => next(); Mod.Implicit()
-      case _: KwLazy => next(); Mod.Lazy()
-      case _: KwOverride if !isLocal => next(); Mod.Override()
+      case _: KwAbstract => atCurPosNext(Mod.Abstract())
+      case _: KwFinal => atCurPosNext(Mod.Final())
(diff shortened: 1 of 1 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2696,69 +2696,69 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+    }
     newLinesOpt()
     mod
   }
 
   def quasiquoteModifier(): Mod = entrypointModifier()
 
   def entrypointModifier(): Mod = {
-    val mod = autoPos {
-      def annot() = annots(skipNewLines = false) match {
```
kitbellew, line 2730: please take a look; changed `skipNewLines` to true, since we are ok skipping newlines.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2696,69 +2696,69 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+      case KwAbstract() => atCurPosNext(Mod.Abstract())
+      case Ident("+") => atCurPosNext(Mod.Covariant())
+      case Ident("-") => atCurPosNext(Mod.Contravariant())
+      case KwLazy() => atCurPosNext(Mod.Lazy())
+      case KwVal() if !dialect.allowUnquotes => atCurPosNext(Mod.ValParam())
+      case KwVar() if !dialect.allowUnquotes => atCurPosNext(Mod.VarParam())
+      case Ident("valparam") if dialect.allowUnquotes => atCurPosNext(Mod.ValParam())
+      case Ident("varparam") if dialect.allowUnquotes => atCurPosNext(Mod.VarParam())
+      case _ =>
+        token.toString match {
```
kitbellew, line 2754: compared to `modifier()` (see line 2715), this `entrypointModifier` doesn't have `opaque`; is that correct?

tgodzik: This might be wrong, thought this was never tested, which might explain why this never popped up.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: This should be ok. Adding skipNewLines just make it a bit more permisive. We anyway usually remove newlines between annots.

tgodzik (approved):
LGTM!

PR #2795, opened 2022-07-25
try to fix release workflow

dos65:
Currently, there is some memory consuming issue - https://github.com/scalameta/scalameta/actions/runs/2721177103
@keynmol suggested to enable batchMode for scala js linker.


tgodzik:
Och, that's weird that the downgrade didn't help, maybe another PR merged in between caused it?

dos65:
@tgodzik no idea. I don't see any suspicious commits into sbt related things.

dos65:
@tgodzik it didn't help too. I will revert it in next PR but need to find what else we can try

Issue #1431, opened 2018-03-11
semanticdb-scalac emits wrong symbol for synthetic toInt 


olafurpg:
```scala
➜  target git:(scopes) ✗ cat tobyte.scala
object a {
  def foo(a: Int): Int = a.toByte
}
➜  target git:(scopes) ✗ scalac tobyte.scala -Xprint:typer
    def foo(a: Int): Int = a.toByte.toInt // <-- toInt was inserted by scalac
➜  target git:(scopes) ✗ metac tobyte.scala
```
```diff
➜  target git:(scopes) ✗ metap META-INF/semanticdb/tobyte.scala.semanticdb
-obtained
+expected
-[1:27..1:33): toByte => _root_.scala.Byte#toInt().
+[1:27..1:33): toByte => _root_.scala.Byte#toByte().
+ Synthetics
+[1:33..1:33): *.toInt
```

PR #2797, opened 2022-07-25
Fix incorrect symbol occurrence for toLong/toInt primitive conversion

olafurpg:
Fixes https://github.com/scalameta/scalameta/issues/1431

Previously, semanticdb-scalac emitted incorrect symbol occurrences whenever
primitive conversion methods were implicitly applied (for example, `scala/Int#toLong`).
This commit mostly fixes this bug, and adds a test case to demonstrate
for what corner cases the bug is still unfixed.


semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolOps.scala
```diff
@@ -295,6 +295,29 @@ trait SymbolOps { self: SemanticdbOps =>
...
+      g.termNames.toShort,
+      g.termNames.toChar,
+      g.termNames.toInt,
+      g.termNames.toLong,
+      g.termNames.toFloat,
+      g.termNames.toDouble
+    )
+    // Returns true if the `this.sym` resolves to a primitive conversion method like toInt/toLong
+    // and the provided mtree name does not match the name of that symbol. This implementation
+    // has a minor bug where it doesn't detect implicit convesions when mtree.value has the name
```
kitbellew, line 310: conveRsions

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.6.2
+sbt.version=1.5.8
```
kitbellew, line 1: should this downgrade be done separately, if it is actually needed?

dos65: not a perfect commit but let's take them together - every check for a possible release workflow fix takes about 1h.
if it will help - that's good.
anyway we will have to find out how to upgrade it the latest in future.

semanticdb/scalac/library/src/main/scala-2.13/scala/meta/internal/semanticdb/scalac/SymbolOpsCompat.scala
```diff
@@ -0,0 +1,30 @@
...
+    // Returns true if the `this.sym` resolves to a primitive conversion method like toInt/toLong
+    // and the provided mtree name does not match the name of that symbol. This implementation
+    // has a minor bug where it doesn't detect implicit convesions when mtree.value has the name
+    // of a conversion method like toInt/toLong, but it's a corner case that can be fixed separately
+    // if it's a big problem.
+    def isImplicitPrimitiveConversion(mtree: m.Name): Boolean = {
+      sym.name.startsWith("to") &&
+        g.definitions.ScalaValueClassesSet.contains(sym.owner) &&
+        implicitPrimitiveConversionNames.contains(sym.name.toTermName) &&
+        !sym.name.startsWith(mtree.value) &&
```
kitbellew, line 26: should this be `||`?

tests/jvm/src/test/scala/scala/meta/tests/semanticdb/TargetedSuite.scala
```diff
@@ -205,4 +205,24 @@ class TargetedSuite extends SemanticdbSuite {
 
     }
   )
+
+  targeted(
```
olafurpg, line 209: I kept this test to ensure that it runs in 2.11/2.12/2.13. The  expect tests only run in 2.12

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolOps.scala
olafurpg: Fixed

project/build.properties
olafurpg: I upgraded to 1.6.2 for this PR while I'm re-using an older 1.5.8 sbt shell locally for edit/test.

semanticdb/scalac/library/src/main/scala-2.13/scala/meta/internal/semanticdb/scalac/SymbolOpsCompat.scala
olafurpg: I refactored this condition into a simpler `sym.name.toString() != mtree.value`

tgodzik (approved):
LGTM!

PR #2801, opened 2022-08-14
Dialect: implement deeper dialect equality check

kitbellew:


scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -577,14 +577,78 @@ final class Dialect private (
   // Dialects have reference equality semantics,
```
tgodzik, line 577: Looks like this comment would not make any sense anymore, but I think specially in case of dialects i doesn't help use to compare with reference.

tgodzik (approved):
LGTM aside from the irrelevant comment.

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
kitbellew: good catch, will remove

PR #2803, opened 2022-08-15
bugfix: Allow single using identifier

tgodzik:


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2221,7 +2221,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
               else if (tail.nonEmpty) syntaxError("repeated argument not allowed here", at = head)
             case _ =>
           }
-          val using = nextIf(x.toString == soft.KwUsing.name)
+          val using = nextIf(x.toString == soft.KwUsing.name && ahead(token.isNot[CloseDelim]))
```
dos65, line 2224: Also possible cases:
- `,` - `foo(using, bar)`
- `:` - `val using ="asdsa"; foo(using: String)`

tgodzik: Ach, good catch! Added a better one.

kitbellew (approved):

dos65 (approved):
LGTM!

PR #2804, opened 2022-08-15
Dialect: remove equals/hashCode overrides

kitbellew:
Originally, Dialect was a case class and we needed them to preserve the reference equality semantics. However, they are default with ordinary classes, and the overrides are no longer necessary.

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
@@ -577,9 +577,9 @@ final class Dialect private (
   // Dialects have reference equality semantics,
   // because sometimes dialects representing distinct Scala versions
   // can be structurally equal to each other.
-  override def canEqual(that: Any): Boolean = this eq that.asInstanceOf[AnyRef]
-  override def equals(other: Any): Boolean = this eq other.asInstanceOf[AnyRef]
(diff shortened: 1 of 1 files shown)


scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -577,9 +577,13 @@ final class Dialect private (
   // Dialects have reference equality semantics,
   // because sometimes dialects representing distinct Scala versions
   // can be structurally equal to each other.
-  override def canEqual(that: Any): Boolean = this eq that.asInstanceOf[AnyRef]
+  override def canEqual(that: Any): Boolean = that.isInstanceOf[Dialect]
+  // As Dialect is no longer a case class, no need to override equals/hashCode
+  /*
   override def equals(other: Any): Boolean = this eq other.asInstanceOf[AnyRef]
   override def hashCode: Int = System.identityHashCode(this)
+   */
```
dos65, line 585: Maybe remove it?

dos65 (approved):
LGTM

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
kitbellew: let's do that :)

PR #2806, opened 2022-08-19
Trees: mark all stats having mods with a trait

kitbellew:


scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -32,6 +32,10 @@ object Tree extends InternalTreeXtensions {
 @branch trait Ref extends Tree
 @branch trait Stat extends Tree
 
+object Stat {
+  @branch trait WithMods extends Stat { def mods: List[Mod] }
```
tgodzik, line 36: I wonder if we couldn't have just a Classifier for this or a helper method? Do we need to add a trait speicifically?

kitbellew: what's the downside?

tgodzik (approved):

Issue #2807, opened 2022-08-19
scala3: incomplete support for `_` in types


kitbellew:
https://dotty.epfl.ch/docs/reference/changed-features/wildcards.html describes this:

- scala 2:
  - type wildcard is `_`
  - type placeholder is unsupported
- scala 3.0:
  - type wildcard is `?` and `_`
  - type placeholder is `*`
- scala 3.1:
  - type wildcard is `?`, `_` is deprecated
  - type placeholder is `*`
- scala 3.2:
  - type wildcard is `?`
  - type placeholder is `_`, `*` is deprecated
- scala 3.3:
  - type wildcard is `?`
  - type placeholder is `_`

In `Dialect`, we have the following today:

- `allowQuestionMarkPlaceholder` (which is misleading since it's a wildcard but not a placeholder)
- `allowPlusMinusUnderscoreAsPlaceholder` (which is a bit misleading since should apply to invariant placeholder types as well, not just covariant/contravariant)
- `allowTypeParamUnderscore` (not sure if this is for a different syntax element and could remain as-is)

I would suggest the following `Dialect` changes (and corresponding parser modifications):

- rename `allowQuestionMarkPlaceholder` to `allowQuestionMarkAsTypeWildcard`
  - keep `withAllowQuestionMarkPlaceholder` method for now but deprecate it (make it redirect to the new `withAllowQuestionMarkAsTypeWildcard`)
- similarly, rename `allowPlusMinusUnderscoreAsPlaceholder` to `allowUnderscoreAsTypePlaceholder` (to distinguish `_` between wildcard in 3.1 and placeholder in 3.2)
- possibly also add `allowStarAsTypePlaceholder`
  - otherwise, allow star if `allowQuestionMarkAsTypeWildcard=true` and `allowUnderscoreAsTypePlaceholder=false`, although this doesn't cover the 3.2 case

kitbellew:
also:
- `Type.Placeholder` is used for wildcard and might need to be split
- `Type.Placeholder` is used instead of a pattern tree (see scalameta/scalafmt#3303)

dos65:
@kitbellew Thanks for collecting the info! Looks reasonable.

tgodzik:
I really hoped we wouldn't need to create a new dialect for each minor version of Scala 3, which will be super difficult for people to handle :/

I wonder if we could just allow each placeholder in every place, from the parser perspective this wouldn't be much of an issue:

type wildcard is ? and _
type placeholder is * and _

[...]

PR #2809, opened 2022-08-22
Dialect: add intermediate scala3 dialects

kitbellew:
Helps with #2807.

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
@@ -102,7 +102,7 @@ package object dialects {
   implicit val ParadiseTypelevel212 = Typelevel212
     .withAllowInlineMods(true)
 
-  implicit val Scala3 = Scala213
+  implicit val Scala30 = Scala213
     // there 3 different ways to specify vargs, some will be removed in future Scala 3 versions
     .withAllowAtForExtractorVarargs(true) // both @ and : work currently for Scala 3
     .withAllowColonForExtractorVarargs(true) // both @ and : work currently for Scala 3
@@ -142,9 +142,15 @@ package object dialects {
     .withAllowStarWildcardImport(true)
     .withAllowProcedureSyntax(false)
     .withAllowDoWhile(false)
-    .withAllowUnderscoreAsTypePlaceholder(true)
     .withUseInfixTypePrecedence(true)
     .withAllowInfixOperatorAfterNL(true)
+
+  implicit val Scala31 = Scala30
+
+  implicit val Scala32 = Scala31
+    .withAllowUnderscoreAsTypePlaceholder(true)
+
+  implicit val Scala3 = Scala32
 
   @deprecated("Use Scala3 instead", "4.4.2")
   implicit val Dotty = Scala3
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
@@ -769,6 +769,9 @@ object Dialect extends InternalDialect {
   }
   private lazy val standardPairs = Seq[sourcecode.Text[Dialect]](
     Dotty,
+    Scala30,
+    Scala31,
+    Scala32,
     Scala3,
     Paradise211,
     Paradise212,
--- a/tests/jvm/src/test/scala-2.13/scala/meta/tests/api/SurfaceSuite.scala
+++ b/tests/jvm/src/test/scala-2.13/scala/meta/tests/api/SurfaceSuite.scala
@@ -102,6 +102,9 @@ class SurfaceSuite extends FunSuite {
       |scala.meta.dialects.Scala213 *
       |scala.meta.dialects.Scala213Source3 *
       |scala.meta.dialects.Scala3 *
+      |scala.meta.dialects.Scala30 *
+      |scala.meta.dialects.Scala31 *
+      |scala.meta.dialects.Scala32 *
       |scala.meta.dialects.Typelevel211 *
       |scala.meta.dialects.Typelevel212 *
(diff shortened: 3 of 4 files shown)


scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -144,10 +144,18 @@ package object dialects {
...
     .withAllowProcedureSyntax(false)
     .withAllowDoWhile(false)
-    .withAllowPlusMinusUnderscoreAsPlaceholder(true)
     .withUseInfixTypePrecedence(true)
     .withAllowInfixOperatorAfterNL(true)
 
+  implicit val Scala31 = Scala30.duplicate
+
+  implicit val Scala32 = Scala31
+    .withAllowPlusMinusUnderscoreAsPlaceholder(true)
```
dos65, line 153: I think we should keep it enabled starting from scala30.

It might be enabled by `-Ykind-projector` in compiler and as I remember we decided to enable it by default to avoid introducing an additional dialect. This syntax anyway won't be parsed without this flag so it doesn't produce any any ambiguity.

kitbellew: I'm planning to rename this flag and replace with the one which supports underscore as placeholder in types overall, not just covariant. and that one was added in 3.2.

kitbellew: also, scala3 points to 33, so it's enabled for that dialect.

dos65: Then maybe do dialect changes together with the changes you planned to do after it?
It will be a bit more clear what will be the difference between these dialects

kitbellew: @dos65 it's all described in the linked #2807. the dialect field is renamed in #2811 (conflict between this pr and that one will be resolved based on which pr gets merged first).

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -144,10 +144,18 @@ package object dialects {
     .withAllowStarWildcardImport(true)
     .withAllowProcedureSyntax(false)
     .withAllowDoWhile(false)
-    .withAllowPlusMinusUnderscoreAsPlaceholder(true)
     .withUseInfixTypePrecedence(true)
     .withAllowInfixOperatorAfterNL(true)
 
+  implicit val Scala31 = Scala30.duplicate
+
+  implicit val Scala32 = Scala31
```
tgodzik, line 152: Actually, it seems it might not be supported by Scala 3.2 yet :thinking: The docs require some updates, I contacted the maintainers.

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -103,7 +103,7 @@ package object dialects {
   implicit val ParadiseTypelevel212 = Typelevel212
     .withAllowInlineMods(true)
 
-  implicit val Scala3 = Scala213
```
tgodzik, line 106: Should it maybe stay as a alias for the latest Scala 3 dialect?

kitbellew: it is, please see below

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
kitbellew: it was actually supported earlier, in [3.0.1](https://github.com/lampepfl/dotty/releases/tag/3.0.1). i kept it under 3.2, though, following the documentation.

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -142,10 +142,16 @@ package object dialects {
...
-    .withAllowUnderscoreAsTypePlaceholder(true)
     .withUseInfixTypePrecedence(true)
     .withAllowInfixOperatorAfterNL(true)
 
+  implicit val Scala31 = Scala30
+
+  implicit val Scala32 = Scala31
+    .withAllowUnderscoreAsTypePlaceholder(true)
+
+  implicit val Scala3 = Scala32
```
kitbellew, line 153: this is the alias

kitbellew (commented):
removed scala 3.3 for now, as 3.2 is still to be released.

tgodzik (approved):
Looks good! With alias for the latest version, we should be fine. What do you think @dos65 ?

dos65:
Honestly, I still not fully understand the point of that at this time. 

I think right now there is no need in splitting it on `3.0` and `3.1`. `withAllowUnderscoreAsTypePlaceholder` (with a different name) was always enabled for scala3 and I think there were no issues with that. 
As I understand `3.2` will be an LTS version. So `3.0` and `3.1` are close to the end of life.
I can't imagine the case when there will be a need in using these dialects. 

However, I think we will have to do the similar thing at the time when there will be `3.3`

tgodzik:
> However, I think we will have to do the similar thing at the time when there will be 3.3

That's mostly to prepare for that, though we don't have to merge it yet.

dos65 (approved):

kitbellew:
> > That's mostly to prepare for that, though we don't have to merge it yet.
> 
> Update - I changed my mind. At least even if there will be no other changes between these dialects I think having them will be good idea from API/scalafmt settings viewpoint. The range of dialects will be consistent and obvious in some rare cases for users - if there is `3.2` and greater why there is no `3.1`?

exactly! even though there was no change, it was for consistency. people may have complained that they couldn't use something they expected...

PR #2811, opened 2022-08-22
Dialect: add allowUnderscoreAsTypePlaceholder

kitbellew:
It replaces allowPlusMinusUnderscoreAsPlaceholder. Helps with #2807.

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
@@ -142,7 +142,7 @@ package object dialects {
     .withAllowStarWildcardImport(true)
     .withAllowProcedureSyntax(false)
     .withAllowDoWhile(false)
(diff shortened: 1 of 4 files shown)


scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -142,7 +142,7 @@ package object dialects {
     .withAllowStarWildcardImport(true)
     .withAllowProcedureSyntax(false)
     .withAllowDoWhile(false)
-    .withAllowPlusMinusUnderscoreAsPlaceholder(true)
+    .withAllowUnderscoreAsTypePlaceholder(true)
```
tgodzik, line 145: Should we also change it for `Scala212Source3` and `Scala213Source3` ?

dos65: for scala2 we use `allowPlusMinusUnderscoreAsIdent`

PR #2812, opened 2022-08-22
Change implicit classes to derive from AnyVal

kitbellew:


project/Mima.scala
```diff
@@ -32,6 +32,19 @@ object Mima {
     ProblemFilters.exclude[A]("scala.meta." + metaType)
 
   val apiCompatibilityExceptions: Seq[ProblemFilter] = Seq(
+    // implicit classes
+    exclude[IncompatibleResultTypeProblem]("package.XtensionDialectTreeSyntax"),
```
tgodzik, line 36: This looks duplicated, no? What message is being printed here? I am curious why it reports `IncompatibleResultTypeProblem`

kitbellew: because of anyval, the ctor returns the type of argument now and not the type of implicit class. the calls do not look duplicated.

tgodzik: `exclude[IncompatibleResultTypeProblem]("package.XtensionDialectTreeSyntax")` is twice, no?

tgodzik: It's a palindrome from what I see :sweat_smile:

kitbellew: i was sleeping when i said i didn't see duplicates 🙂 will remove

tgodzik (approved):
LGTM

PR #2810, opened 2022-08-22
Dialect: add allowQuestionMarkAsTypeWildcard

kitbellew:
It replaces misnamed allowQuestionMarkPlaceholder. Helps with #2807.

Issue #405, opened 2016-05-27
consider changing Term.Apply to take Seq[Seq[Term.Arg]]


xeno-by:
Otherwise, we're going to need unsatisfying workarounds like `SyntacticXXX` in scala.reflect. E.g. see #277. I find this prospect suboptimal.

This is probably not 1.0 material, because the fallout may be too strong at the moment.

xeno-by:
@olafurpg @johannegger Could you comment on this please?

olafurpg:
My preference would be to have Term.Apply take a Seq[ArgList] and an ArgList would have a Seq[Term.Arg].

Why? In scalafmt I often have an opening `(` owned by a Term.Apply. The problem is that it gets quite tricky to find out which argument lists "owns" that `(`. If there was something like an ArgList tree node then I could trivially find the owner of `(` and query stuff like how many arguments are in that list.

jvican:
I agree with @olafurpg, his proposal is neat. But introducing the proposed change breaks consistency of Meta tree nodes. The convention of using `Seq[Seq[Term.Arg]]` for representing parameter lists is very present in the Meta trees, as well as Scalac and Dotty. Doing so could lead to confusion, so there's probably a tradeoff here.

olafurpg:
Example problem I've struggled with: given a `Defn.Def`, find the closing parenthesis (if one exists) of the last param list.

If we attach a tree node to each param lists, this would be trivial but currently I to do all sorts of collections gymnastics to cover a bunch of corner cases. However, I agree with @jvican that changing the tree structure may not be the best choice here. A `contrib` module with utilities to do stuff like the problem above might be enough to satisfy me.

xeno-by:
We're on the verge of releasing 2.0, and afterwards we plan to start being really conservative about breaking changes. The matter discussed in the ticket can probably be improved, but there's not enough interest to justify potential breakages.

kitbellew:
@olafurpg @xeno-by wanted to revisit this decision, especially given #2820.

olafurpg:
[...]

PR #2815, opened 2022-08-26
Trees: split off Type.Wildcard from Placeholder

kitbellew:
The `Placeholder` type (with bounds) we used earlier will be replaced with:

- `AnonymousParam`, with invariant/covariant/contravariant prefix indicator (for `"_"`, `"+_"` and `"-_"`)
- `Wildcard`, with bounds

Helps with #2807.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -925,21 +925,32 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
     def simpleType(): Type = {
       val startPos = auto.startTokenPos
+      def wildcardType(): Type = {
+        next()
+        Type.Wildcard(typeBounds())
+      }
+      def anonymousParamWithVariant(value: String): Type = {
+        val variant = if (value(0) == '+') Mod.Covariant() else Mod.Contravariant()
+        Type.AnonymousParam(Some(atPos(startPos)(variant)))
+      }
       val res = token match {
         case LeftParen() => makeTupleType(startPos, inParensOnOpen(types()))
-        case Underscore() => next(); Type.Placeholder(typeBounds())
         case MacroSplicedIdent(ident) =>
           Type.Macro(macroSplicedIdent(ident))
         case MacroSplice() =>
           Type.Macro(macroSplice())
+        case _: Underscore if dialect.allowUnderscoreAsTypePlaceholder =>
+          next(); Type.AnonymousParam(None)
+        case _: Underscore =>
+          wildcardType()
         case Ident("?") if dialect.allowQuestionMarkAsTypeWildcard =>
-          next(); Type.Placeholder(typeBounds())
+          wildcardType()
         case Ident(value @ ("+" | "-"))
             if (dialect.allowPlusMinusUnderscoreAsIdent || dialect.allowUnderscoreAsTypePlaceholder) &&
               tryAhead[Underscore] =>
           next() // Ident and Underscore
           if (dialect.allowUnderscoreAsTypePlaceholder)
-            Type.Placeholder(typeBounds())
+            anonymousParamWithVariant(value)
           else
             Type.Name(s"${value}_")
         case _: Literal =>
@@ -1047,8 +1058,12 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           val underlying1 = loop(underlying, convertTypevars = false)
           val annots1 = annots
           Type.Annotate(underlying1, annots1)
-        case Type.Placeholder(bounds) =>
-          val bounds1 = bounds
+        case t: Type.Wildcard =>
+          Type.Wildcard(t.bounds)
+        case t: Type.AnonymousParam =>
+          Type.AnonymousParam(t.variant)
+        case t: Type.Placeholder =>
+          val bounds1 = t.bounds
           Type.Placeholder(bounds1)
         case tpe: Lit =>
           tpe
@@ -3373,7 +3388,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
    */
   private def givenDecl(mods: List[Mod]): Stat = autoEndPos(mods) {
     accept[KwGiven]
-    val anonymousName = autoPos(scala.meta.Name.Anonymous())
+    val anonymousName = autoPos(Name.Anonymous())
 
     val forked = in.fork
     val (sigName, sigTparams, sigUparamss) =
--- a/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
@@ -238,7 +238,23 @@ object Type {
   @ast class Method(paramss: List[List[Term.Param]], tpe: Type) extends Type {
     checkParent(ParentChecks.TypeMethod)
   }
-  @ast class Placeholder(bounds: Bounds) extends Type
+  @deprecated("Placeholder replaced with AnonymousParam and Wildcard", ">4.5.13")
+  @branch trait Placeholder extends Type {
+    def bounds: Bounds
+    def copy(bounds: Bounds = this.bounds): Placeholder
+  }
+  @deprecated("Placeholder replaced with AnonymousParam and Wildcard", ">4.5.13")
+  object Placeholder {
+    @ast private[meta] class Impl(bounds: Bounds) extends Placeholder
+    @inline def apply(bounds: Bounds): Placeholder = Impl(bounds)
+    @inline final def unapply(tree: Placeholder): Option[Bounds] = Some(tree.bounds)
+  }
+  @ast class Wildcard(bounds: Bounds) extends Placeholder
+  @ast class AnonymousParam(variant: Option[Mod.Variant]) extends Placeholder {
+    @deprecated("Placeholder replaced with AnonymousParam and Wildcard", ">4.5.13")
+    override final def bounds: Bounds = Bounds(None, None)
+    override def copy(bounds: Bounds): Placeholder = Placeholder(bounds)
+  }
   @ast class Bounds(lo: Option[Type], hi: Option[Type]) extends Tree
   @ast class ByName(tpe: Type) extends Type {
     checkParent(ParentChecks.TypeByName)
@@ -304,7 +320,7 @@ object Pat {
   }
   @ast class Typed(lhs: Pat, rhs: Type) extends Pat {
     checkFields(rhs match {
-      case _: Type.Var | _: Type.Placeholder => false
+      case _: Type.Var | _: Type.Placeholder | _: Type.Wildcard | _: Type.AnonymousParam => false
       case _ => true
     })
   }
--- a/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/NamerMacros.scala
+++ b/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/NamerMacros.scala
@@ -91,13 +91,12 @@ trait CommonNamerMacros extends MacroHelpers {
 
     paramss.foreach(_.foreach(x => addStubbedMemberWithName(x.name)))
     extraStubs.foreach(x => addStubbedMemberWithName(TermName(x)))
-    val qcopyParamss = paramss.map(_.map { case ValDef(mods, name, tpt, _) =>
-      q"val $name: $tpt = this.$name"
-    })
+    val qcopyParamss = paramss.map(_.map(t => q"val ${t.name}: ${t.tpt}"))
     qstats += q"def copy(...$qcopyParamss): $name = $stub"
 
     extraAbstractDefs.foreach {
-      case x: ValOrDefDefApi if x.mods.hasFlag(Flag.ABSTRACT) || x.rhs.isEmpty =>
+      case x: ValOrDefDefApi
+          if x.mods.hasFlag(Flag.ABSTRACT) || x.mods.hasFlag(Flag.OVERRIDE) || x.rhs.isEmpty =>
         addStubbedOverrideMember(x)
       case _ =>
     }
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
@@ -685,6 +685,19 @@ object TreeSyntax {
           Type,
           s(p(AnyInfixTyp, t.tpe), " ", kw("match"), " {", r(t.cases.map(i(_)), ""), n("}"))
         )
+      case t: Type.AnonymousParam =>
+        m(SimpleTyp, o(t.variant), "_")
+      case t: Type.Wildcard =>
+        /* In order not to break existing tools `.syntax` should still return
+         * `_` instead `?` unless specifically used.
+         */
+        def questionMarkUsed = t.origin match {
+          case o: Origin.Parsed => !o.tokens.headOption.exists(_.is[Token.Underscore])
+          case _ => false
+        }
+        val useQM = dialect.allowQuestionMarkAsTypeWildcard &&
+          (dialect.allowUnderscoreAsTypePlaceholder || questionMarkUsed)
+        m(SimpleTyp, s(kw(if (useQM) "?" else "_")), t.bounds)
       case t: Type.Placeholder =>
         /* In order not to break existing tools `.syntax` should still return
          * `_` instead `?` unless specifically used.
--- a/tests/jvm/src/test/scala-2.13/scala/meta/tests/api/SurfaceSuite.scala
+++ b/tests/jvm/src/test/scala-2.13/scala/meta/tests/api/SurfaceSuite.scala
@@ -395,6 +395,7 @@ class SurfaceSuite extends FunSuite {
       |scala.meta.Type.And
       |scala.meta.Type.Annotate
       |scala.meta.Type.AnonymousName
+      |scala.meta.Type.AnonymousParam
       |scala.meta.Type.Apply
       |scala.meta.Type.ApplyInfix
       |scala.meta.Type.Bounds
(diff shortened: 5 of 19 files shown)


kitbellew:
@tgodzik is this too controversial? 🙂

kitbellew:
> I will take a look at it, just didn't have the time yet. Sorry about that!

thank you, wasn't sure if you saw that.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -902,19 +902,24 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
         case LeftParen() => makeTupleType(startPos, inParensOnOpen(types()))
-        case Underscore() => next(); Type.Placeholder(typeBounds())
         case MacroSplicedIdent(ident) =>
           Type.Macro(macroSplicedIdent(ident))
         case MacroSplice() =>
           Type.Macro(macroSplice())
+        case _: Underscore =>
+          next()
+          if (dialect.allowUnderscoreAsTypePlaceholder)
+            Type.Placeholder("")
```
tgodzik, line 912:
```suggestion
            Type.Placeholder("_")
```
?

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -237,7 +237,9 @@ object Type {
   @ast class Method(paramss: List[List[Term.Param]], tpe: Type) extends Type {
     checkParent(ParentChecks.TypeMethod)
   }
-  @ast class Placeholder(bounds: Bounds) extends Type
+  @ast class WildcardRef extends scala.meta.Ref
+  @ast class Wildcard(ref: WildcardRef, bounds: Bounds) extends Type
```
tgodzik, line 241: Why do we need `ref` ? Is it only to get the position for `_`?

tgodzik (changes requested):
I think this is a good idea, thought I left some comments. Thanks for working on this!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: the parameter doesn't refer to the wildcard but the variant prefix (covariant or contravariant).

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
kitbellew: yes. similar to how we have `Pat.Wildcard`.

i am actually thinking of replacing the variant prefix in Placeholder with an `Option[Mod.Covariant | Mod.Contravariant]`, similarly.

project/Mima.scala
```diff
@@ -32,6 +32,13 @@ object Mima {
     ProblemFilters.exclude[A]("scala.meta." + metaType)
 
   val apiCompatibilityExceptions: Seq[ProblemFilter] = Seq(
+    // Type.Placeholder
```
kitbellew, line 35: it didn't pass mima :)

tgodzik: ach right, I thought it's only for the class. This is a candidate for 5.x series then. For added methods minor is usually ok (though we might not have a perfect track record her), but for sure anything breaking should be 5.x

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -237,7 +237,9 @@ object Type {
...
+  @ast class Placeholder(variant: Option[Mod.Variant]) extends Type
```
kitbellew, line 242: made variant into a type

kitbellew:
@tgodzik generally, i'd prefer to release it sooner, without waiting for #2820, as that one is a massive change, with a comment in #405 indicating that it should possibly be done by introducing new top-level trees instead.

so, while this change modifies the `Placeholder` type, there are two ways ahead:

- assume that it's a minor change, about a type which is used infrequently, and release in 4.5.14 or 4.6.0
- deprecate the `Placeholder` type and create a completely new one, say, `VariantPlaceholder` or `TypePlaceholder` or something else and thus make it completely backwards-compatible.

Thoughts?

kitbellew:
Kept old `Placeholder`, added new one as `AnonymousParam` (since in the linked document it's referred to as _anonymous type parameter_).

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -302,7 +306,7 @@ object Pat {
     checkFields(parts.length == args.length + 1)
   }
   @ast class Typed(lhs: Pat, rhs: Type) extends Pat {
-    checkFields(!rhs.is[Type.Var] && !rhs.is[Type.Placeholder])
+    checkFields(!rhs.is[Type.Var] && !rhs.is[Type.Wildcard])
```
kitbellew, line 309: actually, here i don't know what should be used :)

kitbellew: added a test in #2835 and rebased.

kitbellew:
rebased on #2842, fixed the test bug.

kitbellew:
is there anything pending for me to address on this one? if i understand correctly, it's no longer triggering mima and thus doesn't require a major version change.

tgodzik:
> is there anything pending for me to address on this one? if i understand correctly, it's no longer triggering mima and thus doesn't require a major version change.

Sorry about that! I was low on time last week and I am catching up. Will take a look again tomorrow. The fewer braces issue also popped up, which I wanted to check if it requires a long refactor.

tgodzik (changes requested):
One thing I am worried about is that this will still cause issue for people using existing code and it will happen in the runtime :thinking: so code could stop working without them knowing about it. For example if you match on placeholder.

In that case removing Placeholder would be better since it would fail in compile time :thinking: but that would cause binary incompatibility issues.

Honestly, this does looks better but I don't think there is a way around it. We could change it only for Scala 3, which would at least mean that all the existing Scala 2 codebases will work without issues, but it's still not perfect. All the rules that depend on Placeholder would have to change.

@bjaglin What do you think about the change? I think this might affect scalafix rules the most.

kitbellew:
> One thing I am worried about is that this will still cause issue for people using existing code and it will happen in the runtime 🤔 so code could stop working without them knowing about it. For example if you match on placeholder.
> 
> In that case removing Placeholder would be better since it would fail in compile time 🤔 but that would cause binary incompatibility issues.
> 
> Honestly, this does looks better but I don't think there is a way around it. We could change it only for Scala 3, which would at least mean that all the existing Scala 2 codebases will work without issues, but it's still not perfect. All the rules that depend on Placeholder would have to change.
> 
> @bjaglin What do you think about the change? I think this might affect scalafix rules the most.

we could keep placeholder for now, parse using a dialect flag. will wait for @bjaglin to chime in before i proceed.

kitbellew:
> > One thing I am worried about is that this will still cause issue for people using existing code and it will happen in the runtime 🤔 so code could stop working without them knowing about it. For example if you match on placeholder.
> > In that case removing Placeholder would be better since it would fail in compile time 🤔 but that would cause binary incompatibility issues.
> > Honestly, this does looks better but I don't think there is a way around it. We could change it only for Scala 3, which would at least mean that all the existing Scala 2 codebases will work without issues, but it's still not perfect. All the rules that depend on Placeholder would have to change.
> > @bjaglin What do you think about the change? I think this might affect scalafix rules the most.
> 
> we could keep placeholder for now, parse using a dialect flag. will wait for @bjaglin to chime in before i proceed.

added a new flag, to preserve legacy scala2 parsing. the change now affects only scala3 dialects but differentiates between 3.1 and 3.2.

kitbellew:
> > > One thing I am worried about is that this will still cause issue for people using existing code and it will happen in the runtime 🤔 so code could stop working without them knowing about it. For example if you match on placeholder.
> > > In that case removing Placeholder would be better since it would fail in compile time 🤔 but that would cause binary incompatibility issues.
> > > Honestly, this does looks better but I don't think there is a way around it. We could change it only for Scala 3, which would at least mean that all the existing Scala 2 codebases will work without issues, but it's still not perfect. All the rules that depend on Placeholder would have to change.
> > > @bjaglin What do you think about the change? I think this might affect scalafix rules the most.
> > 
> > 
> > we could keep placeholder for now, parse using a dialect flag. will wait for @bjaglin to chime in before i proceed.
> 
> added a new flag, to preserve legacy scala2 parsing. the change now affects only scala3 dialects but differentiates between 3.1 and 3.2.

one other option: not introduce the new dialect flag, parse anything before 3.2 as before, and parse 3.2 and after differently. wdyt, @tgodzik ?

tgodzik:
> > > > One thing I am worried about is that this will still cause issue for people using existing code and it will happen in the runtime thinking so code could stop working without them knowing about it. For example if you match on placeholder.
> > > > In that case removing Placeholder would be better since it would fail in compile time thinking but that would cause binary incompatibility issues.
> > > > Honestly, this does looks better but I don't think there is a way around it. We could change it only for Scala 3, which would at least mean that all the existing Scala 2 codebases will work without issues, but it's still not perfect. All the rules that depend on Placeholder would have to change.
> > > > @bjaglin What do you think about the change? I think this might affect scalafix rules the most.
> > > 
> > > 
> > > we could keep placeholder for now, parse using a dialect flag. will wait for @bjaglin to chime in before i proceed.
> > 
> > 
> > added a new flag, to preserve legacy scala2 parsing. the change now affects only scala3 dialects but differentiates between 3.1 and 3.2.
> 
> one other option: not introduce the new dialect flag, parse anything before 3.2 as before, and parse 3.2 and after differently. wdyt, @tgodzik ?

I think it should be fine to just change the behaviour for Scala 3 altogether.

kitbellew:
> I think it should be fine to just change the behaviour for Scala 3 altogether.

then that's what the latest patch is doing. there is no clear existing type-related exclusively Scala 3 dialect flag; the one to allow `?` for wildcards is now also used for Scala 2. the type placeholder flag is only enabled for 3.2.

kitbellew:
> > I think it should be fine to just change the behaviour for Scala 3 altogether.
> 
> then that's what the latest patch is doing. there is no clear existing type-related exclusively Scala 3 dialect flag; the one to allow `?` for wildcards is now also used for Scala 2. the type placeholder flag is only enabled for 3.2.

instead of adding a new dialect flag, we could also reuse `allowTypeLambdas`, as closely related.

tgodzik:
> > > I think it should be fine to just change the behaviour for Scala 3 altogether.
> > 
> > 
> > then that's what the latest patch is doing. there is no clear existing type-related exclusively Scala 3 dialect flag; the one to allow `?` for wildcards is now also used for Scala 2. the type placeholder flag is only enabled for 3.2.
> 
> instead of adding a new dialect flag, we could also reuse `allowTypeLambdas`, as closely related.

:+1: 

@bjaglin do you think this change can cause any issues in scalafix? Sorry for pinging you, but want to avoid any issues.

bjaglin:
Thanks for the heads up!

As you pointed out, this suggested change impacting pattern matching at runtime will need a (early-semver) minor bump in scalafix to highlight false negatives for rules like https://github.com/xuwei-k/scalafix-rules/blob/main/rules/src/main/scala/fix/Scala3Placeholder.scala. This is the same discussion as https://github.com/scalacenter/scalafix/pull/1556.

I would favor deprecating `Placeholder` in 4.5.x in favor of `AnonymousParam` & `Wildcard` for both Scala 2 & Scala 3, ahead of a clean, binary-breaking removal in 4.6.x. I am not familiar with the `@ast` macro, but would it be possible to do without it by inlining generated code, so that we can tweak `Placeholder#unapply` to be aware of `AnonymousParam` & `Wildcard` in 4.5.x?

kitbellew:
@bjaglin so what you're proposing is:
- do not emit Placeholder at all, regardless of any dialect flags
- implement a new Placeholder.unapply method which returns empty bounds for AnonymousParam and whatever is used inside Wildcard?

tgodzik:
> @bjaglin so what you're proposing is:
> 
> * do not emit Placeholder at all, regardless of any dialect flags
> * implement a new Placeholder.unapply method which returns empty bounds for AnonymousParam and whatever is used inside Wildcard?

That might work, but it might be hard to impllement :thinking: 

@kitbellew could we release the current version with the support for fewer braces? Would be great to get that into Metals before the next release. I will try also to implement that in scalafmt (hopefully that wouldn't be a huge change :fearful: )

kitbellew:
> That might work, but it might be hard to impllement 🤔

Why?

> @kitbellew could we release the current version with the support for fewer braces? Would be great to get that into Metals before the next release. I will try also to implement that in scalafmt (hopefully that wouldn't be a huge change 😨 )

#2810, which was committed shortly after 4.5.13 was released, started this sequence of changes in #2807 (with a few coming after, like #2811); if at all possible, i'd finish it before releasing the next version.

however, if you believe this Placeholder split is not getting enough traction, then we can release now, but might have to stipulate in release notes that the wildcard changes are not yet usable.

tgodzik:
> Why?

Just guessing really, since everything touching the macros is complicated :sweat: 

> however, if you believe this Placeholder split is not getting enough traction, then we can release now, but might have to stipulate in release notes that the wildcard changes are not yet usable.

I think that this change is quite difficult and might still take us a while to agree on a final design :thinking: so that's why I was thinking of doing the release before hand

kitbellew:
> > Why?
> 
> Just guessing really, since everything touching the macros is complicated 😓

I think this one is simpler since all we (likely) need to do is implement `object Placeholder { def unapply(...) = ... }` explicitly, and the macros would fallback on that.

> > however, if you believe this Placeholder split is not getting enough traction, then we can release now, but might have to stipulate in release notes that the wildcard changes are not yet usable.
> 
> I think that this change is quite difficult and might still take us a while to agree on a final design 🤔 so that's why I was thinking of doing the release before hand

if that's so, then please go ahead with release.

bjaglin:
> so what you're proposing is:
>   * do not emit Placeholder at all, regardless of any dialect flags
>   * implement a new Placeholder.unapply method which returns empty bounds for AnonymousParam and whatever is used inside Wildcard?

Precisely. `Placeholder` would also remain available (under deprecation warnings) for programs pretty-printing their own trees. That's another use-case in scalafix rules, although not as common as pattern matching.

> I think this one is simpler since all we (likely) need to do is implement object Placeholder { def unapply(...) = ... } explicitly, and the macros would fallback on that.

Indeed, it seems that the macro code already skips generating `unapply` if it's explicitly defined so hopefully we would just need to capture what's currently generated via `-Ymacro-debug-lite` to port it to an explicit source definition, and extend it with mappings for `AnonymousParam` & `Wildcard` :crossed_fingers: 
https://github.com/scalameta/scalameta/blob/e06a7720551d3b24edf2e6901f0ce751d322ae30/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala#L431-L448

kitbellew:
> > so what you're proposing is:
> > 
> > * do not emit Placeholder at all, regardless of any dialect flags
> > * implement a new Placeholder.unapply method which returns empty bounds for AnonymousParam and whatever is used inside Wildcard?
> 
> Precisely. `Placeholder()` would remain available in 4.5.x (under deprecation warnings) for programs pretty-printing their own trees. That's another use-case in scalafix rules, although not as common as pattern matching.

Done. @tgodzik PTAL.

project/Mima.scala
```diff
@@ -35,6 +35,8 @@ object Mima {
     ProblemFilters.exclude[A]("scala.meta." + metaType)
 
   val apiCompatibilityExceptions: Seq[ProblemFilter] = Seq(
+    // Placeholder
+    exclude[IncompatibleMethTypeProblem]("Type#Placeholder.unapply"),
```
kitbellew, line 39: @bjaglin earlier, `unapply` took a `Placeholder` parameter whereas now it's a `Type` instead. this makes it a source-compatible change but perhaps not a binary-compatible one.

bjaglin: Good point. It does look like https://github.com/xuwei-k/scalafix-rules/blob/v0.2.8/rules/src/main/scala/fix/Scala3Placeholder.scala would fail at runtime against newer versions of scalafix :disappointed: 

```
➜  javap -p -c -cp $(cs fetch com.github.xuwei-k::scalafix-rules:0.2.8 | head -1) 'fix.Scala3Placeholder$$anonfun$fix$1' | grep Placeholder\$\.unapply
      23: invokevirtual #49                 // Method scala/meta/Type$Placeholder$.unapply:(Lscala/meta/Type$Placeholder;)Lscala/Option;
      20: invokevirtual #49                 // Method scala/meta/Type$Placeholder$.unapply:(Lscala/meta/Type$Placeholder;)Lscala/Option;
```

What about adding back that overload for binary compatibility? Would scalac handle it well and discard it in favor of the wider overload when pattern matching on a tree?

kitbellew: i tried this. it complained about unapply overloads. what else can be tried?

bjaglin: Interestingly, it seems that the [error](https://github.com/scala/scala/blob/d978429e66b0c93ee804883abd693d5651c17eef/src/compiler/scala/tools/nsc/typechecker/PatternTypers.scala#L324) is only raised when pattern matching is used from synthetics, not from source. Not sure why...

https://github.com/scalameta/scalameta/blob/a4319baa16618bd848bbcffbb328a56ba1aa362c/scalameta/common/shared/src/main/scala/scala/meta/internal/transversers/transverser.scala#L64 could be rewritten to use the right overload explicitly I guess? But that would mean clients pattern matching on `Placeholder` from macros would get the same problem.

I see you are now exploring the subclassing option :+1: Eventually removing `Placeholder` from the hierarchy will impact binary compatibility for `AnonymousParam` & `Wildcard`, but that can't happen until 5.x anyway, so binary compatibility is best-effort.

bjaglin:
A note about deprecation: I was looking at https://github.com/scalameta/scalameta/blob/main/COMPATIBILITY.md, and it effectively says that deprecation warnings can only be introduced in a major release. Is that enforced? It seems overly strict.

kitbellew:
> A note about deprecation: I was looking at https://github.com/scalameta/scalameta/blob/main/COMPATIBILITY.md, and it effectively says that deprecation warnings can only be introduced in a major release. Is that enforced? It seems overly strict.

no, it's not enforced.

kitbellew:
@bjaglin @tgodzik looks to be working now, no mima errors.

tgodzik (approved):
LGTM! Thanks for working on this, the final solution doesn't actually look that complex

PR #2820, opened 2022-08-29
WIP - Add support for Clause Interleaving

Sporarum:
This is still work in progress, as such it might be better to wait to review

This PR aims to add support for a potentially upcoming feature of Scala 3: Clause Interleaving
This would allow methods as complicated as: `def foo[A][B <: A](using a: A)(b: B)[C <: a.type, D](cd: (C, D))[E]: Foo[A, B, C, D, E]`
For more information, notably motivations, consult: 
SIP: https://github.com/scala/improvement-proposals/pull/47
implementation: https://github.com/lampepfl/dotty/pull/14019

To make sure tooling catches up rapidly to these changes, I am trying to update scalameta in advance, this is this PR

This PR should not be merged until Clause Interleaving makes it into the language !

### DIFF
--- a/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
@@ -340,8 +340,7 @@ object Decl {
   @ast class Def(
       mods: List[Mod],
       name: Term.Name,
-      tparams: List[scala.meta.Type.Param],
-      paramss: List[List[Term.Param]],
+      paramss: List[Clause],
       decltpe: scala.meta.Type
   ) extends Decl with Member.Term with Stat.WithMods
   @ast class Type(
--- a/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
@@ -428,8 +428,11 @@ object Defn {
       decltpe: Option[scala.meta.Type],
       body: Term
   ) extends Defn with Member.Term with Stat.WithMods {
-    val termParamss = paramss.collect{ case Clause.TermClause(termParams) => termParams }
-    checkFields(paramss.collect{ case Clause.TermClause(termParams) => termParams }.forall(onlyLastParamCanBeRepeated))
+    checkFields(paramss.forall {
+      case term: Clause.TermClause =>
+        onlyLastParamCanBeRepeated(term.params)
+      case _ => true
+    })
   }
   @ast class Macro(
       mods: List[Mod],
--- a/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
@@ -324,6 +324,13 @@ object Member {
   }
 }
 
+
+@branch trait Clause extends Tree
+object Clause {
+  @ast class TermClause(params: List[Term.Param]) extends Clause
+  @ast class TypeClause(params: List[Type.Param]) extends Clause
+}
+
 @branch trait Decl extends Stat
 object Decl {
   @ast class Val(mods: List[Mod], pats: List[Pat] @nonEmpty, decltpe: scala.meta.Type)
@@ -417,12 +424,12 @@ object Defn {
   @ast class Def(
       mods: List[Mod],
       name: Term.Name,
-      paramss: List[Either[List[scala.meta.Type.Param],List[Term.Param]]],
+      paramss: List[Clause],
       decltpe: Option[scala.meta.Type],
       body: Term
   ) extends Defn with Member.Term with Stat.WithMods {
-    val termParamss = paramss.collect{ case Right(termParams) => termParams }
-    checkFields(paramss.collect{ case Right(termParams) => termParams }.forall(onlyLastParamCanBeRepeated))
+    val termParamss = paramss.collect{ case Clause.TermClause(termParams) => termParams }
+    checkFields(paramss.collect{ case Clause.TermClause(termParams) => termParams }.forall(onlyLastParamCanBeRepeated))
   }
   @ast class Macro(
       mods: List[Mod],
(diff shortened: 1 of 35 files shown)


kitbellew:
which language requires that? is there an ebnf doc for it? any tests?

Sporarum:
> which language requires that? is there an ebnf doc for it? any tests?
I wasn't expecting reviews yet, as this is a draft PR
I'll edit this the description tomorrow, but here's the SIP: https://github.com/scala/improvement-proposals/pull/47

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -417,12 +417,12 @@ object Defn {
   @ast class Def(
       mods: List[Mod],
       name: Term.Name,
-      tparams: List[scala.meta.Type.Param],
-      paramss: List[List[Term.Param]],
+      paramss: List[Either[List[scala.meta.Type.Param],List[Term.Param]]],
```
kitbellew, line 420: I think this is more likely to be
```
List[(List[Type.Param], @nonEmpty List[Term.Param])]
```
or
```
List[(Option[List[Type.Param]], @nonEmpty List[Term.Param])]
```

since it shouldn't allow `[A][B1, B2](a: A)`, which then doesn't require `Either` handling.

Sporarum: It does allow the above !
And since `[A]: A` is allowed, `@nonEmpty` can't be used
So we really need something like `List[List | List]`, that's more or less how it is implemented in the candidate implementation for clause interleaving

kitbellew: > It does allow the above ! And since `[A]: A` is allowed, `@nonEmpty` can't be used So we really need something like `List[List | List]`, that's more or less how it is implemented in the candidate implementation for clause interleaving

what is `[A]: A`? is that `def foo([...]: ...)`?

If `[...][...]` is not allowed (as it shouldn't, i think), either will not be structured enough.

Sporarum: I meant `def foo[A]: A`, which is already valid

With clause interleaving, `def bar[A][B]` will be allowed

Therefore we really need some form of disjunct union, I chose `Either`, but it could also be done with two `case class`es for example

kitbellew: i would forgo Either and define a new Tree which contains just tparams and params, so that you can capture positions, tokens etc.

however, you must also make changes to parsing of Term.Apply and ApplyType.

kitbellew:
@Sporarum looks like quasiquotes won't work... :(

Sporarum:
@kitbellew What do you mean ?

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -417,12 +424,12 @@ object Defn {
...
       name: Term.Name,
-      tparams: List[scala.meta.Type.Param],
-      paramss: List[List[Term.Param]],
+      paramss: List[Clause],
       decltpe: Option[scala.meta.Type],
       body: Term
   ) extends Defn with Member.Term with Stat.WithMods {
-    checkFields(paramss.forall(onlyLastParamCanBeRepeated))
+    val termParamss = paramss.collect{ case Clause.TermClause(termParams) => termParams }
+    checkFields(paramss.collect{ case Clause.TermClause(termParams) => termParams }.forall(onlyLastParamCanBeRepeated))
```
tgodzik, line 432:
```suggestion
    checkFields(paramss.forall {
      case term: Clause.TermClause =>
        onlyLastParamCanBeRepeated(term.params)
      case _ => true
    })
```
Macros seem to be breaking if unapply is used inside here

Sporarum: Thank you so much, it was that !

Sporarum:
The `q"..$mods def $name[..$tparams](...$paramss): $tpe"` don't work, @kitbellew is that what you meant by quasiquotes ?
How can I fix it ?

kitbellew:
> The `q"..$mods def $name[..$tparams](...$paramss): $tpe"` don't work, @kitbellew is that what you meant by quasiquotes ? How can I fix it ?

yes, that's what i meant. i don't know how you'd fix it.

kitbellew:
@Sporarum fyi, the recommendation is for you to preserve the old parsing and create new top-level trees and put the new parsing under a dialect flag: https://github.com/scalameta/scalameta/issues/405#issuecomment-1237307603

tgodzik:
> @Sporarum fyi, the recommendation is for you to preserve the old parsing and create new top-level trees and put the new parsing under a dialect flag: [#405 (comment)](https://github.com/scalameta/scalameta/issues/405#issuecomment-1237307603)

I can finish that once the SIP is approved and we know in which version this ends up in.

kitbellew:
thanks, @tgodzik .

@Sporarum i would also like to point out that naming (both here and in SIP) is a bit confusing and broad. this is not "clause interleaving" but "parameter clause interleaving". and since we don't have `Term.IfClause` etc, I'd suggest using something simple like `Term.Params`.

Sporarum:
@tgodzik Could you take a look at the PR in its current state ? I think the logic is done, the rest seems to be to fix the tests using q"...def..."

tgodzik:
@Sporarum looks ok! We might actually need to rework it based on the SIP feedback, but let's wait until the proposal is passed. Thanks for the great work!

Sporarum:
Hello,
I wanted to make sure:
Are all changes by this PR superseded by other ones, and if so which ones ?

( I'm assuming yes, and by these ones: https://github.com/scalameta/scalameta/issues/2869 and https://github.com/scalameta/scalameta/pull/2929 )

Sporarum:
Oh and fyi the implementation for interleaving is in the final-final stage, it should be merged today or next week

kitbellew:
> Hello, I wanted to make sure: Are all changes by this PR superseded by other ones, and if so which ones ?
> 
> ( I'm assuming yes, and by these ones: #2869 and #2929 )

yes. since i stopped following discussions around the sip a while ago, please help ensure that the last one implements the final approved version.

Sporarum:
As this has been subsumed, I'll close it
(And edit the PR message to redirect to the correct PR)

PR #2825, opened 2022-09-02
Add Scala 2.13.9 support

dos65:
For now just a draft with snapshot version to check that everything is fine

### DIFF
--- a/semanticdb/scalac/library/src/main/scala-2.13/scala/meta/internal/semanticdb/scalac/CommandLineParser.scala
+++ b/semanticdb/scalac/library/src/main/scala-2.13/scala/meta/internal/semanticdb/scalac/CommandLineParser.scala
@@ -1,0 +1,5 @@
+package scala.meta.internal.semanticdb.scalac
+
+object CommandLineParser {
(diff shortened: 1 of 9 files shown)


semanticdb/scalac/library/src/main/scala-2.13/scala/meta/internal/semanticdb/scalac/CommandLineParser.scala
```diff
@@ -0,0 +1,5 @@
+package scala.meta.internal.semanticdb.scalac
+
+object CommandLineParser {
+  def tokenize(line: String): List[String] = scala.sys.process.Parser.tokenize(line)
```
tgodzik, line 4: Why did this pop up only in 2.13.9? :thinking:

dos65: In scala/scala this class was deleted and replaced by sys.process one after 2.13.8 - https://github.com/scala/scala/pull/10057

tgodzik (approved):

PR #2830, opened 2022-09-04
bugfix: Show better error when no `{` is before case

tgodzik:
Related to https://github.com/scalacenter/scastie/issues/636

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1393,7 +1393,9 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     if (acceptOpt[Indentation.Indent]) {
       autoEndPos(t)(Term.Match(t, indentedAfterOpen(caseClauses())))
     } else {
-      autoEndPos(t)(Term.Match(t, inBracesOrNil(caseClauses())))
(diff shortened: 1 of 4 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1393,7 +1393,15 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
-      autoEndPos(t)(Term.Match(t, inBracesOrNil(caseClauses())))
+      autoEndPos(t)(
+        Term.Match(
+          t,
+          inBracesOr(
+            caseClauses(),
+            syntaxError(s"'{' expected but '$token' found. ", at = token.pos)
+          )
+        )
+      )
```
kitbellew, line 1404:
```suggestion
      autoEndPos(t)(Term.Match(t, inBracesOr(caseClauses(), syntaxErrorExpected[LeftBrace])))
```

kitbellew (changes requested):
minor request: `syntaxErrorExpected` already does exactly how you invoke `syntaxError`.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Thanks! Fixed

PR #2832, opened 2022-09-05
ScalametaParser: allow newline after pat infix op

kitbellew:
Helps with scalameta/scalafmt#3313.


kitbellew:
hmm... intellij had no problem with this syntax but the compiler did. never mind.

PR #2829, opened 2022-09-03
ScalametaParser: look for one ellipsis in params

kitbellew:

PR #2833, opened 2022-09-05
ScalametaParser: minor refactor in params parsing

kitbellew:
Replaces #2829.


tests/shared/src/test/scala/scala/meta/tests/quasiquotes/ErrorSuite.scala
```diff
@@ -768,7 +768,6 @@ class ErrorSuite extends FunSuite {
       import scala.meta.dialects.Scala211
       ??? match {
         case q"$_(...$argss)(...$_)" =>
-        case q"$_(...$argss)(foo)(...$_)" =>
```
tgodzik, line 771: Why did you remove it from the error suite?

PR #2836, opened 2022-09-06
fix: allow export in extesion block

dos65:
Reported in Discord: https://discord.com/channels/632642981228314653/632652693013528589/1016780987658682398

Syntax added in 3.2.0: https://scala-lang.org/blog/2022/09/05/scala-3.2.0-released.html#exports-in-extension-clauses

### DIFF
--- a/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
@@ -114,8 +114,7 @@ object Term {
     })
   }
   @ast class Block(stats: List[Stat]) extends Term {
-    // extension group block can have declarations without body too
-    checkFields(stats.forall(st => st.isBlockStat || st.is[Decl]))
+    checkParent(ParentChecks.TermBlock)
   }
   @ast class EndMarker(name: Term.Name) extends Term
   @ast class If(cond: Term, thenp: Term, elsep: Term) extends Term {
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/ParentChecks.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/ParentChecks.scala
@@ -107,4 +107,13 @@ object ParentChecks {
   def TypeMethod(tree: Type.Method, parent: Tree, destination: String): Boolean = {
     parent.is[Type] || parent.is[Defn.Type]
   }
+
+  def TermBlock(tree: Term.Block, parent: Tree, dest: String): Boolean = {
+    tree.stats.forall {
+      case _: Decl => true
+      case _: Export => parent.is[Defn.ExtensionGroup]
+      case x => x.isBlockStat
+    }
+  }
+
 }
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/ExtensionMethodsSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/ExtensionMethodsSuite.scala
@@ -399,6 +399,40 @@ class ExtensionMethodsSuite extends BaseDottySuite {
     )
   }
 
+  test("private-export") {
+    runTestAssert[Stat](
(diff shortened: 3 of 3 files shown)


scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -114,8 +114,8 @@ object Term {
     })
   }
   @ast class Block(stats: List[Stat]) extends Term {
-    // extension group block can have declarations without body too
-    checkFields(stats.forall(st => st.isBlockStat || st.is[Decl]))
+    // extension group block can have declarations or export without body too
+    checkFields(stats.forall(st => st.isBlockStat || st.is[Decl] || st.is[Export]))
```
kitbellew, line 118: if this is allowed **only** in extensions, then i think it's time to change this to a `checkParent` call instead. adding something like this to `ParentChecks.scala`:
```
def TermBlock(tree: Term.Block, parent: Tree, dest: String): Boolean = {
  val isExportOK = parent.is[Defn.ExtensionGroup]
  tree.stats.forall {
    case _: Decl => true
    case _: Export => isExportOK
    case x => x.isBlockStat
  }
}
```
and then here
```
checkParent(ParentChecks.TermBlock)
```

dos65: @kitbellew Thx, applied your suggestion!

tgodzik: It seems that has broken the code :thinking: I think you might need to inline isExportOK

tgodzik: ` scala.meta.tests.parsers.dotty.MinorDottySuite.type-in-block` is failing

dos65: fixed

scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/ParentChecks.scala
```diff
@@ -107,4 +107,13 @@ object ParentChecks {
   def TypeMethod(tree: Type.Method, parent: Tree, destination: String): Boolean = {
     parent.is[Type] || parent.is[Defn.Type]
   }
+
+  def TermBlock(tree: Term.Block, parent: Tree, dest: String): Boolean = {
+    val isExportOK = parent.is[Defn.ExtensionGroup]
+    tree.stats.forall {
+      case x if x.isBlockStat => x.isBlockStat
```
kitbellew, line 114: it's always true on the right, no?

also, forall will now throw a match exception instead of returning false.

kitbellew (changes requested):

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
kitbellew: > if this is allowed **only** in extensions, then i think it's time to change this to a `checkParent` call instead. adding something like this to `ParentChecks.scala`:
> 
> ```
> def TermBlock(tree: Term.Block, parent: Tree, dest: String): Boolean = {
>   val isExportOK = parent.is[Defn.ExtensionGroup]
>   tree.stats.forall {
>     case _: Decl  | _: Export => isExportOK
>     case x => x.isBlockStat
>   }
> }
> ```
> 
> and then here
> 
> ```
> checkParent(ParentChecks.TermBlock)
> ```

looks like i relied on the comment too much (the one that @kpbochenek wrote a white ago, before @tgodzik added support for types anywhere). the comment needs to be changed to something like
```
// dotty block can have type declarations, or exports if within extension group, without body too
```

would this work:
```
def TermBlock(tree: Term.Block, parent: Tree, dest: String): Boolean = {
  val isExportOK = parent.is[Defn.ExtensionGroup]
  tree.stats.forall {
    case _: Decl  => true
    case _: Export => isExportOK
    case x => x.isBlockStat
  }
}
```

PR #2837, opened 2022-09-07
ScalametaParser: in param, pass Mod tree, not flag

kitbellew:
Currently, we indicate whether we saw a `using` or `implicit` token in parameter parsing, and generate a new Mod tree that doesn't necessarily point to that keyword but to the current position.

Instead, let's generate the Mod tree over the keyword and pass it along.


tgodzik (approved):
LGTM!

PR #2839, opened 2022-09-07
Trees: become[] to declare base type, not Quasi

kitbellew:


tgodzik (approved):
Looks ok, but I am not that familiar with quasi quotes to see exactly why these were needed. Looks simpler for sure and doesn't seem to break any tests.

kitbellew:
> Looks ok, but I am not that familiar with quasi quotes to see exactly why these were needed. Looks simpler for sure and doesn't seem to break any tests.

i am not too familiar either but this change paves the way for a few additional simplifications.

PR #2840, opened 2022-09-08
ScalametaParser: quasi() returns T with Quasi

kitbellew:
Since it's now guaranteed that the returned type is a Quasi, it allows us to simplify the code a bit and remove some unnecessary checks and associated failure conditions.


kitbellew:
Follow-on to #2839.

tgodzik (approved):
The change looks ok, but could you add a bit more info to the PR why is this needed?

kitbellew:
> The change looks ok, but could you add a bit more info to the PR why is this needed?

since it's now known that the result is a quasi, it simplified the code a little, removed some unnecessary checks and failure conditions.

PR #2842, opened 2022-09-08
Add test with underscores in match type

kitbellew:
It demonstrates the current bug in placeholder-vs-wildcard handling.


kitbellew:
this bug is partially fixed in #2815

dos65 (approved):

PR #2846, opened 2022-09-09
adt/Reflection: extract checking hierarchy method

kitbellew:
Removes code duplication and maintenance overhead.


tgodzik (approved):
LGTM!

PR #2848, opened 2022-09-12
feature: Add support for fewer braces

tgodzik:
More information: https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html#variant-indentation-marker--for-arguments

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
@@ -150,7 +150,9 @@ final class Dialect private (
     // Scala 3 uses proper precedence rules for infix types, unlike Scala 2
     val useInfixTypePrecedence: Boolean,
     // Scala213Source3 and Scala3 allow infix operator being placed after nl
-    val allowInfixOperatorAfterNL: Boolean
+    val allowInfixOperatorAfterNL: Boolean,
+    // Scala 3 will allow dropping braces for block arguments such as `list.map: a =>`
+    val allowFewerBraces: Boolean
 ) extends Product with Serializable {
 
   // NOTE(olafur) checklist for adding a new dialect field in a binary compatible way:
@@ -240,7 +242,8 @@ final class Dialect private (
       allowUnderscoreAsTypePlaceholder = false,
       allowGivenImports = false,
       useInfixTypePrecedence = false,
-      allowInfixOperatorAfterNL = false
+      allowInfixOperatorAfterNL = false,
+      allowFewerBraces = false
       // NOTE(olafur): declare the default value for new fields above this comment.
     )
   }
@@ -451,6 +454,10 @@ final class Dialect private (
     privateCopy(allowInfixOperatorAfterNL = newValue)
   }
 
+  def withAllowFewerBraces(newValue: Boolean): Dialect = {
+    privateCopy(allowFewerBraces = newValue)
+  }
+
   // NOTE(olafur): add the next `withX()` method above this comment. Please try
   // to use consistent formatting, use `newValue` as the parameter name and wrap
   // the body inside curly braces.
@@ -514,7 +521,8 @@ final class Dialect private (
       allowUnderscoreAsTypePlaceholder: Boolean = this.allowUnderscoreAsTypePlaceholder,
       allowGivenImports: Boolean = this.allowGivenImports,
       useInfixTypePrecedence: Boolean = this.useInfixTypePrecedence,
-      allowInfixOperatorAfterNL: Boolean = this.allowInfixOperatorAfterNL
+      allowInfixOperatorAfterNL: Boolean = this.allowInfixOperatorAfterNL,
+      allowFewerBraces: Boolean = this.allowFewerBraces
       // NOTE(olafur): add the next parameter above this comment.
   ): Dialect = {
     new Dialect(
@@ -575,7 +583,8 @@ final class Dialect private (
       allowUnderscoreAsTypePlaceholder,
       allowGivenImports,
       useInfixTypePrecedence,
-      allowInfixOperatorAfterNL
+      allowInfixOperatorAfterNL,
+      allowFewerBraces
       // NOTE(olafur): add the next argument above this comment.
     )
   }
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
@@ -151,7 +151,9 @@ final class Dialect private (
     val useInfixTypePrecedence: Boolean,
     // Scala213Source3 and Scala3 allow infix operator being placed after nl
     val allowInfixOperatorAfterNL: Boolean,
-    // Scala 3 will allow dropping braces for block arguments such as `list.map: a =>`
+    /* Scala 3 allows dropping braces for block arguments such as `list.map: a =>`
+     * It wasn't available in Scala 3.0 and got introduced later.
+     */
     val allowFewerBraces: Boolean
 ) extends Product with Serializable {
 
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2166,7 +2166,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
            * .filter: x =>
            *   x > 2
            * ```
-           * Witout manual handling here, filter would be included for `(a+1).filter`
+           * Without manual handling here, filter would be included for `(a+1).filter`
            */
           val param = simpleExpr(allowRepeated = false)
           val contextFunction = token.is[ContextArrow]
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1561,7 +1561,6 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           val arguments = addPos { Term.Apply(t, List(args)) }
           t = simpleExprRest(arguments, canApply = true, startPos = startPos)
         } else if (acceptOpt[Colon]) {
-
           if (token.is[At] || (token.is[Ellipsis] && ahead(token.is[At]))) {
             t = addPos(Term.Annotate(t, annots(skipNewLines = false)))
           } else if (token.is[Underscore] && ahead(isStar)) {
@@ -1723,8 +1722,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       syntaxError(s"not a legal formal parameter", at = other)
   }
   private def convertToParams(tree: Term): List[Term.Param] = tree match {
-    case Term.Tuple(ts) => ts.toList flatMap convertToParam
-    case _ => List(convertToParam(tree)).flatten
+    case Term.Tuple(ts) => ts.flatMap(convertToParam)
+    case _ => convertToParam(tree).toList
   }
 
   def implicitClosure(location: Location): Term.Function = {
@@ -2157,6 +2156,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           in.observeIndented()
           blockExpr(allowRepeated = false)
         } else {
+
           /**
            * We need to handle param and then open indented region, otherwise only the block will be
            * handles and any `.` will be accepted into the block:
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -153,6 +153,10 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   /* ------------- TOKEN STREAM HELPERS -------------------------------------------- */
 
   def isColonEol(token: Token): Boolean = {
+    token.is[Colon] && isEolAfterColon(token)
+  }
+
+  def isEolAfterColon(token: Token): Boolean = {
 
     @tailrec
     def isNextEOL(t: Token): Boolean = {
@@ -163,7 +167,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       }
     }
 
-    dialect.allowSignificantIndentation && token.is[Colon] && isNextEOL(token)
+    dialect.allowSignificantIndentation && isNextEOL(token)
   }
 
   /* ------------- PARSER-SPECIFIC TOKENS -------------------------------------------- */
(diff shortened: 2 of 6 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2146,6 +2153,36 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     }
   }
 
+  private def followingIsLambdaAfterColon(): Boolean = {
+    /** Skip matching pairs of `(...)` or `[...]` parentheses.
+     *  @pre  The current token is `(` or `[`
+     */
+    def skipParens[T1: TokenClassifier, T2: TokenClassifier]: Unit = {
```
kitbellew, line 2160: slightly rewritten, let me know if it's more readable:
```
def skipParens[T1: TokenClassifier, T2: TokenClassifier]: Unit = {
  // starts on left delimiter
  @tailrec
  def iter = {
    next()
    if (token.is[EOF]) false
    else if (token.is[T2]) true
    else  (token.isNot[T1] || skipParens[T1, T2]) && iter
  }
  // stops on right delimiter (true), or end (false)
  iter
}

ahead {
  val okParams = token match {
    case _: Ident | _: Underscore => true
    case _: LeftParen => skipParens[LeftParen, RightParen]
    case _: LeftBracket => skipParens[LeftBracket, RightBracket]
    case _ => false
  }
  okParams && {
    next()
    token.is[RightArrow] || token.is[ContextArrow]
  } && {
    next()
    token.is[Indentation.Indent]
  }
}
```

kitbellew (commented):
lgtm. a few comments or suggestions.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2146,6 +2153,36 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     }
   }
 
+  private def followingIsLambdaAfterColon(): Boolean = {
```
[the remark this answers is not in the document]
tgodzik: That's taken care of by checking the indentation token

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Thanks! I looks better, I already rewrote it a bit from the Dotty version, but it was still lacking :sweat_smile:

tgodzik:
> lgtm. a few comments or suggestions.

I realized, we need to handle infix also and I want to copy over some tests from dotty. Will continue tomorrow!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: could also do this, perhaps, without skipParens recursion:
```
def skipParens[T1: TokenClassifier, T2: TokenClassifier]: Unit = {
  // starts on left delimiter
  @tailrec
  def iter(nest: Int) = {
    next()
    if (token.is[EOF]) false
    else if (token.is[T2]) nest == 0 || iter(nest - 1)
    else (token.is[T1]) iter(nest + 1)
    else iter(nest)
  }
  // stops on right delimiter (true), or end (false)
  iter(0)
}
```

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/FeverBracesSuite.scala
```diff
@@ -0,0 +1,236 @@
+package scala.meta.tests.parsers.dotty
+
+import scala.meta.tests.parsers._
+import scala.meta._
+import scala.meta.tests.tokenizers.TokenizerSuite
+
+class FeverBracesSuite extends BaseDottySuite {
+
+  override implicit val dialect = dialects.Scala3.withAllowFewerBraces(true)
```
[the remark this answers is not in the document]
tgodzik: > I'm wondering if we are going to enable it by default for scala3.

We might, but I was thinking a safer approach would be to do it gradually.

> Otherwise we will have to add new options to scalafmt settings.

We can already change a single dialect options thanks to the changes by @kitbellew 

> Also, as it might be enabled by import we would need the same mechanic

We can enable it by default for nightly versions first and change it to stable later on.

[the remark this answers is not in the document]
dos65: > We can already change a single dialect options thanks to the changes by @kitbellew

oh, cool. didn't know that!

Then we would need only import mechanic but we can do it separate PR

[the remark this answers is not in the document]
dos65: I mean the case when you parse the source using std Scala3 dialect and if it contains:
`import import language.experimental.fewerBraces` then parser treats  it as  it was called with `Scala3.withAllowFewerBraces(true)` dialect

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -150,7 +150,9 @@ final class Dialect private (
     // Scala 3 uses proper precedence rules for infix types, unlike Scala 2
     val useInfixTypePrecedence: Boolean,
     // Scala213Source3 and Scala3 allow infix operator being placed after nl
-    val allowInfixOperatorAfterNL: Boolean
+    val allowInfixOperatorAfterNL: Boolean,
+    // Scala 3 will allow dropping braces for block arguments such as `list.map: a =>`
+    val allowFewerBraces: Boolean
```
kitbellew, line 155: one additional point: currently, "fewer braces" is the name of the experimental import to enable it in the code but at some point it will be unnecessary, and then this name might become confusing. should we make it something like allowLambdaOptionalBraces?

tgodzik: I think it should be good enough to keep the information in the comment and it will be easier for people currently to enable the dialect option. In the future, it might only be a problem for contributors, but the comment should suffice.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2138,6 +2138,13 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           }
         }
         simpleExprRest(arguments, canApply = true, startPos = startPos)
+      case _: Colon if dialect.allowFewerBraces && (isColonEol(token) || followingIsLambdaAfterColon()) =>
+        accept[Colon]
+        in.observeIndented()
+        val args = blockExpr(allowRepeated = false)
+        val arguments = addPos { Term.Apply(t, List(args)) }
+        simpleExprRest(arguments, canApply = true, startPos = startPos)
+
```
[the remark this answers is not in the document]
tgodzik: Removed

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2138,6 +2138,13 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           }
         }
         simpleExprRest(arguments, canApply = true, startPos = startPos)
+      case _: Colon if dialect.allowFewerBraces && (isColonEol(token) || followingIsLambdaAfterColon()) =>
+        accept[Colon]
```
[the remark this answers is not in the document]

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2146,6 +2153,36 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+          next()
+          token.is[Indentation.Indent]
+        }
+      if (token.is[Ident] || token.is[Underscore]) {
+        next()
+        isArrowIndent()
+      } else if (token.is[LeftParen]) {
+        skipParens[LeftParen, RightParen]
+        isArrowIndent()
+
```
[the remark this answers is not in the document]
tgodzik: Removed

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2138,6 +2138,13 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           }
         }
         simpleExprRest(arguments, canApply = true, startPos = startPos)
+      case _: Colon if dialect.allowFewerBraces && (isColonEol(token) || followingIsLambdaAfterColon()) =>
```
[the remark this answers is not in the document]
tgodzik: Done! Thanks for the suggestion!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Good idea, I also added a test which has multiple `(` and `[`. I had to slightly rewrite TreeSyntax since it would always change to `{` if type was defined for parameter, but it's actually not needed since we already wrap it in `()` for all cases beside implicit and using

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2138,6 +2149,12 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           }
         }
         simpleExprRest(arguments, canApply = true, startPos = startPos)
+      case _: Colon if allowFewerBraces =>
```
kitbellew, line 2152: last question: do we need to check for `canApply` here (and is there a test which confirms that)?

kitbellew (approved):
one last comment below

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Added, it shouldn't break, but it should always be in an apply.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1637,60 +1648,6 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
             val contextFunction = token.is[ContextArrow]
             next()
             t = addPos {
-              def convertToParam(tree: Term): Option[Term.Param] = tree match {
```
tgodzik, line 1640: Moved outside

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1550,7 +1554,14 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
             case _ =>
           }
+        } else if (token.is[Colon] && allowFewerBraces) {
+          next()
+          in.observeIndented()
+          val args = blockExpr(allowRepeated = false)
+          val arguments = addPos { Term.Apply(t, List(args)) }
+          t = simpleExprRest(arguments, canApply = true, startPos = startPos)
         } else if (acceptOpt[Colon]) {
+
```
kitbellew, line 1564: extra empty line

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1716,6 +1672,61 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+        )
+      )
+    case Lit.Unit() =>
+      None
+    case other =>
+      syntaxError(s"not a legal formal parameter", at = other)
+  }
+  private def convertToParams(tree: Term): List[Term.Param] = tree match {
+    case Term.Tuple(ts) => ts.toList flatMap convertToParam
+    case _ => List(convertToParam(tree)).flatten
```
kitbellew, line 1727: or
```
  case Term.Tuple(ts) => ts.flatMap(convertToParam) // if tuple args are already a list; or toList at the end
  case _ => convertToParam(tree).toList // no need to flatten
```

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -1140,9 +1140,11 @@ object TreeSyntax {
     // Multiples and optionals
     implicit def syntaxArgs: Syntax[List[Term]] = Syntax {
       case (b: Term.Block) :: Nil => s(" ", b)
-      case (f @ Term.Function(params, _)) :: Nil if !params.exists(_.decltpe.isEmpty) =>
+      case (f @ Term.Function(params, _)) :: Nil
+          if params.exists(_.mods.exists(m => m.is[Mod.Implicit] || m.is[Mod.Using])) =>
         s(" { ", f, " }")
-      case args => s("(", r(args, ", "), ")")
+      case args =>
+        s("(", r(args, ", "), ")")
```
kitbellew, line 1147: what's the change here?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2138,6 +2149,39 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           }
         }
         simpleExprRest(arguments, canApply = true, startPos = startPos)
+      case _: Colon if canApply && allowFewerBraces =>
+        val isEol = isEolAfterColon(token)
+        next()
+        // map:
+        val args = if (isEol) {
+          in.observeIndented()
```
kitbellew, line 2157: can this be
```
  map: /*
    line
    line */
   indentedCode
```

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -1140,7 +1140,8 @@ object TreeSyntax {
     // Multiples and optionals
     implicit def syntaxArgs: Syntax[List[Term]] = Syntax {
       case (b: Term.Block) :: Nil => s(" ", b)
-      case (f @ Term.Function(params, _)) :: Nil if !params.exists(_.decltpe.isEmpty) =>
```
kitbellew, line 1143: forgot to ask: why this change?

tgodzik: It was adding {} for `apply( (a: String) => a )` which is not neccessary and was breaking one of the added tests, since {} would be a different tree

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Works!

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
tgodzik: Removed

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Changed!

kitbellew (approved):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2138,6 +2148,40 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+          /**
+           * We need to handle param and then open indented region, otherwise only the block will be
+           * handles and any `.` will be accepted into the block:
+           * ```
+           * .map: a =>
+           *   a+1
+           * .filter: x =>
+           *   x > 2
+           * ```
+           * Witout manual handling here, filter would be included for `(a+1).filter`
```
dos65, line 2169:
```suggestion
           * Without manual handling here, filter would be included for `(a+1).filter`
```

dos65 (approved):
:fire:

tgodzik:
Now we need to handle it in `scalafmt` :sweat_smile: I will take a look, there is one more fix I want to include in the release.

kitbellew:
> Now we need to handle it in `scalafmt` 😅 I will take a look, there is one more fix I want to include in the release.

and i have a couple of additional fixes on top of #2815.

PR #2849, opened 2022-09-12
ScalametaParser: param ellipses in extension group

kitbellew:
Currently, quasiquote ellipses are not supported when parsing parameters of extension groups.

Instead, let's move the rank-1 ellipsis parsing into the param() method and add the rank-2 parsing into the extension group invocation.

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/ExtensionMethodsSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/ExtensionMethodsSuite.scala
@@ -7,6 +7,9 @@ class ExtensionMethodsSuite extends BaseDottySuite {
 
   implicit val parseBlock: String => Stat = code => blockStat(code)(dialects.Scala3)
 
+  private final val cparam = tparam("c", "Circle")
+  private final val cparamss = List(List(cparam))
+
   /**
    * For checking examples in repl declare:
    * {{{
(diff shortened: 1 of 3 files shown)


tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/ExtensionMethodsSuite.scala
```diff
@@ -433,7 +435,40 @@ class ExtensionMethodsSuite extends BaseDottySuite {
     )
   }
 
-  final val defcrc = Defn.Def(Nil, tname("crc"), Nil, Nil, Some(pname("Int")), int(2))
+  test("extension quasiquotes 1") {
+    import dialects.Scala3
+
+    val q"extension [..$tparams](...$params) { ..$stats }" =
```
tgodzik, line 441: Could we add a tests with `tparams`?

tgodzik (approved):
Just one nit, but otherwise LGTM!

dos65 (approved):

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/ExtensionMethodsSuite.scala
kitbellew: added. the code parsing tparams is the same as elsewhere, unlike the paramss parsing, but you are right, tests don't need to know that.

tgodzik: Thanks!

PR #2854, opened 2022-09-14
bugfix: Allow refinement after with in type

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/2853

Previously, we would not accept `type T = AnyRef with<Indentation>`. Now, we accept the statements in after indentation as refinements.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
@@ -133,6 +133,10 @@ private[parsers] class LazyTokenIterator private (
       foundIndentation
   }
 
+  def previousIndentation: Int = {
+    sepRegions.lift(1).fold(0)(_.indent)
+  }
+
   def observeOutdented(): Boolean = {
     if (!dialect.allowSignificantIndentation) false
     else {
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -881,9 +881,15 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     def compoundTypeRest(typ: Type): Type = {
       val startPos = typ.startTokenPos
       var t = typ
-      while (acceptOpt[KwWith]) {
+      // Indentation means a refinement and we cannot join refinements this way
+      while (acceptOpt[KwWith] && !token.is[Indentation.Indent]) {
         val rhs = annotType()
         t = atPos(startPos, rhs)(Type.With(t, rhs))
+      }
+      token match {
+        case Indentation.Indent() =>
+          t = autoPos(Type.Refine(Some(t), indented(refineStatSeq())))
+        case _ =>
       }
       if (isAfterOptNewLine[LeftBrace]) refinement(innerType = Some(t))
       else t
@@ -4211,7 +4217,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   def refineStatSeq(): List[Stat] = listBy[Stat] { stats =>
     while (!token.is[StatSeqEnd]) {
       refineStat().foreach(stats += _)
-      if (token.isNot[RightBrace]) acceptStatSep()
+      if (token.isNot[RightBrace] && token.isNot[Indentation.Outdent]) acceptStatSep()
     }
   }
 
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -880,17 +880,25 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
     def compoundTypeRest(typ: Type): Type = {
       val startPos = typ.startTokenPos
-      var t = typ
-      // Indentation means a refinement and we cannot join refinements this way
-      while (acceptOpt[KwWith] && !token.is[Indentation.Indent]) {
-        val rhs = annotType()
-        t = atPos(startPos, rhs)(Type.With(t, rhs))
+
+      @tailrec
+      def gatherWithTypes(previousType: Type): Type = {
+        if (acceptOpt[KwWith]) {
+          /* Indentation means a refinement and we cannot join
+           * refinements this way so stop looping.
+           */
+          if (token.is[Indentation.Indent]) {
+            autoPos(Type.Refine(Some(previousType), indented(refineStatSeq())))
+          } else {
+            val rhs = annotType()
+            val t = atPos(startPos, rhs)(Type.With(previousType, rhs))
+            gatherWithTypes(t)
+          }
+        } else {
+          previousType
+        }
       }
-      token match {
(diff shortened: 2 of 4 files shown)


dos65 (approved):
:+1:

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -877,10 +877,16 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
       val startPos = typ.startTokenPos
       var t = typ
-      while (acceptOpt[KwWith]) {
+      // Indentation means a refinement and we cannot join refinements this way
+      while (acceptOpt[KwWith] && !token.is[Indentation.Indent]) {
         val rhs = annotType()
         t = atPos(startPos, rhs)(Type.With(t, rhs))
       }
+      token match {
+        case Indentation.Indent() =>
```
kitbellew, line 886: can indentation come **without** a `with`? the code allows that.

tgodzik: We shouldn't have an indetation without a `with`, but it's a good idea to make sure!

tgodzik: Added a test and refactored the function to use tail recursion. This way we don't check any tokens twice.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -880,11 +880,25 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+      def gatherWithTypes(previousType: Type): Type = {
+        if (acceptOpt[KwWith]) {
+          /* Indentation means a refinement and we cannot join
+           * refinements this way so stop looping.
+           */
+          if (token.is[Indentation.Indent]) {
+            autoPos(Type.Refine(Some(previousType), indented(refineStatSeq())))
+          } else {
+            val rhs = annotType()
+            val t = atPos(startPos, rhs)(Type.With(previousType, rhs))
```
kitbellew, line 894: I'd use autoEndPos here, getting pos from tree requires a number of steps but getting current end position from token stream is easy.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -880,11 +880,25 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+            gatherWithTypes(t)
+          }
+        } else {
+          previousType
+        }
       }
+      val t = gatherWithTypes(typ)
```
kitbellew, line 901: this is about the next line (github is still pretty bad).

can left brace follow significant indentation like that? can you mix refinement with `with`/indentation and that with left brace?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -880,11 +880,25 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
     def compoundTypeRest(typ: Type): Type = {
       val startPos = typ.startTokenPos
-      var t = typ
-      while (acceptOpt[KwWith]) {
-        val rhs = annotType()
-        t = atPos(startPos, rhs)(Type.With(t, rhs))
+
+      @tailrec
```
kitbellew, line 884: thank you for this solution. also avoids `var`....

kitbellew (changes requested):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: ok, sure! I just copied it over really.

kitbellew: ah, sorry, didn't realize. never mind.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Ok, now it should work the same as in the compiler, though the code got a bit more ugly :/

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -880,13 +884,35 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+        }
       }
-      if (isAfterOptNewLine[LeftBrace]) refinement(innerType = Some(t))
-      else t
+
+      val t = gatherWithTypes(typ)
+      val wasLF = token.is[LF]
+
+      // the indentation needs to be higher than the containing one
+      def canAddBracesRefinement = !dialect.allowSignificantIndentation || !wasLF ||
```
tgodzik, line 910: Had to add this condition to differentiate between:
```
type A = AnyRef with
    type T = Int
 { 
    type D = Double
 }
```

and 

```
type A = AnyRef with
    type T = Int
{ 
    type D = Double
}
```

Seems an edge case, but might be better to be correct.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
```diff
@@ -133,6 +133,10 @@ private[parsers] class LazyTokenIterator private (
       foundIndentation
   }
 
+  def previousIndentation: Int = {
+    sepRegions.lift(1).fold(0)(_.indent)
```
kitbellew, line 137: what's lift(1)? is it headOption?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -880,13 +884,35 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
+        dialect.allowSignificantIndentation && currentIndentation > previousIndentation
```
kitbellew, line 911: you don't need to check the dialect flag again

kitbellew (changes requested):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
tgodzik: it takes the second element of the list if it exists

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Ach, right! Thanks!

kitbellew (approved):
lgtm. sorry i nitpicked long after @dos65 approved :)

tgodzik:
> lgtm. sorry i nitpicked long after @dos65 approved :)

No worries! That was all good comments!

PR #2866, opened 2022-09-16
bugfix: Don't allow open for enum cases

tgodzik:
Follow up from https://github.com/lampepfl/dotty/pull/15961

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -3295,7 +3295,12 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       case KwExtension() =>
         extensionGroupDecl(mods)
       case KwCase() if dialect.allowEnums && enumCaseAllowed && ahead(token.is[Ident]) =>
-        enumCaseDef(mods)
+        mods.find(mod => !mod.isAccessMod && !mod.is[Mod.Annot]) match {
+          case Some(mod) =>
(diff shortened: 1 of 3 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3295,7 +3295,12 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       case KwExtension() =>
         extensionGroupDecl(mods)
       case KwCase() if dialect.allowEnums && enumCaseAllowed && ahead(token.is[Ident]) =>
-        enumCaseDef(mods)
+        mods.find(_.is[Mod.Open]) match {
```
tgodzik, line 3298: I wonder if we should put it in checks, but that cause invariant exceptions, which are not pretty :thinking:

dos65 (approved):
:+1:

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3295,7 +3295,12 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       case KwExtension() =>
         extensionGroupDecl(mods)
       case KwCase() if dialect.allowEnums && enumCaseAllowed && ahead(token.is[Ident]) =>
-        enumCaseDef(mods)
+        mods.find(mod => !mod.isAccessMod && !mod.is[Mod.Annot]) match {
```
tgodzik, line 3298: Actually, only access mods and annotations are allowe here.

PR #2884, opened 2022-09-23
Trees: add AnonymousLambda

kitbellew:
Wraps types referring to AnonymousParam. Helps with #2807.

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
@@ -274,9 +274,11 @@ class MatchTypeSuite extends BaseDottySuite {
                 Type.Name("L"),
                 List(
                   TypeCase(
-                    Type.Apply(
-                      Type.Name("Foo"),
-                      List(Type.AnonymousParam(None))
+                    Type.AnonymousLambda(
+                      Type.Apply(
+                        Type.Name("Foo"),
+                        List(Type.AnonymousParam(None))
+                      )
                     ),
                     Type.Name("L")
                   ),
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/TypeSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/TypeSuite.scala
@@ -317,13 +317,13 @@ class TypeSuite extends BaseDottySuite {
 
   test("F[_]") {
     assertTpe("F[_]") {
-      Apply(TypeName("F"), AnonymousParam(None) :: Nil)
+      AnonymousLambda(Apply(TypeName("F"), List(AnonymousParam(None))))
     }
     assertTpe("F[+_]") {
-      Apply(TypeName("F"), AnonymousParam(Some(Mod.Covariant())) :: Nil)
+      AnonymousLambda(Apply(TypeName("F"), List(AnonymousParam(Some(Mod.Covariant())))))
     }
     assertTpe("F[-_]") {
-      Apply(TypeName("F"), AnonymousParam(Some(Mod.Contravariant())) :: Nil)
+      AnonymousLambda(Apply(TypeName("F"), List(AnonymousParam(Some(Mod.Contravariant())))))
     }
   }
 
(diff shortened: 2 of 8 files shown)


tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
```diff
@@ -274,9 +274,11 @@ class MatchTypeSuite extends BaseDottySuite {
                 Type.Name("L"),
                 List(
                   TypeCase(
-                    Type.Apply(
-                      Type.Name("Foo"),
-                      List(Type.AnonymousParam(None))
+                    Type.AnonymousLambda(
```
tgodzik, line 277: Isn't this actually just a placholder? `Foo[_] ` it's not a lambda

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/TypeSuite.scala
```diff
@@ -317,13 +317,13 @@ class TypeSuite extends BaseDottySuite {
 
   test("F[_]") {
     assertTpe("F[_]") {
-      Apply(TypeName("F"), AnonymousParam(None) :: Nil)
+      AnonymousLambda(Apply(TypeName("F"), List(AnonymousParam(None))))
```
tgodzik, line 320: Will this not be a Lambda only after we disable the possibility to use `_` as placeholder?

tgodzik (changes requested):

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
kitbellew: but for scala32, `_` is not a wildcard anymore, hence `F[_]` is actually only valid as a lambda.

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/TypeSuite.scala
kitbellew: in scala 3.2 (and scala3 points to 3.2), `_` *is* a placeholder (aka, anonymous parameter for type lambda) but not a _wildcard_.

kitbellew (commented):
there's still confusion about the meaning of placeholder (it's anonymous parameter in type lambdas, not a substitute-any-type wildcard).

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
tgodzik: I don't think it happened yet despite what is in the plans.

tgodzik: IF I reprint the statement with Dotty I get back:
```
object match_types:
  type Combine[L, R] = L match
    case Foo[?] => L
    case Bar[?] => L
    case _ => R
    case ? => L
```

kitbellew: > I don't think it happened yet despite what is in the plans.

in that case, `AnonymousParam` that we added in the previous commit is also invalid (that and AnonymousLambda should go together).

we might need to disable this field in scala3.2 and just set it in tests specifically, or add a new scala3.3 dialect:
```
    // Dotty uses `_` for placeholder for types since 3.2
    val allowUnderscoreAsTypePlaceholder: Boolean,
```

kitbellew: > IF I reprint the statement with Dotty I get back:
> 
> ```
> object match_types:
>   type Combine[L, R] = L match
>     case Foo[?] => L
>     case Bar[?] => L
>     case _ => R
>     case ? => L
> ```

so `_` is still a wildcard, not a placeholder, then. is it with unreleased nightly 3.2, or a prior version?

kitbellew: hmm, looks like 3.2.0 has already been released, and doesn't mention this. should we attach `allowUnderscoreAsTypePlaceholder` to a new 3.3 dialect instead?

tgodzik: > in that case, AnonymousParam that we added in the previous commit is also invalid (that and AnonymousLambda should go together).

Ach sorry, I didn't realize that :fearful: 

> hmm, looks like 3.2.0 has already been released, and doesn't mention this. should we attach allowUnderscoreAsTypePlaceholder to a new 3.3 dialect instead?

I think that might be good enough for now.

tgodzik: I will follow up with the compiler team to see when they plan to deprecate it.

kitbellew:
any updates on this? anything i should change?

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
tgodzik: I didn't get any response, but maybe it's fine to disable it for now until we know in which dilect it will be introduced?

kitbellew: will do, thank you!

kitbellew: last question: do you know if this `_` as lambda placeholder (not wildcard) is supported in any of the scala2[source3] variants?

tgodzik: There is apparently the flag `-source:future` that can be used, but that's apparently a bunch different features, which will be introduced later on. We could possibly introduce the dialect `Scala3Future` that could contain them though I don't know if there is a list of all the future features.

PR #2886, opened 2022-09-23
ScalametaParser: constrain match type pattern

kitbellew:
Currently, unlike pattern matching for Terms which follows a different parsing logic and produces Pat trees, pattern matching for Types uses exactly the same logic as regular Types.

This occasionally leads to incorrect parsing, and here we try to address the specific problem of parsing a wildcards and anonymous params which is different in a type pattern match context.

Helps with #2807.

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
@@ -222,7 +222,7 @@ class MatchTypeSuite extends BaseDottySuite {
                 Type.Name("Left"),
                 List(
                   TypeCase(Type.Name("Unit"), Type.Name("Right")),
-                  TypeCase(Type.Wildcard(Type.Bounds(None, None)), Type.Name("Left"))
+                  TypeCase(Type.Name("?"), Type.Name("Left"))
                 )
               ),
               Type.Bounds(None, None)
@@ -288,11 +288,11 @@ class MatchTypeSuite extends BaseDottySuite {
                     Type.Name("L")
                   ),
                   TypeCase(
-                    Type.AnonymousParam(None),
+                    Type.PatWildcard(),
                     Type.Name("R")
                   ),
                   TypeCase(
-                    Type.Wildcard(Type.Bounds(None, None)),
+                    Type.Name("?"),
                     Type.Name("L")
                   )
                 )
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -792,7 +792,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
     def typeCaseClause(): TypeCase = autoPos {
       accept[KwCase]
-      val pat = infixTypeOrTuple(allowFunctionType = false)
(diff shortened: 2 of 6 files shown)


tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
```diff
@@ -288,11 +288,11 @@ class MatchTypeSuite extends BaseDottySuite {
                     Type.Name("L")
                   ),
                   TypeCase(
-                    Type.AnonymousParam(None),
+                    Type.PatWildcard(),
                     Type.Name("R")
                   ),
                   TypeCase(
-                    Type.Wildcard(Type.Bounds(None, None)),
+                    Type.Name("?"),
```
tgodzik, line 295:
```suggestion
                    Type.PatWildcard("?"),
```

Both should be wildcards.

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
```diff
@@ -222,7 +222,7 @@ class MatchTypeSuite extends BaseDottySuite {
                 Type.Name("Left"),
                 List(
                   TypeCase(Type.Name("Unit"), Type.Name("Right")),
-                  TypeCase(Type.Wildcard(Type.Bounds(None, None)), Type.Name("Left"))
+                  TypeCase(Type.Name("?"), Type.Name("Left"))
```
tgodzik, line 225: This should be a wildcard still.

tgodzik (changes requested):

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
kitbellew: not really. `case ?` is not a wildcard, at least not in my testing. unless the `?` type is defined earlier, the compiler complains. could you check, please?

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
kitbellew: same answer as before. `?` is a generic type parameter wildcard, but not a pattern.

tgodzik: Currently, `case _ => R` means exactly the same as `case ? => R`, so why would it treat differently?

kitbellew: means where? i created a Scala 3 test project and tried using `case ? => ...`, and the compiler complained that type `?` was not defined. it didn't look at it as a wildcard. i must misunderstand what you said, then.

tgodzik: Actually, you are right. Didn't realize that `?` is limited to parameter types.

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/MatchTypeSuite.scala
tgodzik: Actually, you are right, I didn't realize it.

tgodzik (approved):
LGTM! Thanks for explaining and checking it properly.

Issue #2880, opened 2022-09-22
ParseException `; expected but . found`


sergiuszkierat:
input :

```scala
    Flow.fromGraph(GraphDSL.create() {
      implicit b =>
        import GraphDSL.Implicits._

        val removeItems       = b.add(RemoveLineItems(settings.lineItems.removal).graph)
        val filtering         = b.add(Filtering(settings.filtering).flow)
        val projection        = b.add(Projection(settings.projection).graph)

        val input_<       = b.add(Broadcast[FileContent](2))

        // format: off
        input_< ~> filtering          ~>       removeItems.in0
        // format: on

        FlowShape(input_<.in, projection.out)
    })
```

dialect : `scala213source3`
scalameta : `v4.5.13`
scala : `2.13.8`

PR #2888, opened 2022-09-27
bugfix: Don't allow multiple infix operators after newline

tgodzik:
Previously, if we had two infix operators after newline they would be treated as whole expression together with the previous line. That, however, is not allowed by the parser. Now, we don't continue the expression if there are two infix operators after newline.

Fixes https://github.com/scalameta/scalameta/issues/2880

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/Scala213Suite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/Scala213Suite.scala
@@ -128,6 +128,38 @@ class Scala213Suite extends ParseSuite {
     )(dialects.Scala213Source3)
   }
 
+  test("issue-2880") {
+    runAssert(
+      """|Flow {
+         |    b.add()
+         |    input_< ~> filtering ~> removeItems.in0
+         |}
+         |""".stripMargin
+    )(
+      Term.Apply(
+        Term.Name("Flow"),
+        List(
+          Term.Block(
+            List(
+              Term.Apply(Term.Select(Term.Name("b"), Term.Name("add")), Nil),
+              Term.ApplyInfix(
+                Term.ApplyInfix(
+                  Term.Name("input_<"),
+                  Term.Name("~>"),
+                  Nil,
+                  List(Term.Name("filtering"))
+                ),
+                Term.Name("~>"),
+                Nil,
+                List(Term.Select(Term.Name("removeItems"), Term.Name("in0")))
+              )
+            )
+          )
+        )
+      )
+    )(dialects.Scala213Source3)
+  }
+
   private def runAssert(code: String)(expected: Tree)(implicit d: Dialect): Unit = {
     assertTree(templStat(code)(d))(expected)
   }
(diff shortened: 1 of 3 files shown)


tests/shared/src/test/scala/scala/meta/tests/parsers/TermSuite.scala
```diff
@@ -1383,20 +1383,6 @@ class TermSuite extends ParseSuite {
     )
   }
 
-  test("using-call-site in scala2") {
```
kitbellew, line 1386: what happened to this test?

tests/shared/src/test/scala/scala/meta/tests/parsers/Scala213Suite.scala
```diff
@@ -128,6 +128,38 @@ class Scala213Suite extends ParseSuite {
     )(dialects.Scala213Source3)
   }
 
+  test("issue-2880") {
+    runAssert(
+      """|Flow {
+         |    b.add()
+         |    input_< ~> filtering ~> removeItems.in0
```
kitbellew, line 135: which part here was "multiple infix operators"? `input_<` and then `~>`?

tests/shared/src/test/scala/scala/meta/tests/parsers/Scala213Suite.scala
```diff
@@ -128,6 +128,38 @@ class Scala213Suite extends ParseSuite {
     )(dialects.Scala213Source3)
   }
 
+  test("issue-2880") {
```
kitbellew, line 131: is there a test which can demonstrate that something used to be parsed one way, but is now parsed differently, or rejected?

kitbellew (changes requested):

tests/shared/src/test/scala/scala/meta/tests/parsers/Scala213Suite.scala
tgodzik: Yes, that was it though I don't neccessarily understand the exact rule: https://docs.scala-lang.org/scala3/reference/changed-features/operators.html#syntax-change

There is a test failing in dotty that I should take a look at.

tests/shared/src/test/scala/scala/meta/tests/parsers/TermSuite.scala
tgodzik: My bad, will return it.

tests/shared/src/test/scala/scala/meta/tests/parsers/Scala213Suite.scala
tgodzik: I think the rule is actually different, need to rework this.

kitbellew:
@tgodzik is this still active?

tgodzik:
> @tgodzik is this still active?

I was meaning to get back to it, but for now I have failed to do so :sweat:

kitbellew:
> > @tgodzik is this still active?
> 
> I was meaning to get back to it, but for now I have failed to do so 😓

should we release without it, for now?

tgodzik:
> > > @tgodzik is this still active?
> > 
> > 
> > I was meaning to get back to it, but for now I have failed to do so sweat
> 
> should we release without it, for now?

I think so, I haven't really got back to it since it's a pretty rare scenario.

PR #2889, opened 2022-09-28
replace binaryCompatVar with a field annotation and full unapply

### LINKED ISSUE #2881: `q"opaque type Foo <: String = String"` loses the `<: String` ?
```scala
//> using scala "2.13.9"
//> using lib "org.scalameta::scalameta::4.5.13"

import scala.meta._

object Bug {

  def main(args: Array[String]): Unit =
    println(q"opaque type Foo <: String = String")

}
```

```
$ scala-cli bug.scala 
opaque type Foo = String
```

kitbellew:
Fixes #2881, by making private vars into "official" fields akin to other ones, and by introducing a full unapply (with all fields) and changing `Liftables` to use that instead.

### DIFF
--- a/project/Mima.scala
+++ b/project/Mima.scala
@@ -35,6 +35,11 @@ object Mima {
     ProblemFilters.exclude[A]("scala.meta." + metaType)
 
   val apiCompatibilityExceptions: Seq[ProblemFilter] = Seq(
+    // newField; these methods should have been package-private
+    exclude[DirectMissingMethodProblem]("Defn#Type.setBounds"),
+    exclude[DirectMissingMethodProblem]("Template.setDerives"),
+    exclude[DirectMissingMethodProblem]("Term#If.setMods"),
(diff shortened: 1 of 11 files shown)


kitbellew:
@tgodzik suggest reviewing one commit at a time, and make sure that `ast.scala` (which has a larger diff) is not hidden.

tgodzik:
Do you think it's ok to do a release before merging this one? I would like to add support for fewer braces at least in Metals for now.  Or would you prefer after this?

kitbellew:
of course, we can release now. i only wanted to get in all the #2807 commits.

kitbellew:
@tgodzik waiting for further instructions :)

tgodzik:
> @tgodzik waiting for further instructions :)

I will take a look tomorrow! I am finally finishing up with the Metals release.

tgodzik (approved):
LGTM!

PR #2892, opened 2022-09-28
Transverser: don't extract/bind fields via unapply

kitbellew:
Access the fields directly via the owner instance. Helps with #2881.


tgodzik (approved):
LGTM

PR #2894, opened 2022-09-28
ast: refactor getting and setting fields

kitbellew:
Helps with #2881.

### DIFF
--- a/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
+++ b/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
@@ -4,8 +4,9 @@ package trees
 
 import scala.language.experimental.macros
 import scala.annotation.StaticAnnotation
-import scala.reflect.macros.whitebox.Context
 import scala.collection.mutable.ListBuffer
+import scala.reflect.macros.whitebox.Context
+
 import org.scalameta.internal.MacroCompat
 
 // @ast is a specialized version of @org.scalameta.adt.leaf for scala.meta ASTs.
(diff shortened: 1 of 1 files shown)


scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
```diff
@@ -106,42 +108,46 @@ class AstNamerMacros(val c: Context) extends Reflection with CommonNamerMacros {
...
         val paramss = rawparamss
 
         // step 4: turn all parameters into vars, create getters and setters
-        def internalize(name: TermName) = TermName("_" + name.toString)
         val fieldParamss = paramss
-        val fieldParams = fieldParamss.flatten.map(p => (p, p.name.decodedName.toString))
-        istats1 ++= fieldParams.map({ case (p, _) =>
-          var getterAnns = List(q"new $AstMetadataModule.astField")
-          if (p.mods.annotations.exists(_.toString.contains("auxiliary")))
-            getterAnns :+= q"new $AstMetadataModule.auxiliary"
```
tgodzik, line 127: So `auxiliary` was never used? Maybe remove it altogether?

tgodzik (approved):
LGTM, just one small comment, but otherwise it looks like simplification of the current code. Thanks for doing it!

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
kitbellew: i assumed that it was looking for AstMetadata.auxilliary and not just any old annotation, and simply kept all annotations instead.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
```diff
@@ -106,42 +108,46 @@ class AstNamerMacros(val c: Context) extends Reflection with CommonNamerMacros {
...
+        val fieldParams = fieldParamss.flatten
+        fieldParams.foreach { p =>
+          val getterAnns = q"new $AstMetadataModule.astField" :: p.mods.annotations
```
kitbellew, line 139: here, all annotations. but i don't know if any specific one is used.

PR #2901, opened 2022-10-07
bugfix: Fix issues with fewer braces and partial functions

tgodzik:
It's not an ideal solution, but regions are a bit of black magic unfortunately. The other solutions might be more complex then this.

Fixes https://github.com/scalameta/scalameta/issues/2900

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
@@ -102,11 +102,19 @@ private[parsers] class LazyTokenIterator private (
 
   def observeIndented(): Boolean = {
     observeIndented0 { (i, prev) =>
+      /* When adding RegionIndent (we wrap the current code block in indentation)
+       * we might no longer need the region added on the current token.
+       *
+       * In case the region was needed, we will add it again as we haven't yet progressed
+       * to the next token.
+       */
       val undoRegionChange =
         prev.headOption match {
           case Some(RegionParen(_)) if token.is[LeftParen] => prev.tail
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
```diff
@@ -107,6 +107,12 @@ private[parsers] class LazyTokenIterator private (
           case Some(RegionParen(_)) if token.is[LeftParen] => prev.tail
           case Some(RegionEnumArtificialMark) if token.is[KwEnum] => prev.tail
           case Some(_: RegionBrace) if token.is[LeftBrace] => prev.tail
+          /* Handle fewer braces and partial function.
+           * RegionArrow will be added first and then we would add RegionIndent
+           * which would cause the RegionArrow not to be removed on `=>`.
+           * If we add RegionIndent, Arrow will be readded.
+           */
+          case Some(RegionArrow) if dialect.allowFewerBraces && token.is[KwCase] => prev.tail
```
kitbellew, line 115: does the colon add the arrow region? i don't see any arrows before the first case in the tested example...

kitbellew: ah, found it. `case` adds the arrow region while `=>` adds the case region :)

kitbellew (approved):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
tgodzik: I also added a bit better explanation of what is happening here.

kitbellew (approved):

PR #2907, opened 2022-10-14
Trees: allow Seq fields instead of just List

kitbellew:


kitbellew:
@tgodzik @dos65 ptal. required to introduce replaced fields later, to avoid erasure problems with List[List[x]] -> List[y].

tgodzik (approved):
LGTM

PR #2914, opened 2022-10-17
ast: check annotations use current major version

kitbellew:
This way, we'll remember to remove older field annotations before a major version release (which allows binary non-compatibility).

### DIFF
--- a/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
+++ b/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
@@ -459,15 +459,23 @@ class AstNamerMacros(val c: Context) extends Reflection with CommonNamerMacros {
     """
   }
 
-  private def parseVersionAnnot(version: Tree, annot: String, field: String): Version = {
-    val versionStr = version match {
+  private def getAnnotAttribute(value: Tree): String =
+    value match {
       case x: AssignOrNamedArg => x.rhs.toString
       case x => x.toString
     }
-    try { parseVersion(versionStr) }
-    catch {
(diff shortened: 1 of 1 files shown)


scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
```diff
@@ -459,15 +459,23 @@ class AstNamerMacros(val c: Context) extends Reflection with CommonNamerMacros {
...
+
+  private def parseVersionAnnot(version: Tree, annot: String, field: String): Version = {
+    val parsed =
+      try { parseVersion(getAnnotAttribute(version)) }
+      catch {
+        case NonFatal(_) => c.abort(version.pos, s"@$annot must contain $field=major.minor.patch")
+      }
+    majorVersion.foreach { major =>
+      if (parsed._1 < major)
+        c.abort(version.pos, s"@$annot: invalid major version (must be $major)")
```
tgodzik, line 476: Shouldn't the message say something like `@$annot: version for $field no longer valid, please update` ?

kitbellew: so, i don't want the version to be updated :) my goal is to make sure that all these binary-compatibility annotations are removed in a new major release (which will also result is a binary incompatibility, but that's what major releases are for). should i revise the message a bit, to say that?

kitbellew: added the word `obsolete`, perhaps that makes it clearer :)

tgodzik: Sounds better, thanks!

PR #2915, opened 2022-10-17
ast: add a private `fullCopy` method

kitbellew:
This method doesn't need to be binary-compatible and hence supports all current fields, unlike the primary copy method (the one with default parameter values) which, for backwards compatibility, must refer to the original order and types of parameters.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1429,7 +1429,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   def inlineMatchClause(inlineMods: List[Mod]) = {
     autoEndPos(inlineMods)(postfixExpr(allowRepeated = false)) match {
       case t: Term.Match =>
-        copyPos(t)(t.copy(mods = inlineMods))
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1429,7 +1429,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   def inlineMatchClause(inlineMods: List[Mod]) = {
     autoEndPos(inlineMods)(postfixExpr(allowRepeated = false)) match {
       case t: Term.Match =>
-        copyPos(t)(t.copy(mods = inlineMods))
+        copyPos(t)(t.fullCopy(mods = inlineMods))
```
tgodzik, line 1432: Why do we need fullCopy here?

tgodzik (approved):
LGTM aside from one question

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: for now, `fullCopy` does exactly the same as the primary `copy`. However, after we introduce a new annotation, for replaced fields, the primary `.copy` in those cases will be producing trees with replaced fields losing their origin. which might be ok for outside usage but not good for this parser.

PR #2917, opened 2022-10-17
ast: allow replacing old fields with new ones

kitbellew:
As an alternative to existing `newFields` which are appended at the end of the list of existing params, let's have the ability to remove a field as long as it's replaced with another.

Current implementation supports cases when the new field value can be determined from the old one via an implicit conversion, similar to how quasiquotes define implicit Lift/Unlift conversions.

This change is backwards-binary-compatible and hence will not trigger MIMA-verification alerts.


tgodzik (approved):
I think the approach is sound. Can't find any issues looking at it currently, but when MiMa is happy I am happy :sweat_smile:

kitbellew:
> I think the approach is sound. Can't find any issues looking at it currently, but when MiMa is happy I am happy 😅

thank you! it took me a while to figure it out, hopefully it's correct indeed. actual replacements are coming in next prs.

Issue #2869, opened 2022-09-16
Encapsulate args or params as a tree, including delimiters


kitbellew:
Also suggested by @olafurpg in [#405](https://github.com/scalameta/scalameta/issues/405#issuecomment-222133672) and is a prerequisite for #2820.

PR #2919, opened 2022-10-21
Trees: use Term.ParamClause, not List[Term.Param]

kitbellew:
Helps with #2869.

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/quasiquotes/ErrorSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/quasiquotes/ErrorSuite.scala
@@ -803,23 +803,6 @@ class ErrorSuite extends FunSuite {
     )
   }
 
-  test("triple-dot mixes with something else in parameter lists") {
-    assertEquals(
-      typecheckError("""
-      import scala.meta._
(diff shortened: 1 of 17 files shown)


kitbellew:
@tgodzik this is the second PR, covering Term.Param clauses.

tests/shared/src/test/scala/scala/meta/tests/quasiquotes/ErrorSuite.scala
```diff
@@ -803,23 +803,6 @@ class ErrorSuite extends FunSuite {
     )
   }
 
-  test("triple-dot mixes with something else in parameter lists") {
```
tgodzik, line 806: Why did you remove the test? We no longer have this limitation?

tgodzik (approved):
LGTM! Just one question from me.

tests/shared/src/test/scala/scala/meta/tests/quasiquotes/ErrorSuite.scala
kitbellew: because it's not broken anymore (it's capable of parsing this quasiquote now). i added a test in success suite instead.

PR #2934, opened 2022-11-01
ast: allow specifying a ctor in replacedField

kitbellew:


scalameta/common/shared/src/main/scala/scala/meta/internal/trees/metadata.scala
```diff
@@ -12,7 +12,7 @@ object Metadata {
   class branch extends StaticAnnotation
   class astClass extends StaticAnnotation
   class newField(since: String) extends StaticAnnotation
-  class replacedField(until: String) extends StaticAnnotation
+  class replacedField(until: String, ctor: Any = null) extends StaticAnnotation
```
tgodzik, line 15: What would be the example usage here?

kitbellew: it will be added in a new PR.

currently, a replacedField explicitly specifies how to convert from new to old, but to support the deprecated methods (like apply with the old field as parameter), we also need to convert old to new, and that was done via implicits.

however, sometimes we don't want to define an implicit (because it might conflict with another implicit) and simply define how to convert old to new explicitly. so this `ctor` field will refer to a method which does that.

tgodzik (approved):
Thanks for explaning!

PR #2938, opened 2022-11-01
TreeSyntax: don't print Name.Anonymous

kitbellew:
Name.Anonymous was designed to be used when the name is not actually present; let's handle the few outlier cases when it stands for `_`.


tgodzik (approved):
LGTM

PR #2943, opened 2022-11-03
Trees: add Name.Placeholder, use as appropriate

kitbellew:
Name.Anonymous was designed to be used when the name is not actually present; thus, using it for `_` is inconsistent and leads to complex handling.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1764,8 +1764,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       case t: Name => Some(t)
       case t: Quasi => Some(t.become[Term.Name])
       case t: Term.Select => Some(t.name)
-      case t: Term.Placeholder => Some(copyPos(t)(Name.Anonymous()))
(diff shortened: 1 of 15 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1764,8 +1764,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       case t: Name => Some(t)
       case t: Quasi => Some(t.become[Term.Name])
       case t: Term.Select => Some(t.name)
-      case t: Term.Placeholder => Some(copyPos(t)(Name.Anonymous()))
-      case t: Term.Eta => Some(atPos(t.endTokenPos)(Name.Anonymous()))
+      case t: Term.Placeholder => Some(copyPos(t)(Name.Placeholder()))
```
tgodzik, line 1767: Let's release it under 4.7, since it's an API addition and can change what people expect from the parser.

tgodzik (approved):
LGTM, just one comment from my side.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: will do, thanks.

PR #2946, opened 2022-11-04
Trees: use common Tuple checks in all tuples

kitbellew:


tgodzik (commented):
Looks good, though could you add a test that benefits from this change?

kitbellew:
> Looks good, though could you add a test that benefits from this change?

reused this PR for a much smaller commit. moved the rest of the logic to #2948, with the actual change which uses it (and hence contains tests).

PR #2948, opened 2022-11-05
Trees: use Type.FuncParamClause in Type.FunctionType

kitbellew:
Helps with #2869.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -729,14 +729,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       }
 
       val openParenPos = in.tokenPos
-      accept[LeftParen]
-      val ts = if (!token.is[RightParen]) listBy[Type] { tsBuf =>
-        do {
-          tsBuf += paramOrType()
-        } while (acceptOpt[Comma] || token.is[Ellipsis])
-      }
-      else Nil
-      accept[RightParen]
+      val ts = inParensOr(commaSeparated(paramOrType()))(Nil)
       // NOTE: can't have this, because otherwise we run into #312
       // newLineOptWhenFollowedBy[LeftParen]
 
@@ -747,10 +740,11 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
         syntaxError(message, at = token)
       }
 
-      if (allowFunctionType && acceptOpt[RightArrow]) {
-        Type.Function(ts, typeIndentedOpt())
-      } else if (allowFunctionType && acceptOpt[ContextArrow]) {
-        Type.ContextFunction(ts, typeIndentedOpt())
+      def funcParams = autoEndPos(openParenPos)(ts.reduceWith(Type.FuncParamClause.apply))
+      if (allowFunctionType && token.is[RightArrow]) {
+        Type.Function(funcParams, { next(); typeIndentedOpt() })
+      } else if (allowFunctionType && token.is[ContextArrow]) {
+        Type.ContextFunction(funcParams, { next(); typeIndentedOpt() })
       } else {
         ts.foreach {
           case t: Type.ByName => syntaxError("by name type not allowed here", at = t)
@@ -787,15 +781,17 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     }
 
     def typ(): Type = autoPosOpt {
+      val startPos = auto.startTokenPos
       val t: Type =
         if (token.is[LeftBracket] && dialect.allowTypeLambdas) typeLambdaOrPoly()
         else infixTypeOrTuple()
 
+      def funcParams = autoEndPos(startPos)((t :: Nil).reduceWith(Type.FuncParamClause.apply))
       token match {
-        case RightArrow() => next(); Type.Function(List(t), typeIndentedOpt())
-        case ContextArrow() => next(); Type.ContextFunction(List(t), typeIndentedOpt())
-        case KwForsome() => next(); Type.Existential(t, existentialStats())
-        case KwMatch() if dialect.allowTypeMatch => next(); Type.Match(t, typeCaseClauses())
+        case RightArrow() => Type.Function(funcParams, { next(); typeIndentedOpt() })
+        case ContextArrow() => Type.ContextFunction(funcParams, { next(); typeIndentedOpt() })
+        case KwForsome() => Type.Existential(t, { next(); existentialStats() })
+        case KwMatch() if dialect.allowTypeMatch => Type.Match(t, { next(); typeCaseClauses() })
         case _ => t
       }
     }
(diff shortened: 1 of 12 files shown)


scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -223,15 +223,35 @@ object Type {
...
+  }
   @branch trait FunctionType extends Type {
-    def params: List[Type]
+    def param: Type
+    def params: List[Type] = FunctionType.paramToParams(param)
     def res: Type
   }
-  @ast class Function(params: List[Type], res: Type) extends FunctionType
+  @ast class Function(param: Type, res: Type) extends FunctionType {
+    @replacedField("4.6.0", FunctionType.paramsToParam) final override def params: List[Type] =
```
tgodzik, line 245: Could you maybe add some docs about it and maybe add @deprecated flag? Otherwise users are getting both `param` and `params` when doing `.` completions on the tree node for Type.Function

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -629,29 +633,36 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
-      case Nil => autoEndPos(lpPos)(zero)
-      case only :: Nil =>
-        only match {
-          case q: Quasi if q.rank == 1 => copyPos(q)(tuple(body))
-          case _ => only
-        }
-      case _ => autoEndPos(lpPos)(tuple(body))
-    }
+  private def makeTuple[A <: Tree](lpPos: Int, body: List[A], zero: => A, tuple: List[A] => A)(
+      single: A => Either[List[A], A]
```
tgodzik, line 637: Couldn't this function be a Boolean?
```suggestion
      isSingleTuple: A => Boolean
```
and/or maybe add some comment what it does?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -629,29 +633,36 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
-    makeTuple(lpPos, body, Lit.Unit(), Term.Tuple.apply)
+    makeTuple(lpPos, body, Lit.Unit(), Term.Tuple.apply)(Right.apply)
   }
 
-  def makeTupleType(lpPos: Int, body: List[Type]): Type = {
-    val args = body match {
-      case singleArg :: Nil => maybeAnonymousLambda(singleArg) :: Nil
-      case _ => body
+  private def makeTupleType(lpPos: Int, body: List[Type], zero: => Type, wrap: Boolean): Type = {
+    makeTuple(lpPos, body, zero, Type.Tuple.apply) {
```
tgodzik, line 654: Everywhere aside from here we just use t, so this is the only place we need to return anything aside from Boolean (so an Either), so it seems like we are complicating the signature of the `single` function just for this one case.

kitbellew: what would you suggest? either we do it like this, or we duplicate the Quasi checks instead.

kitbellew:
@tgodzik

- moved tuple handling logic into #2964
- changed to explicit `FuncParamClause` type rather than `Unit/Type/Tuple` depending on the situation, which also allows proper quasiquotes usage.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -226,6 +226,10 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     accept[LeftParen]
     inParensAfterOpen(body)
   }
+  @inline final def inParensOr[T](body: => T)(alt: => T): T = {
+    accept[LeftParen]
+    inParensAfterOpenOr(body)(alt)
```
tgodzik, line 231:
```suggestion
    if (acceptOpt[LeftParen]) inParensAfterOpenOr(body)(alt)
    else alt
```
?
Otherwise, the method seems misleading since it doesn't accept no parens.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -716,14 +720,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
       val openParenPos = in.tokenPos
-      accept[LeftParen]
-      val ts = if (!token.is[RightParen]) listBy[Type] { tsBuf =>
-        do {
-          tsBuf += paramOrType()
-        } while (acceptOpt[Comma] || token.is[Ellipsis])
-      }
-      else Nil
-      accept[RightParen]
+      val ts = inParensOr(commaSeparated(paramOrType()))(Nil)
```
tgodzik, line 723:
```suggestion
      val ts = inParensOr(commaSeparated(paramOrType()))(syntaxError(...))
```
? since we expect them and we can give a meaningful error message here? Parens are expected and if there are not there we should show error.

tgodzik (changes requested):
Just two small comments, thanks for the other PRs!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: for both comments: the method requires _parens_ but it doesn't require anything _in_ them (that is, allows the closing paren right away).

tgodzik: Maybe then `inParens` instead `inParensOr` : 
```
   @inline final def inParens[T](body: => T)(whenEmpty: => T): T = {
```

kitbellew: there is already a method called that way, for cases when there's no "or", meaning that an immediate closing paren is not expected... let me review.

kitbellew: so, here's what we have:

- both parameters (_body_ and _alt_) are of the same type, so we can't reorder them to disambiguate from regular `inParens`
- there are several methods named similarly (two sets right below) and behaving similarly.

kitbellew: thanks, will rename the `alt` parameter.

kitbellew: added a separate PR for that, #2967.

kitbellew: @tgodzik that PR has been merged. please let me know if there are any other adjustments to make.

tgodzik (approved):
LGTM

PR #2958, opened 2022-11-08
feature: Allow fewer braces for Scala 3.3.x

tgodzik:
Follow up from https://github.com/lampepfl/dotty/pull/16297


scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -150,7 +150,9 @@ package object dialects {
 
   implicit val Scala32 = Scala31
 
-  implicit val Scala3 = Scala32
+  implicit val Scala33 = Scala31.withAllowFewerBraces(true)
```
kitbellew, line 153: and i was recently wondering whether we needed to add this to the future dialect... thank you!

kitbellew (approved):

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
kitbellew: ah, wait, it should be Scala32.withXxx.

tgodzik: Ach my bad, should be fixed now.

PR #2963, opened 2022-11-11
ScalametaParser: keep delims around body

kitbellew:
Partial undo of #2941. Turns out it's very hard to handle enclosed trees when their enclosing parens belong to a parent.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -226,6 +226,10 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     accept[LeftParen]
     inParensAfterOpen(body)
   }
+  @inline final def inParensOr[T](body: => T)(alt: => T): T = {
+    accept[LeftParen]
+    inParensAfterOpenOr(body)(alt)
+  }
   @inline private def inParensOnOpen[T](body: => T): T = {
     next()
     inParensAfterOpen(body)
@@ -359,10 +363,6 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   def atPosTry[T <: Tree](start: StartPos, end: EndPos)(body: => Try[T]): Try[T] = {
     val startTokenPos = start.startTokenPos
     body.map(atPos(startTokenPos, end))
-  }
-  def atPosTryOpt[T <: Tree](start: StartPos, end: EndPos)(body: => Try[T]): Try[T] = {
-    val startTokenPos = start.startTokenPos
-    body.map(atPosOpt(startTokenPos, end))
   }
 
   def autoPos[T <: Tree](body: => T): T = atPos(start = auto, end = auto)(body)
@@ -678,7 +678,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
    * context or not. Formerly, this was threaded through numerous methods as boolean isPattern.
    */
   trait PatternContextSensitive {
-    private def tupleInfixType(allowFunctionType: Boolean = true): Type = autoPosOpt {
+    private def tupleInfixType(allowFunctionType: Boolean = true): Type = autoPos {
       // NOTE: This is a really hardcore disambiguation caused by introduction of Type.Method.
       // We need to accept `(T, U) => W`, `(x: T): x.U` and also support unquoting.
       var hasParams = false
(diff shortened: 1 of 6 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -226,6 +226,10 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     accept[LeftParen]
     inParensAfterOpen(body)
   }
+  @inline final def inParensOr[T](body: => T)(alt: => T): T = {
```
tgodzik, line 229: Seems this is being modified in the other PR? Let's merge that first

tests/shared/src/test/scala-2.13/scala/meta/tests/tokenizers/TokensPositionSuite.scala
```diff
@@ -158,30 +158,37 @@ class TokensPositionSuite extends BasePositionSuite(dialects.Scala213) {
   )
   checkPositions[Stat](
     "(f)((((a))))",
-    """|Term.ArgClause ((((a))))
+    """|Term.Name (f)
```
tgodzik, line 161: I don't think name should contain `()`. Might be problematic if anyone wants to get the identifier position.

tgodzik (changes requested):

tests/shared/src/test/scala-2.13/scala/meta/tests/tokenizers/TokensPositionSuite.scala
kitbellew: then let me try once again. i had a really, really hard time getting `scalafmt` to work when parens became part of the parent. perhaps we just need to make a few changes in the formatter not to rely on these.

kitbellew: question: in `(a).b`, which tree should own the parens? (i'd prefer to include in `a` as that's what they are designed to encompass)

tgodzik: I would say `(a).b` since it's not really part of anything else, but the select. Was it previously containing (a) for Term.Name? If that is the case we can revert if it's causing issues

kitbellew: let me try once again. i spent a week trying to fix all the scalafmt failures with the current scalameta, but couldn't do it. generally, commentary around the formatter codebase talks about scalameta helping the formatter rather than requiring massive special-case processing, so since i (possibly, misguidedly but with same logic as yours) removed these positions in #2941, this PR was an attempt to restore some of them.

tgodzik: I am fine with reverting, especially if it was that way before.

tgodzik:
@kitbellew can we release a new version without this PR? Or is it blocking for scalafmt?

kitbellew:
> @kitbellew can we release a new version without this PR? Or is it blocking for scalafmt?

I'm trying to get the formatter working without this change, it might not be needed -- and perhaps not ideal to depend on positions of parentheses.

but since the next release will not be a patch, should we perhaps finalize the decision -- and implementation -- of the interleaved change (#2929)?

tgodzik:
> > @kitbellew can we release a new version without this PR? Or is it blocking for scalafmt?
> 
> I'm trying to get the formatter working without this change, it might not be needed -- and perhaps not ideal to depend on positions of parentheses.

Ok, I can wait for that :+1: 

> but since the next release will not be a patch, should we perhaps finalize the decision -- and implementation -- of the interleaved change (#2929)?

I don't think we need to hurry with interleaved, we can do another minor later, it's not a problem. As for decisions, did any of my suggestions make sense?

kitbellew:
> > > @kitbellew can we release a new version without this PR? Or is it blocking for scalafmt?
> > 
> > 
> > I'm trying to get the formatter working without this change, it might not be needed -- and perhaps not ideal to depend on positions of parentheses.
> 
> Ok, I can wait for that 👍

thank you, i might be ready next week.

> > but since the next release will not be a patch, should we perhaps finalize the decision -- and implementation -- of the interleaved change (#2929)?
> 
> I don't think we need to hurry with interleaved, we can do another minor later, it's not a problem. As for decisions, did any of my suggestions make sense?

the desire made sense but i haven't tried implementing it -- and thus seeing what transpires. and yes, let's postpone it until next release.

PR #2987, opened 2022-12-06
Trees: param clause group in Def/Given/Extension

kitbellew:


tests/jvm/src/test/scala-2.13/scala/meta/tests/trees/ReflectionSuite.scala
```diff
@@ -76,6 +76,7 @@ class ReflectionSuite extends FunSuite {
       |Byte
       |Char
       |Int
+      |List[meta.Term.ParamClause]
```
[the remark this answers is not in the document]
tgodzik: Och, that's weird, but I guess it's fine

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -444,6 +448,44 @@ object Member {
   @branch trait ParamClause extends SyntaxValuesClause {
     def values: List[Param]
   }
+
+  @ast class ParamClauseGroup(
+      tparamClause: sm.Type.ParamClause,
+      paramClauses: List[sm.Term.ParamClause]
+  ) extends Tree with Tree.WithTParamClause with Tree.WithParamClauses {
```
tgodzik, line 455: Do we need `WithParamClauses` still? It's only used here, or is it still used somewhere?

tgodzik (approved):
LGTM, just one small comment

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
kitbellew: in the formatter 🙂 rather than listing a number of various specific trees, i try to group them via a relevant interface.

tgodzik: Sure makes sense!

Issue #2993, opened 2022-12-15
Diagnostics missing in semanticdb file generated with semanticdb-scalac_2.13.10:4.7.0


bjaglin:
https://github.com/scalacenter/scalafix/blob/v0.10.4/scalafix-tests/input/src/main/scala-2/test/noAutoTupling/NoAutoTupling.scala

```diff
 $ metap ./scalafix-tests/input/target/jvm-2.13/meta/META-INF/semanticdb/scalafix-tests/input/src/main/scala-2/test/noAutoTupling/NoAutoTupling.scala.semanticdb
--- 4.7.0
+++ 4.6.0
@@ -9,7 +9,6 @@
 Language => Scala
 Symbols => 137 entries
 Occurrences => 158 entries
-Diagnostics => 22 entries
 Synthetics => 35 entries
 
 Symbols:
@@ -311,141 +310,6 @@
 [48:4..48:7) => test/noAutoTupling/NoAutoTupling#NoFalsePositives.Foo.
 [48:8..48:13) => test/noAutoTupling/NoAutoTupling#NoFalsePositives.Foo.apply().
 
-Diagnostics:
-[7:2..7:12) [warning] adapted the argument list to the expected 2-tuple: add additional parens instead
-        signature: NoAutoTupling.a(x: (Int, Boolean)): (Int, Boolean)
-  given arguments: 2, true
- after adaptation: NoAutoTupling.a((2, true): (Int, Boolean))
-[13:2..13:34) [warning] adapted the argument list to the expected 2-tuple: add additional parens instead
-        signature: NoAutoTupling.c(x: Int, y: Boolean)(z: (String, List[Int])): (Int, Boolean, (String, List[Int]))
-  given arguments: "foo", {
-  final <synthetic> <artifact> val rassoc$2: Int(1) = 1;
-  {
-  final <synthetic> <artifact> val rassoc$1: Int(2) = 2;
-  Nil.$colon$colon(rassoc$1)
-}.$colon$colon(rassoc$2)
-}
- after adaptation: NoAutoTupling.c(("foo", {
-  final <synthetic> <artifact> val rassoc$2: Int(1) = 1;
-  {
-  final <synthetic> <artifact> val rassoc$1: Int(2) = 2;
-  Nil.$colon$colon(rassoc$1)
-}.$colon$colon(rassoc$2)
-}): (String, List[Int]))
-[16:2..16:12) [warning] adapted the argument list to the expected 2-tuple: add additional parens instead
[...]

PR #2995, opened 2022-12-19
chore: Test ExpectSuite under Scala 2.13

tgodzik:
I split the expect file for Scala 2.12 and Scala 2.13 and it seems we get proper diagnostics for both :thinking:


tgodzik:
@bjaglin we haven't been testing diagnostics under 2.13 it seems (though I could have sworn we had at least a couple of tests), but it seems diagnostics are present in the expected output. Anything specific about those tests in https://github.com/scalameta/scalameta/issues/2993 ?

Maybe we broke the parser for that file?

tgodzik:
On the other hand I can reproduce that locally :thinking: Will need to dig in more

tgodzik:
Ach, actually I used the wrong file :/  We can check if the results here change after https://github.com/scalameta/scalameta/pull/2996

kitbellew (approved):

kitbellew:
@tgodzik is it possible that this should have been rebased on top of #2996 (which was merged a day earlier)? i am getting test errors in `ExpectSuite` from the commit tagged with release v4.7.1.

tgodzik:
> @tgodzik is it possible that this should have been rebased on top of #2996 (which was merged a day earlier)? i am getting test errors in `ExpectSuite` from the commit tagged with release v4.7.1.

Wait, I checked that and it seemed to work ok locally. I should have rebased :sweat: I will raise a PR

PR #2997, opened 2022-12-22
bugfix: Fix expect test suite

tgodzik:
The number of diagnostics are overriden in the recently updated SemanticdbReporter, which is why we have more of them here.


kitbellew (approved):
looks simple enough :)

PR #3007, opened 2022-12-29
Fix end marker handling

kitbellew:
Use the right dialect flag, reuse soft keyword KwEnd properly handling `end` identifiers in backquotes.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
@@ -69,7 +69,7 @@ class SoftKeywords(dialect: Dialect) {
   @classifier
   trait KwEnd {
     val name = "end"
-    @inline def isEnabled = dialect.allowSignificantIndentation
+    @inline def isEnabled = dialect.allowEndMarker
     @inline final def unapply(token: Token): Boolean = isEnabled && name == token.toString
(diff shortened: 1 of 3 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
```diff
@@ -69,7 +69,7 @@ class SoftKeywords(dialect: Dialect) {
   @classifier
   trait KwEnd {
     val name = "end"
-    @inline def isEnabled = dialect.allowSignificantIndentation
+    @inline def isEnabled = dialect.allowEndMarker
```
tgodzik, line 72: Actually, do we need a separate dialect option for this? I don't think it will be removed in the future.

kitbellew: that i don't know 🙂 should we remove it instead?

kitbellew: to be clear: i haven't added the flag, I'm simply making its use consistent.

tgodzik: Maybe let's leave it, just in case Martin decides against end markers in the future :sweat_smile:

tgodzik (approved):

PR #3019, opened 2023-01-05
ScannerTokens: require token index as param

kitbellew:
Remove the logic which caches token indices and tries to find the index for a given token.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
@@ -258,7 +258,7 @@ private[parsers] class LazyTokenIterator private (
         (nextRegions, currRef)
       case _: KwEnum =>
         (RegionEnumArtificialMark :: sepRegions, currRef)
-      case CaseIntro() =>
+      case _ if isCaseIntro(currPos) =>
         val nextRegions = sepRegions.headOption match {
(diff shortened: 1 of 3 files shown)


kitbellew:
@tgodzik after this, i was thinking to release. are you planning to add anything else?

tgodzik (commented):
Why do we need to use index everywhere, in some places we already have the token, so this seems an added complication.

I thought the idea is to use index in some places where needed.

I think I have trouble understanding why is this PR helping us? I quite liked the `@classifier` approach :sweat_smile:

kitbellew:
> Why do we need to use index everywhere, in some places we already have the token, so this seems an added complication.

in order to look at the tokens before or after the given one, we need to have -- or find -- the index. however, the reality is that we always have the position that we're currently parsing or iterating over, so finding that is an unnecessary overhead.

on the other hand, we have cases when the given token is not really one of the original ones produced by the tokenizer (such as indent and outdent), so having just the index is not sufficient.

> I thought the idea is to use index in some places where needed.
> 
> I think I have trouble understanding why is this PR helping us? I quite liked the `@classifier` approach 😅

this is an optimization, removing an unnecessary operation at the beginning of every lookahead or lookbehind sequence. if you believe that it provides a questionable gain at the expense of readability, we can skip this.

tgodzik (approved):
If you think it's an improvement over the previous approach it's fine, though not sure if the readability is impacted. Maybe not?

> @tgodzik after this, i was thinking to release. are you planning to add anything else?

I don't have anything from my side.

PR #3020, opened 2023-01-06
LazyTokenIterator: don't outdent before `=>`

kitbellew:
In #3003, along with checking for backquote in possible leading infix, we accidentally added an otherwise valid check for the Ident token. (Proves once again that each commit should be about a single explicit atomic change.)

Unfortunately, it looks like the "leading infix" check happened to cover some unrelated cases, such as the `=>`, and that commit broke the logic.

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/NewFunctionsSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/NewFunctionsSuite.scala
@@ -637,6 +637,60 @@ class NewFunctionsSuite extends BaseDottySuite {
     )
   }
 
+  test("dependent-type-arrow-after-nl") {
+    val expectedSyntax = "type T = (e: Entry) => e.Key"
+    val expectedTree = Defn.Type(
+      Nil,
+      pname("T"),
+      Nil,
+      Type.Function(
+        List(Type.TypedParam(pname("e"), pname("Entry"))),
+        Type.Select(tname("e"), pname("Key"))
+      ),
+      Type.Bounds(None, None)
+    )
+
+    runTestAssert[Stat](
+      """|type T = 
+         |  (e: Entry) 
+         |  => e.Key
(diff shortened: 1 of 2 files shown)


kitbellew:
@tgodzik might need your help. i found this case in one of scalafmt tests you added: https://github.com/scalameta/scalafmt/blob/master/scalafmt-tests/src/test/resources/scala3/DependentFunction.stat#L9

however, section 2.2 here seems to suggest that we don't need to allow this case at all (at least not when indent is strictly smaller): https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html#optional-braces

could you please double-check my logic, and perhaps check which of the 4 new tests is actually allowed by the scala3 compiler?

tgodzik:
> however, section 2.2 here seems to suggest that we don't need to allow this case at all (at least not when indent is strictly smaller):

It seems that it is not supported, the compiler is not happy about it. Not sure if that was an error at the moment I added the test or was it changed afterwards :thinking:

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/NewFunctionsSuite.scala
```diff
@@ -637,6 +637,62 @@ class NewFunctionsSuite extends BaseDottySuite {
...
+        List(Type.TypedParam(pname("e"), pname("Entry"))),
+        Type.Select(tname("e"), pname("Key"))
+      ),
+      Type.Bounds(None, None)
+    )
+
+    runTestAssert[Stat](
+      """|type T = 
+         |  (e: Entry) 
+         |  => e.Key
```
tgodzik, line 656: This one is valid, the ones with lower indentation are not from what I checked.

kitbellew: thank you for checking. in that case, looks like the other solution (#3021) is more appropriate here, please take a look.

PR #3025, opened 2023-01-07
LazyTokenIterator: move regions into TokenRef

kitbellew:
Later on, it will allow us to buffer subsequent tokens for lookahead.

Also, move TokenRef out as a toplevel class.

For #3031.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
@@ -10,50 +10,37 @@ import scala.meta.tokens.Tokens
 object LazyTokenIterator {
 
   def apply(st: ScannerTokens)(implicit dialect: Dialect): LazyTokenIterator =
-    new LazyTokenIterator(st, Nil, TokenRef(st.tokens(0), 0), -1)
-
-  private class TokenRef private (
-      val token: Token,
-      val pos: Int,
-      val nextPos: Int,
-      val pointPos: Int
-  )
-
-  private object TokenRef {
-    def apply(token: Token, pos: Int) = new TokenRef(token, pos, pos + 1, pos)
-    def apply(token: Token, pos: Int, nextPos: Int, pointPos: Int) =
-      new TokenRef(token, pos, nextPos, pointPos)
-  }
+    new LazyTokenIterator(st, TokenRef(Nil, st.tokens(0), 0), -1)
 
 }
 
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
```diff
@@ -63,17 +50,17 @@ private[parsers] class LazyTokenIterator private (
    * Deals with different rules for indentation after self type arrow.
    */
   def undoIndent(): Unit = {
-    sepRegions match {
+    curr.regions match {
       case (region: SepRegionIndented) :: others if curr.token.is[Indentation.Indent] =>
         next()
-        sepRegions = sepRegions match {
+        curr.regions = curr.regions match {
```
tgodzik, line 56:
```suggestion
        val reducedRegions = curr.regions match {
```
and then:
```
curr = curr.withRegions(newRegions)
```

we would need to keep regions a var in that case and have just `curr` to be mutable

tgodzik (changes requested):
Looks good, though I have a small suggestion

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
kitbellew: done

kitbellew: @tgodzik one question about this change: i am planning to add another `var` to `TokenRef` (to make it into a linked list, for caching computations), and the whole point is not to make a copy when setting `.next` on it.

with that, do you still prefer `regions` to be a `val`?

tgodzik: Or we could make it a method like:
```
curr.reduceRegions{
  case ...
}

```
so that we keep the var private, what do you think?

kitbellew: sure, i can do that. although we don't really do any sort of vetting of the values assigned in `reduceRegions`, which makes it basically the same as a blind assignment. and `TokenRef` is a package-private class...

kitbellew: i'll keep regions a `val` here.

PR #3026, opened 2023-01-07
LazyTokenIterator: move methods to ScannerTokens

kitbellew:
These methods do not require any of the instance variables provided by the LazyTokenIterator class, except tokens from ScannerTokens. For #3031.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
@@ -25,21 +25,6 @@ object LazyTokenIterator {
       new TokenRef(token, pos, nextPos, pointPos)
   }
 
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/LazyTokenIterator.scala
```diff
@@ -155,43 +140,6 @@ private[parsers] class LazyTokenIterator private (
       case _ => false
     })
 
-  private def mkIndentToken(pointPos: Int): Token = {
```
tgodzik, line 158: We don't use these aside from LazyTokenIterator, why does it make sense to move them?

kitbellew: everything will move, in steps

tgodzik (approved):

PR #3027, opened 2023-01-07
LazyTokenIterator: pass current token to nextToken

kitbellew:
That way, nextToken no longer depends on any instance variables except tokens. This will allow us to move this method to ScannerTokens next. For #3031.


tgodzik (approved):
:rocket:

PR #3028, opened 2023-01-07
ScalametaParser: for fewer braces, avoid lookahead

kitbellew:
Instead, try parsing the function and back out if that fails. For #3031.


tgodzik (approved):
LGTM

Issue #2405, opened 2021-07-02
[Scala 3] Parser issues


tgodzik:
I've taken a look at parsing the Dotty pos test cases and I found a bunch of errors:

Issues:

- [x] https://github.com/scalameta/scalameta/pull/2408

```scala 
type Macro[X] = (=> Quotes) ?=> Expr[X]
```

- [x] https://github.com/scalameta/scalameta/pull/2409
```scala
given noCaller: Caller(???) with
   override def computeValue() = ()
```

```scala
given C(1) with {}
``` 

```scala     
given t1[T]: E[T]("low") with {}
```

- [x] https://github.com/scalameta/scalameta/pull/2404
```scala
given [T](using e: Exp[T]): Exp[NCtx => T] with
    import NCtx.*
    def lit(i: Int) = {
      case Pos => e.lit(i)
      case Neg => e.neg(e.lit(i))
    }
    def neg(x: NCtx => T): NCtx => T = {
     case Pos => x(Neg)
     case Neg => x(Pos)
    }
    def add(l: NCtx => T, r: NCtx => T): NCtx => T =
      c => e.add(l(c), r(c))
```
      
- [x] https://github.com/scalameta/scalameta/pull/2422
```scala
println(s"\"Hello\", $person")
```
 
- [x] https://github.com/scalameta/scalameta/pull/2419
```scala
check_success("\"\\u0024\" == \"$\"", s"\u0024", "$")
```

- [x] https://github.com/scalameta/scalameta/pull/2410
```scala
'{ val x: Int = ${ (q2) ?=> a } }
``` 

- [x] https://github.com/scalameta/scalameta/pull/2418
```scala
 def f(x: Int) =
  assert(
    if x > 0 then
      true
    else
      false
  )
  assert(
    if x > 0 then
      true
    else
      false)
  assert(
    if x > 0 then
      true
    else
      false, "fail")
  assert(
    if x > 0 then
      true
    else
      if x < 0 then
        true
      else
        false, "fail")
  (
    if x > 0 then
      println(x)
      x
    else
      s"""foo${
        if x > 0 then
          println(x)
          x
        else
          -x
        }"""
  )
```

- [x] https://github.com/scalameta/scalameta/pull/2417
```scala
  object Test:
  try List(1, 2, 3) match
  case x :: xs => println(x)
  case Nil => println("Nil")
  catch
  case ex: java.io.IOException => println(ex)
  case ex: Throwable => throw ex
  end try
```

[...]

PR #3034, opened 2023-01-08
ScalametaParser: parse annot args before params

kitbellew:
Allow annotations before class constructor, meaning annotation could have its arg clauses but has to stop before ctor param clauses. For #2405.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2947,6 +2947,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   def annotsBuf[T >: Mod.Annot](
       annots: ListBuffer[T],
       skipNewLines: Boolean,
+      insidePrimaryCtorAnnot: Boolean = false,
       allowArgss: Boolean = true
   ): Unit = {
     while (token match {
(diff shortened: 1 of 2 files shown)


kitbellew:
@tgodzik how does the compiler handle this case? it can probably tell literal goes to annotation, implicit goes to constructor, default values as well...

except, specifically, the ambiguity between ascribed annotation argument and constructor parameter `(foo: Foo)`? and what's the proper way to use this as a parameter, and not annotation argument?

tgodzik:
> except, specifically, the ambiguity between ascribed annotation argument and constructor parameter (foo: Foo)? and what's the proper way to use this as a parameter, and not annotation argument?

It doesn't accept type annotated parameters there, this will always be treated as a param. It seems it parses it annotation and those might not accept typed params?

It's somewhere around here:
https://github.com/lampepfl/dotty/blob/main/compiler/src/dotty/tools/dotc/parsing/Parsers.scala#L3702

tgodzik:
Actually, it seems everything is explained here https://github.com/lampepfl/dotty/blob/main/compiler/src/dotty/tools/dotc/parsing/Parsers.scala#L2586

tgodzik (changes requested):
Thanks for looking into this! I did some digging and I think I found how it's done in the compiler. The Scala 2 one should handle it the same.

tgodzik (approved):
LGTM!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3828,7 +3832,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   def primaryCtor(owner: TemplateOwner): Ctor.Primary = autoPos {
     if (owner.isClass || (owner.isTrait && dialect.allowTraitParameters) || owner.isEnum) {
       val mods = listBy[Mod] { buf =>
-        annotsBuf(buf, skipNewLines = false, allowArgss = false)
+        annotsBuf(buf, skipNewLines = false, allowArgss = false, insidePrimaryCtorAnnot = true)
```
kitbellew, line 3835: @tgodzik btw, what is `allowArgss = false` here? if i read the code that you shared correctly, at least dotty allows multiple, not sure about scala2.

PR #3041, opened 2023-01-13
TokenIterator: add peekToken

kitbellew:
That way, we substantially reduce overhead of looking ahead. This also allows avoiding forking in ScalametaParser in many cases. For #3031.


tgodzik (approved):
Looks good!

PR #3042, opened 2023-01-13
ScannerTokens: for outdent, precompute next tokens

kitbellew:
Rather than requiring a separate call to nextToken which will in reality repeat the same calculations but with remaining SepRegions, let's add those subsequent token refs via `.next` field. For #3031.


tgodzik (approved):
LGTM

Issue #3045, opened 2023-01-17
Turn TokenIterator into pure context-free iterator


kitbellew:
This means removing all `observe{In,Out}dented` methods. This is needed to unblock #2303.

PR #3047, opened 2023-01-18
ScannerTokens: fix `case` handling for all cases

kitbellew:
Introduce new regions for `try` and `catch` and remove case-specific flags in RegionIndent and RegionBrace.

Let's add RegionCaseMark following `match` and `catch`, to avoid special handling for other uses of the `case` keyword, such as `case class`, `enum` cases, partial functions or case generators.

For #3045.


tgodzik (approved):
Looks good! Just two questions.

PR #3048, opened 2023-01-18
ScalametaParser: support coloneol in refined types

kitbellew:
For #3045.

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/TypeSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/TypeSuite.scala
@@ -98,14 +98,47 @@ class TypeSuite extends BaseDottySuite {
   }
 
   test("coloneol-type3") {
-    runTestError[Stat](
+    runTestAssert[Stat](
       """|type A = Product:
          |  type T>: Null:
          |    type D <: Product
          |""".stripMargin,
-      """|error: ; expected but : found
-         |type A = Product:
-         |                ^""".stripMargin
+      assertLayout = Some(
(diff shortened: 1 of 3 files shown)


tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/TypeSuite.scala
```diff
@@ -98,14 +98,47 @@ class TypeSuite extends BaseDottySuite {
   }
 
   test("coloneol-type3") {
-    runTestError[Stat](
+    runTestAssert[Stat](
       """|type A = Product:
```
tgodzik, line 102: This doesn't seem to compile with Scala 3:
```
type A = Product:
[error]    |                  ^
[error]    |                  end of statement expected but ':' found
```

I don't think we can have colon for refined types?

kitbellew: interesting... i found the Refinement definition here https://dotty.epfl.ch/docs/internals/syntax.html, not sure if I interpreted it correctly. or is it not out yet?

kitbellew: in fact, there's no mention of `with` to refine...

tgodzik: Not sure, I will ask around.

tgodzik: It seems this will work from 3.3.x so this might have been a recent change.

tgodzik (approved):
LGTM

PR #3057, opened 2023-01-20
Trees: turn Defn.Var into proper Tree.WithBody

kitbellew:
Do not hide placeholder assignment.

### DIFF
--- a/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
@@ -580,19 +580,27 @@ object Defn {
       decltpe: Option[sm.Type],
       rhs: Term
   ) extends Defn with Stat.WithMods with Tree.WithDeclTpeOpt with Tree.WithBody {
+    checkFields(!rhs.is[Term.Placeholder])
     checkFields(pats.forall(!_.is[Term.Name]))
     override def body: Tree = rhs
+  }
+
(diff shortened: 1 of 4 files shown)


scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -574,25 +574,36 @@ object Decl {
...
+    def apply(rhs: Option[Term]): Term = rhs.getOrElse(Term.Placeholder())
+    def toOpt(body: Term): Option[Term] = if (isPlaceholder(body)) None else Some(body)
+  }
   @ast class Var(
       mods: List[Mod],
       pats: List[Pat] @nonEmpty,
       decltpe: Option[sm.Type],
-      rhs: Option[Term]
+      @replacesFields("4.7.2", VarRhsCtor)
+      body: Term
```
tgodzik, line 600: Does `var a: Int` and `var a: Int = _` become the same thing? Should we have something like an `EmptyTree` insteadto models the first case?

kitbellew: `var a: Int` becomes De**cl**.Var whereas the other one is De**fn**.Var

tgodzik: Ach, right! I forgot about it!

tgodzik (approved):
LGTM

PR #3061, opened 2023-01-21
YML: regroup CI tests

kitbellew:
Extracts structured tests into additional groups. Increases the number of tests by 5, by adding jdk11 tests for the two latest scala versions and jvm/js/native.


tgodzik (approved):
LGTM!

PR #3062, opened 2023-01-23
SBT: include scala 3 as well

kitbellew:


.github/workflows/ci.yml
```diff
@@ -38,6 +38,7 @@ jobs:
         scala:
           - '2.12.17'
           - '2.13.10'
+          - '3.2.1'
```
tgodzik, line 41: This seemed overly optimistic :laughing: We need to get rid or rewrtite the macros for it to work unfortunately :/

Issue #3067, opened 2023-02-06
Make Tree version migration easier and backwards-compatible


kitbellew:
@bjaglin pointed out, on scalameta discord, some deficiencies around the current tree versioning.

PR #3069, opened 2023-02-06
ast: make version and its parsing public

kitbellew:
Helps with #3067.

### DIFF
--- a/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
+++ b/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
@@ -5,8 +5,8 @@ package trees
 import scala.language.experimental.macros
 import scala.annotation.{StaticAnnotation, tailrec}
 import scala.collection.mutable.ListBuffer
+import scala.math.Ordered.orderingToOrdered
 import scala.reflect.macros.whitebox.Context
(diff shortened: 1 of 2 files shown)


scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
```diff
@@ -722,15 +719,14 @@ class AstNamerMacros(val c: Context) extends Reflection with CommonNamerMacros {
...
+  type Version = (Int, Int, Int)
+  val versionOrdering = implicitly[Ordering[Version]]
 
   private def versionToString(v: Version): String =
     s"${v._1}.${v._2}.${v._3}"
 
-  private def parseVersion(v: String): Version = {
-    val since = v.stripPrefix("\"").stripSuffix("\"")
-    val versions = since.split('.').map(_.toInt)
+  def parseVersion(v: String, delim: Char = '.'): Try[Version] = Try {
```
tgodzik, line 728: Actually, why do we need to make it public? Could we move it outside of  AstNamerMacros to some kind of Version class/object?

kitbellew: done, thank you

tgodzik (approved):
LGTM!

PR #3072, opened 2023-02-08
ast: preserve `apply/unapply` from older versions

### LINKED ISSUE #3067: Make Tree version migration easier and backwards-compatible

kitbellew:
Fixes #3067.

### DIFF
--- a/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
+++ b/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
@@ -79,6 +79,7 @@ class AstNamerMacros(val c: Context) extends Reflection with CommonNamerMacros {
         val replacedFields = versionedParams.flatMap(_.replaced.flatMap { field =>
           field.oldDefs.map { case (oldDef, _) => field.version -> oldDef }
         })
+        val mstatsPerVersion = paramsVersions.map { ver => (ver, ListBuffer[Tree]()) }
         def paramsForVersion(v: Version): List[ValDef] =
           positionVersionedParams(versionedParams.flatMap(_.getApplyDeclDefnBefore(v)._1))
 
@@ -351,7 +352,7 @@ class AstNamerMacros(val c: Context) extends Reflection with CommonNamerMacros {
         // generate new applies for each new field added
         // with field A, B and additional binary compat ones C, D and E, we generate:
         // apply(A, B, C), apply(A, B, C, D), apply(A, B, C, D, E)
-        paramsVersions.foreach { v =>
+        mstatsPerVersion.foreach { case (v, verMstats) =>
           val applyParamsBuilder = List.newBuilder[(ValDef, Int)]
           val applyBodyBuilder = List.newBuilder[Tree]
           versionedParams.foreach { vp =>
@@ -362,6 +363,7 @@ class AstNamerMacros(val c: Context) extends Reflection with CommonNamerMacros {
           val params = positionVersionedParams(applyParamsBuilder.result())
           applyBodyBuilder += q"$mname.apply(..$internalArgs)"
           val applyBody = applyBodyBuilder.result()
+          verMstats += q"def apply(..$params): $iname = { ..$applyBody }"
           mstats1 += q"@${getDeprecatedAnno(v)} def apply(..$params): $iname = { ..$applyBody }"
         }
 
@@ -381,10 +383,18 @@ class AstNamerMacros(val c: Context) extends Reflection with CommonNamerMacros {
           }
           if (params.nonEmpty) {
             val latestTree = getUnapply(params)
-            mstats1 += paramsVersions.headOption.fold {
-              latestTree
-            } { ver =>
-              getUnapply(paramsForVersion(ver), getDeprecatedAnno(ver))
+            mstatsPerVersion match {
+              case (headVer, headMstats) :: tail =>
+                val headParams = paramsForVersion(headVer)
+                headMstats += getUnapply(headParams)
+                tail.foreach { case (ver, verMstats) =>
+                  verMstats += getUnapply(paramsForVersion(ver))
+                }
+                val afterLastVer = getAfterVersion(mstatsPerVersion.last._1)
+                val anno = getDeprecatedAnno(headVer, s"; use `.$afterLastVer`")
+                mstats1 += getUnapply(headParams, anno)
+              case Nil =>
+                mstats1 += latestTree
             }
             mstatsLatest += latestTree
           } else {
@@ -421,13 +431,14 @@ class AstNamerMacros(val c: Context) extends Reflection with CommonNamerMacros {
           )
         }
 
-        mstats1 += q"""
-          object internal { // to be ignored by Mima
-            object Latest {
-              ..$mstatsLatest
-            }
-          }
-        """
+        val latestTermName = mstatsPerVersion.foldLeft(TermName(initialName)) {
+          case (afterPrevVer, (ver, verMstats)) =>
+            mstats1 += q"object $afterPrevVer { ..$verMstats }"
+            getAfterVersion(ver)
+        }
+        mstats1 += q"object $latestTermName { ..$mstatsLatest }"
+        // to be ignored by Mima, use "internal"
+        mstats1 += q"object internal { final val Latest = $latestTermName }"
 
         mstats1 += q"$mods1 class $name[..$tparams] $ctorMods(...${bparams1 +: paramss1}) extends { ..$earlydefns } with ..$parents1 { $self => ..$stats1 }"
         val cdef1 = q"$imods1 trait $iname extends ..$iparents1 { $iself => ..$istats1 }"
@@ -551,13 +562,13 @@ class AstNamerMacros(val c: Context) extends Reflection with CommonNamerMacros {
   private def getVersionedParams(
       params: List[ValDef],
       stats: List[Tree]
-  ): (List[VersionedParam], Seq[Version]) = {
+  ): (List[VersionedParam], List[Version]) = {
     val appendedFields: Map[String, Version] = getNewFieldVersions(params)
     val replacedFields: Map[String, Seq[ReplacedField]] = ReplacedField.getMap(params, stats)
     val versionsBuilder = Set.newBuilder[Version]
     appendedFields.values.foreach(versionsBuilder += _)
     replacedFields.values.foreach(_.foreach(versionsBuilder += _.version))
-    val versions = versionsBuilder.result().toSeq.sorted
+    val versions = versionsBuilder.result().toList.sorted
     val versionedParams = params.map { p =>
       val pname = p.name.toString
       val appended = appendedFields.get(pname)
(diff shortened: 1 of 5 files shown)


kitbellew:
@bjaglin can you please take a look? this should introduce persistent versioned objects with `unapply`, and quasiquotes should be compiled with a specific version.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
```diff
@@ -697,8 +709,11 @@ class AstNamerMacros(val c: Context) extends Reflection with CommonNamerMacros {
     }
   }
 
-  private def getDeprecatedAnno(v: Version) =
-    q"new scala.deprecated(${Literal(Constant(v.toString))})"
+  private def getDeprecatedAnno(v: Version, why: String = "") =
+    q"new scala.deprecated(${Literal(Constant(v.toString + why))})"
```
bjaglin, line 713: Slightly beyond the scope of this PR, but why not advertize the "since" version through the argument of the same name? It can be useful for filtering with `-Wconf`

tests/shared/src/test/scala/scala/meta/tests/parsers/TermSuite.scala
```diff
@@ -112,6 +112,18 @@ class TermSuite extends ParseSuite {
     assertTerm("f(x = xs: _*)") {
       Term.Apply(Term.Name("f"), List(Assign(Term.Name("x"), Repeated(Term.Name("xs")))))
     }
+    val Term.Apply.Until_4_6_0(
+      Term.Name("f"),
+      List(Assign(Term.Name("x"), Repeated(Term.Name("xs"))))
+    ) = term("f(x = xs: _*)")
+    val Term.Apply.After_4_6_0(
```
bjaglin, line 119: Considering the constraint "[Every Scala program linkable against x.y.z1 must be linkable against x.y.z2.](https://github.com/scalameta/scalameta/blob/main/COMPATIBILITY.md)", there shouldn't be any change in the 4.6.x releases, so we could omit the patch version
```suggestion
    val Term.Apply.After_4_6(
```

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
```diff
@@ -720,4 +735,25 @@ object AstNamerMacros {
     if (majorIdx > 0) Some(buildVersion.substring(0, majorIdx).toInt) else None
   }
 
+  private val initialName = "Initial"
```
bjaglin, line 738: what's the need for that extractor? isn't `case Foo.Initial(xxx)` the same as `case Foo(xxx)`?

tests/shared/src/test/scala/scala/meta/tests/parsers/TermSuite.scala
```diff
@@ -112,6 +112,18 @@ class TermSuite extends ParseSuite {
     assertTerm("f(x = xs: _*)") {
       Term.Apply(Term.Name("f"), List(Assign(Term.Name("x"), Repeated(Term.Name("xs")))))
     }
+    val Term.Apply.Until_4_6_0(
```
bjaglin, line 115: I am wondering what would be the use case for this one, as source compiling against 4.6.0 would rather use the existing/previous extractor with warnings or the new `After` one.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
```diff
@@ -421,13 +431,15 @@ class AstNamerMacros(val c: Context) extends Reflection with CommonNamerMacros {
...
-        """
+        val latestTermName = mstatsPerVersion.foldLeft(TermName(initialName)) {
+          case (afterPrevVer, (ver, verMstats)) =>
+            mstats1 += q"object $afterPrevVer { ..$verMstats }"
+            mstats1 += q"final val ${getUntilVersion(ver)} = $afterPrevVer"
+            getAfterVersion(ver)
+        }
+        mstats1 += q"object $latestTermName { ..$mstatsLatest }"
+        // to be ignored by Mima, use "internal"
+        mstats1 += q"object internal { final val Latest = $latestTermName }"
```
bjaglin, line 442: Quasiquotes expanded by 4.6.[0-3] depend on this, so it would be nice to have some MiMa to honor "Every Scala program linkable against x.y.0 must be linkable against x.{y+1}.z."? That won't be an issue for Scalafix as it didn't ship with 4.6.[0-3], but that might bite other downstream projects.

Maybe it can be deprecated, frozen to `After_4_6_0` (or the historical `unapply` if it does not exist).

tests/shared/src/test/scala/scala/meta/tests/parsers/TermSuite.scala
bjaglin: Is `Since` less ambiguous?
```suggestion
    val Term.Apply.Since_4_6(
```

bjaglin (commented):
Thanks for looking into that! A note about that versioning strategy might be useful in https://scalameta.org/docs/trees/guide.html#with-normal-constructors-1.

tests/shared/src/test/scala/scala/meta/tests/parsers/TermSuite.scala
kitbellew: @bjaglin while it might be correct in terms of semver, i don't want to overload this logic with those assumptions/requirements. i am simply using the version, as-is, without any semantic parsing, that is specified in the `@newField` or `@replacedField` annotation on a tree field.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
kitbellew: The whole point of `internal` was that it was not subject to mima (and it was named `Impl` before and has already been renamed anyway). The leak via quasiquotes was unintentional, but I'd like to keep it this way, and incur minor incompatibility if anyone has already used it.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
kitbellew: it is today but in 5.0.0 it won't be. in fact, I'd like to discourage `case Foo(xxx)` and might suggest completely removing the non-versioned `unapply` completely in next major.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
kitbellew: see comment above

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
```diff
@@ -381,10 +383,18 @@ class AstNamerMacros(val c: Context) extends Reflection with CommonNamerMacros {
...
-              getUnapply(paramsForVersion(ver), getDeprecatedAnno(ver))
+            mstatsPerVersion match {
+              case (headVer, headMstats) :: tail =>
+                val headParams = paramsForVersion(headVer)
+                headMstats += getUnapply(headParams)
+                tail.foreach { case (ver, verMstats) =>
+                  verMstats += getUnapply(paramsForVersion(ver))
+                }
+                val afterLastVer = getAfterVersion(mstatsPerVersion.last._1)
+                val anno = getDeprecatedAnno(headVer, s"; use `.$afterLastVer`")
```
kitbellew, line 394: isn't this the "since" version we are advertising? (see your comment before next to `def getDeprecatedAnno`)

tests/shared/src/test/scala/scala/meta/tests/parsers/TermSuite.scala
kitbellew: i thought about it, too. added `Until_` "for completeness", as an alias for the earlier `After_` but can remove it if it's confusing or unnecessary.

tests/shared/src/test/scala/scala/meta/tests/parsers/TermSuite.scala
kitbellew: sorry, didn't comment on `Since`. to me, "since" implies `>=` whereas this is strictly `>` which is what I tried to convey with "after".

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
kitbellew: another option is to remove Initial, remove deprecation from original Term.Apply but then remove the unapply method from it in the next major (if there are any intervening versions). would that work better?

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
kitbellew: in case i misunderstood it the first time: i can't always provide the "since" parameter.

kitbellew (commented):
1. removed `Until_`
2. amended `guide.md`

tests/shared/src/test/scala/scala/meta/tests/parsers/TermSuite.scala
bjaglin: Oh, I didn't realize `replacedField` had a `until` parameter, unlike `newField` and `replacesFields` which have a `since` one. Is there a reason for that?

I find `Since_4_7_0` more useful than `After_4_6_0` (which only tells me this syntax won't build against 4.6.0). Also, if I understand correctly, having the `since` would allow for a more precise/canonical deprecation message (see other comment above).

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
bjaglin: Fair enough, especially since I guess the upcoming release will remain in the 4.7.x serie, so there is absolutely no reason for clients to stick to the earlier 4.7.x versions with an non-guaranteed compatibility.

bjaglin (approved):
Still curious to hear about whether exposing the since version is a possibility but LGTM either way. Thanks again for finding ways to move forward without breaking compatibility!

tests/shared/src/test/scala/scala/meta/tests/parsers/TermSuite.scala
kitbellew: 1. this reminds me, I should rename `since` parameter in `newField` and `replacesFields` to `after`.
2. `replacedField` means this field was replaced after the given version and is no longer available, the parameter shows the last version it was present in
3. the other two are marked with the version after which they were added.

when we add the code which changes these fields, the only thing we know for a fact is which version was the last released one (and is exactly how `-SNAPSHOT` version is computed). similarly, we have a check that verifies that none of the version tags are later than the current build version. of course, this check will not be possible otherwise.

i can make an educated *guess* about the next release tag but since it hasn't happened yet, it's just a guess. and here i prefer not to do that :) (sort of like some parks in this area which state that they close 30 minutes _before_ dusk.)

if you change your code to depend on version 4.7.4, does it matter if the unapply is called `After_4_6_0` or `Since_4_7_0`? neither of them is `AsOf_4_7_4` anyway...

bjaglin: > i can make an educated guess about the next release tag but since it hasn't happened yet, it's just a guess.

Considering https://github.com/scalameta/scalameta/blob/main/COMPATIBILITY.md, any new or replaced field needs to be introduced in the next minor release (or major one, but I believe these are so rare that they could be considered the exception), so that helps educating your guess further ;) On a side note, forward compatibility within patch versions does not seem to be checked with MiMa though (but I believe it would be possible to run it in forward mode?).

> if you change your code to depend on version 4.7.4, does it matter if the unapply is called After_4_6_0 or Since_4_7_0? neither of them is AsOf_4_7_4 anyway...

Agreed! It's more that from a consumer perspective, it just takes a bit of thinking to understand why you get a deprecation warning mentioning 4.6.0 when bumping from 4.6.0 to 4.7.x and end up having code mentioning the *prior* version to get rid of it.

I wish there was more tooling available to help developers provide a "since-centric" view (since that's what deprecation annotation uses). From the projects I know, it's mostly handled manually (based on educated guesses).

PR #3088, opened 2023-02-21
ScalametaParser: no repeated initial implicit mod

kitbellew:
The existing code already checks for repeated modifiers on parameters except for the implicit/using modifier on the first parameter.


kitbellew:
combined with #3087.

Issue #3093, opened 2023-03-10
Anonymous class without braces in if block with braces fails to parse


tanishiking:
Related https://github.com/scalameta/metals/pull/4877

Scalameta >= 4.7.2 fails to parse the following program with Scala3 dialect

```scala
if(x) {
  new TreeMap:
    override def foo = ???
}
```

while the following works 

```scala
if(x)
  new TreeMap:
    override def foo = ???
```

```
scala> import scala.meta._
import scala.meta._

scala> val program = """
     | if(x) {
     |   new TreeMap:
     |     def foo = ???
     | }
     | """
val program: String =
"
if(x) {
  new TreeMap:
    def foo = ???
}
"

scala> dialects.Scala3(program).parse[Stat]
val res0: scala.meta.parsers.Parsed[scala.meta.Stat] =
<input>:3: error: expected template body
  new TreeMap:
              ^

```

PR #3096, opened 2023-03-11
Allow Optional Braces Around Template Bodies anyway

tanishiking:
fix https://github.com/scalameta/scalameta/issues/3093

It seems like we should accept the optional braces around template body when we accept `:`, even inside the block with braces.

I'm trying to confirm this is valid change from syntax specification https://docs.scala-lang.org/scala3/reference/other-new-features/indentation.html (that's why it's still draft)...

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -4175,14 +4175,13 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     } else if (isColonEol()) {
       accept[Colon]
 
-      val nextIndented =
-        if (enumCaseAllowed)
-          in.observeIndentedEnum()
-        else
-          in.observeIndented()
+      if (enumCaseAllowed)
+        in.observeIndentedEnum()
+      else
+        in.observeIndented()
 
-      if (nextIndented)
-        indentedOnOpen(templateStatSeq(enumCaseAllowed, secondaryConstructorAllowed))
+      if (acceptOpt[Indentation.Indent])
+        indentedAfterOpen(templateStatSeq(enumCaseAllowed, secondaryConstructorAllowed))
       else if (!enumCaseAllowed && isEndMarkerIntro(tokenPos))
         (selfEmpty(), Nil)
       else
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -4181,12 +4181,12 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
-        indentedOnOpen(templateStatSeq(enumCaseAllowed, secondaryConstructorAllowed))
-      else if (!enumCaseAllowed && isEndMarkerIntro(tokenPos))
+      if (!enumCaseAllowed && isEndMarkerIntro(tokenPos) && !nextIndented)
         (selfEmpty(), Nil)
-      else
-        syntaxError("expected template body", token)
+      else 
+        // https://docs.scala-lang.org/scala3/reference/other-new-features/indentation.html#optional-braces-around-template-bodies
+        // https://github.com/scalameta/scalameta/issues/3093
+        indentedOnOpen(templateStatSeq(enumCaseAllowed, secondaryConstructorAllowed))
```
kitbellew, line 4189: `indentedOnOpen` may be called only if it's definitely known that the current token is `Indentation.Indent`, and in this case it isn't.

I'd change to this:
```
      if (enumCaseAllowed)
        in.observeIndentedEnum()
      else
        in.observeIndented()

      if (acceptOpt[Indentation.Indent])
        indentedAfterOpen(templateStatSeq(enumCaseAllowed, secondaryConstructorAllowed))
```
and the rest could remain unchanged (`else if (!enumCaseAllowed...`).

kitbellew (changes requested):

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tanishiking: Thanks! that makes sense :+1:

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -4181,8 +4181,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
         else
           in.observeIndented()
 
-      if (nextIndented)
```
kitbellew, line 4184: the nextIndented variable is no longer used.

kitbellew (approved):
one minor comment

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tanishiking: Oops, right, thanks!

kitbellew (approved):

PR #3104, opened 2023-03-28
Fix #3103

### LINKED ISSUE #3103: Scalameta crash with Hydra
One of our customers has failing builds caused by a race condition in the semanticdb serialization code. This race condition does _not_ exist in the open-source Scala version, but I hope we can find a convenient solution that can help our customer.

[Hydra](https://www.triplequote.com/hydra/compilation/) works by splitting source files in several batches, and distributing each batch to a Scala compiler worker. These workers are, roughly speaking, instances of the Scala compiler and they run in parallel, including their compiler plugins.

The race condition occurs between two threads, one that tries to write the first semanticdb file in a new directory, and another thread that finished and runs `RemoveOrphanSemanticdbFiles.process`. The second one sees an empty directory before the first thread has a chance to write the first file, removes it, and the first thread crashes.

I believe the change in https://github.com/scalameta/scalameta/pull/2605/files introduced this race condition, or at least, it made it much more likely to occur in practice. In our tests, it looks like protobuf generation may take >1s (wall clock time) for the project in question, and that's plenty of time to run into the issue.

I think a simple workaround that fixes the issue for "most" people would be to generate the protobuf message before creating the parent directory, and minimize the window where the directory can be deleted. 

I'll follow up with a PR to that effect.

dragos:
Move protobuf serialization before directory creation. This reduces the chance of a race condition with the file-deletion code, when running inside Hydra, the parallel Scala compiler.

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/package.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/package.scala
@@ -13,10 +13,10 @@ package object scalac {
         if (append) StandardOpenOption.APPEND
         else StandardOpenOption.TRUNCATE_EXISTING
       val out = SemanticdbPaths.toSemanticdb(sdocument, targetroot)
+      val bytes = s.TextDocuments(sdocument :: Nil).toByteArray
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
LGTM! Let us know if this fixes issues with Hydra.

PR #3109, opened 2023-04-23
Correct occurences in for comprehensions containing `withFilter`

susliko:
Fixes https://github.com/scalameta/metals/issues/4548

### DIFF
--- a/tests/jvm/src/test/resources/metacp.expect_2.12
+++ b/tests/jvm/src/test/resources/metacp.expect_2.12
@@ -1764,12 +1764,16 @@ Schema => SemanticDB v4
 Uri => example/ForComprehension.class
 Text => empty
 Language => Scala
-Symbols => 2 entries
+Symbols => 3 entries
 
 Symbols:
-example/ForComprehension# => class ForComprehension extends AnyRef { +1 decls }
+example/ForComprehension# => class ForComprehension extends AnyRef { +2 decls }
   AnyRef => scala/AnyRef#
 example/ForComprehension#`<init>`(). => primary ctor <init>()
+example/ForComprehension#x. => val method x: Option[Tuple2[Int, Int]]
+  Option => scala/Option#
+  Tuple2 => scala/Tuple2#
+  Int => scala/Int#
 
 example/ImplicitConversion.class
 --------------------------------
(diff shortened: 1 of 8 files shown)


tests/jvm/src/test/resources/metacp.expect_2.12
```diff
@@ -1487,8 +1487,7 @@ com/javacp/Test#staticField. => private[javacp] static field staticField: Int
 com/javacp/Test#staticMethod(). => private[javacp] static method staticMethod(): Unit
   javacp => com/javacp/
   Unit => scala/Unit#
-com/javacp/Test#strictfpMethod(). => @strictfp private[javacp] method strictfpMethod(): Unit
```
tgodzik, line 1490: Why change this? It seems to be failing now. You can use the `save-expect` command to update the expected file.

susliko: That's the result of  `save-expect` (even if invoked on main branch) with jvm 17+
Compilation also gives a warning:
```
[warn] /.../scalameta/semanticdb/integration/src/main/java/com/javacp/Test.java:69:1:  [strictfp] as of release 17, all floating-point expression
s are evaluated strictly and 'strictfp' is not required
[warn]     strictfp void strictfpMethod() {}
[warn]                   ^1 warning
```

tgodzik: Maybe the issue is that we run the tests with 8/11 then :thinking:

susliko: Yep
Another option is to remove this `strictfpMethod` from tests 😃

tgodzik: We can for sure change that, I think this was added as a test for annotations if I am not mistaken

susliko: Not sure about annotations, it's defined [here]:(https://github.com/scalameta/scalameta/blob/main/semanticdb/integration/src/main/java/com/javacp/Test.java#L69) 
```java
strictfp void strictfpMethod() {}
```

tgodzik: ach, that might be something different then, sorry I didn't look exactly. The best case here is to run sbt with JDK 11, if it's a problem for you I can do it later on.

susliko: Done

tgodzik (approved):
LGTM!

PR #3120, opened 2023-05-05
chore: Add support for Scala 2.12.18

tgodzik:


SethTisue:
2.12.18 is on Maven Central

PR #3129, opened 2023-05-22
Fixed .syntax on ifs with an inner if without an else

### LINKED ISSUE #3128: Syntax function does not return the correct value for Ifs with an inner if without an else
Using:
  - Scalameta version 4.7.7
  - Scala 2.13.10

Steps to reproduce:
```scala
import scala.meta._

val expr1 = q"if (someCondition1) println(123)"
val expr2 = q"if (someCondition2) ($expr1) else println(321)"

println(expr2.syntax) // this is incorrect
```

actual:
```scala
if (someCondition2) if (someCondition1) println(123) else println(321)
```

expected:
```scala
if (someCondition2) (if (someCondition1) println(123)) else println(321)

OR

if (someCondition2) if (someCondition1) println(123) else () else println(321)
```

---

The `.syntax` is changing the semantics of the expression which shouldn't happen.

RCMartins:
Fixes #3128

### DIFF
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
@@ -86,7 +86,13 @@ object TreeSyntax {
     import SyntacticGroup.Type._, SyntacticGroup.Term._, SyntacticGroup.Pat._,
       SyntacticGroup.Literal, SyntacticGroup.Path
 
-    def p(og: SyntacticGroup, t: Tree, left: Boolean = false, right: Boolean = false) = {
+    def p(
+        og: SyntacticGroup,
+        t: Tree,
+        left: Boolean = false,
+        right: Boolean = false,
+        force: Boolean = false
+    ) = {
       def opNeedsParens(oo: String, io: String, customPrecedence: Boolean = true): Boolean = {
         def precedence(op: String) = if (customPrecedence) op.precedence else 0
         require(left != right)
@@ -111,7 +117,8 @@ object TreeSyntax {
         }
       }
       s(t) match {
-        case Show.Meta(ig: SyntacticGroup, res) if groupNeedsParens(og, ig) => s("(", res, ")")
+        case Show.Meta(ig: SyntacticGroup, res) if force || groupNeedsParens(og, ig) =>
+          s("(", res, ")")
         case res => res
       }
     }
@@ -423,7 +430,7 @@ object TreeSyntax {
             " (",
             t.cond,
             ") ",
-            if (needParens) s("(", t.thenp, ")") else p(Expr, t.thenp),
+            p(Expr, t.thenp, force = needParens),
             if (guessHasElsep(t)) s(" ", kw("else"), " ", p(Expr, t.elsep)) else s()
           )
         )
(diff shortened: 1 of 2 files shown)


tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
```diff
@@ -464,6 +464,8 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
   test("some tricky parenthesization") {
     assertEquals(templStat("if (1) 2 else 3 + 4").syntax, "if (1) 2 else 3 + 4")
     assertEquals(templStat("(if (1) 2 else 3) + 4").syntax, "(if (1) 2 else 3) + 4")
+    assertEquals(templStat("if (1) (if (2) 3) else 4").syntax, "if (1) (if (2) 3) else 4")
```
kitbellew, line 467: can you please split your PR into two commits, one adding the tests (and the current, pre-fix, behaviour), and the second fixing the problem and showing the difference in the tests?

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -411,6 +411,11 @@ object TreeSyntax {
           case _ => block(s(), t.stats)
         }
       case t: Term.If =>
+        val needParens: Boolean =
+          guessHasElsep(t) && (t.thenp match {
+            case innerIf: Term.If => !guessHasElsep(innerIf)
+            case _ => false
+          })
```
kitbellew, line 418:
```suggestion
        val needParens: Boolean = t.thenp match {
          case innerIf: Term.If => !guessHasElsep(innerIf) && guessHasElsep(t)
          case _ => false
        }
```

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -419,7 +424,7 @@ object TreeSyntax {
             " (",
             t.cond,
             ") ",
-            p(Expr, t.thenp),
+            if (needParens) s("(", t.thenp, ")") else p(Expr, t.thenp),
```
kitbellew, line 427:
```suggestion
            w("(", p(Expr, t.thenp), ")", needParens),
```

kitbellew (changes requested):

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
kitbellew: actually, maybe never mind, as `p()` could add own parens.

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
RCMartins: Ok, done

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
RCMartins: if `p()` had a `force: Boolean` parameter we could use `p(Expr, t.thenp, force = needParens)`, maybe that would be better?

kitbellew: > if `p()` had a `force: Boolean` parameter we could use `p(Expr, t.thenp, force = needParens)`, maybe that would be better?

that would be nice :)

kitbellew (approved):
thank you!

PR #3137, opened 2023-05-27
ScalametaParser: allow block as catch handler

### LINKED ISSUE #3136: Parse error on Term.TryWithHandler when using a Term.Block in the catch clause
Using:
  - Scalameta version 4.7.7
  - Scala 2.13.10

Valid scala code:
```scala
def handler(n: Int)(t: Throwable): Unit =
  println(s"handled! (with n=$n)")
try ??? catch {
  val a = 10
  handler(a)
}
// prints:
// handled! (with n=10)
```

scalameta can't parse the `try`
```scala
import scala.meta._

q"""try ??? catch {
  val a = 10
  handler(a)
}"""
```

---

actual:
```bash
cmd31.sc:2: illegal start of simple expression
  val a = 10
  ^
Compilation Failed
```

expected:
```scala
Term.TryWithHandler(
  expr = Term.Name(value = "???"),
  catchp = Term.Block(
    stats = List(
      Defn.Val(
        mods = List(),
        pats = List(Pat.Var(name = Term.Name(value = "a"))),
        decltpe = None,
        rhs = Lit.Int(value = 10)
      ),
      Term.Apply(
        fun = Term.Name(value = "handler"),
        argClause = Term.ArgClause(values = List(Term.Name(value = "a")), mod = None)
      )
    )
  ),
  finallyp = None
)
```

kitbellew:
Fixes #3136.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1520,30 +1520,29 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           case LeftBrace() | Indentation.Indent() => block()
           case _ => expr()
         }
-        def caseClausesOrExpr = caseClausesIfAny().getOrElse(expr())
-        val catchopt =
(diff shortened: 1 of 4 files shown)


tgodzik (approved):
LGTM!

Issue #3138, opened 2023-05-27
Parsing issue in ApplyInfix with a tuple and braces in a new line


RCMartins:
Using:
  - Scalameta version 4.7.7
  - Scala 2.13.10

Steps to reproduce:
```scala
import scala.meta._

val term = 
  q"""
    Foo bar (2, 3)
    { a }
  """

term.structure
```

The above expression should be parsed has an Term.ApplyInfix, but it is being parsed as a Term.Block.
It should be equivalent to parsing `Foo bar (2, 3) { a }` or `Foo bar ((2, 3){a})`.

actual:
```scala
Term.Block(
  stats = List(
    Term.ApplyInfix(
      lhs = Term.Name(value = "Foo"),
      op = Term.Name(value = "bar"),
      targClause = Type.ArgClause(values = List()),
      argClause = Term.ArgClause(values = List(Lit.Int(value = 2), Lit.Int(value = 3)), mod = None)
    ),
    Term.Block(stats = List(Term.Name(value = "a")))
  )
)
```

expected:
```scala
Term.ApplyInfix(
  lhs = Term.Name(value = "Foo"),
  op = Term.Name(value = "bar"),
  targClause = Type.ArgClause(values = List()),
  argClause = Term.ArgClause(
    values = List(
      Term.Apply(
        fun = Term.Tuple(args = List(Lit.Int(value = 2), Lit.Int(value = 3))),
        argClause = Term.ArgClause(values = List(Term.Block(stats = List(Term.Name(value = "a")))), mod = None)
      )
    ),
    mod = None
  )
)
```

---

Here is a small Scala 2.13 example that uses this syntax:
```scala
object Foo {
  def bar(tuple: (Int, Int)): List[Int] = List(tuple._1, tuple._2)
}
implicit class Extension(tuple: (Int, Int)) {
  def apply(n: Int): (Int, Int) = (tuple._1 + n, tuple._2 + n)
}
val a = 5
println(
  Foo bar (2, 3)
  { a }
)
// prints
// List(7, 8)
```

PR #3139, opened 2023-05-27
PrettyPrinter bug fix on Term.Block

### LINKED ISSUE #2046: Pretty print of block
Hey folks, 

I encounter incorrect behavior of pretty print from scala meta in case of blocks:

```
import scala.meta._

val a = q"val a = 1; a"

val tree = q"val b = 1; $a"

println(tree.syntax)
```

produce: 

```
{
  val b = 1
  {
    val a = 1
    a
  }
}
```
which is incompatible with scala syntax.

https://scastie.scala-lang.org/GaU07vEdSxOMuUArWk8Zyw

RCMartins:
Fixes #2046 

I tried to create a complete solution that works in all cases, not just the ones reported in the issue.
I added tests for all `Stat` combinations that I could find in https://scalameta.org/docs/trees/quasiquotes.html. Let me know if there is any relevant one that I missed.

---

### Code style
In `guessNeedsLineSep` function I "expanded" the pattern matching of the stats (e.g. `Defn.Val(_, _, _, rhs) =>`), but maybe you would prefer a more concise solution.
Maybe something like this instead?
```scala
...
case t: Decl.Type => false
case t: Defn.Val => guessNeedsLineSep(t.rhs)
case t: Defn.Var => guessNeedsLineSep(t.body)
case t: Defn.Def => guessNeedsLineSep(t.body)
...
```

---

Note:
I think it would be better to not merge until the bugs I found while working on this (#3136 and #3138) are resolved.

### DIFF
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
@@ -296,6 +296,30 @@ object TreeSyntax {
       }
       !isOnlyChildOfOnlyChild(t)
     }
+    @annotation.tailrec
+    def guessNeedsLineSep(t: Tree): Boolean = t match {
+      case t: Term.ApplyUnary => guessNeedsLineSep(t.arg)
+      case _: Decl.Val | _: Decl.Var | _: Decl.Def | _: Defn.Type | _: Term.Ref | _: Member.Apply |
+          _: Term.Ascribe | _: Term.Tuple | _: Term.New | _: Term.Interpolate | _: Term.Xml |
+          _: Lit =>
+        true
+      case t: Term.Do => false
+      case t: Tree.WithBody => guessNeedsLineSep(t.body)
+      case t: Stat.WithTemplate => t.templ.self.isEmpty && t.templ.stats.isEmpty
+      case t: Term.ApplyInfix =>
+        val args = t.argClause.values
+        args.lengthCompare(1) != 0 || guessNeedsLineSep(args.head)
+      case t: Term.Return => guessNeedsLineSep(t.expr)
+      case t: Term.Throw => guessNeedsLineSep(t.expr)
+      case t: Term.If => guessNeedsLineSep(if (guessHasElsep(t)) t.elsep else t.thenp)
+      case t: Term.Try =>
+        t.finallyp match {
+          case Some(x) => guessNeedsLineSep(x)
+          case _ => t.catchp.isEmpty && guessNeedsLineSep(t.expr)
+        }
+      case t: Term.TryWithHandler => guessNeedsLineSep(t.finallyp.getOrElse(t.catchp))
+      case _ => false
+    }
 
     // Branches
     implicit def syntaxTree[T <: Tree]: Syntax[T] = Syntax {
@@ -398,10 +422,24 @@ object TreeSyntax {
       case t: Term.Tuple => m(SimpleExpr1, s("(", r(t.args, ", "), ")"))
       case t: Term.Block =>
         import Term.{Block, Function}
+        def withBlankLines(x: Seq[Stat]) = {
+          val builder = List.newBuilder[Show.Result]
+          var prevStat: Stat = null
+          x.foreach { stat =>
+            if (stat.is[Block] && null != prevStat && guessNeedsLineSep(prevStat)) builder += EOL
+            builder += i(stat)
+            prevStat = stat
+          }
+          builder.result()
+        }
         def block(pre: Show.Result, x: Seq[Stat]) = {
           val res =
             if (pre == Show.None && x.isEmpty) s("{}")
-            else s("{", w(" ", pre), r(x.map(i(_))), n("}"))
+            else {
+              val stats =
+                if (dialect.allowSignificantIndentation) x.map(i(_)) else withBlankLines(x)
+              s("{", w(" ", pre), r(stats), n("}"))
+            }
           m(SimpleExpr, res)
         }
         t match {
--- a/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
@@ -1933,4 +1933,36 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
     assertSyntax(tree, "\"\"\"\n\\\"\\\"\\\"\"\"\"")(tree)
   }
 
+  test("#2046 new line separator required before some Term.Block") {
+    assertEquals(
+      blockStat("{{a}; val b = 1; val c = 2; {d}; {e}}").syntax,
+      s"""|{
+          |  {
+          |    a
+          |  }
+          |  val b = 1
+          |  val c = 2
+          |
+          |  {
+          |    d
+          |  }
+          |  {
+          |    e
+          |  }
+          |}
+        """.stripMargin.trim.replace("\n", EOL)
+    )
+    assertEquals(
+      blockStat("{if (a) while (b) for (c <- d) -e; {f}}").syntax,
+      s"""|{
+          |  if (a) while (b) for (c <- d) -e
+          |
+          |  {
+          |    f
+          |  }
+          |}
+        """.stripMargin.trim.replace("\n", EOL)
+    )
+  }
+
 }
--- a/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/TreeSyntaxSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/TreeSyntaxSuite.scala
@@ -18,7 +18,7 @@ class TreeSyntaxSuite extends scala.meta.tests.parsers.ParseSuite {
   ): Unit = {
     val stat = statStr.trim // make sure no trailing newlines
     test(s"${loc.line}: $stat") {
-      val sep = "" // XXX: after fixing TreeSyntax, modify this
+      val sep = if (needNL) "\n" else ""
       val statSyntax = Option(syntaxStr).getOrElse(stat).replace("\n", "\n  ")
 
       val expectedSyntax =


scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -398,10 +457,24 @@ object TreeSyntax {
...
-            else s("{", w(" ", pre), r(x.map(i(_))), n("}"))
+            else
+              s(
+                "{",
+                w(" ", pre),
+                r(if (dialect.allowSignificantIndentation) x.map(i(_)) else requireNewLines(x)),
+                n("}")
+              )
           m(SimpleExpr, res)
         }
```
kitbellew, line 479: I think this is simpler and checks only on demand:
```
def withBlankLines(x: Seq[Stat]) = {
  val builder = List.newBuilder[Show.Result]
  var prevStat = null
  x.foreach { stat =>
    if (stat.is[Block] && null != prevStat && guessNeedsLineSep(prevStat)) builder += EOL
    builder += i(stat)
    prevStat = stat
  }
  builder.result()
}

def block(pre: Show.Result, x: Seq[Stat]) = {
  val res =
    if (pre == Show.None && x.isEmpty) s("{}")
    else {
      val stats = if (dialect.allowSignificantIndentation) x.map(i) else withBlankLines(x)
      s("{", w(" ", pre), r(stats), n("}"))
    }
  m(SimpleExpr, res)
}
```

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -296,6 +296,65 @@ object TreeSyntax {
...
+      case _: Term.New => true
+      case _: Term.NewAnonymous => false
+      case _: Term.Placeholder => unreachable
+      case _: Term.Eta => false
+      case _: Term.Repeated => false
+      case _: Term.Param => unreachable
+      case _: Term.Interpolate => true
+      case _: Term.Xml => true
+      case _: Import => false
+      case _: Lit => true
```
kitbellew, line 355: why here?

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
```diff
@@ -1933,4 +1933,161 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
     assertSyntax(tree, "\"\"\"\n\\\"\\\"\\\"\"\"\"")(tree)
   }
 
+  test("#2046 new line separator required before some Term.Block") {
```
kitbellew, line 1936: a request: could you see if it's possible to move `runTestAssert` from BaseDottySuite to its parent, ParseSuite, along with three implicit `parseXxxx` methods at the top of the class?

and then here, use it:

```
val code = "{val ..."
val syntax = s""" ... """
runTestAssert[Stat](code, Some(syntax)) {
  Term.Block(...)
}
```
that way, we not only test syntax but also that both the original code and the syntax are parsed exactly the same.

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
```diff
@@ -1933,4 +1933,161 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
...
+        }
+    }
+
+    def sep(t: Stat, expected: String): Unit = testBlock(t, expected, needSep = true)
+    def noSep(t: Stat, expected: String): Unit = testBlock(t, expected, needSep = false)
+
+    def declTest(f: String => Unit): Unit = Seq("val", "var", "def").foreach(f)
+    def classTest(f: String => Unit): Unit = Seq("class", "object", "trait").foreach(f)
+
+    declTest(k => sep(templStat(s"$k foo: Int"), s"$k foo: Int"))
```
kitbellew, line 2017: for every case where `guessXxx` returns true, could we add a test, please (unless it's already been done)?

kitbellew (changes requested):
thank you for this contribution! below are some suggestions and comments.

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
kitbellew: moved it myself in #3140, perhaps you could use it.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -296,6 +296,65 @@ object TreeSyntax {
...
+      case _: Decl.Type => false
+      case Defn.Val(_, _, _, rhs) => guessNeedsLineSep(rhs)
+      case Defn.Var.After_4_7_2(_, _, _, body) => guessNeedsLineSep(body)
+      case Defn.Def.After_4_7_3(_, _, _, _, body) => guessNeedsLineSep(body)
+      case _: Defn.Macro => true
+      case _: Defn.Type => true
+      case Defn.Class.After_4_6_0(_, _, _, _, template) => guessNeedsLineSep(template)
+      case Defn.Trait.After_4_6_0(_, _, _, _, template) => guessNeedsLineSep(template)
+      case Defn.Object(_, _, template) => guessNeedsLineSep(template)
+      case _: Term.This => true
```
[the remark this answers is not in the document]
RCMartins: I checked every single case by hand on a Scala 2.13.10 test project. Unless I'm missing something `this` with a block after does not compile:
```scala
object Main9 {
    this
    {
      val a: Int = 0
      a
    }
  }
}
```
```
sbt:test> compile
[info] compiling 10 Scala sources and 1 Java source to C:\Scala\test\target\scala-2.13\classes ...
[error] C:\Scala\test\src\main\scala\Main9.scala:4:5: Main9.type does not take parameters
[error]     {
[error]     ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
```

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -296,6 +296,65 @@ object TreeSyntax {
...
+      t.stats.isEmpty
+    @annotation.tailrec
+    def guessNeedsLineSep(t: Stat): Boolean = t match {
+      case _: Decl.Val => true
+      case _: Decl.Var => true
+      case _: Decl.Def => true
+      case _: Decl.Type => false
+      case Defn.Val(_, _, _, rhs) => guessNeedsLineSep(rhs)
+      case Defn.Var.After_4_7_2(_, _, _, body) => guessNeedsLineSep(body)
+      case Defn.Def.After_4_7_3(_, _, _, _, body) => guessNeedsLineSep(body)
```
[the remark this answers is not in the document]
RCMartins: I used the versioned `unapply` because I was using Intellij at first and the default apply has showing as deprecated, that was the only reason. I'll use your sugestions to improve the code, thank you!

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -296,6 +296,65 @@ object TreeSyntax {
...
+      case _: Defn.Macro => true
+      case _: Defn.Type => true
+      case Defn.Class.After_4_6_0(_, _, _, _, template) => guessNeedsLineSep(template)
+      case Defn.Trait.After_4_6_0(_, _, _, _, template) => guessNeedsLineSep(template)
+      case Defn.Object(_, _, template) => guessNeedsLineSep(template)
+      case _: Term.This => true
+      case _: Term.Super => unreachable
+      case _: Term.Name => true
+      case _: Term.Select => true
+      case Term.ApplyUnary(_, arg) => guessNeedsLineSep(arg)
```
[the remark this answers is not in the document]
RCMartins: Sorry, are you asking why this particular case `ApplyUnary` requires the separation between blocks? I created at least one test for every case in the `guessNeedsLineSep` function if you want to check the particular test cases.
The tests are in the same order as the cases (for now, before the refactoring), so it's easy to find them. The cases that are constant (true or false) there is only one case, for the remaining cases I created a positive and a negative test.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -296,6 +296,65 @@ object TreeSyntax {
...
+      case _: Term.Apply => true
+      case _: Term.ApplyType => true
```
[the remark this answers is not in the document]
RCMartins: I'm not familiar with the `Member` object. For what I can see it seems useful for pattern matching on common fields between the elems. I'll try to use it

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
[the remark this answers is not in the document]
kitbellew: understood. yes, for some classes, `unapply` was preserved for MIMA compatibility and made deprecated. but let's not use `unapply` :)

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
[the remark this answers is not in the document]
kitbellew: is this interpreted as an invocation of a secondary ctor? :)

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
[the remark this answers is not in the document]
kitbellew: yes, about this particular case. somehow it wasn't clear to me initially, but perhaps i can see it now :)

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
RCMartins: Checking only on demand is definitely better, as this Block cases are rare. The list builder is also better than my concatenation + reverse thing I had.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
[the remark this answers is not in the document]
RCMartins: I think it's something like that, but q"this; {a}" requires a new line (or the ';').
Because even if there is a way for `this\n{a}` to compile, it's still not the original expression for sure :)

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
RCMartins: Ok, I'll try to use it, thanks!

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -296,6 +296,65 @@ object TreeSyntax {
...
     }
+    def guessNeedsLineSep(t: Template): Boolean =
+      t.stats.isEmpty
+    @annotation.tailrec
+    def guessNeedsLineSep(t: Stat): Boolean = t match {
+      case _: Decl.Val => true
+      case _: Decl.Var => true
+      case _: Decl.Def => true
+      case _: Decl.Type => false
+      case Defn.Val(_, _, _, rhs) => guessNeedsLineSep(rhs)
```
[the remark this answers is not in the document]
RCMartins: `Tree.WithBody` has a `body: Tree` so it's incompatible.
I can use `case t: Tree.WithBody => guessNeedsLineSep(t.body.asInstanceOf[Stat])` or just change the recursive function to `def guessNeedsLineSep(t: Tree)`. I'm more inclined to changing the function, what do you think?

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
RCMartins: Literal + \n + { expr } is an apply, so
```scala
"foo"
{ 0 }
```
is `"foo"(0) == 'f'`, so it requires a new line here.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -296,6 +296,65 @@ object TreeSyntax {
...
+      case _: Term.ApplyType => true
+      case Term.ApplyInfix.After_4_6_0(_, _, _, Term.ArgClause(term :: Nil, _)) =>
+        guessNeedsLineSep(term)
+      case _: Term.ApplyInfix => true
+      case Term.Assign(_, rhs) => guessNeedsLineSep(rhs)
+      case Term.Return(expr) => guessNeedsLineSep(expr)
+      case Term.Throw(expr) => guessNeedsLineSep(expr)
+      case _: Term.Ascribe => true
+      case _: Term.Annotate => false
+      case _: Term.Tuple => true
```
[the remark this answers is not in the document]
RCMartins: Without the new line it calls .apply() on the tuple:
```scala
implicit class Extension(tuple: (Int, Int)) {
  def apply(n: Int): (Int, Int) = (tuple._1 + n, tuple._2 + n)
}
println {
  (1, 2)
  {
    5
  }
}
```
```
(6,7)
```

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
RCMartins: I tried to use this:
```scala
def testBlock(stat: String, expectedStr: String, needSep: Boolean): Unit = {
  def expected(sep: String): String =
    ...
  val code = s"$stat;{a}"
  val newBlock = templStat(code)
  val syntax = expected(sep = if (needSep) "\n" else "")
  runTestAssert[Stat](code, Some(syntax))(newBlock)
  ...
}
def noSep(t: String, expected: String): Unit = testBlock(t, expected, needSep = false)
noSep("type foo", "type foo")
```
But I'm having a somewhat cryptic error (at least for me). I don't see anything different in the other usages of `runTestAssert`, so I'm a little lost.
This error below has for `type foo`, but I think it's happening in all the tests:
```
Expected all trees to have non-empty `.origin`.
To fix this failure, update ScalametaParser to use `autoPos()` where the trees below got constructed.
Pro tip: you may also want to add a PositionSuite test for this tree node to verify that the position you set is correct.
=> Obtained
List(
  "Decl.Type(Nil, Type.Name(\"foo\"), Type.ParamClause(Nil), Type.Bounds(None, None))",
  "Type.Name(\"foo\")",
  "Type.ParamClause(Nil)",
  "Type.Bounds(None, None)"
)
=> Diff (- obtained, + expected)
-List(
-  "Decl.Type(Nil, Type.Name(\"foo\"), Type.ParamClause(Nil), Type.Bounds(None, None))",
-  "Type.Name(\"foo\")",
-  "Type.ParamClause(Nil)",
-  "Type.Bounds(None, None)"
-)
+Nil
    at munit.Assertions$.assertEquals(Assertions.scala:14)
    at scala.meta.tests.parsers.MoreHelpers$.requireNonEmptyOrigin(ParseSuite.scala:151)
    at scala.meta.tests.parsers.ParseSuite.runTestAssert(ParseSuite.scala:132)
    at scala.meta.tests.prettyprinters.SyntacticSuite.testBlock$1(SyntacticSuite.scala:1997)
    ...
```

RCMartins:
Besides using the new test function `runTestAssert` and waiting on issue #3138, I think I did all your sugestions. Let me know if there are more improvements to add in the meanwhile

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
RCMartins: The current tests should cover every case. The cases that are constant (true or false) there is only one case, for the remaining cases I created a positive and a negative test.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -296,6 +296,29 @@ object TreeSyntax {
       }
       !isOnlyChildOfOnlyChild(t)
     }
+    @annotation.tailrec
+    def guessNeedsLineSep(t: Tree): Boolean = t match {
+      case _: Decl.Val | _: Decl.Var | _: Decl.Def | _: Defn.Macro | _: Defn.Type | _: Term.This |
```
kitbellew, line 301: why do you include `Decl` types here? they can't be confused with definitions because they all require an `=` before the body, so a block after shouldn't be confused. (as mentioned below, that includes the procedure syntax because a declaration must have a type and procedure syntax doesn't.)

if I am mistaken about that, there are 5 `Decl` classes, you didn't include `Given` and `Type`. why? should we match simply on `_: Decl`, to simplify matters (just like with `Tree.WithBody`)?

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -296,6 +296,29 @@ object TreeSyntax {
...
+          _: Term.Name | _: Term.Select | _: Member.Apply | _: Term.Ascribe | _: Term.Tuple |
```
kitbellew, line 302: please add `Term.Ref`, and remove its children `Term.This`, `Term.Name`, `Term.Select`, while moving up `Term.ApplyUnary` from further down.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -296,6 +296,29 @@ object TreeSyntax {
...
+          _: Term.New | _: Term.Interpolate | _: Term.Xml | _: Lit =>
+        true
+      case t: Term.Do => false
+      case t: Tree.WithBody => guessNeedsLineSep(t.body)
```
kitbellew, line 306: please remove `Defn.Macro` and `Defn.Type` above, they are already covered as they are both `Tree.WithBody`.

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
kitbellew: hmm, this _possibly_ means you found a bug in the parser. every tree that parser **exposes** has to have its positions within source code assigned (via an object called `origin`). it doesn't do it in all methods but in trees accessible via a public method the origin must be set.

let me check out your code and try.

kitbellew (changes requested):

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
```diff
@@ -1933,4 +1933,167 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
...
+    def testBlock(t: String, expectedStr: String, needSep: Boolean): Unit = {
+      def expected(sep: String): String =
+        s"""
+        |{
+        |${expectedStr.split("\n").map("  " + _).mkString("", "\n", sep)}
+        |  {
+        |    a
+        |  }
+        |}
+        """.trim.stripMargin.replace("\n", EOL)
```
kitbellew, line 1991: please put `.stripMargin` first, as `scalafmt` will not properly format this string otherwise (a bug in scalafmt, turns out).

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -296,6 +296,29 @@ object TreeSyntax {
...
+      case t: Stat.WithTemplate => t.templ.stats.isEmpty
```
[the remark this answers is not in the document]
RCMartins: I completely overlooked that the `self` field also changes the braces logic, I'll look into that!

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -296,6 +296,29 @@ object TreeSyntax {
...
+      case t: Term.ApplyUnary => guessNeedsLineSep(t.arg)
+      case Term.ApplyInfix(_, _, _, t :: Nil) => guessNeedsLineSep(t)
+      case t: Term.ApplyInfix => true
```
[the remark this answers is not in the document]
RCMartins: Ok, got it, so as a rule of thumb always avoid using the <Tree>.unapply?

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
RCMartins: The Decl.{Val, Var, Def} require the line separator, while Decl.Type does not. I don't know the reason but the Scala 2.13 parser fails on those. The tests that I created check that the line separator is actually necessary (because it either fails to parse without it or because it changes the tree structure).

The Decl.Given AFAIK is Scala3 only so it should never reach the `guessNeedsLineSep` function. In the call site we are checking for `dialect.allowSignificantIndentation`:
`if (dialect.allowSignificantIndentation) x.map(i(_)) else withBlankLines(x)`

Maybe we should add the Decl.Given (and other scala3 only Stats) anyways for future-proof? Is there an easy way to parse Scala3 without this `allowSignificantIndentation` flag on? If this makes sense then I'll have to brush up on my Scala 3 syntax, I don't use it on my day-to-day job :)

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
RCMartins: Ok, I "copied" this `.trim.stripMargin ...` from other tests in this file, should I also fix those?

kitbellew: perhaps for others we don't need to fix it. i will try to fix the formatter later.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
RCMartins: The `Defn.Macro` was indeed redundant now.
The `Defn.Type` has a `body: Type`, so I can only remove it from here if I replace it with a new pattern `_: Type` instead (that should only be called from a recursive call of `Defn.Type` bodies).

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
kitbellew: in that case, let's leave it as-is.

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
kitbellew: no bug :) this particular test suite intentionally resets origins, to force syntax reprinting. i moved all your tests into a new suite (and put all of them at class level, so that each case is a separate test, so that each failure is reported independently): #3143.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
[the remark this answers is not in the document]
kitbellew: 1. yes, avoid unapply on trees, they change.
2. in local testing it seems that i found another corner case with two blocks following each other, so the rule should be:
```
      case t: Term.ApplyInfix =>
        t.args match {
          case x :: Nil if !x.is[Term.Block] => guessNeedsLineSep(x)
          case _ => true
        }

```

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
kitbellew: if `Defn.Type` (or `Defn.Macro`) should always be excluded (regardless of its body), then please feel free to keep it here. it looks like your tests are quite comprehensive, that should become clear right away.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -296,6 +296,29 @@ object TreeSyntax {
       }
       !isOnlyChildOfOnlyChild(t)
     }
+    @annotation.tailrec
+    def guessNeedsLineSep(t: Tree): Boolean = t match {
+      case _: Decl.Val | _: Decl.Var | _: Decl.Def | _: Defn.Macro | _: Defn.Type | _: Term.This |
+          _: Term.Name | _: Term.Select | _: Member.Apply | _: Term.Ascribe | _: Term.Tuple |
+          _: Term.New | _: Term.Interpolate | _: Term.Xml | _: Lit =>
+        true
+      case t: Term.Do => false
```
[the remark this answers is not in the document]
kitbellew: ah, sorry. `Term.Do` has a body which comes first (and the expression comes in parens so no confusion there). please keep it as-is :)

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
RCMartins: > and put all of them at class level, so that each case is a separate test, so that each failure is reported independently
That is useful, the tests for this issue are quite extensive.

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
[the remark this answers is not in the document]
RCMartins: Can you provide the test case for this change?

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
```diff
@@ -1933,4 +1933,167 @@ class SyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
...
+    sep("Foo", "Foo")
+    sep("Foo.bar", "Foo.bar")
+    sep("Foo.bar", "Foo.bar")
+    sep("-10", "-10")
+    noSep("-{10}", "-{\n  10\n}")
+    sep("foo(bar)", "foo(bar)")
+    sep("foo[Bar]", "foo[Bar]")
+    // sep("foo + ()", "foo + ()") // TODO bug #3138 ?
+    sep("foo + bar", "foo + bar")
+    noSep("foo + {bar}", "foo + {\n  bar\n}")
```
kitbellew, line 2052: about the `ApplyInfix` change checking for `Block`: I think it was this one. at least, one of the tests failed for me. not sure how my `testBlock` is different from yours (they look similar) :)

kitbellew:
@RCMartins #3143 is merged.

RCMartins:
@kitbellew Thanks, I'll use the new test class, and update the PR.
In the meanwhile I think I found another bug :) #3142

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
RCMartins: Now after the rebase it's failing for me too, it's probably a regression from #3141, do you want me to create a new issue for this?
The case is something like this:
```scala
  test("#3138: infix with block with another following block") {
    val codeOnNextLine =
      """|{
         |  Foo bar { a }
         |  { b }
         |}
         |""".stripMargin
    val syntaxOnNextLine =
      """|{
         |  Foo bar {
         |    a
         |  }
         |  {
         |    b
         |  }
         |}
         |""".stripMargin
    val treeOnNextLine = Term.Block(
      Term.ApplyInfix(
        Term.Name("Foo"),
        Term.Name("bar"),
        Nil,
        List(Term.Block(List(Term.Name("a"))))
      ) :: Term.Block(List(Term.Name("b"))) :: Nil
    )
    runTestAssert[Stat](codeOnNextLine, Some(syntaxOnNextLine))(treeOnNextLine)
  }
```

Right now in the main branch is returning:
```
{
  Foo bar ({
    a
  }) {
    b
  }
}
```

RCMartins:
I think the only thing remaining is the regression from https://github.com/scalameta/scalameta/pull/3141

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/SyntacticSuite.scala
kitbellew: > Now after the rebase it's failing for me too, it's probably a regression from #3141, do you want me to create a new issue for this? In v4.7.7 it was parsing this correctly.

Why is it a regression? Why can you do
```
foo bar (1)
{ 2 }
```
but not
```
foo bar { 1 }
{ 2 }
```
or
```
foo bar { (1, 2) }
{ 3 }
```

in fact, looks like both
```
foo bar (1)
{ 2 }
foo bar { 1 }
{ 2 }
```
should be allowed in scala2 but not scala3.

see https://dotty.epfl.ch/docs/internals/syntax.html
and https://www.scala-lang.org/files/archive/spec/2.11/13-syntax-summary.html

RCMartins: I think I know the anwser:
If you go back to the initial issue I opened: #3138, you can see that:
```scala
foo bar (1)
{ 2 }
```
is **not** the same as `foo bar (1, 2)` but rather `foo bar ((1){2})` (just one param: `1.apply(2)`)

---

This expression:
```scala
foo bar { 1 }
{ 2 }
```
is actually a block with two terms, `foo bar { 1 }` and `{ 2 }`, no ambiguity.
And you can't place them in the same line (without a `;`)

---

If you try something like `{ (1, 2) } { 3 }` is actually invalid syntax (you would have to place a `;` between), but
```scala
{ (1, 2) }
{ 3 }
```
Are just two blocks in a row (probably inside another block)

---

I think this is just an issue with the scalameta parser, what is your opinion?

kitbellew: @RCMartins thank you for that but you are not explaining _WHY_ you think that way.

I ran a little experiment here:
https://scastie.scala-lang.org/HkPoBGmNSwGxSz81IeNBOw

```
def foo1(a: Int)(b: Int) = a + b

def foo2(a: Int) = a + 1

// no errors
0 + foo1 { 1 }
{ 2 }

// error: missing argument list for method foo1
0 + foo1 { 1 }

{ 2 }

// error: Int doesn't take parameters
0 + foo2 { 1 }
{ 2 }
```

which means that `{ 2 }` on a separate line, after `{ 1 }`, was not actual treated as a separate expression.

RCMartins: # Part 0 - my useless previous "example"

> If you try something like `{ (1, 2) } { 3 }` is actually invalid syntax (you would have to place a `;` between), but
> ```scala
> { (1, 2) }
> { 3 }
> ```
> Are just two blocks in a row (probably inside another block)

I can see now that this previous example didn't really matter and just added more confusion, so just completely ignore it please :)

# Part 1 - It's just Term.Apply

In my original [issue](https://github.com/scalameta/scalameta/issues/3138) the incorrect expression is a `Term.ApplyInfix`, and I think the parsing problem is only with that particular Term.

I think your experiment tested a different tree that has no issue now or before:
```scala
def foo1(a: Int)(b: Int) = a + b
```

In your example:
```scala
0 + foo1 { 1 }
{ 2 }
```
Is the same as
```scala
0 + ( ( foo1 { 1 } ) { 2 } )
```

So, the "potential problemetic code" is just two `Term.Apply`:
`Term.Apply(Term.Apply(Term.Name("foo1"), List(Term.Block(List(Lit.Int(1))))), List(Term.Block(List(Lit.Int(2)))))`

And with that I have no issue with.

# Part 2 - My example and Infix parsing logic

My example:
```scala
foo bar { 1 }
{ 2 }
```
Seems similar in quick glance to to your example but they are not:
```scala
0 + foo1 { 1 }
{ 2 }
```

With an infix op the right-hand side will only accept one of two things:
1. One Term, that can have braces around (in that case the term is a Term.Block, and any code after `}` should not be part of the ApplyInfix term)
2. Multiple Terms, with a `()` around, and `,` between

In my example, `bar` is the infix op, it will only accept one of two things, in my case its case 1. `{ 1 }` is the term for the right-hand side of the operation. After the right-hand side there is a new line a new `Stat` begins: `{ 2 }`.

In your example, `+` is the infix op, it will only accept one of two things, in your case its case 1. `foo1 { 1 }\n{ 2 }` is the term for right-hand side of the operation. After the right-hand side there's no more code.
*WHY* it parses that way? it first parses `Term.Name("foo1")` then it checks that it has an `{` and creates an `Term.Apply` and then it checks that it has another `{` and creates another `Term.Apply`.

So there is no issue in your example above, because it's just a diferent tree.

# Part 3 - Generating a new example that causes the issue

I can create generate the issue again from your example if I add some braces (and an aditional block `{ 3 }` to trigger the issue):

```scala
0 + {
  foo1 { 1 }
  { 2 }
}
{ 3 }
```

This expression is completely fine in scala, and parses correctly in scalameta 4.7.7, but parses the wrong tree with the latest scalameta version. You can try this new expression in: https://scastie.scala-lang.org/yY7TuHrEQC6K23I5Hj2Btw

As you can see, it's just two Stats, the `+` operator picks up the term on the right side which is: `{ foo1 { 1 } { 2 } }` and finishes parsing that term, it should **NOT** continue into the next line (`{ 3 }`).

# Conclusion

For completeness sake I'll add this two examples in my tests:
```scala
testBlockAddNL("foo {bar}", "foo {\n  bar\n}")
testBlockAddNL("foo\n{bar}\n{baz}", "foo {\n  bar\n} {\n  baz\n}")
```

I hope this helps clarifing the issue, it's not trivial to explain, but I tried my best!
if you need any help generating more problematic examples, let me know, I have less time during the week, but I'll try :)

kitbellew: does #3148 address this?

RCMartins: @kitbellew yes, I tried merging with the branch locally and it fixes the issue! Thank you :)

kitbellew (approved):

PR #3153, opened 2023-06-01
ScalametaParser: allow indent after self-type

kitbellew:
That way, we can correctly process template with self-type and remove the `undoIndent` work-around. For #3045.


tgodzik (approved):
LGTM!

PR #3163, opened 2023-06-06
ScalametaParser: higher-kinded are not wildcard

### LINKED ISSUE #3162: scalafmt replaces `Resource[F,_]` with `Resource[F,?]` which does not compile

kitbellew:
Fixes #3162.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -809,7 +809,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
 
     def entrypointType(): Type = paramType()
 
-    def typeArgs(): Type.ArgClause = autoPos(inBrackets(types()).reduceWith(Type.ArgClause.apply))
+    def typeArgs(): Type.ArgClause =
+      TypeBracketsContext.within(autoPos(inBrackets(types()).reduceWith(Type.ArgClause.apply)))
 
     def infixTypeOrTuple(inMatchType: Boolean = false): Type = {
       if (token.is[LeftParen]) tupleInfixType(allowFunctionType = !inMatchType)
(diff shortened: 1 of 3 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -959,7 +960,10 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           Type.Macro(macroSplice())
         case _: Underscore if inMatchType =>
           next(); Type.PatWildcard()
-        case _: Underscore if dialect.allowUnderscoreAsTypePlaceholder =>
+        case _: Underscore
+            if dialect.allowUnderscoreAsTypePlaceholder ||
```
kitbellew, line 964: @tgodzik this parameter is currently added to Scala3Future dialect and not Scala33 or earlier. I wonder if it has in fact already been released, which is why the submitter sees this problem in his Scala33 codebase. I tried searching dotty release notes but couldn't find anything since 2021.

tgodzik: This never happened :/ It was supposed to happen in 3.2, but now the site is heavily outdated 

https://docs.scala-lang.org/scala3/reference/changed-features/wildcards.html

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/TypeSuite.scala
```diff
@@ -527,4 +527,94 @@ class TypeSuite extends BaseDottySuite {
     assertNoDiff(err.shortMessage, "can't have multiple parameter lists in function types")
   }
 
+  test("#3162 [scala30] higher-kinded is not wildcard 1") {
+    implicit val dialect: Dialect = dialects.Scala30
+    runTestAssert[Stat]("def foo[A <: C[_]] = bar.baz[_, F[_]]")(
```
tgodzik, line 532: Shouldn't we expect `bar.baz[?, F[_]]` as default reprinted?

tgodzik (approved):
LGTM!

PR #3167, opened 2023-06-08
ScalametaParser: fail if no fewer-braces body

### LINKED ISSUE #3164: Allow removeOptionalBraces for test cases with Scala 3.3.0 Version
With the new Scala 3.3.0 version, we can now replace
```
test ("………") {
}
```
with
```
test ("……….") :
```
The complier is happy.
However, Scala fmt sbt plugin is not happy with it. Does anyone know how to add the such rule for tests in `.scalafmt.conf` file? It’s not to format the test codes automatically to that style, it’s to allow such change by human.

kitbellew:
Fixes #3164. By making it explicitly invalid...


tgodzik (approved):
LGTM!

PR #3169, opened 2023-06-10
ScalametaParser: change like-lambda param parsing

kitbellew:
- change convertToParam, to make sure that type is only parsed in one place, and that not just `using` is allowed as a possible Mod
- also, return Option[] in convertToParam, to avoid checking that it looks like lambda before parsing it as lamda parameter
For #2774.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1638,7 +1638,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     }
 
     // Note the absense of `else if` here!!
-    if (token.isAny[RightArrow, ContextArrow]) {
+    val contextFunction = token.is[ContextArrow]
+    if (contextFunction || token.is[RightArrow]) {
       // This is a tricky one. In order to parse lambdas, we need to recognize token sequences
       // like `(...) => ...`, `id | _ => ...` and `implicit id | _ => ...`.
       //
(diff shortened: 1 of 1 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1732,41 +1700,66 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
     }
   }
 
-  private def convertToParamClause(tree: Term): Term.ParamClause = (tree match {
-    case Term.Tuple(ts) => ts.flatMap(convertToParam)
-    case _ => convertToParam(tree).toList
-  }).reduceWith(toParamClause(None))
+  private def convertToParamClause(tree: Term)(
+      okNameLike: => Boolean,
+      okParamLike: => Boolean
```
tgodzik, line 1737:
```suggestion
      isNameLike: => Boolean,
      isParamLike: => Boolean
```
?

tgodzik (approved):
Super minor comment, otherwise LGTM!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: i think the meaning here is not "is this a name" but "is a name allowed here". should i rename to `isNameAllowed"?

tgodzik: I think that would be clearer, thanks!

kitbellew: done

tgodzik: Thanks!

PR #3174, opened 2023-06-12
ScalametaParser: keep provided name for Anonymous

kitbellew:


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1139,7 +1139,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   def termName(): Term.Name = name(Term.Name(_))
   def typeName(): Type.Name = name(Type.Name(_))
   @inline private def anonNameEmpty(): Name.Anonymous = autoPos(Name.Anonymous())
-  @inline private def anonName(): Name.Anonymous = atCurPosNext(Name.Anonymous())
+  @inline private def anonName(): Name.Anonymous = atCurPosNext(Name.Anonymous(token.name))
```
tgodzik, line 1142: Shouldn't we have a separate name like `Name.This` and `Name.Underscore` instead of using token.name? Since those are actual keywords?

kitbellew: interesting idea. also, should we use super as well?

but underscore in self type seems different, no? in fact, underscore _is the anonymous name_, while the current usage mostly applies when there's no name 🙂

i initially wanted to use Name.Indeterminate for them but it gets escaped with backquotes.

tgodzik: `Name.super` probably also? Since it is used similar to normal identifiers. You're right about `_` that is an actual anonymous name,

kitbellew: added Name.This, reused Name.Placeholder since it was used for a very similar purpose.

tgodzik: Looks like this broke some tests :scream:

kitbellew: fix tests, or find another way?

tgodzik: I would try to fix it if possible, these new names should be actually treated the same as anonymous was

kitbellew:
@tgodzik fixed.

tgodzik (changes requested):

tests/jvm/src/test/resources/example/Flags.scala
```diff
@@ -8,7 +8,7 @@ package object p/*<=flags.p.package.*/ {
   def z/*<=flags.p.package.z().*/(pp/*<=flags.p.package.z().(pp)*/: Int/*=>scala.Int#*/) = 3
   def m/*<=flags.p.package.m().*/[TT/*<=flags.p.package.m().[TT]*/]: Int/*=>scala.Int#*/ = macro ???/*=>scala.Predef.`???`().*/
   abstract class C/*<=flags.p.package.C#*/[+T/*<=flags.p.package.C#[T]*/, -U/*<=flags.p.package.C#[U]*/, V/*<=flags.p.package.C#[V]*/](x/*<=flags.p.package.C#x.*/: T/*=>flags.p.package.C#[T]*/, y/*<=flags.p.package.C#y.*/: U/*=>flags.p.package.C#[U]*/, z/*<=flags.p.package.C#z.*/: V/*=>flags.p.package.C#[V]*/) {
-    def this/*<=flags.p.package.C#`<init>`(+1).*/() = this(???/*=>scala.Predef.`???`().*/, ???/*=>scala.Predef.`???`().*/, ???/*=>scala.Predef.`???`().*/)
```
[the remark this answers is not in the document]
kitbellew: fixed

tgodzik (approved):
LGTM!

PR #3184, opened 2023-06-16
ScalametaParser: use TemplateOwner in parsing

kitbellew:
Add attribute methods to TemplateOwner.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2141,7 +2141,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           canApply = false
           Success(autoPos {
             next()
-            template(enumCaseAllowed = false, secondaryConstructorAllowed = false) match {
+            template(OwnedByTrait) match {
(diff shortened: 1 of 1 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2141,7 +2141,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
           canApply = false
           Success(autoPos {
             next()
-            template(enumCaseAllowed = false, secondaryConstructorAllowed = false) match {
+            template(OwnedByTrait) match {
```
tgodzik, line 2144: Why use trait here? Can it not be a class also?

tgodzik (approved):
Looks good, just one question.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Actually nevermind, makes sense!

kitbellew: i think because class allowed secondary constructor 🙂

PR #3192, opened 2023-06-21
ScalametaParser: allow `erased` keyword in lambda

kitbellew:
For #2774.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1718,16 +1718,50 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   private def convertToParam(tree: Tree): Option[Term.Param] = {
     def getModFromName(name: Name): Option[Mod] = name.value match {
       case soft.KwUsing() => Some(copyPos(name)(Mod.Using()))
+      case soft.KwErased() => Some(copyPos(name)(Mod.Erased()))
       case _ => None
     }
+    @tailrec
     def getMod(t: Tree, mods: List[Mod] = Nil): Option[List[Mod]] = t match {
       case t: Term.Name => getModFromName(t).map(_ :: mods)
+      case t: Term.Select =>
+        getModFromName(t.name) match {
+          case Some(mod) => getMod(t.qual, mod :: mods)
+          case _ => None
+        }
+      case t: Term.ApplyInfix =>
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1718,16 +1718,50 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
...
     }
+    @tailrec
     def getMod(t: Tree, mods: List[Mod] = Nil): Option[List[Mod]] = t match {
       case t: Term.Name => getModFromName(t).map(_ :: mods)
+      case t: Term.Select =>
+        getModFromName(t.name) match {
+          case Some(mod) => getMod(t.qual, mod :: mods)
+          case _ => None
+        }
+      case t: Term.ApplyInfix =>
```
tgodzik, line 1732: I wonder if we could detect the mods early instead of converting apply/select? I don't see it though currently :/

The only option I see is doing `t.tokens` and using those tokens to create mods. But that would reparse that part of tree :/

kitbellew: perhaps i don't understand what you mean :)

this `convertToParamClause` -> `convertToParam` logic is applied on a previous parsed expression tree _if and only if_ the parser finds a function arrow token after that.

so by then there's no "early", the parsing is already done, and all you need to do is convert.

if i misunderstood, could you please explain?

tgodzik: By early I meant that I would hope no converting would be needed at all. But I don't see that it would be possible.What do you think about doing `tree.tokens` and reparsing that? Wouldn't it allow us to make the code shorter? Just the same as we would have with other erased places.

Alternatively, we could retry parsing with a forked parser, but this time create a param.

Essentially, I feel like the parsed tree that we get here is not useful for the new param term, so converting it is adding unnecessary work.

kitbellew: would you mind it if we worked on that separately? the conversion logic has been here for a while, and i don't mind revisiting it but perhaps not in the context of this particular feature.

tgodzik: Sure!

tgodzik (approved):

PR #3196, opened 2023-06-23
ScannerTokens: infer "fewer braces" indent

kitbellew:
For #3045.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -150,16 +150,6 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
   @inline private def isColonIndent(): Boolean = token.is[Colon] && isIndentAfter()
   @inline private def isIndentAfter(): Boolean = peekToken.is[Indentation.Indent]
 
-  @tailrec
-  private def isEolAfter(index: Int): Boolean = {
-    val nextIndex = getNextSafeIndex(index)
-    tokens(nextIndex) match {
-      case _: AtEOLorF | MultilineComment() => true
-      case _: Trivia => isEolAfter(nextIndex)
-      case _ => false
-    }
-  }
-
   /* ------------- PARSER-SPECIFIC TOKENS -------------------------------------------- */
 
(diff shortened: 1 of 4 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1609,6 +1599,12 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
     }))(location)
   }
 
+  private def isEolAfterColonFewerBracesBody(): Boolean = peekToken match {
```
tgodzik, line 1602: We don't check for colon at all here, so this is really just `isNextIndentation`, no?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScannerTokens.scala
```diff
@@ -762,16 +776,37 @@ final class ScannerTokens(val tokens: Tokens)(implicit dialect: Dialect) {
...
+                 *   - after variable and before its type:
+                 *     - within `def`, `val`, `var` declaration or definition: excluded with
+                 *       RegionDefMark
+                 *     - cast within an argument expression: doesn't allow newline as it can then be
+                 *       confused with a fewer-braces invocation
+                 *     - within self-type: not allowed by the compiler
+                 *   - after fewer-braces method call: will apply if not handled otherwise
+                 */
+                def couldBeFewerBraces(): Boolean =
+                  dialect.allowFewerBraces && !sepRegions.contains(RegionDefMark) &&
```
tgodzik, line 791: `!sepRegions.headOption.contains(RegionDefMark) `

Shouldn't it be at the head?

tgodzik (approved):
Looks good, just some minor comments.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScannerTokens.scala
kitbellew: not really. here's an example:

```
def foo[A : Xxx](bar: Int): Int = 0
```

the first colon and second colon will have a RegionParen/RegionBracket first, before RegionDefMark, but they cannot start fewer braces.

PR #3197, opened 2023-06-23
ScalametaParser: allow `erased` keyword in types

### LINKED ISSUE #2774: Support erased definitions
This template is mostly a guideline, not a strict requirement, except
the sections marked `(required)`.

Please make sure that the issue is observable using the latest version of `scalafmt` and
within the `scalafmt` **CLI**: `https://scalameta.org/scalafmt/docs/installation.html#cli`

Using the CLI is _mandatory_, in one of these configurations:
- latest version of CLI and latest version in `.scalafmt.conf`
- previous version of CLI and latest version in `.scalafmt.conf`
- latest version of CLI and previous version in `.scalafmt.conf`

```
version = 3.5.8
binPack.parentConstructors = true
maxColumn = 120
includeCurlyBraceInSelectChains = true
align.preset = most
trailingCommas = preserve
runner.dialect = scala3
project.excludeFilters = ["kaivessel/package.scala"]

rewrite.scala3.insertEndMarkerMinLines = 5
...
```

NB: before submitting, please confirm that the problem is observed in the
*latest published* version of the formatter! We don't publish hotfixes for older
versions, and the problem you have observed in an older version may have already
been fixed.

When I run scalafmt via CLI like this: `<command-line parameter>`

If you were not using CLI initially:
- please check using `scalafmt` CLI before submitting here (required)
- if the problem does not present via CLI, DO NOT SUBMIT
  here: this is NOT a formatter issue
  - it could be an issue with your integration (sbt-scalafmt, metals, intellij etc.)
  - instead please submit the issue in the appropriate tool's repository
  - for instance, if the tool is `sbt` with the `sbt-scalafmt` plugin: please
    submit in `sbt-scalafmt` and include scalafmt parameters from `build.sbt`
    and the `sbt` command used
  - conversely, do not submit formatting issues into other tools; you must first
    verify using `scalafmt` CLI

## Steps

Given code like this:
```scala
package kuraga.datum

import scala.language.experimental.erasedDefinitions

import Repr.Holder

object Repr:
  trait Holder:
    type A

trait Repr[-P[-_, +_], +X]:
  def apply(erased holder: Holder)(p: P[X, holder.A]): holder.A
```

## Problem

Scalafmt formats raises an error like this:
```scala
Repr.scala:12: error: [dialect scala3] : expected but identifier found
  def apply(erased holder: Holder)(p: P[X, holder.A]): holder.A
```

## Workaround

I've found that by...

## Notes

See also...

kitbellew:
Fixes #2774.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -654,7 +654,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       var hasTypes = false
 
       @tailrec
-      def paramOrType(): Type = token match {
+      def paramOrType(modsBuf: mutable.Builder[Mod, List[Mod]]): Type = token match {
         case t: Ellipsis =>
           ellipsis[Type](t)
         case t: Unquote =>
@@ -662,7 +662,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
         case KwImplicit() if !hasImplicits =>
           next()
           hasImplicits = true
-          paramOrType()
+          paramOrType(modsBuf)
         case t: Ident if tryAhead[Colon] =>
           if (hasTypes)
             syntaxError(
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -654,7 +654,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
       var hasTypes = false
 
       @tailrec
-      def paramOrType(): Type = token match {
+      def paramOrType()(implicit modsBuf: mutable.Builder[Mod, List[Mod]]): Type = token match {
```
tgodzik, line 657: Is making it implicit necessary?

tgodzik (approved):
One minor comment, otherwise LGTM!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: of course, not 😀 old habits, reducing size of diff.

kitbellew: done

PR #3205, opened 2023-06-28
ScalametaParser: don't fail on `override` in class

### LINKED ISSUE #3201: Bug: `override' modifier not allowed for classes
```scala
trait A:
   type E <: Throwable
   
object B extends A:
   override class E(m: String) extends Exception(m)

@main def entryPoint = try
   throw B.E("😱")
   catch
      case e: Exception => println(e.getMessage)

```

The above code compiles and runs fine but it triggers an error in Metals.

(Try switching "worksheet" mode on in [Scastie](https://scastie.scala-lang.org/oUeY5oDHT0yOoehJvniSig).)

Looks like a bug to me as according to the spec "TemplateStat[ements]" may have "Modifier"s including `override`:

https://docs.scala-lang.org/scala3/reference/syntax.html

kitbellew:
Fixes #3201.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/Messages.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/Messages.scala
@@ -29,8 +29,6 @@ object Messages {
   val InvalidImplicit =
     "`implicit' modifier can be used only for values, variables, methods and classes"
   val InvalidImplicitTrait = "traits cannot be implicit"
-  val InvalidImplicitClass = "classes cannot be implicit"
   val InvalidAbstract = "`abstract' modifier can be used only for classes"
-  val InvalidOverrideClass = "`override' modifier not allowed for classes"
   val InvalidLazyClasses = "classes cannot be lazy"
 }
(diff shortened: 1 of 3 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/Messages.scala
```diff
@@ -29,8 +29,6 @@ object Messages {
   val InvalidImplicit =
     "`implicit' modifier can be used only for values, variables, methods and classes"
   val InvalidImplicitTrait = "traits cannot be implicit"
-  val InvalidImplicitClass = "classes cannot be implicit"
   val InvalidAbstract = "`abstract' modifier can be used only for classes"
-  val InvalidOverrideClass = "`override' modifier not allowed for classes"
```
tgodzik, line 34: Looks like this is not allowed for Scala 2.13.x, but maybe it's not really useful to add a new dialect option.

tgodzik (approved):
LGTM

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/Messages.scala
kitbellew: we could check for significant indentation flag, or, like we frequently do, just err on the side of more permissive parsing, assuming the compiler is ok.

PR #3243, opened 2023-07-09
sbt: Update sbt to 1.9.1

eed3si9n:


kitbellew:
@eed3si9n looks like it doesn't work:

https://github.com/scalameta/scalameta/actions/runs/5501810663/jobs/10025654763

it runs `sbt publish` (or some version of it), and there's no `!`... what to do?

PR #3247, opened 2023-07-11
ScannerTokens: handle `case` after `= <indent>`

kitbellew:


kitbellew:
MIMA failure should be fixed in #3248.

kitbellew:
> MIMA failure should be fixed in #3248.

@tgodzik not sure if you have ideas. it seems that `sbt mima` runs on jars which are before shading. if that's true, how would we change that?

kitbellew:
@tgodzik with this change, here's where case clause is recognized: 
- explicitly: after catch/match and indent/brace
- implicitly (as partial function):
  - after colon/equals and indent
  - inside left brace region except template and for

are there any other cases?

tgodzik (approved):
I can't think of another case, I guess it will show up with a bug report if there is anything.

As for shadowing vs mima, I am not sure how exactly to do it, but we would probably have to point `mimaCurrentClassfiles` at the locally published jar. We could probably run `publishLocal` first and then we would probably be able to point to that jar that was published

PR #3248, opened 2023-07-11
SBT: remove geny dependency in testkit

kitbellew:
Apparently, geny.Generator is the result type of a public method within this library, and since it's used only for testing, and only in `scalafmt` benchmarks, changing that result type should be ok.


kitbellew:
hmm... doesn't do what we want

tgodzik (approved):

PR #3250, opened 2023-07-13
SBT: build only for the last 4 versions of scala

kitbellew:
Let's include, for each minor version, up to 4 patch versions.


tgodzik (approved):
LGTM

PR #3251, opened 2023-07-13
SBT: publish JVM/JS/Native separately

kitbellew:


build.sbt
```diff
@@ -865,3 +879,17 @@ lazy val shadingSettings = Def.settings(
   ).map(ShadingRule.moveUnder(_, "scala.meta.shaded.internal")),
   validNamespaces ++= Set("scala.meta", "org.scalameta", "scala", "java")
 )
+
+def platformPublishSettings(platform: Platforms.Platform) =
+  if (Platforms.buildPlatform(platform))
+    Def.settings()
+  else
+    Def.settings(
```
kitbellew, line 887: I don't know how much of it is needed, just copied, with some mods, from https://gist.github.com/joescii/4431cb2185eddb699dfb79b9c910e333

olafurpg (approved):
LGTM 👍🏻 The best way to test this is by running an actual release. If this works, it's possible to expand it even further by splitting the matrix by Scala version (2.11.x, 2.12.x, 2.13.x)

kitbellew:
> LGTM 👍🏻 The best way to test this is by running an actual release. If this works, it's possible to expand it even further by splitting the matrix by Scala version (2.11.x, 2.12.x, 2.13.x)

i think a release will be triggered after each commit, so we don't need to tag a public release.

as to the scala version, do you mean running the literal `++2.13.x +publish...`, or is there a different way to constrain the version?

olafurpg:
You can `skip in publish` based on the env variable, which allows you to do `+publish` but it will only work for the Scala version that's defined by the env variable.

PR #3253, opened 2023-07-14
bugfix: Fix nested types with underscore

tgodzik:
Fixes https://github.com/scalameta/metals/issues/5437

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -963,7 +963,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
         case _: Underscore if inMatchType =>
           next(); Type.PatWildcard()
         case _: Underscore
-            if dialect.allowUnderscoreAsTypePlaceholder ||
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -963,7 +963,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) { parser =>
         case _: Underscore if inMatchType =>
           next(); Type.PatWildcard()
         case _: Underscore
-            if dialect.allowUnderscoreAsTypePlaceholder ||
+            if dialect.allowUnderscoreAsTypePlaceholder &&
```
tgodzik, line 966: I don't think we can use underscore in type lambdas if we don't allow it as type palceholder, no?

kitbellew: this was my attempt to solve #3162, perhaps with the assumption that the flags we set for our dialect do not always reflect objective reality of compiler parameters. it's possible that the fix should have been more about the context of what we're parsing...

tgodzik: Ok seems it's not as quick as thoguht to fix, will get back to it next week

PR #3262, opened 2023-07-20
ScalametaParser: don't use exprMaybeIndented()

### LINKED ISSUE #3261: can't parse poorly-indented chained match expressions without braces
Scala 3.3.1-RC1

`.scalafmt.conf`
```
version = 3.7.10
runner.dialect = Scala3
indent.significant = 2
```

```scala
object small:
  val value =
    Nil match
    case Nil => "empty"
    case _   => "nonempty"
  match
    case "empty"    => 0
    case "nonempty" => 1

```

The code is compilable but scala fmt fails with error:
`Scalafmt parse error (small.scala:95): [dialect scala3] ; expected but match found`

kitbellew:
An expression which starts with an indent doesn't have to be indented but instead could start with an indented part and continue past it. Fixes #3261.


tgodzik (approved):
LGTM!

PR #3263, opened 2023-07-21
chore: Remove files for no longer used versions

tgodzik:


tgodzik:
No need to compile or test it if we don't release for those versions.

agboom:
Hi @tgodzik, sorry for bumping an old PR, but could you maybe give some context as to why these old Scala versions are removed?

Recently some of our builds started breaking, indirectly because of the removal of these old versions. We use [scalac-scoverage-plugin](https://github.com/scoverage/scalac-scoverage-plugin) for measuring coverage, which depends on Scalameta. A few days ago this plugin removed these versions [in this PR](https://github.com/scoverage/scalac-scoverage-plugin/pull/604), which was causing to break our builds on Scala versions older than 2.13.10.
The reason this happened is because we use an Ivy revision range (i.e. `"[2.0.6,3)"`) in the scoverage plugin to make sure to support a broad enough range of Scala versions (since every patch needs a new plugin release, this reduced the maintenance burden). Keeping the old versions supported would mitigate this, but of course there may be a good reason to remove old versions, hence my question.

kitbellew:
scoverage does the same, perhaps because of us: https://github.com/scoverage/scalac-scoverage-plugin/blob/main/build.sbt#L16

agboom:
Thanks for explaining @kitbellew, that makes sense. I should've looked on discord first :see_no_evil: 
You're right, scoverage explained they follow suite: https://github.com/scoverage/scalac-scoverage-plugin/pull/604#issuecomment-1961244487

We do try to keep our Scala versions up-to-date, but we have a lot of projects with different Scala patch versions, and a single sbt plugin that needs to support all those Scala versions. We either need to step up our Scala upgrade process or find a way to match the right scoverage plugin version to the right Scala version. But that's kind of out of scope of this discussion

tgodzik:
We generate a list of supported versions here https://github.com/scalameta/metals/blob/main/project/SemanticDbSupport.scala#L3

So we can match semanticdb version to Scala version. Might be useful for you.

PR #3270, opened 2023-07-23
NestedContext: move all contexts to parser

### LINKED ISSUE #3063: Wrong `pattern must be a value`
Found an issue when working on https://github.com/scalameta/metals/issues/4896
```
${ success.tree.asExprOf[InterpolationConverterType] }
```

looks like an issue with Scala 3 dialect

kitbellew:
The current implementation was mistakenly sharing context with any other instantiated parsers, which would cause problems in a multi-threaded use such as within scalafmt. Fixes #3063.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/NestedContext.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/NestedContext.scala
@@ -14,13 +14,3 @@ private[parsers] trait NestedContext {
   @inline def isInside() = isDeeper(0)
   @inline def isDeeper(level: Int) = nested > level
 }
-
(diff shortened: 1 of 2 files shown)


tgodzik (approved):

tgodzik:
Interesting, do you think this was indeed the cause of #3063?

kitbellew:
> Interesting, do you think this was indeed the cause of #3063?

i checked out one of repos mentioned in that ticket and ran the formatter several dozen times. not sbt, which caches results and doesn't always rerun, but the cli.

sometimes no error, sometimes multiple errors, and always in random files, and almost always in macros. since the error now includes the code and position, the error would never really make sense.

so it seemed a thread-safety error. these contexts are definitely a bug. after applying this change, i reran the format on the same repo many times again, and not a single error.

so, it's possible that this is the fix 😀

PR #3272, opened 2023-07-24
semanticdb: use canonical path for sourceroot

### LINKED ISSUE #2518: sourceroot parameter doesn't work with symlinks or relative paths
If a compiling file's path doesn't "startWith" the `sourceroot` parameter, then it's skipped ([here](https://github.com/scalameta/scalameta/blob/main/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SemanticdbPipeline.scala#L59), and [here](https://github.com/scalameta/scalameta/blob/267c5966bc343ee36683ca36bd19cf6e9b825381/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/InputOps.scala#L33)).

This prevents relative paths and symlinks from being used. The `sourceroot` should be converted to a real path before usage.

kitbellew:
Fixes #2518.

### DIFF
--- a/scalameta/io/js/src/main/scala/java/nio/file/Path.scala
+++ b/scalameta/io/js/src/main/scala/java/nio/file/Path.scala
@@ -25,4 +25,5 @@ trait Path extends Iterable[Path] {
   def toUri: URI
   def toAbsolutePath: Path
   def toFile: java.io.File
+  def toRealPath(): Path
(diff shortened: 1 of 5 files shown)


tgodzik (approved):
LGTM!

PR #3276, opened 2023-07-24
TreeSyntax: separate block in scala3 stats

kitbellew:
Follow-on to #3139 and #3269.


scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -302,7 +302,7 @@ object TreeSyntax {
       case _: Decl.Val | _: Decl.Var | _: Decl.Def | _: Defn.Type | _: Term.Ref | _: Member.Apply |
           _: Term.Ascribe | _: Term.Tuple | _: Term.New | _: Term.Interpolate | _: Term.Xml |
           _: Lit =>
-        true
+        !dialect.allowSignificantIndentation
```
kitbellew, line 305: currently, the only case that applies for scala3 is 3 lines down (`Stat.WithTemplate`).

another option is to leave this line (305) as `true` and separate blocks for scala2 and scala3 exactly the same. a few more blank lines but perhaps it makes it a bit more readable :)

tgodzik (approved):
Looks fine for me, we can add the line separators , that's not an issue. Thanks for looking into this!

Issue #3259, opened 2023-07-18
Shaded classes unavailable in trees_sjs jar versions >= 4.8.3


robby-phd:
Shaded classes such as `scala.meta.shaded.internal.fastparse.ParserInputSource` could not be found in trees_sjs jar in versions >= 4.8.3 (but available in non-sjs/jvm trees jar):

```console
> javap -cp trees_2.13-4.8.3.jar scala.meta.shaded.internal.fastparse.ParserInputSource     
Compiled from "ParserInput.scala"
public interface scala.meta.shaded.internal.fastparse.ParserInputSource {
  public static <T> scala.meta.shaded.internal.fastparse.ParserInputSource$fromParserInput<T> fromParserInput(T, scala.Function1<T, scala.meta.shaded.internal.fastparse.ParserInput>);
  public static scala.meta.shaded.internal.fastparse.ParserInputSourceLowPri$FromReadable$ FromReadable();
  public static <T> scala.meta.shaded.internal.fastparse.ParserInputSourceLowPri$FromReadable fromReadable(T, scala.Function1<T, scala.meta.shaded.internal.geny.Readable>);
  public abstract <T> T parseThrough(scala.Function1<scala.meta.shaded.internal.fastparse.ParserInput, T>);
}
> javap -cp trees_sjs1_2.13-4.8.3.jar scala.meta.shaded.internal.fastparse.ParserInputSource
Error: class not found: scala.meta.shaded.internal.fastparse.ParserInputSource
```

This breaks some sjs compilations when upgrading to >= 4.8.3, e.g., https://jitpack.io/org/sireum/kekinian/8d2094d30f/build.log (to reproduce: https://github.com/sireum/kekinian/blob/8d2094d30ffdfe4ebe5c04646da11de647d9a17c/.jitpack.yml), log excerpt:

```log
/home/jitpack/build/slang/parser/shared/src/main/scala/org/sireum/lang/parser/SlangParser.scala:832: error: exception during macro expansion:
java.lang.ClassNotFoundException: scala.meta.shaded.internal.fastparse.ParserInputSource
	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:445)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:588)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
	at scala.meta.internal.tokenizers.ScalametaTokenizer.uncachedTokenize(ScalametaTokenizer.scala:22)
[...]

PR #3277, opened 2023-07-25
SBT: shade only for JVM

kitbellew:
Until sbt-shading plugin is released with the cross-platform fixes, let's limit shading to JVM only. Helps with #3259.


tgodzik (approved):
LGTM!

PR #3280, opened 2023-07-25
ConfigOps: sourceroot as relative to targetroot

### LINKED ISSUE #2515: Option for sourceroot to be relative to targetroot (or another parameter that is)
Current the `sourceroot` option is treated as being relative to the cwd, rather than the `targetroot`. Unfortunately this causes issues in Gradle.

Sorry it's a little complicated to explain, but reason is:

* Gradle doesn't use the project root as it's cwd. Instead it has N workers that run under their own directories, but point the `scalac` task to a particular scala file and output location (which is under the project).

* If you set `sourceroot` as a compiler option this breaks caching in Gradle, as any differences in compiler options (between machines) forces Gradle to recompile, rather than to download existing compiled artifacts. In our particular case, this is a signifiant overhead. We rely on Gradle's caching to make compiling our very large (almost mono) repo tractable.   
 
However, having an option (either `sourceroot`, or a new option) that you can set relative to `targetroot` solves this. As the `targetroot` is automatically derived from `scalac`'s output directory, which is already set to the right place.

kitbellew:
That is, one could specify `-P:semanticdb:sourceroot:targetroot:../xyz`. Fixes #2515.

### DIFF
--- a/docs/semanticdb/guide.md
+++ b/docs/semanticdb/guide.md
@@ -234,6 +234,10 @@ and the "Value" column describes the `<value>` component):
     <td>
       Used to relativize source file paths into
       <code>TextDocument.uri</code>.
+      <br/>
+      If the value starts with `targetroot:`, the remaining part is 
+      interpreted as relative to `targetroot` (see below; must be
+      specified first) instead of current working directory.
     </td>
     <td>Current working directory</td>
   </tr>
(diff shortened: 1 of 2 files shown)


tanishiking (approved):
LGTM from my side :+1:

PR #3304, opened 2023-07-31
SemanticDB: keep only 2.x code, remove 2.x.y

kitbellew:


tgodzik (changes requested):
We can remove the 2.12.x reporters, but in 2.13.12 we will get some changes that will again require this :thinking:

kitbellew:
@tgodzik are you saying that we should preserve the 2.13 patch versions?

tgodzik:
> @tgodzik are you saying that we should preserve the 2.13 patch versions?

yes, otherwise I will be adding them back very soon :sweat_smile:

kitbellew:
> > @tgodzik are you saying that we should preserve the 2.13 patch versions?
> 
> yes, otherwise I will be adding them back very soon 😅

are you referring to this code (as `VersionCompilerOps` is the only class that was specialized for 2.13.x):
```
package scala.meta.internal.semanticdb.scalac

import scala.tools.nsc.interactive.Global

trait VersionCompilerOps {

  def forceWarnings(global: Global) = {}

}
```

kitbellew:
for now, removed changes to 2.13, kept only 2.12 and 2.12.x.

tgodzik:
> > > @tgodzik are you saying that we should preserve the 2.13 patch versions?
> > 
> > 
> > yes, otherwise I will be adding them back very soon sweat_smile
> 
> are you referring to this code (as `VersionCompilerOps` is the only class that was specialized for 2.13.x):
> 
> ```
> package scala.meta.internal.semanticdb.scalac
> 
> import scala.tools.nsc.interactive.Global
> 
> trait VersionCompilerOps {
> 
>   def forceWarnings(global: Global) = {}
> 
> }
> ```

Hmm... that might be SemanticdbReporter, so these are probably safe to remove then. Sorry about the misinformation :sweat:

tgodzik (approved):
Looks good, you can also remove 2.13.x then. I will need to add SemanticdbReporter anyway

PR #3313, opened 2023-08-01
SemanticDB: allow silencing info/warning/error

kitbellew:
Capture them only in semanticdb and not report further.

The reason why this is useful is because scalac's `-Xreporter` option is not yet supported in scala3 -- and it's also ignored by sbt, which uses the `compilerReporter` setting instead (which isn't readily documented).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -322,6 +322,7 @@ lazy val semanticdbIntegration = project
         "-Xplugin-require:semanticdb",
         warnUnusedImports,
         "-Yrangepos",
+        "-P:semanticdb:reporter_severity:error", // don't report below error
         "-P:semanticdb:text:on", // include text to print occurrences in expect suite
         "-P:semanticdb:failures:error", // fail fast during development.
         "-P:semanticdb:exclude:Exclude.scala",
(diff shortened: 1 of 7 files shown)


tgodzik:
I am not sure about the exact use case here. Why would we want to silence messages coming from the compiler? Shouldn't we silence them differently via `-Wconf`? 

Or the case is that because we are using our own reporter users are getting messages which should not pop up?

kitbellew:
honestly, I'm not sure if there's a different way. i was simply a bit unnerved by all the warnings coming from the compiler when building the integration examples in this build.

and it turns out, that those warnings are needed by semanticdb itself (it intercepts them) so we can't silence them completely but if, other than semanticdb, we wouldn't have wanted to see them, it's not clear how to silence them _after_.

thoughts?

tgodzik:
> honestly, I'm not sure if there's a different way. i was simply a bit unnerved by all the warnings coming from the compiler when building the integration examples in this build.
> 
> and it turns out, that those warnings are needed by semanticdb itself (it intercepts them) so we can't silence them completely but if, other than semanticdb, we wouldn't have wanted to see them, it's not clear how to silence them _after_.
> 
> thoughts?

This would be useful if we decided to always add things like unused warnings and hide them from the user, but still be able to run scalafix rules etc. :thinking: But honestly I don't see a lot of scenarios beside the one in our tests that we wouldn't want the warning to show up for the user. We want them to know something can be fixed.

Maybe it's fine to just filter out those warnings in the IDE for your use case?

Issue #3314, opened 2023-08-13
Include generated source in source jars


Arthurm1:
In `common_2.13-4.8.3.jar` there are `platform`, `semanticdb` and `semanticidx` packages under the `scala/meta/internal`

These don't exist in the `common_2.13-4.8.3-sources.jar`

I assume this is because they are generated sources by protobuf?

Can they be included in the sources jar or do they get generated directly to `.class` files?

tgodzik:
Ok, I think I managed to release them in https://github.com/scalameta/scalameta/pull/3315 but it's in a protobuf dir (which shouldn't be an issue for metals since we index the actual package anyway)

PR #3315, opened 2023-08-16
improvement: Publish semanticdb generated sources

tgodzik:
Should solve https://github.com/scalameta/scalameta/issues/3314


kitbellew (approved):
do you understand why this is needed? the original ticket was about access to internal code.

Arthurm1:
@kitbellew I was debugging Metals and wanted to trace into the `TextDocument#parseFrom(Bytes)` to see why I was getting errors.  Having the source in the jar would have made it quicker to diagnose the problem.

PR #3316, opened 2023-08-17
bugfix: Don't shade for 2.11.12

tgodzik:
Looks like there are some issues with shading for Scala 2.11.12, so since this is a barely supported version I don't think we need to shade it. It's currently only used in Metals I think.

Related to https://github.com/scalameta/metals-vscode/issues/1420

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -865,10 +865,14 @@ lazy val docs = project
   .enablePlugins(BuildInfoPlugin, DocusaurusPlugin)
 
 lazy val shadingSettings = Def.settings(
-  shadedModules ++= Set(
-    "com.lihaoyi" %% "geny",
-    "com.lihaoyi" %% "fastparse"
-  ),
+  shadedModules ++= {
+    if (!isScala211.value)
+      Set(
+        "com.lihaoyi" %% "geny",
+        "com.lihaoyi" %% "fastparse"
+      )
+    else Set.empty
+  },
   shadingRules ++= Seq(
     "geny",
     "fastparse"


kitbellew:
@tgodzik i think this needs to be rebased on top of #3317.

also, a bit surprised why 2.11 doesn't work; do you think it's a problem with a library that sbt-shading is using? (and do you think we can ever get sbt-shading released?)

tgodzik:
> @tgodzik i think this needs to be rebased on top of #3317.
> 
> also, a bit surprised why 2.11 doesn't work; do you think it's a problem with a library that sbt-shading is using? 

Not sure, I assumed no one is using shading for 2.11 and it's not really worth investigating.

> (and do you think we can ever get sbt-shading released?)

There were some issues with secrets, I don't think Alex managed to fix them :/

kitbellew (changes requested):
1. i'd rebase on top of `main`, to make sure it continues working
2. as it stands right now, we enable the shading plugin and some rules but there are no modules to shade; instead, why not make the following change (line 886 in `main`), without modifying `shadingSettings`:

```
def crossPlatformShading(project: sbtcrossproject.CrossProject) =
  if (!isScala211.value && shadingSettings.nonEmpty)
    project.enablePlugins(ShadingPlugin).settings(shadingSettings)
  else project
```

tgodzik:
> i'd rebase on top of main, to make sure it continues working 

Rebased!

> as it stands right now, we enable the shading plugin and some rules but there are no modules to shade; instead, why not make the following change (line 886 in main), without modifying shadingSettings:

Tried something like that, but:
```
/home/tgodzik/Documents/scalameta/build.sbt:892: error: `value` can only be used within a task or setting macro, such as :=, +=, ++=, Def.task, or Def.setting.
  if (!isScala211.value && shadingSettings.nonEmpty)
                  ^
sbt.compiler.EvalException: Type error in expression
```
we need to be in a setting.

kitbellew (approved):
sbt magic, again...

PR #3320, opened 2023-08-23
bugfix: Don't allow * as type lambda by default

tgodzik:
This currently needs to be enabled with a compiler flag, so I think it makes sense not to have it on as default. In Metals we can properly detect the flag, while in scalafmt people will most likely need to set it themselves.

Related to https://github.com/scalameta/metals/issues/5573

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
@@ -142,7 +142,6 @@ package object dialects {
     .withAllowStarWildcardImport(true)
     .withAllowProcedureSyntax(false)
     .withAllowDoWhile(false)
-    .withAllowStarAsTypePlaceholder(true)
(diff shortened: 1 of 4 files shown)


kitbellew:
should we try to detect instead whether the `*` is used as a standalone type? `.type` seems a strong indication against.

i think we did something similar with `_` recently.

tgodzik:
> should we try to detect instead whether the `*` is used as a standalone type? `.type` seems a strong indication against.
> 
> i think we did something similar with `_` recently.

That's one option, but I fear that there will always be edge cases that we didn't account for.

kitbellew (approved):
@tgodzik if you prefer this approach, i am good with that. the only caveat is that a number of people won't know how to change the dialect in scalafmt (despite https://scalameta.org/scalafmt/docs/configuration.html#runnerdialectoverride) and will start sending bugs.

as an alternative, i proposed #3321. let me know what you think.

PR #3323, opened 2023-08-24
chore: Add support for Scala 2.13.12

tgodzik:
Closes https://github.com/scalameta/scalameta/issues/3329


kitbellew:
@tgodzik do you need to add resolvers, like https://github.com/scalameta/scalameta/pull/3318/files#diff-8fd74420ca55aa91d4c53775d2a55a999c405b5a4fafeaf73b7f0ef83e1f3b0c

tgodzik:
> @tgodzik do you need to add resolvers, like https://github.com/scalameta/scalameta/pull/3318/files#diff-8fd74420ca55aa91d4c53775d2a55a999c405b5a4fafeaf73b7f0ef83e1f3b0c

They are already added, we are just missing releases for native and js, which will need to come first. This is basically ready and when Scala is released we should be good to update the version to 2.13.12

kitbellew:
@tgodzik @bjaglin does this change (the fact that some older scala versions are now excluded) mean the next scalameta version is 4.8.11 or 4.9.0?

tgodzik:
Hmm... I don't think this fits into semantic versioning at all so either way is fine.

PR #3347, opened 2023-09-29
Migrate quasiquote macros to Scala 3

jchyb:
This draft PR is meant to serve as a signal of progress made in migrating scalameta quasiquote macros to Scala3. Apologies for taking up the PR tracker with another draft PR. I will be unavailable for a few weeks, but I intend to come back to this PR after that time and finish it. It is very much not ready for review.

The aim here was to migrate only the macro methods used for defining quasiquotes, leaving all of the codegen-by-macro-annotations stuff in scala 2 for now. So far, I managed to get some quasiquote apply methods working, although sometimes the resolved type is incorrect (requiring asInstanceOf, as seen in the tests). The unapply methods do need some more work. Of course, things like the build structure etc. are not final.


tgodzik:
I tried to come up with a way to publish just the parsers module, but that still has a bunch of errors when using fastparse and also there is a small macro that we would need to rewrite :thinking: 

https://github.com/scalameta/scalameta/compare/main...tgodzik:scalameta:cross-publish?expand=1

If successful it would be possible to just depend on `parsers_3` and get basically everything we need. I will try and see if I can fix things there.

jchyb:
Progress update:
All the apply methods should work now, and unapply methods for quasiquotes are also implemented. There are some issues, however:
* unapply methods generated with trees involving `Some` crash the compiler for now. I've submitted the issue into the compiler (and will probably look more closely into it later on, https://github.com/lampepfl/dotty/issues/19362). I have a workaround in mind, as previously when I had an implicit dialect (that I had to remove for unrelated reasons) in unapply macro method entry point, the crash didn't trigger.
* making transparent inline unapply methods work is a bit difficult for now, although I've made some progress there, also for some reason they get executed twice (I've still yet to minimize and submit this issue into the compiler)
* since unapply is not transparent for now, the compiler is not able to type the returned values correctly (we always obtain Any and have to cast it ourselves). I've adjusted the tests to reflect that (pretty much every test).
* I've never managed to obtain type ascriptions from unapply methods in Scala 3. There was a feature here, where we could return the correct type depending on the type ascription (e.g. we could either return Term.ArgParam or List[Term]). The feature is ported, but unused since obtaining those ascriptions is the missing part. I managed to make this work only for refutable extractors like `val q"def method(args: Term.ArgClause)" = q"def method(a: Int, b: Double)"`, but it will only work for non-transparent inline unapply (which we want to make transparent, for reasons stated in the bullet points above). It also does not work for unapplies in pattern matches at all. Nevertheless, I've left this experiment in the code, commented out (with the "EXPERIMENT" label). 
* The above also means that we are not able to throw any compilation errors if users use the wrong ascription in a quasiquote

tgodzik:
Thanks for the great work! I will have to go over it and ask some questions since it's something I haven't dealt with yet.

I almost finished with making scalameta release without quasiquotes for Scala 3, I need to tidy it up and figure out some smaller sbt issues.

> unapply methods generated with trees involving Some crash the compiler for now.

So only Boolean ones were working previously? There were no tests for that?

> since unapply is not transparent for now, the compiler is not able to type the returned values correctly (we always obtain Any and have to cast it ourselves). I've adjusted the tests to reflect that (pretty much every single test was changed because of that).

Looks like we should wait for 3.3.2 for that? It would be super useful to have the correct types here without a need for casts. I think we can wait for the next release since it should be soon.

> I've never managed to obtain type ascriptions from unapply methods in Scala 3. 

Which feature do you mean? Do we have it in a test suite somewhere?

quasiquotes3/shared/src/main/scala/scala/meta/internal/quasiquotes/ExprLifts.scala
```diff
@@ -0,0 +1,513 @@
...
+  def liftOptionTree[T: Type](maybeTree: Option[MetaTree]): internalQuotes.reflect.Tree
+  def liftTrees[T: Type](trees: Seq[MetaTree]): internalQuotes.reflect.Tree
+  def liftTreess(treess: List[List[MetaTree]]): internalQuotes.reflect.Tree
+  def liftQuasi0(quasi: Quasi, optional: Boolean = false): internalQuotes.reflect.Tree
+
+  protected def unquotesName(q: scala.meta.internal.trees.Quasi): Boolean
+}
+
+// TODO should be generated by a custom codegen script
+trait ExprLifts(using override val internalQuotes: Quotes)(isPatternMode: Boolean) extends ReificationMacros with InternalTrait {
```
tgodzik, line 27: Was this generated previously? Why do we need so much more in Scala 3?

quasiquotes3/shared/src/main/scala/scala/meta/quasiquotes/Api.scala
```diff
@@ -0,0 +1,196 @@
...
+    transparent inline def template(inline args: Any*): Template = ${ ReificationMacros.templateImpl('sc, 'args) }
+    transparent inline def mod(inline args: Any*): Mod = ${ ReificationMacros.modImpl('sc, 'args) }
+    transparent inline def enumerator(inline args: Any*): Enumerator = ${ ReificationMacros.enumeratorImpl('sc, 'args) }
+    transparent inline def importer(inline args: Any*): Importer = ${ ReificationMacros.importerImpl('sc, 'args) }
+    transparent inline def importee(inline args: Any*): Importee = ${ ReificationMacros.importeeImpl('sc, 'args) }
+    transparent inline def source(inline args: Any*): Source = ${ ReificationMacros.sourceImpl('sc, 'args) }
+  }
+
+  // unapply methods
+  extension (stringContext: scala.StringContext)
```
tgodzik, line 31: Why do we need to repeat the methods here?

quasiquotes3/shared/src/main/scala/scala/meta/quasiquotes/Api.scala
```diff
@@ -0,0 +1,196 @@
...
+      val parse = implicitly[Parse[Importee]]
+      parse(input, dialect)
+    } match {
+      case x: scala.meta.parsers.Parsed.Success[_] => x.tree
+      case x: scala.meta.parsers.Parsed.Error => throw x.details
+    }
+  }
+  object XTensionQuasiquoteSource {
+    private[meta] def parse(input: scala.meta.inputs.Input, dialect: scala.meta.Dialect) = {
+      val parse = implicitly[Parse[Source]]
```
tgodzik, line 181: Since this is all repeated could we not have a method:
```
def parseImpl[T](using parse: Parse[T] )(input: scala.meta.inputs.Input, dialect: scala.meta.Dialect) =
{ 
      parse(input, dialect)
    } match {
      case x: scala.meta.parsers.Parsed.Success[_] => x.tree
      case x: scala.meta.parsers.Parsed.Error => throw x.details
    }
```
and use it everwhere?

quasiquotes3/shared/src/main/scala/scala/meta/quasiquotes/Api.scala
```diff
@@ -0,0 +1,196 @@
...
+    @annotation.compileTimeOnly(".source should not be called directly. Use source\"...\" string interpolation.")
+    def source: QuasiquoteUnapply = ???
+  
+  extension (inline sc: QuasiquoteUnapply)
+    transparent inline def unapplySeq(scrutinee: Any): Option[Seq[Any]] =
+      ${ ReificationMacros.unapplyImpl('sc, 'scrutinee) }
+
+
+  // parsers
+  object XTensionQuasiquoteTerm {
```
tgodzik, line 65: How are those parsers used? I don't see any explicit usage.

quasiquotes3/shared/src/test/scala/scala/meta/tests/SuccessSuite.scala
```diff
@@ -0,0 +1,2955 @@
...
+    val xs = List(q"x", q"y").asInstanceOf[List[Term]]
+    assertEquals(q"foo(..${xs})".syntax, "foo(x, y)")
+    val xss = List(List(q"x", q"y")).asInstanceOf[List[List[Term]]]
+    assertEquals(q"foo(...${xss})".syntax, "foo(x, y)")
+    val rhs = q"x".asInstanceOf[Term]
+    assertEquals(q"var foo = ${rhs}".syntax, "var foo = x")
+  }
+
+  // test("deconstruction ascriptions") {
+  //   val q"foo(..${xs: List[Term]})" = q"foo(x, y)"
```
tgodzik, line 40: Ach, I see where we have the type ascriptions, I wasn't even sure it was possible. Would be good to support but if it's too hard we can leave it out.

quasiquotes3/shared/src/test/scala/scala/meta/tests/SuccessSuite.scala
```diff
@@ -0,0 +1,2955 @@
...
+import munit.Clue.generate
+import scala.meta.trees._
+import meta.prettyprinters.XtensionStructure
+
+import compat.Platform.EOL
+
+import meta.prettyprinters.XtensionSyntax
+
+class Success extends TreeSuiteBase {
+  test("rank-0 liftables") {
```
tgodzik, line 17: Isn't this the same test suite like for Scala 2? Could we reuse it? Or would it need my PR first I guess?

quasiquotes3/shared/src/test/scala/scala/meta/tests/SuccessSuite.scala
jchyb: without a new SIP it is impossible unfortunately

quasiquotes3/shared/src/main/scala/scala/meta/quasiquotes/Api.scala
jchyb: In `ReificationMacros` in `instantiateParser()`, they are classloaded and invoked (just like in scala 2). In scala 2 this boilerplate was hidden inside classes with a macro annotation

quasiquotes3/shared/src/main/scala/scala/meta/quasiquotes/Api.scala
jchyb: they have different names, and those names point to different parsers in `ReificationMacros`

quasiquotes3/shared/src/main/scala/scala/meta/internal/quasiquotes/ExprLifts.scala
jchyb: It was all generated by a macro triggered inside ReificationMacros.scala scalameta/common/shared/src/main/scala/scala/meta/internal/trees/Liftables.scala (`materializeAst[]')

jchyb:
1. It's unrelated to what we return from the unapply, but concerns how the unapply method is implemented. In our unapply macro method we create a custom pattern match using the reflection API (where we basically create the program tree ourselves). The crash seems to trigger when we use an UnApply node for `Some` in that program tree as part of a macro expansion of an `unapply` method (pretty specific case, and finding the cause here was sure something...).
Curiously, the crash does not trigger when rewriting the `Some` pattern match into a quoted expression (`'{...}`) (by converting I mean hardcoding a single macro method output, in general it is impossible for us to redesign the macro to be able to use quoted expressions like this). When converting that quoted expression to a program tree, the compiler wraps `UnApply` nodes with additional `Typed` nodes, which I suspect are related why it is stopping crashing in that case. We cannot use the Typed nodes in the same way in the reflection API by design, but other cases of `UnApply` nodes without the `Typed` node (and even Some in other contexts) work.
Even more curiously, this crash did not trigger before, when I had `implicit dialect: Dialect` in the `unapplySeq` method signature (I had to get rid of it on Tuesday for other reasons, which is when I noticed the crash), so I want to try having some other fake implicit in that signature as a workaround. But right now I'm looking directly into the compiler in case the fix for the underlying issue is trivial.
2. Agreed
3. Yes it is at the beginning of the `SuccessSuite.scala`. The other issue here is the fact when using incorrect ascriptions, the macro was able to error out in compiletime, now we get a match error on runtime. I believe this is something which will be able to be addressed by the compiler itself after making the method `transparent inline unapply`.

quasiquotes3/shared/src/test/scala/scala/meta/tests/SuccessSuite.scala
jchyb: For now it has tests slightly changed to accommodate the `Any` types returned from the unapply. I hope to use the old suite once we can get `transparent inline unapply` working, outside of those 2 tests with non-standard type ascriptions of course

quasiquotes3/shared/src/test/scala/scala/meta/tests/SuccessSuite.scala
tgodzik: It's fine to drop that functionality for Scala 2 then

kitbellew:
@tgodzik is this logic close enough, to be included in 4.9.0?

tgodzik:
> @tgodzik is this logic close enough, to be included in 4.9.0?

I think it needs more work still, especially on the cross publishing branch

mio-19:
How is the status of this pr?

tgodzik:
We need to first add cross publish for relevant modules. I almost managed to fix it, but I need to rewrite the XMLParser we have to something simpler.

tgodzik:
That's a bit more complex case, but we can't yet cross compile the trees module as it contains overly complex macro code. We can also extract XmlParser since it uses the scanner, which it again is intertwined with the macros. And if we compile tree with 2.13 we get fastparse_2.13 on the classpath, which doesn't work with modules compiled with Scala 3 as it also uses macros.

So the simplest solution is to change the XmlParser to a simpler one as we don't really need to be as exact as it is now.

mio-19:
If XmlParser has its own subproject, the fastparse_2.13 problem would be gone

tgodzik:
> If XmlParser has its own subproject, the fastparse_2.13 problem would be gone

Unfortunately, XmlParser actually uses the scalameta scanner, so both need each other

jchyb:
Fixed most of the problems on my end here. Quasiquotes unapply now finally works, and I also managed to have them all correctly automatically add type ascriptions, eg. in :
```scala
    val q"1 match { case _: List[..$tpes] => }" = q"1 match { case _: List[t] => }"
```
`tpes` correctly gets typed as ArgClause (previously it was Any).
This means that I could make the test for scala 2 and 3 reasonably similar enough that they can be reused. I will not be able to reasonably implement the custom `deconstruction ascriptions`, as discussed before, eg:
```scala
val q"1 match { case _: List[..${tpes: List[Type]}] => }" = q"1 match { case _: List[t] => }"
```
where ArgClause gets automatically converted to List[Type], but I left empty spots to implement that if getting the ascribed types will become possible. However the user-facing workarounds should be simple.

Only parts left to do her is rebase and possibly rewrite ExprLifts into a macro/codegen script to simplify maintainance (in scala 2 this part was a macro, that was hard to read snd seemed like a lot of work at the time, so for now I generated the code into a string and manually adapted it for scala 3). Not sure if that is worth doing, or if ExprLifts is maintainable enough for now.

tgodzik:
> so for now I generated the code into a string and manually adapted it for scala 3)

Och interesting, I wonder if this approach would work at some point for trees module :thinking: How did you do it?

Maybe it's better to leave it as non macro, but I will take a look (also historically) if this will be a problem.

We are still blocked on my PR, but I think I have another idea how to get that merged.

quasiquotes3/shared/src/main/scala/scala/meta/quasiquotes/Api.scala
tgodzik: What do you think about this one? Is it legit to rewrite those into single function?

quasiquotes3/shared/src/test/scala/scala/meta/tests/TestHelpers.scala
```diff
@@ -0,0 +1,21 @@
...
+// import scala.meta.Token
+
+object TestHelpers {
+
+  def getMessageWithExtraClue(msg: String, extraClue: String = ""): String =
+    if (extraClue.isEmpty) msg else s"$msg ($extraClue)"
+
+  def getSyntax(code: => String, syntax: String = null): String = Option(syntax).getOrElse(code)
+
+  // trait Tokenize {
```
tgodzik, line 13: If not needed let's remove those

quasiquotes3/shared/src/main/scala/scala/meta/internal/quasiquotes/ConversionMacros.scala
```diff
@@ -0,0 +1,87 @@
...
+                case Some(liftable) => '{$liftable.apply(${outsideExpr.asExprOf[t]})}
+                case None => report.errorAndAbort(typeMismatchMessage(outsideTpe, insideTpe))
+        }
+  }
+
+  def liftUnapply[I: Type](outside: Expr[Any]): Expr[Option[I]] = {
+    // NOTE: Here's an interesting idea that I'd like to explore.
+    // How about we allow things like `42 match { case q"$x" => x }`?
+    // For that to work, we just need to wrap the reification result into `Lift.unapply`!
+    ???
```
tgodzik, line 64: Let's not leave it at `???` better to throw an exception with an explanation.

quasiquotes3/shared/src/main/scala/scala/meta/internal/quasiquotes/ConversionMacros.scala
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  def unliftApply[O: Type](inside: Expr[Any]): Expr[O] = {
+    // NOTE: here we just disregard the expected outside type, because I can't find uses for it
```
tgodzik, line 68: Best to mark these by your login so people in the future will know who commented on it, what do you think?

jchyb: None of those are actually mine, I copied those from the Scala 2 counterpart implementation. I'll add that information to the note.

quasiquotes3/shared/src/main/scala/scala/meta/internal/quasiquotes/ConversionMacros.scala
jchyb: Same here as below, copied from the counterpart implementation for Scala 2. I can add a legitimate message, but then it would propably be better to also add it into the Scala 2 version of the file

quasiquotes3/shared/src/test/scala/scala/meta/tests/TestHelpers.scala
jchyb: Due to the current build structure, I had to copy some stuff from the tests module, including this (since I could not have it depend on the tests module, since it contains some macro implementations I believe). I could not copy everything (aforementioned macro and structure issues), so I commented out the problematic and unnecessary parts). I needed these files so as to not have to change the actual test file (for easier comparison between scala 2 and 3). What I am trying to say is that I think it will be better for me to polish this part later, when the build structure is set in stone (perhaps I will be able to completely get rid of this copy)

jchyb:
I wanted to rebase on top of the latest build but happened upon what seems to be a very niche issue in the compiler: https://github.com/scala/scala3/issues/22177. 

Unless I will manage to find a workaround, I might not be able to put the scala 3 quasiquotes into the `parsers` module (where the scala 2 version resides). Basically what happens is the Scala 3 compiler is not able to have the `scala.meta.internal.trees.Quasi.become` which is done while we are parsing the quasiquote. It did not happen before, because I had extruded a part of parsers (running in scala 2.13, this is the part of the code that directly calls `become`) and made a separate module for scala 3 quasiquotes (and I might have to keep it that way for now).

tgodzik:
>  It did not happen before, because I had extruded a part of parsers (running in scala 2.13, this is the part of the code that directly calls become)

Which part is that?

If possible I would avoid a separate module or at least make sure that people don't have to add it as a separate dependency.

Any idea how hard it would be to fix the issue in the compiler?

jchyb:
This is where it would throw the exception: https://github.com/scalameta/scalameta/blob/55b9ce167870ff29d21d51cd4fe8a35b055b4746/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L4351
It's in ScalaMetaParser, which we call inside a macro (to parse the quasiquote) from Scala 3, but it's called on an object from Scala 2. (like in the linked issue minimisation).
Not sure how much time it would take to fix that in the compiler, I'll try to investigate today. 
Good news is there is a simple workaround, but it probably breaks bincompat (removing `with Quasi` part from `def become[T <: Tree: AstInfo]: T with Quasi` method definition and adding one asInstanceOf).

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -1513,7 +1513,9 @@ package internal.trees {
     def rank: Int
     def tree: Tree
     def pt: Class[_]
-    def become[T <: Tree: AstInfo]: T with Quasi
+    def become[T <: Tree: AstInfo]: T
+    // `become` should be returning `T with Quasi`,
+    // workaround for https://github.com/scala/scala3/issues/22177
```
jchyb, line 1518: The workaround in question. We could also add another method, which we then would use instead in ScalametaParser, keeping the more type safe `become` for the users (or at least Scala 2 users).

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/QuasiquoteType.scala
```diff
@@ -0,0 +1,19 @@
+package scala.meta
+
+private[meta] enum QuasiquoteType:
+  case Term // q""
```
kitbellew, line 4: i thought `q""` covered any `Stat`, no? not just terms but also  definitions etc.

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/quasiquotes/Api.scala
```diff
@@ -0,0 +1,196 @@
...
+
+
+  // parsers
+  object XTensionQuasiquoteTerm {
+    private[meta] def parse(input: scala.meta.inputs.Input, dialect: scala.meta.Dialect) = {
+      val parse = implicitly[Parse[Ctor]]
+      parse(input, dialect)
+    }.orElse{
+      val parse = implicitly[Parse[Stat]]
+      parse(input, dialect)
```
kitbellew, line 71: can this not be generalized a bit, say, like this:
```
private[meta] object Api { // companion; not sure how much has to be part of `trait Api`

  def parse[A <: Tree](input: ..., dialect: ...)(implicit parser: Parse[A]) = {
    parser(input, dialect)
  }

  def asParsedAny[A <: Tree, B <: Tree](input: ..., dialect: ...)(implicit parser: Parse[A]) = {
    parse[A](input, dialect).orElse(parse[B](input, dialect))
  }
}

// here
object XtensionQuasiquoteXxx {
  private[meta] parse(input, dialect) = parseAny[Ctor, Stat](input, dialect).get
}
```

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/quasiquotes/Api.scala
```diff
@@ -0,0 +1,196 @@
...
+    } match {
+      case x: scala.meta.parsers.Parsed.Success[_] => x.tree
+      case x: scala.meta.parsers.Parsed.Error => throw x.details
+    }
```
kitbellew, line 75: isn't this a simple `.get` on the `Parsed` instance?

tests/jvm/src/test/scala/scala/meta/tests/parsers/BootstrapSuite.scala
```diff
@@ -20,6 +20,7 @@ class BootstrapSuite extends ParseSuite {
       val name = dir.getName
       if (dir.isDirectory && name == "target" || name == "community-projects") return
       def bootstrapTest(src: File): Unit = test("tokenize " + src.getAbsolutePath) {
+        implicit val dialect: Dialect = dialects.Scala33
```
kitbellew, line 23: why is this needed now?

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
kitbellew: it's under `internal`, so should be fine. otherwise, you'd get mima errors.

kitbellew (changes requested):
caveat: i didn't read the larger portions related to actual scala3 macros, as i don't have any experience with them.

the only suggestion, here and in other parts of this process, is it would be great to mention (if necessary) in scala-2 and scala-3 codebases where in the *other* language the same logic is implemented; this would help greatly to ensure that any future changes are in sync as otherwise an implementer might forget about the other side. and somehow make sure everything is tested under both, so that changing scala-2 but not -3 will trigger test failures.

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/quasiquotes/Api.scala
jchyb: Ah! That's true! This was one of the cases where I reused what was generated via macro annotation in scala 2, I missed this similarity completely. I'll try to simplify it.

tests/jvm/src/test/scala/scala/meta/tests/parsers/BootstrapSuite.scala
jchyb: This test would fail while parsing the added scala 3-only files with macros. Alternatively we can filter those out and keep the test as it was.

kitbellew: as i mentioned earlier, i'd prefer to have all, or most, tests cover both scala 2 and 3, so their function as regression test will continue to work.

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/quasiquotes/Api.scala
```diff
@@ -57,136 +57,62 @@ private[meta] trait Api {
...
+private[meta] object Api {
+  def parse[A <: Tree](implicit input: inputs.Input, dialect: Dialect, parser: Parse[A]) =
+    parser(input, dialect)
+
+  def parseAny[A <: Tree, B <: Tree](
+      implicit input: inputs.Input, dialect: Dialect, parserA: Parse[A], parserB: Parse[B]
+    ) = parse[A](input, dialect).orElse(parse[B](input, dialect)).get
+  
+  def parseAny[A <: Tree](implicit input: inputs.Input, dialect: Dialect, parser: Parse[A]) =
+    parse[A](input, dialect).get
```
kitbellew, line 113: i'd simplify a little:
```
  def parseAny[A <: Tree : Parser](implicit input: inputs.Input, dialect: Dialect) =
    parse[A].get

  def parseAny[A <: Tree : Parser, B <: Tree : Parser](implicit input: inputs.Input, dialect: Dialect) =
    parse[A].orElse(parse[B]).get
```

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ExprLifts.scala
```diff
@@ -0,0 +1,542 @@
...
+      case (y @ (_: scala.meta.Type.CasesBlock)) => liftTypeCasesBlock(y.asInstanceOf[scala.meta.Type.CasesBlock])
+      case (y @ (_: scala.meta.Term.CasesBlock)) => liftTermCasesBlock(y.asInstanceOf[scala.meta.Term.CasesBlock])
+      case (y @ (_: scala.meta.Term.EnumeratorsBlock)) => liftTermEnumeratorsBlock(y.asInstanceOf[scala.meta.Term.EnumeratorsBlock])
+      case (y @ (_: scala.meta.Stat.Block)) => liftStatBlock(y.asInstanceOf[scala.meta.Stat.Block])
+      case (y @ (_: scala.meta.Ctor.Block)) => liftCtorBlock(y.asInstanceOf[scala.meta.Ctor.Block])
+      case (Y @ (_: scala.meta.Pkg.Body)) => liftPkgBody(y.asInstanceOf[scala.meta.Pkg.Body])
+      case _ => sys.error("none of leafs matched " + (y.getClass))
+    }
+  }
+}
```
kitbellew, line 542: missing EOL at EOF

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ExprLifts.scala
```diff
@@ -0,0 +1,542 @@
...
+  def liftTreess(treess: List[List[MetaTree]]): internalQuotes.reflect.Tree
+  def liftQuasi0(quasi: Quasi, optional: Boolean = false): internalQuotes.reflect.Tree
+  def liftOrigin(origin: Origin): internalQuotes.reflect.Tree
+
+  protected def unquotesName(q: scala.meta.internal.trees.Quasi): Boolean
+}
+
+// In Scala 2 version, this part of the codebase was automatically generated by a macro,
+// with materializeAst[MetaTree].
+// Probably would be better if it was also generated here, by a similar macro or by a custom codegen script.
```
kitbellew, line 30: probably would be better. since we keep adding trees, this code would have to continue evolving, and it might become high-maintenance. you probably explained why it was hard or impossible to do, i need to go over your earlier comments.

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ExprLifts.scala
kitbellew, line 1: just discovered this file, GH hid it so it looked like a simple renamed file.

kitbellew (commented):
discovered more code, sorry about the back-and-forth.

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
jchyb: It might, but requires way more changes (e.g. adding with Quotes to where AstInfo is generated, and specifying `with Quotes` for every typeBound where we pass Tree).

jchyb:
I think I am cursed this week - I wanted to add the codegen module and halfway through I ran into another issue: https://github.com/scala/scala3/issues/22200. This one I'll try to fix in the compiler, since that might be faster (and since it's for an unreleased codegen module, it's possible for us to use the latest versions of the compiler)

jchyb:
> > I think I am cursed this week - I wanted to add the codegen module and halfway through I ran into another issue: [scala/scala3#22200](https://github.com/scala/scala3/issues/22200). This one I'll try to fix in the compiler, since that might be faster (and since it's for an unreleased codegen module, it's possible for us to use the latest versions of the compiler)
> 
> Could we add the codegen later on?

From my perspective it should not be a problem. One issue is that in the meantime adding new stuff to trees could be a bit difficult in the meantime (but I don't know if there is anything needed to add in the near future).

jchyb:
I did find a workaround (the underlying issue is a bit harder than I anticipated), but perhaps I should just swallow my pride and write this codegen in Scala 2... Either way by Monday it should be done

scala3-tree-lifts/impl/src/main/scala/org/scalameta/adt/Main.scala
```diff
@@ -0,0 +1,21 @@
+package org.scalameta.adt
+
+import scala.meta.{Tree => MetaTree}
+
+import java.io.File
+import java.nio.file.Files
+
+object Main {
+  def main(args: Array[String]): Unit = {
+    println(args.toList)
```
tgodzik, line 10:
```suggestion
```

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ConversionMacros.scala
```diff
@@ -0,0 +1,93 @@
...
+          liftableMaybe match
+            case Some(liftable) => '{ $liftable.apply(${ outsideExpr.asExprOf[t] }) }
+            case None => report.errorAndAbort(typeMismatchMessage(outsideTpe, insideTpe))
+
+  def liftUnapply[I: Type](outside: Expr[Any]): Expr[Option[I]] =
+    // Note copied from the Scala 2 counterpart file.
+    // NOTE: Here's an interesting idea that I'd like to explore.
+    // How about we allow things like `42 match { case q"$x" => x }`?
+    // For that to work, we just need to wrap the reification result into `Lift.unapply`!
+    ???
```
tgodzik, line 74: Let's throw an exception with some better message

tgodzik (commented):
Looks like this is ready or almost? I skimmed over the macro code, but probably not the best person to look into it. Any suggestions who could do it?

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ConversionMacros.scala
jchyb: I've added one here and in the Scala 2 counterpart

jchyb:
Yes, I consider it ready

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ConversionMacros.scala
tgodzik: Thanks!

scala3-tree-lifts/macro/src/main/scala/org/scalameta/adt/TreeLiftsGenerate.scala
```diff
@@ -0,0 +1,240 @@
...
+          |  def liftableSubTree0[T <: MetaTree](y: T)(using Quotes): Tree = {
+          |    y match {
+          |      ${clauses.mkString("\n      ")}
+          |      case _ => sys.error("none of leafs matched " + (y.getClass))
+          |    }
+          |  }
+          |}
+          |""".stripMargin
+    val retStrList = Literal(Constant(retStr))
+    c.Expr[String](retStrList)
```
kitbellew, line 171: is this a way to turn a string into a compiled code in scala3?

scala3-tree-lifts/macro/src/main/scala/org/scalameta/adt/TreeLiftsGenerate.scala
```diff
@@ -0,0 +1,240 @@
...
+       |      case mType @ MethodType(_, _, res) => getResultType(res, hasImplicit || mType.isImplicit)
+       |      case res => (res, hasImplicit)
+       |    val apply = term.symbol.methodMember("apply").head
+       |    val (resType, hasImplicit) = getResultType(expr.asTerm.tpe.memberType(apply), hasImplicit = false)
+       |    resType.asType match
+       |      case '[t] =>
+       |        if isPatternMode then
+       |          TypedOrTest(Unapply(Select.unique(term, "unapply"), Nil, args.toList), TypeTree.of[t])
+       |        else 
+       |          val applied = Select.overloaded(term, "apply", Nil, (origin.map(List(_)).getOrElse(Nil) ++ args.toList).asInstanceOf[List[Term]])
```
kitbellew, line 204:
```
origin.map(List(_)).getOrElse(Nil) ++ args.toList
```
can be replaced with
```
origin.fold(args.toList)(_ :: args.toList)
```
or
```
origin.toList :: args.toList
```

scala3-tree-lifts/macro/src/main/scala/org/scalameta/adt/TreeLiftsGenerate.scala
```diff
@@ -0,0 +1,240 @@
...
+          |$treeByMode
+          |
+          |$termMethods
+          |
+          |  ${liftAdts.mkString("\n  ")}
+          |
+          |  def liftableSubTree0[T <: MetaTree](y: T)(using Quotes): Tree = {
+          |    y match {
+          |      ${clauses.mkString("\n      ")}
+          |      case _ => sys.error("none of leafs matched " + (y.getClass))
```
kitbellew, line 165: `.getClass` and then `.getSimpleName`, perhaps?

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ReificationMacros.scala
```diff
@@ -0,0 +1,688 @@
...
+    val (dialect, dialectExpr) = instantiateDialect(mode)
+    val parser = instantiateParser(qType)
+    val skeleton = parseSkeleton(parser, input, dialect)
+    reifySkeleton(skeleton, mode, qType, dialectExpr, input)
+  }
+
+  private def metaInput() = {
+    val pos = Position.ofMacroExpansion
+    val reflectInput = pos.sourceFile
+    val content = new String(reflectInput.content.get)
```
kitbellew, line 171: why do you need `new String`? isn't `.content` an `Option[String]` already?

also, what to do if it's None?

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ReificationMacros.scala
```diff
@@ -0,0 +1,688 @@
...
+    val start = {
+      var i = 0
+      while (content(pos.start + i) != '"') i += 1 // skip method name
+      while (content(pos.start + i) == '"') i += 1 // skip quotations
+      pos.start + i
```
kitbellew, line 176: why not:
```
      var i = pos.start
      while (content(i) != '"') i += 1 // skip method name
      while (content(i) == '"') i += 1 // skip quotations
      i
```

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ReificationMacros.scala
```diff
@@ -0,0 +1,688 @@
...
+    }
+    val end = {
+      var i = 0
```
kitbellew, line 179: same:
```
      var i = pos.end - 1
      while (content(i) == '"') i -= 1 // skip quotations
      i + 1
```

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ReificationMacros.scala
```diff
@@ -0,0 +1,688 @@
...
+      .head
+    (input: Input, dialect: Dialect) =>
+      try parserMethod.invoke(parserModuleInstance, input, dialect).asInstanceOf[MetaTree]
+      catch { case ex: java.lang.reflect.InvocationTargetException => throw ex.getTargetException }
+  }
+  private def parseSkeleton(parser: MetaParser, input: Input, dialect: Dialect): MetaTree =
+    try parser(input, dialect)
+    catch {
+      case TokenizeException(pos, message) => report.errorAndAbort(message)
+      case ParseException(pos, message) => report.errorAndAbort(message)
```
kitbellew, line 287: `pos` is not used. is it possible to avoid unapply then and just do
```
  case ex @ (_: TokenizeException | _: ParseException) => report.errorAndAbort(ex.getMessage)
```

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ReificationMacros.scala
```diff
@@ -0,0 +1,688 @@
...
+              else {
+                Predef.require(prefix.isEmpty && debug(trees, acc, prefix))
+                if mode.isTerm then
+                  val tree = Apply(
+                    TypeApply(Select.unique(acc.get.asInstanceOf[Term], "++"), List(TypeTree.of[T])),
+                    List(liftQuasi(quasi).asInstanceOf[Term])
+                  )
+                  loop(rest, Some(tree), Nil)
+                else report.errorAndAbort(Messages.QuasiquoteAdjacentEllipsesInPattern(quasi.rank))
+              }
```
kitbellew, line 394: I would strongly prefer to avoid `Option.get` (like you do on line 389 here, and in some other parts of the code as well). why not:

```
              acc match {
                case None if prefix.isEmpty => loop(rest, Some(liftQuasi0(quasi)), Nil)
                case None =>
                  val acc = prefix.foldRight(Option.empty[Tree]) { (curr, acc) =>
                    // Note copied from the Scala 2 counterpart file.
                    // NOTE: We cannot do just q"${liftTree(curr)} +: ${liftQuasi(quasi)}"
                    // because that creates a synthetic temp variable that doesn't make any sense in a pattern.
                    // Neither can we do q"${liftQuasi(quasi)}.+:(${liftTree(curr)})",
                    // because that still wouldn't work in pattern mode.
                    // Finally, we can't do something like q"+:(${liftQuasi(quasi)}, (${liftTree(curr)}))",
                    // because would violate evaluation order guarantees that we must keep.
                    val currElement = liftTree(curr)
                    val alreadyLiftedList = acc.getOrElse(liftQuasi(quasi))
                    Some(
                      if (mode.isTerm) Apply(
                        TypeApply(
                          Select.unique(alreadyLiftedList.asInstanceOf[Term], "+:"),
                          List(TypeTree.of[T])
                        ),
                        List(currElement.asInstanceOf[Term])
                      )
                      else Unapply(
                        TypeApply(
                          Select.unique('{ +: }.asTerm, "unapply"),
                          List(TypeTree.of[T], TypeTree.of[List], TypeTree.of[List[T]])
                        ),
                        Nil,
                        List(currElement, alreadyLiftedList)
                      )
                    )
                  }
                  loop(rest, acc, Nil)
                case Some(x: Term) if mode.isTerm =>
                  Predef.require(prefix.isEmpty && debug(trees, acc, prefix))
                  val tree = Apply(
                    TypeApply(Select.unique(x, "++"), List(TypeTree.of[T])),
                    List(liftQuasi(quasi).asInstanceOf[Term])
                  )
                  loop(rest, Some(tree), Nil)
                case _ =>
                  report.errorAndAbort(Messages.QuasiquoteAdjacentEllipsesInPattern(quasi.rank))
              }
```

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ReificationMacros.scala
```diff
@@ -0,0 +1,688 @@
...
+                else
+                  TypedOrTest(
+                    Unapply(
+                      TypeApply(Select.unique('{ List }.asTerm, "unapplySeq"), List(TypeTree.of[T])),
+                      Nil,
+                      args
+                    ),
+                    TypeTree.of[List[T]]
+                  )
+              } else acc.get
```
kitbellew, line 440: this one is a simple `acc.getOrElse { ... }`

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ReificationMacros.scala
```diff
@@ -0,0 +1,688 @@
...
+                          Select.unique('{ :+ }.asTerm, "unapply"),
+                          List(TypeTree.of[T], TypeTree.of[List], TypeTree.of[List[T]])
+                        ),
+                        Nil,
+                        List(acc.get, otherTree)
+                      ),
+                      TypeTree.of[List[Any]]
+                    )
+                loop(rest, Some(tree), Nil)
+              }
```
kitbellew, line 419: same comment as above:
```
              acc match {
                case None => loop(rest, acc, prefix :+ other)
                case Some(x) =>
                  Predef.require(prefix.isEmpty && debug(trees, acc, prefix))
                  val otherTree = liftTree(other)
                  val tree =
                    if mode.isTerm then
                      Apply(
                        TypeApply(Select.unique(x.asInstanceOf[Term], ":+"), List(TypeTree.of[T])),
                        List(otherTree.asInstanceOf[Term])
                      )
                    else
                      TypedOrTest(
                        Unapply(
                          TypeApply(
                            Select.unique('{ :+ }.asTerm, "unapply"),
                            List(TypeTree.of[T], TypeTree.of[List], TypeTree.of[List[T]])
                          ),
                          Nil,
                          List(x, otherTree)
                        ),
                        TypeTree.of[List[Any]]
                      )
                  loop(rest, Some(tree), Nil)
              }
```

tests/jvm/src/test/scala/scala/meta/tests/parsers/BootstrapSuite.scala
kitbellew: was this comment addressed?

kitbellew (changes requested):

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ReificationMacros.scala
```diff
@@ -0,0 +1,688 @@
...
+        val (returned, isBooleanExtractor) =
+          if (n == 0)
+            // boolean extractor
+            ('{ true }, true)
+          else if (n == 1)
+            // returning Tuple1 requires us to manually unapply later, so we return the raw value instead
+            ('{ ${ args.head.asExprOf[Option[Any]] }.get }, false)
+          else (
+            Select.overloaded(
+              Ref(Symbol.classSymbol("scala.Tuple" + n).companionModule),
```
Florian3k, line 637: Maybe use `TupleClass`?
```suggestion
              Ref(defn.TupleClass(n).companionModule),
```

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ReificationMacros.scala
jchyb: I ended up changing this so that we do use the position now (forgot to do that before)

scala3-tree-lifts/macro/src/main/scala/org/scalameta/adt/TreeLiftsGenerate.scala
jchyb: We return the String in a scala 2 macro so that the runtime code (scala3-tree-lifts/impl/src/main/scala/org/scalameta/adt/Main.scala) can generate the scala 3 file.

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/registry.scala
```diff
@@ -29,7 +29,7 @@ class RegistryMacros(val c: Context) extends AstReflection with MacroHelpers {
       ) = mdef
       val enclosingUnit = c.asInstanceOf[{ def enclosingUnit: { def body: Tree } }].enclosingUnit
         .body
-      val anns1 = anns :+ q"new $AstMetadataModule.registry(${enclosingUnit.detectAst})"
+      val anns1 = anns :+ q"new $AstMetadataModule.registry(..${enclosingUnit.detectAst})"
```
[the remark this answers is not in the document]
jchyb: Apologies, this and the other changes to scala 2 trees (outside of the `Quasi` one), were leftovers from attempts to write the codegen part in Scala 3 (instead of how it ended up now, in Scala 2) that somehow got committed.

tests/jvm/src/test/scala/scala/meta/tests/parsers/BootstrapSuite.scala
jchyb: In the most recent changes, I ended up removing this dialect, and just filtering scala-3 files, preserving previous behavior of the test.

scala3-tree-lifts/macro/src/main/scala/org/scalameta/adt/TreeLiftsGenerate.scala
kitbellew: another comment yet to be replied to :)

scalameta/common/shared/src/main/scala/scala/meta/internal/trees/Reflection.scala
```diff
@@ -35,7 +35,7 @@ trait Reflection extends AdtReflection {
   private lazy val scalaMetaRegistry: Map[Symbol, List[Symbol]] =
     AllModule.initialize.annotations match {
       case List(ann) if ann.tree.tpe =:= RegistryAnnotation.toType =>
-        val q"new $_($_.$_[..$_](..${astPaths: List[String]}))" = ann.tree
+        val q"new $_(..${astPaths: List[String]})" = ann.tree
```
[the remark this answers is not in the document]
jchyb: Sorry, just noticed the comment. I think it should be ok to merge this separately (I initially make the change for a workaround for a bug, when reading scala2 annotations in scala 3, when writing the TreeLifts codegen, but I ended up having to rewrite that with scala 2 anyway). The PR is here: https://github.com/scalameta/scalameta/pull/4121

kitbellew:
@jchyb as i mentioned earlier, i am not proficient enough to do a conclusive review, perhaps @tgodzik has more insight (and if not, we can try merging).

in either case, i would propose to release v4.12.4 without this change, and then fast-follow with v4.12.5 containing just this change. that way, in the unlikely event there's a problem, people have a workable and updated version while we are figuring out a fix.

kitbellew:
> > as i mentioned earlier, i am not proficient enough to do a conclusive review, perhaps @tgodzik has more insight (and if not, we can try merging).
> 
> I don't think I can find any issues right away, we might need to ship and get people using it to find issues.

Yep. Some people have been requesting this for a while, they could lend a hand.

> > in either case, i would propose to release v4.12.4 without this change, and then fast-follow with v4.12.5 containing just this change. that way, in the unlikely event there's a problem, people have a workable and updated version while we are figuring out a fix.
> 
> After #4112 maybe?

Agreed.

tgodzik (approved):
Thanks for the huge amount of work @jchyb ! I think we are good to merge

kitbellew (approved):

PR #3356, opened 2023-10-02
ScalametaParser: fix warnings, inherited vs outer

kitbellew:
Most recently, compiler started displaying warnings about invocations of a few methods that they are defined both in an outer class as well as a parent class and that such ambiguity will not be allowed in scala3.


tgodzik (approved):
I just saw them, thanks!

PR #3358, opened 2023-10-03
improvement: Remove macro usage from Scalameta parser

tgodzik:
I was trying to get the parsers code to compile with Scala 3, but the classifier macros were a bit of a problem. Rewriting them would require other modules to also cross compile, while removing makes it possible to just use other modules as Scala 2.13 ones.

The changes are not terrible, so I feel like the macros weren't really neccessary, while obviously useful. 

The worst part is syntaxErrorExpected but I haven't figured out how to do that better and that is just error messages, so nothing should break


kitbellew:
i haven't read it fully yet but: don't you need to start with tokens and trees? or perhaps i didn't understand the comment about other modules and 2.13.

tgodzik:
> i haven't read it fully yet but: don't you need to start with tokens and trees? or perhaps i didn't understand the comment about other modules and 2.13.

My plan is to leave tokens and trees at 2.13 for the time being. 2.13.x can be used from 3 aside from macros, so if they're all generated in other modules we should be fine.

kitbellew (approved):

PR #3360, opened 2023-10-05
improvement: Generate implicits for AstInfo class instead of from macros

tgodzik:
Thanks to this change I managed to compile the parsers module with Scala 3, which is pretty close to getting it cross published.


kitbellew:
@tgodzik how are you testing compilation with scala3? the reason i am asking is, there are still invocations of AstInfo in ScalametaParser (reelipsis, which uses Quasi.become), so that part must somehow still work... so why that but not this?

tgodzik:
> @tgodzik how are you testing compilation with scala3? the reason i am asking is, there are still invocations of AstInfo in ScalametaParser (reelipsis, which uses Quasi.become), so that part must somehow still work... so why that but not this?

`AstInfo` is still in use, but I created it manually for almost all the types, so there is no need for macros to generate them.

You can test it out on the branch here: https://github.com/scalameta/scalameta/compare/main...tgodzik:scalameta:cross-publish2?expand=1

I still have the tests to compile, but they should be simple enough.

PR #3362, opened 2023-10-05
AstInfo: define implicit explicitly for each tree

kitbellew:


kitbellew:
@tgodzik ptal, this seems to work with scala3 as well. i tried it with your [tgodzik/scalameta/cross-publish2](https://github.com/tgodzik/scalameta/tree/cross-publish2) branch.

tgodzik (approved):
Makes sense! Though we will have to migrate off macros at one point if Scala 3 drops compatibility with 2.13, but that might not happen any time soon.

PR #3365, opened 2023-10-06
improvement: Don't pattern match in tests in vals

tgodzik:
This is showing up as warnings in Scala 3 since the left side is more specialized and need an explicit unchecked annotation.

There is still plenty to fix unfortunately, but will continue next week.

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/DeclSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/DeclSuite.scala
@@ -6,23 +6,30 @@ import scala.meta.dialects.Scala211
 
 class DeclSuite extends ParseSuite {
   test("val x: Int") {
-    val Decl.Val(Nil, List(Pat.Var(Term.Name("x"))), Type.Name("Int")) = templStat("val x: Int")
+    assertTree(templStat("val x: Int"))(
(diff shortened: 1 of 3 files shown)


tests/shared/src/test/scala/scala/meta/tests/parsers/DeclSuite.scala
```diff
@@ -6,23 +6,30 @@ import scala.meta.dialects.Scala211
 
 class DeclSuite extends ParseSuite {
   test("val x: Int") {
-    val Decl.Val(Nil, List(Pat.Var(Term.Name("x"))), Type.Name("Int")) = templStat("val x: Int")
+    assertTree(Decl.Val(Nil, List(Pat.Var(Term.Name("x"))), Type.Name("Int")))(
```
kitbellew, line 9: i thought the first parameter is "obtained" and the second is "expected", i.e. swap them? not at the computer right now, but could check the parameter names to assertTree when I get to it.

tgodzik: Ach right, I tried automating the changes a bit, so this was done wrong. Fixed now!

kitbellew (approved):

PR #3367, opened 2023-10-08
ScannerTokens: end marker must be preceded by NL

### LINKED ISSUE #3366: Match on result of invocation of method called `end` requires braces
```
version = "3.7.14"
runner.dialect = scala3
```

When I run scalafmt via CLI like this: `scalafmt Test.scala`

## Steps

Given code like this:
```scala
val x = 
  x.end match
    case _ => ()
```

## Problem

Scalafmt doesn't format the code but states:
```
xxxxx/Test.scala:2: error: [dialect scala3] { expected but \n found
  x.end match
             ^
```

## Expectation

It should format the code normally

## Workaround

You can put parentheses around the expression, i.e. `(x.end) match` which works and is my current workaround.

kitbellew:
Fixes #3366.


tgodzik (approved):
LGTM!

PR #3368, opened 2023-10-09
improvement: Get rid of more pattern matches in vals

tgodzik:
I started rewriting the test to have full trees in expect, but that proved to take a long while so instead I created matchSubStructure method and used Chat GPt to help out rewrite it quicker


tests/shared/src/test/scala/scala/meta/tests/parsers/ParseSuite.scala
```diff
@@ -93,6 +93,14 @@ class ParseSuite extends TreeSuiteBase with CommonTrees {
     )
   }
 
+  protected def matchSubStructure[T <: Tree](
```
kitbellew, line 96: this is elegant! :)

tests/shared/src/test/scala/scala/meta/tests/parsers/LitSuite.scala
```diff
@@ -5,93 +5,104 @@ import scala.meta._
 import scala.meta.dialects.Scala211
 
 class LitSuite extends ParseSuite {
+
+  // Structure does not always return consisten results
```
kitbellew, line 9: consistenT

tests/shared/src/test/scala/scala/meta/tests/parsers/LitSuite.scala
```diff
@@ -5,93 +5,104 @@ import scala.meta._
...
   }
 
   test("2147483647") {
-    val Lit(2147483647) = term("2147483647")
+    assertTree(term("2147483647"))(Lit.Int(2147483647))
   }
 
   test("-2147483648") {
-    val Lit(-2147483648) = term("-2147483648")
+    assertTree(term("-2147483648"))(Lit.Int(-2147483648))
```
kitbellew, line 38: this works...

tests/shared/src/test/scala/scala/meta/tests/parsers/LitSuite.scala
```diff
@@ -5,93 +5,104 @@ import scala.meta._
...
+    assertTree(term("2147483648L"))(Lit.Long(2147483648L))
   }
 
   test("9223372036854775808L") {
     intercept[ParseException] { term("9223372036854775808L") }
   }
 
   test("9223372036854775807L") {
-    val Lit(9223372036854775807L) = term("9223372036854775807L")
+    assertTree(term("9223372036854775807L"))(Lit.Long(9223372036854775807L))
```
kitbellew, line 58: and this works...

tests/shared/src/test/scala/scala/meta/tests/parsers/LitSuite.scala
```diff
@@ -5,93 +5,104 @@ import scala.meta._
...
 
   test("()") {
-    val Lit(()) = term("()")
+    assertTree(term("()"))(Lit.Unit())
   }
 
   test("0xCAFEBABE") {
-    val Lit(-889275714) = term("0xCAFEBABE")
-    val Lit(889275714) = term("-0xCAFEBABE")
+    assertLiteral(term("0xCAFEBABE"), Lit.Int(-889275714))
```
kitbellew, line 98: and this doesn't?

tests/shared/src/test/scala/scala/meta/tests/parsers/LitSuite.scala
```diff
@@ -105,12 +116,15 @@ class LitSuite extends ParseSuite {
   }
 
   test("#1382") {
-    val Lit("\"\"") = term("\"\"\"\"\"\"\"\"")
-    val Lit("\"\"\"\"\"\"\"") = term("\"\"\"\"\"\"\"\"\"\"\"\"\"")
+    assertLiteral(term("\"\"\"\"\"\"\"\""), Lit.String("\"\""))
```
kitbellew, line 119: and this doesn't either?

tests/shared/src/test/scala/scala/meta/tests/parsers/LitSuite.scala
```diff
@@ -5,93 +5,104 @@ import scala.meta._
 import scala.meta.dialects.Scala211
 
 class LitSuite extends ParseSuite {
+
+  // Structure does not always return consisten results
+  private def assertLiteral(obtained: Term, expected: Lit) {
```
kitbellew, line 10: I am not really sure what's happening here. why is regular assertTree not working on 
```
assertLiteral(term("0xCAFEBABE"), Lit.Int(-889275714))
assertLiteral(term("42.42"), Lit.Double(42.42))
assertLiteral(term("42.42f"), Lit.Double(42.42f))
```

kitbellew (changes requested):
one cosmetic change, and a few questions.

kitbellew:
> ... and used Chat GPT to help out rewrite it quicker

i didn't see this 🙂 how does this work? i haven't used it before for code.

tests/shared/src/test/scala/scala/meta/tests/parsers/ParseSuite.scala
tgodzik: Thanks!

tests/shared/src/test/scala/scala/meta/tests/parsers/LitSuite.scala
tgodzik: It works if printing via `.structure` is the same. But if it prints differently yet it's a correct value it will be reported as an error.

tgodzik: Previously, we would do an unapply which only compares values, so that would be correct. When using `Lit.Int(0xCAFEBABE)` the int inside will be printed normally, so we would get a mismatch.

tgodzik:
> > ... and used Chat GPT to help out rewrite it quicker
> 
> i didn't see this 🙂 how does this work? i haven't used it before for code.

I tried changing the cases to manually construct trees, but that proved a bit tedious, so instead I realized it would be quickes if the change was purely syntactic. In ChatGPT I wrote the expected input (asserts to change) and an example of what I want to turn it to and it managed to change all the cases correctly, so things were much faster this way.

I tried it to also generate full trees, but it's not really good with anything semantic in the code. Syntactic stuff it can do perfectly and regexes seemed rather complex here :s

kitbellew (approved):
didn't mean for you to remove the literal assert, was mostly curious.

PR #3371, opened 2023-10-22
avoid deprecated `java.net.URL` constructor

xuwei-k:
https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/net/URL.html#%3Cinit%3E(java.lang.String)


tgodzik (approved):
Thanks!

PR #3373, opened 2023-10-25
Fix FuncParamClause not adding parentheses for Type.Function

hugo-vrijswijk:
Before:

```scala
q"def f: (B => B) => A".syntax
// "def f: B => B => A"
```

After:

```scala
q"def f: (B => B) => A".syntax
// "def f: (B => B) => A"
```

See the added test for more examples. Other cases are welcome as I might have missed something


tgodzik (approved):
LGTM!

PR #3375, opened 2023-10-26
refactor: Don't use pattern match on val in TokenSuite

tgodzik:
This shows a lot of warning when compiling with Scala 3 (I am down to like 80!) and also don't use quasiquotes in Scala3SyntaxSuite as it's not needed and not supported in Scala 3


kitbellew (approved):

kitbellew:
you didn't ask me for a review, but i assumed it was an oversight :)

tgodzik:
> you didn't ask me for a review, but i assumed it was an oversight :)

Och, I was supposed to do it one the CI passes, thanks!

PR #3406, opened 2023-12-14
TokenIterator: replace {current,peek}Indentation

kitbellew:
Instead, add the `indenting` method which specifically validates whether there's an increase in significant indentation at given newline. Helps with scalameta/scalafmt#3720.


tgodzik (approved):
LGTM!

PR #3411, opened 2023-12-16
ScannerTokens: make `isLeadingInfix` compliant

kitbellew:
The implementation was incorrect, didn't check the case when the infix op was on a line by itself.

However, the fix is incomplete as it doesn't handle the very rare case when the infix expression is at the file level and not nested within a block of any kind. This will be fixed in the next commit.

Also, add a correlated fix in `getOutdentIfNeeded`.

Helps with scalameta/scalafmt#3720.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScannerTokens.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScannerTokens.scala
@@ -96,19 +96,6 @@ final class ScannerTokens(val tokens: Tokens)(implicit dialect: Dialect) {
       case _ => true
     })
 
-  def getTokenAtLineStart(idx: Int): Token =
-    getTokenAtLineStart(idx, tokens(idx))
-
-  @tailrec
-  private def getTokenAtLineStart(idx: Int, nextNonSpace: Token): Token =
-    if (idx == 0) nextNonSpace
-    else
-      tokens(idx) match {
-        case _: AtEOL => nextNonSpace
-        case _: HSpace => getTokenAtLineStart(idx - 1, nextNonSpace)
-        case t => getTokenAtLineStart(idx - 1, t)
-      }
-
   val soft = new SoftKeywords(dialect)
 
   object TypeIntro {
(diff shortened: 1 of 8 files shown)


kitbellew:
@tgodzik when you have some time ...

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/FewerBracesSuite.scala
```diff
@@ -990,4 +990,563 @@ class FewerBracesSuite extends BaseDottySuite {
...
+  test("scalafmt #3720 2 complex, break after op, top-level, indented") {
+    val code =
+      """
+        |  foo.bar:
+        |    arg
+        |  ++
+        |  baz:
+        |    arg
+        |  ++
+        |  qux
```
tgodzik, line 1110: Why lost qux in this case?

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/FewerBracesSuite.scala
```diff
@@ -990,4 +990,563 @@ class FewerBracesSuite extends BaseDottySuite {
...
+  test("scalafmt #3720 2 complex, break after op, top-level, not indented") {
+    val code =
+      """
+        |foo.bar:
+        |  arg
+        |++
+        |  baz:
+        |    arg
+        |  ++
+        |  qux
```
tgodzik, line 1083: Here also, `qux` is lost

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScannerTokens.scala
```diff
@@ -929,6 +955,21 @@ object ScannerTokens {
     loop(str.length - 1, 0)
   }
 
+  // return None if single line, or Some(all indents are at least minIndent)
+  private def multilineCommentMinIndent(t: Comment, minIndent: Int): Option[Boolean] = {
```
tgodzik, line 959: Should we add some tests for multiline comments? Should we even accept multiline comments between operators, don't they count as empty line?

tgodzik (commented):
Sorry for the delay!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScannerTokens.scala
kitbellew: 1. we can add tests, yes; i will revise.
2. i used scastie to test out the behaviour; handling of comments is not mentioned on the page i linked on line 712 above, unfortunately.

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/FewerBracesSuite.scala
kitbellew: it's a bug. i mentioned it in the commit (and pr) comment:

_However, the fix is incomplete as it doesn't handle the very rare case when the infix expression is at the file level and not nested within a block of any kind. This will be fixed in the next commit._

tgodzik: Ach, ok. Sorry about that then!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScannerTokens.scala
tgodzik: just a couple of test to make sure we don't regress in the future would be awesome.

kitbellew: done. also, after additional scastie testing, changed how the infix argument is validated.

also, added a new `EOL` token called `InfixLF` so that we don't need to check this arg a second time in `ScalametaParser`.

tgodzik (approved):
LGTM

PR #3412, opened 2023-12-16
fix: don't emit implicit conversion occurrence when qualifier in synthetic apply

kasiaMarek:
https://github.com/scalameta/metals/issues/5939


semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -736,7 +736,7 @@ trait TextDocumentOps { self: SemanticdbOps =>
                   //    for { (_, _) <- foo } yield ()```
                   // where `for` expands to `foo.withFilter(...).map(...)`
                   case g.Apply(s: g.Select, _) if isSyntheticName(s) => traverse(s)
-                  case _ => tryFindMtree(select.qualifier)
+                  case _ => traverse(select.qualifier)
```
kitbellew, line 739: 1. is that it? :)
2. what about this minor change:
```
  case qualifier => traverse(qualifier)
```

kitbellew (approved):

Issue #3409, opened 2023-12-15
Scala211 doesn't support inline modifiers when printing tree with `inline` modifier


majk-p:
**Issue type:** Bug/Inconsistency
**Scala version:** 2.13.12
**Scala meta version:** 4.8.14 (latest)
**Reproduction snippets:** 
- Straightforward: https://gist.github.com/majk-p/f64713be5e910475e2e8c3ea03ee3e77
- In weaver use case: https://gist.github.com/majk-p/cbceb0125b2083e3a10dac6abcfdfb23

**Workaround exists:** true

## Context
I'm using Scala Meta to generate Scala 3 source code and test the code with [weaver](https://github.com/disneystreaming/weaver-test). Here's a simplified version of a test, where `object Codegen` is something that would be defined by my library.

## Problem

```scala
import scala.meta._
import scala.meta.contrib._
import scala.meta.dialects.Scala3
import weaver._

object Codegen {
  def codeWithInline =
    source"""
    object Generated { 
      inline def dummy = scala.compiletime.error("this method shuold not be invoked")
    }
    """

}

object SampleSpec extends FunSuite {

  val result = Codegen.codeWithInline

  test("Should generate code with inline") {
    val expectation = source"""
    object Generated { 
      inline def dummy = scala.compiletime.error("DIFFERENT MESSAGE TO FAIL TEST")
    }
    """
    assert(result.isEqual(expectation))
  }
}
```
**Expected behavior:** Test fails with an error saying that the values are not equal, printing out both sources and highlighting the difference.

**Actual behavior:** Test fails with runtime error: `UnsupportedOperationException: Scala211 doesn't support inline modifiers`. For full stacktrace run `scala-cli test https://gist.github.com/majk-p/cbceb0125b2083e3a10dac6abcfdfb23`.

## Root cause

In the example above weaver finds out that the assertion is failed and tries to print out the code. Unfortunately it attempts to do it using `toString` (rather than using `syntax`) which is defined as `override def toString = internal.prettyprinters.TreeToString(this)`. `TreeToString` then does [select the dialect](https://github.com/scalameta/scalameta/blob/main/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeToString.scala#L13), builds `TreeSyntax` and calls it. 

```scala
object TreeToString {
  def apply(tree: Tree) = {
    val dialect = tree.origin match {
      case x: Origin.Parsed => x.dialect
[...]

PR #3413, opened 2023-12-17
Allow for modification of dialect on generated tree

majk-p:
Address issue with controlling dialect on a generated tree as described in https://github.com/scalameta/scalameta/issues/3409

### DIFF
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/InternalTrees.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/InternalTrees.scala
@@ -72,6 +72,10 @@ trait InternalTree extends Product {
     }
   }
 
+  def withDialect(dialect: Dialect): Tree = {
+    this.privateCopy(origin = Origin.Generated(dialect))
+  }
+
   // ==============================================================
   // Setters for pieces of internal state defined above.
   // Everyone except for scala.meta's core should be using "private"-less versions of these methods,
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeToString.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeToString.scala
@@ -9,6 +9,7 @@ object TreeToString {
   def apply(tree: Tree) = {
     val dialect = tree.origin match {
       case x: Origin.Parsed => x.dialect
+      case x: Origin.Generated => x.dialect
       case Origin.None if tree.isInstanceOf[Quasi] => QuasiquoteTerm(Scala211, multiline = true)
       case Origin.None => Scala211 // this dialect is as good as any as a default
     }
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/Origin.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/Origin.scala
@@ -31,4 +31,7 @@ object Origin {
     }
   }
 
+  @adt.leaf
+  class Generated(dialect: Dialect) extends Origin
+
 }


kitbellew:
how does this help with quasiquotes? they still don't have any origin.

instead of obsolete `TreeToString(tree)`, simply use `TreeSyntax.reprint(tree)`.

tgodzik:
> how does this help with quasiquotes? they still don't have any origin.

I think it would be good to be able to at least to specify a dialect for the tree if it's created manually. There is no option to do that currently.

 Another problem, would be to make the quasiquotes not losing their origin :thinking:

majk-p:
I'd love to use `reprint`, but in my test case I've got no control over it. This is a matter of a default implementation of `override def toString = internal.prettyprinters.TreeToString(this)`. In the attached weaver test case you can see there's no direct call to `toString` but weaver calls it upon failed test to show failure reason: https://gist.github.com/majk-p/cbceb0125b2083e3a10dac6abcfdfb23

With the proposed changes, one can just attach dialect to their source tree and the test no longer fails with the dialect issue:

```scala
object Codegen {
  def codeWithInline =
    source"""
    object Generated { 
      inline def dummy = scala.compiletime.error("this method shuold not be invoked")
    }
    """.withDialect(Scala3)

}

object SampleSpec extends FunSuite {

  val result = Codegen.codeWithInline

  test("Should generate code with inline") {

    val expectation = source"""
    object Generated { 
      inline def dummy = scala.compiletime.error("DIFFERENT MESSAGE TO FAIL TEST")
    }
    """.withDialect(Scala3)

    assert(result.isEqual(expectation))
  }
}
```

I agree the solution is far from perfect, thus it's just a draft. I still prefer the `expect.eql` workaround (described in the initial issue) as it's much more informational.

kitbellew:
> > how does this help with quasiquotes? they still don't have any origin.
> 
> I think it would be good to be able to at least to specify a dialect for the tree if it's created manually. There is no option to do that currently.
> 
> Another problem, would be to make the quasiquotes not losing their origin 🤔

that would be best. do you understand how a quasiquote-compiled object is persisted? why do the fields of the tree remain but not the origin?

tgodzik:
I am not sure, I know that when mapping the tree the origin is dropped since the Tree itself no longer corresponds to it. Would could instead use the Generated one here.

scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/InternalTrees.scala
```diff
@@ -72,6 +72,10 @@ trait InternalTree extends Product {
     }
   }
 
+  def withDialect(dialect: Dialect): Tree = {
+    this.privateCopy(origin = Origin.Generated(dialect))
```
kitbellew, line 76: even if we wanted to do this, two points:
- the new `Origin` class is not really generated, it's a `DialectOnly`
- this particular method should only modify the origin if it's `Origin.None`

majk-p: One other problem I see with this implementation is that it always returns the `Tree` type instead of sticking to the sub-type.

majk-p:
> I think it would be good to be able to at least to specify a dialect for the tree if it's created manually. There is no option to do that currently.

@tgodzik @kitbellew do you think it's valuable to iterate on this approach?

Do you see other use cases assuming there's a way to prevent quasiquotes from losing their origin?

kitbellew:
> Do you see other use cases assuming there's a way to prevent quasiquotes from losing their origin?

not really.

another possibility is to derive the dialect from the tree itself; if it uses `Mod.Inline`, it's likely some version of scala3. however, then it's unclear whether to use `args: _*` vs `args*`, or `[_]` vs `[*]`.

tgodzik:
> do you think it's valuable to iterate on this approach?

I think it's valuable for at least a case where we construct the tree manually. We can do instead of `Generated` `DialectOnly` like suggested to be more explicit.

tgodzik:
As for quasiquotes loosing their origin, this is most likely a bug somewhere, but I couldn't find it in a reasonable time.

kitbellew:
@majk-p @tgodzik why don't we postpone the decision until we figure out what's happening with origin?

meanwhile, i have found a solution which seems to preserve origin but it breaks when quasiquotes are used for matching (the pattern doesn't contain the origin, naturally, but the matched tree does).

kitbellew:
@majk-p @tgodzik looks like I may have figured it out, opened #3425 and will be submitting a sequence of PRs.

kitbellew:
@majk-p do you believe you still need this, given the recent changes? qq now preserve full origin if they have no interpolations, or the dialect-only origin otherwise; in both cases, the dialect is preserved.

majk-p:
I've just tested the snippet against the master ([this commit](https://github.com/scalameta/scalameta/commit/2059e3645ac8523fa93ca477877368678eab9a5c) to be specific). Unfortunately it seems like regression. What on 4.8.15 outputs:
```scala
$ scala-cli test snippet.test.scala
SampleSpec                                                                                                             
- Should generate code with inline 97ms                                                                                
*************FAILURES**************                                                                                    
SampleSpec                                                                                                             
- Should generate code with inline 97ms                                                                                
  UnsupportedOperationException: Scala211 doesn't support inline modifiers
                                                           
  TreeSyntax.scala:989           scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances#$anonfun$syntaxTree$1
  Syntax.scala:9                 scala.meta.prettyprinters.Syntax$$anon$1#apply
  Show.scala:81                  scala.meta.prettyprinters.Show$#$anonfun$repeat$1
  List.scala:246                 scala.collection.immutable.List#map
  List.scala:79                  scala.collection.immutable.List#map
  Show.scala:81                  scala.meta.prettyprinters.Show$#repeat
```

On `4.8.15-SNAPSHOT` (built with `publishLocal`) crashes with:
```scala
$ scala-cli test snippet.test.scala
Compiling project (test, Scala 2.13.12, JVM (17))
[error] ./snippet.test.scala:13:5
[error] exception during macro expansion: 
[error] java.lang.ClassNotFoundException: scala.meta.shaded.internal.sourcecode.Text
[error]         at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:445)
[error]         at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:592)
[error]         at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:525)
[error]         at scala.meta.Dialect$.standardPairs$lzycompute(Dialect.scala:809)
[error]         at scala.meta.Dialect$.standardPairs(Dialect.scala:809)
[error]         at scala.meta.Dialect$.standards$lzycompute(Dialect.scala:834)
[error]         at scala.meta.Dialect$.standards(Dialect.scala:833)
[error]         at scala.meta.internal.quasiquotes.ReificationMacros.instantiateStandardDialect$1(ReificationMacros.scala:148)
[error]         at scala.meta.internal.quasiquotes.ReificationMacros.instantiateDialect(ReificationMacros.scala:152)
[error]         at scala.meta.internal.quasiquotes.ReificationMacros.expand(ReificationMacros.scala:57)
[error]         at scala.meta.internal.quasiquotes.ReificationMacros.apply(ReificationMacros.scala:53)
[error] ./snippet.test.scala:26:23
[error] exception during macro expansion: 
[error] java.lang.ClassNotFoundException: scala.meta.shaded.internal.sourcecode.Text
[error]         at scala.meta.Dialect$.standardPairs$lzycompute(Dialect.scala:809)
[error]         at scala.meta.Dialect$.standardPairs(Dialect.scala:809)
[error]         at scala.meta.Dialect$.standards$lzycompute(Dialect.scala:834)
[error]         at scala.meta.Dialect$.standards(Dialect.scala:833)
[error]         at scala.meta.internal.quasiquotes.ReificationMacros.instantiateStandardDialect$1(ReificationMacros.scala:148)
[error]         at scala.meta.internal.quasiquotes.ReificationMacros.instantiateDialect(ReificationMacros.scala:152)
[error]         at scala.meta.internal.quasiquotes.ReificationMacros.expand(ReificationMacros.scala:57)
[error]         at scala.meta.internal.quasiquotes.ReificationMacros.apply(ReificationMacros.scala:53)

```
It's the [same snippet](https://gist.github.com/majk-p/cbceb0125b2083e3a10dac6abcfdfb23) as above

tgodzik:
That's weird, looks like shading didn't produce any artifact for sourced library :thinking: 

I wonder if worked without my change if you revert  https://github.com/scalameta/scalameta/pull/3468/files ?

majk-p:
Just switched off my laptop, can do it tomorrow. 

By the way what do you think about including this kind of test as a regression suite?

tgodzik:
> Just switched off my laptop, can do it tomorrow.
> 
> By the way what do you think about including this kind of test as a regression suite?

We could have a separate job that would release locally and just run a couple of Scala CLI snippets sure, though this should really be caught in the tests. However, if this is a problem with release we need a way to test that and maybe this would be simplest?

kitbellew:
@majk-p i can't reproduce the problem. i built the most recent `main` commit using `sbt publishLocal`, then modified the local build of `scalafmt` to use it, added your snippet as a formatting test (which means it has to be parsed first), and it all worked.

zainab-ali:
I see the same resuts as @majk-p when testing on the [same commit to main](https://github.com/scalameta/scalameta/commit/2059e3645ac8523fa93ca477877368678eab9a5c). I also checked this [commit](https://github.com/scalameta/scalameta/commit/e48e3cce328a9a9d312470d6dcd63582933a8c0c) prior to the shading bug fix in  https://github.com/scalameta/scalameta/pull/3468/files and got the same error:
```
[error] exception during macro expansion:
[error] java.lang.ClassNotFoundException: scala.meta.shaded.internal.sourcecode.Text
```

This error isn't present in `4.8.15`, so is most likely a regression since then.

kitbellew:
@majk-p @zainab-ali is it possible that it's specific to scala-cli integration somehow?

majk-p:
There's no scala-cli integration. It's just a code snippet from the initial issue: https://gist.github.com/majk-p/cbceb0125b2083e3a10dac6abcfdfb23 which I ran against the latest `main` and `4.8.15` to compare

```scala
//> using scala "2"
//> using dep "org.scalameta::scalameta:4.8.14" <- Version changed here
//> using test.dep "com.disneystreaming::weaver-cats:0.8.3"
//> using testFramework "weaver.framework.CatsEffect"

import scala.meta._
import scala.meta.contrib._
import weaver._

object Codegen {
  def codeWithInline =
    source"""
    object Generated { 
      inline def dummy = scala.compiletime.error("this method shuold not be invoked")
    }
    """

}

object SampleSpec extends FunSuite {

  val result = Codegen.codeWithInline

  test("Should generate code with inline") {
    val expectation = source"""
    object Generated { 
      inline def dummy = scala.compiletime.error("DIFFERENT MESSAGE TO FAIL TEST")
    }
    """
    assert(result.isEqual(expectation))
  }
}
```

kitbellew:
@tgodzik with the sourcecode commit reverted, the problem is gone. very strange. wonder if geny/fastparse are correctly shaded.

kitbellew:
@majk-p @zainab-ali can you please try your test again, by checking out #3482? in my local testing, the problem seemed to go away.

kitbellew:
never mind :( tests fail.

kitbellew:
> @majk-p @zainab-ali can you please try your test again, by checking out #3482? in my local testing, the problem seemed to go away.

same pr, different solution. please try it now.

majk-p:
Thank you @kitbellew, I think we are on the safe side now! The code doesn't crash anymore:

```scala
SampleSpec
- Should generate code with inline 78ms
*************FAILURES**************
SampleSpec
- Should generate code with inline 78ms
  assertion failed (../../../../Code/personal/scalameta-inline-codegen/snippet.test.scala:34)

  assert(result.isEqual(expectation))
                |      ||
                false  |...
                       scala.meta.contrib.equality.Structurally$$anon$1@77cf140a
```

This is tested against latest main since https://github.com/scalameta/scalameta/pull/3482 has been merged. With that I think we are good to close this PR. The last outstanding thing to do would be to add tests for shading as @tgodzik mentioned, but this is for another PR.

Thank you to all involved!

PR #3422, opened 2023-12-28
ScannerTokens: introduce a line-indent region

kitbellew:
Use it when an LF token is emitted, to preserve the indent. It is useful in order to identify when to outdent a fewer-braces region, especially when leading-infix operators are used with smaller indentation than the significant-indentation region they belong to but larger than the outer.

Helps with scalameta/scalafmt#3720.


tgodzik (approved):
:rocket:

Issue #3425, opened 2024-01-01
quasiquotes: preserve origin or dialect if possible


kitbellew:

bjaglin:
FTR, the discussion in https://github.com/scalameta/scalameta/pull/3413 helped me getting the rationale behind this effort

kitbellew:
@bjaglin we are probably ready to release but before that, wanted to request for you to check the current `main` against scalafix, to see if there are any regressions. 

looks like the most recent version published is `4.8.15+74-aed70110-SNAPSHOT` (https://github.com/scalameta/scalameta/actions/runs/7480387396/job/20359694450)

kitbellew:
@bjaglin also, wanted to get your opinion: do these changes merit a 4.9.0, or should we continue to 4.8.16?

bjaglin:
Scalafix tests look good :+1: 

CI shows that mdoc will need to be republished as it uses [internals](https://github.com/search?q=repo%3Ascalameta%2Fmdoc%20enquote&type=code) that have been [removed](https://github.com/scalameta/scalameta/pull/3469/files).

Quasiquotes macros compiled against scalameta master won't link against existing versions because expanded most likely references `scala.meta.trees.Origin` directly (although we haven't confirmed it in https://github.com/scalameta/scalameta/pull/3454), which is not available in existing versions. So I would suggest 4.9.0 to indicate that, although it's not clear how we should treat macro compatibility... In any case, if you decide to go for 4.8.16, that's fine - I'll check myself if I see references to `scala.meta.trees.Origin` and increment the minor in Scalafix if that's the case.

kitbellew:
@tgodzik should i release at will, or would you like to test it against metals or another repo first?

tgodzik:
> @tgodzik should i release at will, or would you like to test it against metals or another repo first?

Good idea, will run it against the nwest snpashot

tgodzik:
Testing it here https://github.com/scalameta/metals/pull/6014

PR #3449, opened 2024-01-02
adt.Reflection: allow private fields

kitbellew:
Handling of them will be added later. Helps with #3425.


scalameta/common/shared/src/main/scala/org/scalameta/adt/Reflection.scala
```diff
@@ -114,12 +121,14 @@ trait Reflection {
...
-    def fields: List[Field] = allFields.filter(_.sym.isPayloadField)
+    def fields: List[Field] = fields(false)
+    def fields(isPrivateOK: Boolean): List[Field] =
+      allFields.filter(_.sym.isPayloadField(isPrivateOK))
     def allFields: List[Field] = sym.allFields.map(_.asField)
     override def toString = s"leaf $prefix"
   }
   class Field(val sym: Symbol) {
-    if (!sym.isField)
+    if (!sym.isField(true))
```
tgodzik, line 131:
```suggestion
    if (!sym.isField(isPrivateOK = true))
```

tgodzik (approved):
Super minor.

PR #3454, opened 2024-01-02
Origin: move out of `internal`

kitbellew:
As origin is now saved as part of quasiquotes, let's make sure it abides by MIMA standards (which are not currently applied to `internal` code).

Similarly, remove TokenStreamPosition and inline in Origin.Parsed. Helps with #3425.

### DIFF
--- a/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/NamerMacros.scala
+++ b/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/NamerMacros.scala
@@ -24,8 +24,8 @@ trait CommonNamerMacros extends MacroHelpers {
   lazy val PositionModule = q"_root_.scala.meta.inputs.Position"
   lazy val PointClass = tq"_root_.scala.meta.inputs.Point"
   lazy val PointModule = q"_root_.scala.meta.inputs.Point"
-  lazy val OriginClass = tq"_root_.scala.meta.internal.trees.Origin"
-  lazy val OriginModule = q"_root_.scala.meta.internal.trees.Origin"
+  lazy val OriginClass = tq"_root_.scala.meta.trees.Origin"
+  lazy val OriginModule = q"_root_.scala.meta.trees.Origin"
 
   def mkClassifier(name: TypeName): List[Tree] = {
     val q"..$classifierBoilerplate" = q"""
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/Pos.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/Pos.scala
@@ -1,8 +1,8 @@
 package scala.meta.internal.parsers
 
 import scala.meta.Tree
-import scala.meta.internal.trees.Origin
 import scala.meta.prettyprinters._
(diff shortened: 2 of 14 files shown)


kitbellew:
@tgodzik @bjaglin i am having second thoughts about this. is my understanding even true? if origin is not a public field, does it really have to abide by any interface when quasiquotes are involved?

tgodzik:
I think we use Origin in quite a few places outside of scalameta so it makes sense to move it out of internal.

bjaglin (approved):
> is my understanding even true? if origin is not a public field, does it really have to abide by any interface when quasiquotes are involved?

Yes, I believe so - even though the field is not public, its type needs to be MiMa-protected if it's referenced in macro expansions, to ensure that in the future, quasiquotes compiled in a previous semver-compatible version can be linked.

The end goal of https://github.com/scalameta/scalameta/issues/3425 and what exactly should be exposed goes beyond my understanding/knowledge, but this PR is definitely a needed follow-up to https://github.com/scalameta/scalameta/pull/3450 to avoid the same challenges we discussed a while ago on [Discord](https://discord.com/channels/632642981228314653/632642981228314657/1069264160044634143):
> I had a quick look with -Vprint:typer: 4.7.x quasiquotes expand to internal.Impl.unapply (since https://github.com/scalameta/scalameta/commit/f8799846020eeeb95518229598c33cae3270f730). Independently of our discussion, that means that there is a risk of breaking backward compatibility for quasiquotes in the future as MiMa does not run on extractors referenced by the expanded code.

bjaglin:
(I see there is very active progress on this massive effort, thanks a lot for keeping backward compatibility in mind despite the complexity!)

kitbellew:
@bjaglin thanks for chiming in. a further question:
- `internal.Impl.unapply` was problematic as that was the final result of this expansion;
- however, origin is "hidden" somewhere inside the tree and not directly exposed
  - after some thought, it looks to me no different from any other `internal` type, at least as far as it is used in the construction of a scalameta tree, so either all of them are bad or none are...
  - for instance, `InternalTree` is a base class of `Tree`, and yet it's  under`internal` package and hence outside of Mima
  - similarly, `.syntax` is implemented using `TreeSyntax` and that is also `internal`; so if you have code which uses this method, is it at risk if TreeSyntax changes?

bjaglin:
I see your point, it's hard to tell what's really contractual at macro expansion time... Did you have a look at code generated by the quasiquotes macros with `-Vprint:typer`? I doubt that there is anything `internal` at the moment, but if there is, then I agree about your statement "either all of them are bad or none are".

kitbellew:
> ... Did you have a look at code generated by the quasiquotes macros with `-Vprint:typer`?

i don't really know how to do that, truth be told... 😊

bjaglin:
There is probably an easier way to do it, but what I did last time was to add `-Vprint:typer` to `scalacOptions` in a test module, run a full compile (ignoring the verbose output), touch a file exercising a macro by adding some dummy code, and check the output of that incremental compilation hunting for references.

And ... now that I write this, I realize `javap -p -v` on class files generated from macro-expanding sources is probably even easier to identify what really gets emitted and thus should be backward compatible. So that would be my recommendation!

PR #3465, opened 2024-01-05
TreeSyntax: fix bug in interpolate syntax

kitbellew:
Helps with #3425.

### DIFF
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
@@ -363,21 +363,14 @@ object TreeSyntax {
           case _ => m(Path, s(p(SimpleExpr, t.qual), if (guessIsPostfix(t)) " " else ".", t.name))
         }
       case t: Term.Interpolate =>
-        def needBraces(id: String, charAfter: Option[Char]): Boolean = {
-          val startsWithUnderscore = id.startsWith("_")
-          startsWithUnderscore || {
-            val isOpId = isOperatorPart(id.head)
-            val charAfterIsOp = charAfter.exists(isOperatorPart)
-            val charAfterIsIdPart =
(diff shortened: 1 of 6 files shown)


scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
```diff
@@ -363,21 +363,15 @@ object TreeSyntax {
...
-          startsWithUnderscore || {
-            val isOpId = isOperatorPart(id.head)
-            val charAfterIsOp = charAfter.exists(isOperatorPart)
-            val charAfterIsIdPart =
-              charAfter.exists(chr => isNameStart(chr) || Character.isDigit(chr))
-            if (isOpId) charAfterIsOp else charAfterIsIdPart
-          }
+        def needBraces(id: String, nextPart: String): Boolean = {
+          val idHead = id.head
+          id == '_' || !Character.isUnicodeIdentifierStart(idHead) ||
```
tgodzik, line 368:
```suggestion
          idHead == '_' || !Character.isUnicodeIdentifierStart(idHead) ||
```
otherwise it's always false, no?

tgodzik (changes requested):

scalameta/trees/shared/src/main/scala/scala/meta/internal/prettyprinters/TreeSyntax.scala
kitbellew: good eye! apparently, `_` is not a valid unicode identifier start:
Screenshot 2024-01-05 at 09 21 28

but that presents an opportunity to simplify further. i have also found another peculiarity, so minor changes in the next push.

kitbellew: done, please take a look

tgodzik (approved):
LGTM!

PR #3481, opened 2024-01-09
Tree: remove privateOrigin, ensure origin is valid

kitbellew:
We could do this now as origin is now a persisted field and could be declared public.


tgodzik (approved):

kitbellew:
@tgodzik thank you for approval, i will merge this later, checking something in the meantime.

PR #3484, opened 2024-01-13
Trees: save tokens and positions as lazy

kitbellew:
Need to make Parsed.position lazy as well, otherwise its own lazy vals would not persist.


tgodzik (approved):

xuwei-k:
- 🤔 https://github.com/scalameta/scalameta/issues/3650

PR #3543, opened 2024-02-06
add binary literals support

kasiaMarek:
SIP: https://docs.scala-lang.org/sips/binary-integer-literals.html
Scala 2: https://github.com/scala/scala/pull/10656
Scala 3: https://github.com/lampepfl/dotty/pull/19551

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@ The exact test command to run tests depends on what you are working on:
 - `testsJVM/slow:test`: run slow integration tests for semantic APIs.
 - `testsJVM/all:test`: run all tests, both fast and slow.
 
-Tips to make your edit/test/debug worflow as productive as possible:
+Tips to make your edit/test/debug workflow as productive as possible:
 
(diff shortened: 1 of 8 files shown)


tgodzik (approved):
Lgtm! @kitbellew do you see any issues? Should be a noop change for scalafmt

kitbellew:
i looked at the two threads on scala repo, it seems that the feature is only in 2.13 but not in dotty (and so definitely not in 3.0-2 but could be added later to whatever next milestone is).

should we disable this flag in scala30 dialect (which likely derives from scala213) and perhaps reenable in scala3future?

kasiaMarek:
> should we disable this flag in scala30 dialect (which likely derives from scala213) and perhaps reenable in scala3future?
I think that is what I did.

Kordyjan:
It is worth noting that this is not something that is present in any released version of Scala 3. It will become part of the language in 3.5.

tgodzik:
It's currently in the Scala3Future dialect, so this should be safe

PR #3561, opened 2024-02-14
SBT: fail build on warnings, treat some as info

kitbellew:
Now that most warnings have been fixed, let's make the build stricter.

Limit these extra checks to scala-2.13 only, as that's sufficient to ensure clean code and avoids running into tooling differences between versions.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -556,6 +556,10 @@ lazy val sharedSettings = Def.settings(
     if (isScala213.value) List("-Ymacro-annotations")
     else Nil
   },
+  scalacOptions ++= {
+    if (isScala213.value) List("-Xfatal-warnings", "-Wconf:cat=deprecation:is")
+    else Nil
+  },
   scalacOptions ++= Seq("-feature", "-unchecked"),
   Compile / doc / scalacOptions ++= Seq("-skip-packages", ""),
   Compile / doc / scalacOptions ++= Seq("-implicits", "-implicits-hide:."),


build.sbt
```diff
@@ -556,6 +556,14 @@ lazy val sharedSettings = Def.settings(
     if (isScala213.value) List("-Ymacro-annotations")
     else Nil
   },
+  scalacOptions ++= {
+    if (isScala213.value)
+      List(
+        "-Xfatal-warnings",
+        "-Wconf:cat=deprecation:is,cat=other-match-analysis:i"
```
kitbellew, line 563: @tgodzik @bjaglin perhaps you have some ideas; the `match-analysis` entry is only for "unreachable code" warnings coming from something in https://github.com/scalameta/scalameta/blob/main/scalameta/common/shared/src/main/scala/scala/meta/internal/transversers/transverser.scala (or the two adjacent implementations, but since both of them show that message, i assumed it was the parent).

since this is a macro, the message doesn't really help in figuring out which actual line of code leads to this. if you have some idea on how to find it, we could fix the problem -- and remove this extra warning exclusion.

kitbellew: i figured it out: #3562.

kitbellew:
now it's all clean, no warnings other than deprecations. ptal.

tgodzik (approved):
LGTM! Thanks for adding those

bjaglin:
Thanks @kitbellew !

PR #3572, opened 2024-02-20
doc: quasiquote now preserve comments

unarist:
Looks like quasiquotes now preserve comments in the source string.

```
>scala-cli repl --scala 2 --dep org.scalameta::scalameta:4.9.0

scala> import scala.meta._
import scala.meta._

scala> println(q"function  (    argument   ) // comment")
function  (    argument   ) // comment
```

### DIFF
--- a/docs/trees/guide.md
+++ b/docs/trees/guide.md
@@ -256,14 +256,6 @@ println(
 )
 ```
 
-> It's important to keep in mind that quasiquotes expand at compile-time into
-> the same program as if you had written normal constructors by hand. This means
-> for example that formatting details or comments are not preserved
-
-```scala mdoc
-println(q"function  (    argument   ) // comment")
-```
-
 Quasiquotes can be composed together like normal string interpolators with
 dollar splices `$`
 
@@ -271,6 +263,16 @@ dollar splices `$`
 val left = q"Left()"
 val right = q"Right()"
 println(q"$left + $right")
+```
+
+> It's important to keep in mind that quasiquotes expand at compile-time into
+> as if you had written normal constructors by hand. This means for example
+> that formatting details or comments are not preserved if interpolated values
+> are used, because the complete source is not available at compile-time.
+
+```scala mdoc
+val argument = 1
+println(q"function  (    $argument   ) // comment")
 ```
 
 A list of trees can be inserted into a quasiquote with double dots `..$`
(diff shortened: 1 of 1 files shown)


docs/trees/guide.md
```diff
@@ -256,14 +256,6 @@ println(
 )
 ```
 
-> It's important to keep in mind that quasiquotes expand at compile-time into
```
kitbellew, line 259: this sentence needs to be modified rather than removed. the original source is only preserved if there are no interpolated values, otherwise it will be reconstructed from the generated tree.

kitbellew (changes requested):
see comment earlier, about modifying the message.

docs/trees/guide.md
unarist: Thanks, I have updated the message, and moved it after explaining interpolations. How about this?
https://github.com/scalameta/scalameta/pull/3572/commits/207ba9c19ea0c323df33aacd9e56217be1ca841e

docs/trees/guide.md
```diff
@@ -256,14 +256,6 @@ println(
...
-> the same program as if you had written normal constructors by hand. This means
```
kitbellew, line 260: you lost `the same program` when copying.

docs/trees/guide.md
```diff
@@ -273,6 +265,16 @@ val right = q"Right()"
 println(q"$left + $right")
 ```
 
+> It's important to keep in mind that quasiquotes expand at compile-time into
+> as if you had written normal constructors by hand. This means for example
```
kitbellew, line 269: could you please add commas on either side of `for example`? i know they weren't there before :)

kitbellew (changes requested):

docs/trees/guide.md
unarist: fixed! https://github.com/scalameta/scalameta/pull/3572/commits/28f654f220623965e0ebb431c0bd8edc0b3fd3b7

docs/trees/guide.md
unarist: fixed! https://github.com/scalameta/scalameta/pull/3572/commits/28f654f220623965e0ebb431c0bd8edc0b3fd3b7

kitbellew (approved):

PR #3579, opened 2024-02-22
Fix invalid check: empty or nested Member.Tuple

kitbellew:
The new version of scala 2.13 has finally discovered these issues, so let's fix them before adding support for that version.

First, we were using a parent check whereas we should have used a field check.

Second, the pattern was incorrect and didn't match the case of a single nested single-element tuple.

Finally, the only case that would have matched if it had been used as a field check was unnecessary, since the `@nonEmpty` notation covered it.


tgodzik (approved):
LGTM

PR #3580, opened 2024-02-22
Fix invalid pattern: quasiquote handling of Apply

kitbellew:
The new version of scala 2.13 has finally discovered these issues, so let's fix them before adding support for that version.

In this particular case, there was an invalid pattern that wouldn't have matched anything as written, but even if we consider the intent of this pattern, it likely came before changes in scalameta parser which re-interpreted a `foo(...$bar)` tree.

That is, this code, which would have been parsed originally as `Apply(foo, ArgClause(List(Term.Quasi(rank=2))))`, would now become, through special `Syntactic.TermApply.ArgList`, `Apply(foo, ArgClause.Quasi(rank=1))`.

Hence, the `List(...)` would always give way to lower-rank quasi of an `ArgClause`, hence this pattern is no longer necessary.


tgodzik (approved):
LGTM! Thanks!

PR #3584, opened 2024-02-23
ScannerTokens: improve handling within `case` body

kitbellew:
- Remove wrong LF check for `case`
- Detect `case class` in `case` body
Fixes scalameta/scalafmt#3790. Fixes scalameta/scalafmt#3789.


tgodzik (approved):
LGTM!

Issue #3650, opened 2024-03-16
OutOfMemoryError since 4.9.0


xuwei-k:
https://github.com/xuwei-k/scalameta-memory-issue-3650/commit/765dccba2154d1bf90d9c8281ec3e2b7aaa5fa04

reproduce code

```scala
package example

import scala.meta._
import scala.meta.inputs.Input
import scala.meta.parsers.Parse
import scala.meta.prettyprinters._
import scala.meta.transversers._
import scala.jdk.StreamConverters._
import java.io.File

object Main {
  def main(args: Array[String]): Unit = {
    // parse many scala code
    // e.g. https://github.com/scala/scala 2.13.x branch
    java.nio.file.Files.find(
      new File("scala", "src").toPath,
      100,
      (p, _) => p.toFile.getName.endsWith(".scala") && (p.toFile.getName != "Singleton.scala")
    ).toScala(List).map { f =>
      val tree = implicitly[Parse[Source]].apply(Input.File(f), scala.meta.dialects.Scala213Source3).get
      tree.traverse { a => () }
      tree
    }
  }
}
```

VisualVM results in my local Mac

- https://visualvm.github.io/

### 4.8.15

<img width="1142" alt="meta-4-8-img-1" src="https://github.com/scalameta/scalameta/assets/389787/434bc1e8-06a4-4837-8380-e114cbd18603">

<img width="1145" alt="meta-4-8-img-2" src="https://github.com/scalameta/scalameta/assets/389787/356faea2-4a37-4e2e-96cc-136a19ce9001">

### 4.9.0

<img width="1141" alt="meta-4-9-img-1" src="https://github.com/scalameta/scalameta/assets/389787/6c5498e7-c151-4f54-8bab-c0266705a17c">

<img width="1146" alt="meta-4-9-img-2" src="https://github.com/scalameta/scalameta/assets/389787/7a6a1530-d0ec-4526-b6a9-25e379f4c5f7">

## GitHub Actions result

- https://github.com/xuwei-k/scalameta-memory-issue-3650/actions/runs/8305399108

### 4.8.15

```
[info] running (fork) example.Main 
[success] Total time: 18 s, completed Mar 16, 2024, 4:43:48 AM
```

### 4.9.0

```
[info] running (fork) example.Main 
[error] Exception in thread "main" java.lang.OutOfMemoryError: Java heap space
```

## note

[...]

PR #3651, opened 2024-03-16
fix OutOfMemoryError

xuwei-k:
- fix https://github.com/scalameta/scalameta/issues/3650


kitbellew:
@xuwei-k thank you for detecting the problem and proposing a fix.

do you think #3652 will also solve the problem? it uses the fact that Input overall has a cache already, and tokenizers use it:
https://github.com/scalameta/scalameta/blob/eeb1ea6e62aea54dee4cab42bd91849704827004/scalameta/inputs/shared/src/main/scala/scala/meta/internal/inputs/InternalInput.scala#L13
https://github.com/scalameta/scalameta/blob/eeb1ea6e62aea54dee4cab42bd91849704827004/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala#L15

PR #3655, opened 2024-03-16
LegacyTokenData: relax float check, max as decimal

### LINKED ISSUE #3654: When I define a Float with value 3.4028235e38f and run scalafmt, an error occurs.
Hi,

When I define a Float with value 3.4028235e38f and run scalafmt, an error occurs.

Running compile or run does not result in an error.
Is this a bug?

scala version: 3.3.3
scalafmt version: 3.8.0

```
version = 3.8.0
runner.dialect = scala3 // Occurs in others such as scala213 and scala212
```

```shell
scalafmt src/main/scala/Main.scala
```

## Steps

Given code like this:
```scala
@main def main(): Unit =
  println(3.4028235e38f)
```

## Problem

Scalafmt reports errors like this.

```scala
error: [dialect scala3] floating point number too large
  println(3.4028235e38f)
```

## Expectation

No errors

## Workaround

Define in scala using Java methods.

```scala
java.lang.Float.intBitsToFloat(2139095039)
```

## Notes

Is this the same kind of problem reported in the old Scala bug?

https://github.com/scala/bug/issues/10148

kitbellew:
Fixes #3654.


tgodzik (approved):
LGTM!

PR #3665, opened 2024-03-22
Replace scalafiddle reference with scastie

sake92:
Seems like the good-old (dead now) ScalaFiddle redirects to https://bee-line.dk/ for some reason...  

https://scalameta.org/docs/trees/scalafiddle.html shows an iframe of this site: https://bee-line.dk/ :sweat: 

This PR replaces it with a Scastie playground.


tgodzik (approved):
Thanks! Good idea!

kitbellew:
should we do the same for scalafmt? https://scalameta.org/scalafmt/docs/configuration.html#aligntokens

sake92:
> should we do the same for scalafmt? https://scalameta.org/scalafmt/docs/configuration.html#aligntokens

https://github.com/scalameta/scalafmt/pull/3843

PR #3710, opened 2024-04-28
Dialects: add explicit `allowQuietSyntax` flag

kitbellew:
Previously, this logic was unlocked with `allowSignificantIndentation` implicitly. Use `.allowQuietSyntax` for `if-then`/`while-do`. Helps with scalameta/scalafmt#3941.


tgodzik (approved):
LGTM!

PR #3715, opened 2024-05-18
bug: Add failing test

tgodzik:


tests/shared/src/test/scala/scala/meta/tests/parsers/TermSuite.scala
```diff
@@ -1436,4 +1436,26 @@ class TermSuite extends ParseSuite {
     runTestAssert[Term](code, layout)(tree)
   }
 
+  test("issue-metals") {
+    runTestAssert(
+      """|
+         | for (thread <- all
+         |        if cond1 
+         |           && thread.cond) {
```
kitbellew, line 1444: this is parsed in scala3 as leading infix, but in scala2 just a newline.

why is it treated here differently from:
https://github.com/scalameta/scalameta/blob/ba4e4c887c181092357d9e69236b15545a46c512/tests/shared/src/test/scala/scala/meta/tests/parsers/TermSuite.scala#L1375

PR #3727, opened 2024-06-07
SIP-62 Better fors - syntax support

KacperFKorban:
This PR attempts to add support for SIP-62 syntax changes.
I'm not sure what the policy for it is, since some language extensions are not supported here e.g. `namedTuples`, `into`, `modularity`. But, when set to false, this one doesn't break anything, so maybe there's no harm in adding it?

SIP implementation: https://github.com/scala/scala3/pull/20522

The syntax changes are pretty simple: if `betterFors` is enabled, a `for` comprehension can start with a sequence of "aliases" before the first generator.


kitbellew:
@KacperFKorban thank you for this submission, good idea, but first...

@tgodzik i am leaning towards not introducing a new dialect flag; we need the flags when parsing is possibly ambiguous but i don't know if we should be as strict as the compiler; if someone wrote this code and the compiler is happy, we should be, too. wdyt?

KacperFKorban:
@kitbellew Should I change this PR, so that the experimental syntax is always accepted?

kitbellew:
@KacperFKorban the pr i submitted, mentioned above, does it already.

tgodzik:
Thanks @KacperFKorban ! The change is merge in #3730

PR #3728, opened 2024-06-07
fix: allow for type apply after splice

kasiaMarek:
resolves: https://github.com/scalameta/metals/issues/6483


kasiaMarek:
> nice and beautiful this time, thank you!

Initially I wanted it to also cover:
```Scala
def fn[A](a: A) = ???
{
  [B] => (c: B) => fn[B](c)
}[Int](0)
```
Since this snippet compiles but I don't think it's ever used and worth the hassle.

PR #3730, opened 2024-06-10
ScalametaParser: allow Enum.Val before generator

kitbellew:
This is described in SIP-62.


kitbellew:
superseeds #3727.

tgodzik (approved):
LGTM!

KacperFKorban:
I'm a bit late to the party, but I think that these changes are incorrect, which is weird since it looks like @kitbellew literally copied over the test cases from my PR. Is this intentional?

The following example should not be accepted under any compiler flag or option:

```scala
for {
  a = 1
  if b > 0
  b <- Some(2)
} yield a + b
```

kitbellew:
@KacperFKorban i know it should not, but that's my point: i don't know if scalameta should really be in the business of policing syntax as long as it's not ambiguous.

i kept the check for guard not being first, and didn't introduce this check partly because (for some reason) there was a test case with generator followed by val and then followed by guard.

there are a lot of cases which we don't detect, such as mismatched levels of significant indentation.

KacperFKorban:
@kitbellew Ok, fair enough. Just making sure that it is intentional and also transparent with @tgodzik

PR #3744, opened 2024-06-23
Move protobuf code into semanticdb, from scalameta

### LINKED ISSUE #3724: Move scalapb to semanticdb
I'm trying to write my own scalapb plugin and cross compiling for Scala 3.

I wanted to use scalameta for some of the code generation (sadly not all as there is [no support for comments](https://github.com/scalameta/scalameta/issues/3372)).

I have:
```
("org.scalameta" %%% "scalameta" % "4.9.5").cross(CrossVersion.for3Use2_13)
```

But the problem is that my code-gen project also depends on `scalapb-runtime_3`.

I end up with:
```
[error] Modules were resolved with conflicting cross-version suffixes in ProjectRef(uri("file:/Users/jason/src/kyo/"), "kyo-grpc-code-gen"):
[error]    org.scala-lang.modules:scala-collection-compat _3, _2.13
[error]    com.thesamet.scalapb:lenses _3, _2.13
[error]    com.thesamet.scalapb:scalapb-runtime _3, _2.13
```

It looks like `scalapb-runtime_2.13` is getting pulled in by `org.scalameta:common_2.13`:
```
sbt:kyoJVM> kyo-grpc-code-gen/whatDependsOn com.thesamet.scalapb lenses_2.13
[info] com.thesamet.scalapb:lenses_2.13:0.11.15 [S]
[info]   +-com.thesamet.scalapb:scalapb-runtime_2.13:0.11.15 [S]
[info]     +-org.scalameta:common_2.13:4.9.5 [S]
[info]       +-org.scalameta:trees_2.13:4.9.5 [S]
[info]         +-org.scalameta:parsers_2.13:4.9.5 [S]
[info]           +-org.scalameta:scalameta_2.13:4.9.5 [S]
[info]             +-io.getkyo:kyo-grpc-code-gen_3:0.10.2-SNAPSHOT
[info]
[success] Total time: 1 s, completed 5/06/2024, 9:46:25 pm
sbt:kyoJVM> kyo-grpc-code-gen/whatDependsOn com.thesamet.scalapb lenses_3
[info] com.thesamet.scalapb:lenses_3:0.11.15
[info]   +-com.thesamet.scalapb:scalapb-runtime_3:0.11.15
[info]     +-io.getkyo:kyo-grpc-code-gen_3:0.10.2-SNAPSHOT
[info]
[success] Total time: 0 s, completed 5/06/2024, 9:58:13 pm
```

Is it necessary for scalapb to be a dependency of common? Isn't it only for semanticdb? Can it be moved from common to semanticdb?

I'm guessing I can exclude the 2.13 version from the transitive dependencies but it'd be nice to not have to worry about that. Even better would be a [Scala 3 version of scalameta](https://github.com/scalameta/scalameta/issues/2485).

kitbellew:
Fixes #3724.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -102,9 +102,14 @@ lazy val semanticdbScalacCore = project.in(file("semanticdb/scalac/library")).se
   sharedSettings,
   publishJVMSettings,
   fullCrossVersionSettings,
+  protobufSettings,
   mimaPreviousArtifacts := Set.empty,
   moduleName := "semanticdb-scalac-core",
   description := "Library to generate SemanticDB from Scalac 2.x internal data structures",
+  Compile / unmanagedSourceDirectories ++= {
+    val base = (ThisBuild / baseDirectory).value / "semanticdb"
+    List(base / "metap", base / "cli", base / "semanticdb", base / "metacp", base / "symtab")
+  },
   libraryDependencies += "org.scala-lang" % "scala-compiler" % scalaVersion.value
 ).dependsOn(scalameta.jvm)
 
@@ -156,7 +161,6 @@ lazy val common = crossProject(JSPlatform, JVMPlatform, NativePlatform).in(file(
     enableMacros,
     buildInfoPackage := "scala.meta.internal",
     buildInfoKeys := Seq[BuildInfoKey](version),
-    protobufSettings,
     crossScalaVersions := AllScalaVersions
   ).configureCross(crossPlatformPublishSettings).jsSettings(commonJsSettings)
   .enablePlugins(BuildInfoPlugin).nativeSettings(nativeSettings)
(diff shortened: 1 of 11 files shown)


kitbellew:
@tgodzik @bjaglin is it ok to ignore mima errors here? the classes were moved to a different package. cc @steinybot.

tgodzik:
Potentially that could break people's code, however unlikely. Could we just keep it in the same packages and move to another project, that should work fine.

kitbellew:
> Potentially that could break people's code, however unlikely. Could we just keep it in the same packages and move to another project, that should work fine.

the packages are the same but were moved from `scalameta` to `semanticdb-scalac-core`.

bjaglin (approved):
Generated and implicit classes are available to Scalafix rule authors, but there is no particular reason for them to use them, so I believe this can be done in a patch release from a Scalafix perspective :+1:

tgodzik:
> > Potentially that could break people's code, however unlikely. Could we just keep it in the same packages and move to another project, that should work fine.
> 
> the packages are the same but were moved from `scalameta` to `semanticdb-scalac-core`.

Ach, that should be fine then if the user has both jars on their classpath.

kitbellew:
@tgodzik any idea why release job keeps failing: https://github.com/scalameta/scalameta/actions/runs/9636793434 ?

tgodzik:
It was first failing due to an issue in sonatype, but later it was due to the minor issue fixed here https://github.com/scalameta/scalameta/pull/3748

PR #3747, opened 2024-06-24
Distribute various semanticdb files to own modules

kitbellew:
Add a new shared module `semanticdb-shared` under `semanticdb/`.

Also, in addition to existing `metac`, add modules `metap` and `metacp` leaving in each only `main`-related classes, while moving out the rest as follows (showing packages relative to `scala.meta/.internal`):

- to `semanticdb-shared` (under `semanticdb/`) from
  - `cli/`: `cli`
  - `metap/`: `metap`, `semanticdb`
- to `semanticdb-scalac-core` (under `scalac/library/`) from
  - `symtab/`: `symtab`
  - `metacp/`: `classpath`, `javacp`, `metacp`, `scalacp`

This way, the shared module contains the data structures only (and doesn't have too many dependencies,
except scalapb perhaps), while the core library module contains compiler-specific code.


build.sbt
```diff
@@ -102,15 +102,20 @@ lazy val semanticdbScalacCore = project.in(file("semanticdb/scalac/library")).se
...
   libraryDependencies += "org.scala-lang" % "scala-compiler" % scalaVersion.value
+).dependsOn(semanticdbShared)
+
+lazy val semanticdbShared = project.in(file("semanticdb/semanticdb")).settings(
+  moduleName := "semanticdb-shared",
+  sharedSettings,
+  publishJVMSettings,
+  fullCrossVersionSettings,
+  protobufSettings,
+  mimaPreviousArtifacts := Set.empty,
```
kitbellew, line 117: does this mean semanticdb stuff is not subject to mima checks?

tgodzik: That would probably make it so, alternative is probably to just disable mima plugin on it

kitbellew: I'm actually curious why disabling is desired...

tgodzik (approved):
I think it's clearer than the ad hoc directories :+1:

build.sbt
```diff
@@ -102,15 +102,20 @@ lazy val semanticdbScalacCore = project.in(file("semanticdb/scalac/library")).se
...
   description := "Library to generate SemanticDB from Scalac 2.x internal data structures",
-  Compile / unmanagedSourceDirectories ++= {
-    val base = (ThisBuild / baseDirectory).value / "semanticdb"
-    List(base / "metap", base / "cli", base / "semanticdb", base / "metacp", base / "symtab")
-  },
   libraryDependencies += "org.scala-lang" % "scala-compiler" % scalaVersion.value
+).dependsOn(semanticdbShared)
+
+lazy val semanticdbShared = project.in(file("semanticdb/semanticdb")).settings(
+  moduleName := "semanticdb-shared",
```
bjaglin, line 112: this is missing on other projects, I don't know if it's intentional:
> [info] 	published semanticdbmetacp_2.13.14 to https://oss.sonatype.org/content/repositories/snapshots/org/scalameta/semanticdbmetacp_2.13.14/4.9.7+12-060037a8-SNAPSHOT/semanticdbmetacp_2.13.14-4.9.7+12-060037a8-SNAPSHOT.jar

https://github.com/scalameta/scalameta/actions/runs/9646040618/job/26604944229

build.sbt
```diff
@@ -102,15 +102,20 @@ lazy val semanticdbScalacCore = project.in(file("semanticdb/scalac/library")).se
...
+  sharedSettings,
+  publishJVMSettings,
+  fullCrossVersionSettings,
```
bjaglin, line 115: is this needed? can we publish only once per scala binary version for this project, as well as the new ones?

build.sbt
bjaglin: Maybe it is/was a MiMa limitation due to the fact that these projects are fully cross-built?

build.sbt
kitbellew: i will fix that. not intentional. i assumed the name implicitly came from the subdir but it seems it comes from the sbt variable name (quite unexpected). thank you for alerting me to this.

build.sbt
kitbellew: i don't know, i was simply copying from scalac-core module. what will happen if we remove `fullCrossVersionSettings`? cc @tgodzik

build.sbt
kitbellew: i am not sure. but it's also true that if someone does use the library (to read/parse generated semanticdb files), it's basically protobuf anyway, and they are backwards-compatible.

build.sbt
tgodzik: not sure, I would keep it especially with the recent change to Scala 2.13 standard library. Pretty sure it will break currently if you remove it.

bjaglin: https://github.com/scalameta/scalameta/pull/3887

bjaglin:
FTR, I opened https://github.com/coursier/apps/pull/240 for `cs install metap` to handle the move

PR #3795, opened 2024-07-04
bugfix: Try and fix parsing large nested code

tgodzik:
This popped up in https://github.com/scalameta/metals/issues/6552 <- need to figure out where it could hang


kitbellew:
@tgodzik thank you for this, i took a look. i can't figure out why _exactly_ it leads to an infinite loop, but i have an approximate idea, will submit a proposed fix.

tgodzik:
> @tgodzik thank you for this, i took a look. i can't figure out why _exactly_ it leads to an infinite loop, but i have an approximate idea, will submit a proposed fix.

Thanks! I started working on this, but the test took longer than expected :sweat_smile:

PR #3803, opened 2024-07-05
ScalametaParser: less backtrack on ctrl cond, body

kitbellew:
If a scala3 if/while actually uses scala2 syntax, we might have parsed part of the body while looking for `then` or `do`, which leads to some backtracking.

Specifically, for nested `if` clauses with their `then` part in explicit braces, each will be parsed exactly twice, once as a continuation of the condition expression and second time as the body, thus leading to an exponential time complexity.

Let's instead reduce backtracking by processing what might be both an argument to a condition expression or an initial part of the body, such as a brace expression, and then using it either for condition or for body.

Fixes scalameta/metals#6552. Supersedes #3795.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -1415,34 +1415,49 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
   }
 
   private def condExprWithBody[T <: Token: ClassTag]: (Term, Term) = {
-    val cond =
+    val (cond, bodyOpt) =
       if (!dialect.allowQuietSyntax) {
         val cond = condExpr()
         newLinesOpt()
-        cond
+        (cond, None)
       } else if (!token.is[LeftParen]) {
         val cond = expr()
         acceptAfterOptNL[T]
-        cond
+        (cond, None)
       } else {
         val startPos = tokenPos
         val simpleExpr = condExpr()
-        if (acceptIfAfterOptNL[T]) simpleExpr
+        if (acceptIfAfterOptNL[T]) (simpleExpr, None)
         else {
+          // let's consider case when something can continue cond or start body
+          val argsOrInitBody = token match {
+            case _: LeftParen => Some(inParensOrTupleOrUnitExpr(allowRepeated = false))
+            case _: LeftBrace => Some(blockExprOnBrace())
+            case _ => None
+          }
           val complexExpr = tryParse {
-            val simpleRest = simpleExprRest(simpleExpr, canApply = true, startPos = startPos)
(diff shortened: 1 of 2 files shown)


tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/ControlSyntaxSuite.scala
```diff
@@ -4066,4 +4066,39 @@ class ControlSyntaxSuite extends BaseDottySuite {
     runTestAssert[Stat](code, layout)(tree)
   }
 
+  test("metals#6552: huge nested `if`") {
+    val nestedNum = 100 // was previously failing around 25-30 deep
```
kitbellew, line 4070: i tried 1000 and got stack overflow. not sure if should worry about that case.

tgodzik: Stack overflow is fine, people can always increase it if needed.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1415,32 +1415,47 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
...
   private def condExprWithBody[T <: Token: ClassTag]: (Term, Term) = {
-    val cond =
+    val (cond, bodyOpt) =
       if (!dialect.allowQuietSyntax)
-        try condExpr()
+        try (condExpr(), None)
         finally newLinesOpt()
       else if (!token.is[LeftParen])
-        try expr()
+        try (expr(), None)
```
tgodzik, line 1423:
```suggestion
        try{ (expr(), None) }
```
looks like Scala 2.12 doesn't like it

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -1415,32 +1415,47 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
   }
 
   private def condExprWithBody[T <: Token: ClassTag]: (Term, Term) = {
-    val cond =
+    val (cond, bodyOpt) =
       if (!dialect.allowQuietSyntax)
-        try condExpr()
+        try (condExpr(), None)
```
tgodzik, line 1420:
```suggestion
        try { (condExpr(), None) }
```
and probably here also

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: done, in #3805.

tgodzik (approved):
LGTM!

PR #3812, opened 2024-07-08
Token: define Shebang token, detect and parse it

kitbellew:


tgodzik (approved):
Cool! We can get rid of weird workarounds in places

PR #3813, opened 2024-07-09
improvement: Don't throw on wrong brace balance

tgodzik:
This started showing up as invariant failed exception and in reality I wonder if it's the role of the tokenizer to verify. This should be much nicer as a parsing error later on instead of the weird invariant error.


scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
```diff
@@ -163,15 +163,9 @@ class ScalametaTokenizer(input: Input, dialect: Dialect) {
       case _ => pushToken(token)
     }
 
-    def loop(braceBalance: Int = 0): Unit = if (nextToken()) {
+    def loop(): Unit = if (nextToken()) {
       emitToken(getCurrToken())
-      lastEmittedToken match {
-        case _: Token.RightBrace if braceBalance > 1 => loop(braceBalance - 1)
-        case _: Token.LeftBrace if braceBalance > 0 => loop(braceBalance + 1)
-        case _: Token.RightBrace => require(braceBalance > 0)
```
kitbellew, line 171: i think it should be sufficient to remove this and next line. other cases are necessary as splice end can't otherwise be detected.

kitbellew:
@tgodzik should we add a simple test to make sure tokenizer, even if modified again, will continue handling it gracefully?

kitbellew:
@tgodzik if we follow this logic, then tokenizer shouldn't throw any exceptions at all, assuming the user might be in the middle of typing.

if so, what should we do with `tokenizers/Reporter` class? how do we recover gracefully from everything that might be wrong with the code (which might no longer look like scala)?

is it possible that the solution is not invoking tokenization in metals until the user has stopped typing?

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
tgodzik: Looks like I actually had to add a case to ignore any additional `}`

tgodzik:
>  should we add a simple test to make sure tokenizer, even if modified again, will continue handling it gracefully?

added!

> is it possible that the solution is not invoking tokenization in metals until the user has stopped typing?

It's not possible to detect reliably and people might stop typing and the code might still be bad, but we still need tokens for a bunch of things.

>  if we follow this logic, then tokenizer shouldn't throw any exceptions at all, assuming the user might be in the middle of typing.

I think so, ideally we would just return `Tokenized.Error` and avoid throwing as we already have a way of providing feedback. InvariantFailed exception should really be only in cases that something totally unexpected happened with our data structures. Anything else should really be reported as an error. For this particular case, the number of braces is not something tokenizer should detect really in my opinion as the parser will already report it nicely.

kitbellew:
if we shouldn't care about the braces at all, then perhaps we should only check them for cases of interpolation and xml, when the last matching brace emits a special end token. which means, we should worry about braceBalance only when it is positive before left or after right 🙂

i can check if there's a way to parse tokens permissively, when the code is half-baked but still resembles scala.

tgodzik:
> f we shouldn't care about the braces at all, then perhaps we should only check them for cases of interpolation and xml,

Sure, makes sense, I just didn't want to change to code too much, we could have separate loop that just ignores braces, but the current behaviour is basically doing this.

> we should worry about braceBalance only when it is positive before left or after right 🙂

What do you mean? Just report errors when we don't have correct string interpolation end? Ideally, we would not throw there either, if it's possible to return something valid from tokenization standpoint. Alternatively, maybe just catch all and return Tokenized.Error. That's also fine.

kitbellew:
looked into permissive tokenization... CharArrayReader, LegacyScanner and ScalametaTokenizer all apply a large number of validations, most resulting in Reporter.syntaxError or readerError. Only a few `require`, like in this case.

I don't think it's viable to remove all these syntaxError instances; how are they handled in your case (i.e., better than invariant exception or the same)? and could you explain why you need the tokens even if parsing fails later?

tgodzik:
> I don't think it's viable to remove all these syntaxError instances; how are they handled in your case (i.e., better than invariant exception or the same)?

We just return empty tokens and not use it. The alternative is to find every usage of tokenization and wrap it in a try, which can easily be forgotten in the future. It's fine to keep syntaxError, it's not really an issue. Maybe we could remove some that can be validated later on, but it's not really a problem. I would rather only return error and not throw at all.

>  could you explain why you need the tokens even if parsing fails later?

We often have data from the last compilation such as for example diagnostics, code lenses or references and we need to adjust their position if the user code changes, but doesn't yet compile (either intentionally or due to a bug). We use tokens to calculate edit distance in that case and move the positions around.

tgodzik:
So TLDR; Remove throwing altogether and remove validation case by case (like this one)

kitbellew (approved):

kitbellew:
> So TLDR; Remove throwing altogether and remove validation case by case (like this one)

for context: the linked PR tries to address this request but it's possible that it will produce complete garbage in very egregious cases as it might be able to interpret slightly incorrect scala but perhaps not if it's severely incorrect. it will not throw, but it will produce tokens which are not really useful in comparing with the previous, valid state as the edit distance might be substantial.

PR #3835, opened 2024-07-19
LegacyToken: add Invalid, capture tokenize errors

kitbellew:


scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
```diff
@@ -177,26 +179,32 @@ class ScalametaTokenizer(input: Input, dialect: Dialect) {
...
+//      case STRINGPART => unreachable
       case TRUE => Token.KwTrue(input, dialect, curr.offset)
       case FALSE => Token.KwFalse(input, dialect, curr.offset)
       case NULL => Token.KwNull(input, dialect, curr.offset)
 
       case INTERPOLATIONID => Token.Interpolation
           .Id(input, dialect, curr.offset, curr.endOffset, curr.strVal)
       case XMLLIT => getXmlPart(curr)
-      case XMLLITEND => unreachable
+//      case XMLLITEND => unreachable
```
tgodzik, line 207:
```suggestion
```
and also other below?

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/PublicSuite.scala
```diff
@@ -367,25 +367,9 @@ class PublicSuite extends TreeSuiteBase {
     // n/a
   }
 
-  test("scala.meta.tokenizers.TokenizeException.toString") {
-    interceptMessage[TokenizeException](
-      """|<input>:1: error: unclosed string literal
-         |"c
-         |^""".stripMargin.lf2nl
```
tgodzik, line 374: Should we instead make sure that invalid tokens are produced? Either that or remove the tests.

tgodzik (changes requested):
Thanks for working on this! Should be super useful!

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
kitbellew: do you mean to remove the lines completely? i kept them to indicate that these particular tokens should not be obtained via this method, and it's not an oversight they are not here.

tests/shared/src/test/scala/scala/meta/tests/prettyprinters/PublicSuite.scala
kitbellew: we can't remove these tests, without removing the entire PublicSuite. why it exists is a bit unclear for me, it is not to test functionality but to ensure that public methods are somehow mentioned here explicitly...

the actual functionality should be well tested in TokenizerSuite and similar ones. i remove the body of these tests because it no longer worked as before.

tests/shared/src/test/scala/scala/meta/tests/tokenizers/TokenizerSuite.scala
```diff
@@ -1957,38 +2000,88 @@ class TokenizerSuite extends BaseTokenizerSuite {
...
-    val error = """|<input>:1: error: double quotes are not allowed in single-line quasiquotes
-                   | s"..." 
-                   |  ^""".stripMargin.lf2nl
-    interceptMessage[TokenizeException](error)(
-      tokenize(code, dialects.Scala213.unquoteTerm(multiline = false))
-    )
+    val struct = """|BOF [0..0)
+                    |Space [0..1)
+                    |Interpolation.Id(s) [1..2)
+                    |Invalid(double quotes are not allowed in single-line quasiquotes) [2..2)
```
kitbellew, line 2077: as to checking that Invalid tokens are produced... here's one example.

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
tgodzik: Should we just add a comment for all those at the end? It's fine either way

tgodzik (approved):

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
kitbellew: done

PR #3837, opened 2024-07-23
bugfix: Use earliest Scala version for scalameta, parsers and trees projects

tgodzik:
If we use latest by default then the semanticdb parts will also depend on newer Scala versions, which is breaking for anything that might be using it as dependency.

The other option here is to change their suffix to full and publish separate for each version, but I would rather avoid it as it would mean more artifacts to publish and a change for any dependency.

This showed up in https://github.com/scalameta/metals/pull/6608

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,7 +15,7 @@ import munit.sbtmunit.BuildInfo.munitVersion
 import sbtcrossproject.CrossPlugin.autoImport.crossProject
 
 lazy val LanguageVersions = LatestScalaVersions
-lazy val LanguageVersion = LanguageVersions.head
+lazy val LanguageVersion = EarliestScala213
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -15,7 +15,7 @@ import munit.sbtmunit.BuildInfo.munitVersion
 import sbtcrossproject.CrossPlugin.autoImport.crossProject
 
 lazy val LanguageVersions = LatestScalaVersions
-lazy val LanguageVersion = LanguageVersions.head
+lazy val LanguageVersion = EarliestScala213
```
tgodzik, line 18: This just make the projects that are not fully cross compiled to be by default earliest 2.13.x

kitbellew (approved):

PR #3845, opened 2024-07-24
bugfix: Set earlier versions as only cross versions for scalameta projects

tgodzik:
Turns out, the previous approach didn't work and we still published with the newest Scala version. I changed it to only use the earliest versions, so that it's not even possible to set newer versions. I also changed sharedSettings to not include cross settings by default as they should be defined per project (or with fullCrossSettings).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
           - '2.13.14'
         java:
           - '8'
-          - '21'
+          - '17'
     steps:
       - uses: actions/checkout@v4
         with:
--- a/build.sbt
+++ b/build.sbt
@@ -15,7 +15,7 @@ import munit.sbtmunit.BuildInfo.munitVersion
 import sbtcrossproject.CrossPlugin.autoImport.crossProject
 
 lazy val LanguageVersions = LatestScalaVersions
-lazy val LanguageVersion = EarliestScala213
+lazy val LanguageVersion = LatestScala213
 def customVersion = sys.props.get("scalameta.version")
 def parseTagVersion: String = {
   import scala.sys.process._
@@ -152,7 +152,6 @@ lazy val semanticdbMetac = project.in(file("semanticdb/metac")).settings(
   sharedSettings,
   publishJVMSettings,
   fullCrossVersionSettings,
-  crossScalaVersions := LanguageVersions,
   mimaPreviousArtifacts := Set.empty,
   description := "Scalac 2.x launcher that generates SemanticDB on compile",
   libraryDependencies += "org.scala-lang" % "scala-compiler" % scalaVersion.value,
@@ -164,7 +163,6 @@ lazy val semanticdbMetap = project.in(file("semanticdb/metap")).settings(
   sharedSettings,
   publishJVMSettings,
   fullCrossVersionSettings,
-  crossScalaVersions := LanguageVersions,
   mimaPreviousArtifacts := Set.empty,
   description := "Prints SemanticDB files",
   mainClass := Some("scala.meta.cli.Metap")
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           - '2.13.14'
         java:
           - '8'
-          - '21'
+          - '17'
```
kitbellew, line 43: is this part of this change?

build.sbt
```diff
@@ -164,7 +163,6 @@ lazy val semanticdbMetap = project.in(file("semanticdb/metap")).settings(
   sharedSettings,
   publishJVMSettings,
   fullCrossVersionSettings,
-  crossScalaVersions := LanguageVersions,
```
kitbellew, line 167: so this uses all, as part of fullCrossVersionSettings?

build.sbt
```diff
@@ -15,7 +15,7 @@ import munit.sbtmunit.BuildInfo.munitVersion
 import sbtcrossproject.CrossPlugin.autoImport.crossProject
 
 lazy val LanguageVersions = LatestScalaVersions
-lazy val LanguageVersion = EarliestScala213
+lazy val LanguageVersion = LatestScala213
```
kitbellew, line 18: this is restored from the previous change? who is using it, btw?

do we even need LanguageVersion(s)? should we explicitly refer to latest or earliest everywhere?

build.sbt
```diff
@@ -282,6 +279,7 @@ lazy val scalameta = crossProject(allPlatforms: _*).in(file("scalameta/scalameta
 lazy val semanticdbIntegration = project.in(file("semanticdb/integration")).settings(
   description := "Sources to compile to build SemanticDB for tests.",
   sharedSettings,
+  crossScalaVersions := AllScalaVersions,
```
kitbellew, line 282: is this just for testing?

build.sbt
```diff
@@ -193,15 +190,15 @@ lazy val common = crossProject(allPlatforms: _*).in(file("scalameta/common")).se
   enableMacros,
   buildInfoPackage := "scala.meta.internal",
   buildInfoKeys := Seq[BuildInfoKey](version),
-  crossScalaVersions := AllScalaVersions
+  crossScalaVersions := EarliestScalaVersions
```
kitbellew, line 193: so common, trees, parsers, and scalameta are all built for the oldest version per major? assuming that older minor will work with newer minor library? and only semantic is building newer ones?

kitbellew (commented):
lgtm but a few questions.

.github/workflows/ci.yml
tgodzik: Yeah, id doesn't compile for 21 on older Scala 2 versions

build.sbt
tgodzik: Yes, which makes sense since it compiles with full version suffix. And I think at some point I had issues with tests, which also use it and are cross compiled to all versions, but I had numerous issues at points, so not 100% sure if needed, but I do think the change is correct.

build.sbt
tgodzik: Yep, it's only used for testing and should really be AllScalaVersions as tests are also AllScalaVersions. I had some issues otherwise with Scala versions and this should be fine.

build.sbt
tgodzik: > assuming that older minor will work with newer minor library

Yes, since the scala library is still binary compatible.

build.sbt
tgodzik: I inlined them, since this is really the default value and one use in 2 test/bench projects

kitbellew (approved):

tgodzik:
Double checked and looks like this working alright now

tgodzik:
I would say it's as expected. I didn't set earlier version for JS and Native as that is not needed for semanticdb. Everything else that has 2.13.11 is a dependency of semanticdb project. Anything in particular that caused your concern?

tgodzik:
Ach and some projects are set under 2.13.12, those are the ones that need to be fully cross compiled.

kitbellew:
> I would say it's as expected. I didn't set earlier version for JS and Native as that is not needed for semanticdb. Everything else that has 2.13.11 is a dependency of semanticdb project. Anything in particular that caused your concern?

what caught my eye is the fact that for that particular build, instead of using one specific version, or at least two (one which matches and another which is compatible), we ended up with three.

PR #3906, opened 2024-08-22
Input: add WithTokenizerOptions wrapper

kitbellew:
Since our public code signatures take Input and Dialect, the only way to provide TokenizerOptions without breaking MIMA compatibility is by using either of those types.

Neither of them in reality is perfect for this task, but since Dialect is a final class and adding tokenizer options as a field to it seems a bit illogical, let's implement it as a wrapper of the Input type.

Follow-on to #3786.

### DIFF
--- a/scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
+++ b/scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
@@ -3,6 +3,7 @@ package scala.meta.inputs
 import scala.meta.common._
 import scala.meta.internal._
 import scala.meta.io._
+import scala.meta.tokenizers.TokenizerOptions
 
 import java.nio.charset.Charset
 import java.nio.charset.StandardCharsets
@@ -11,6 +12,11 @@ import java.nio.{file => nio}
 sealed trait Input extends Product with Serializable with inputs.InternalInput {
   def chars: Array[Char]
   def text: String = new String(chars)
+
+  private[meta] def tokenizerOptions: Option[TokenizerOptions] = None
+  def withoutTokenizerOptions: Input = this
+  def withTokenizerOptions(implicit options: Option[TokenizerOptions]): Input = Input
+    .WithTokenizerOptions(this)
 }
 
 object Input {
@@ -25,6 +31,7 @@ object Input {
     val input: Input
     override def chars: Array[Char] = input.chars
     override def text: Predef.String = input.text
+    override private[meta] def tokenizerOptions: Option[TokenizerOptions] = input.tokenizerOptions
   }
 
   case object None extends Text {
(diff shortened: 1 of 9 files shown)


tgodzik (approved):
I guess neither is really perfect and putting it into dialect would be very misleading

tgodzik:
Thanks for working on this!

scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
```diff
@@ -111,6 +118,19 @@ object Input {
...
+  final case class WithTokenizerOptions private (input: Input)(options: TokenizerOptions)
+      extends Proxy {
+    override def toString = s"Input.WithTokenizerOptions($input, $tokenizerOptions)"
+    override private[meta] def tokenizerOptions = Some(options)
+    override def withoutTokenizerOptions: Input = input.withoutTokenizerOptions
+    override def withTokenizerOptions(implicit options: Option[TokenizerOptions]): Input = options
+      .fold(withoutTokenizerOptions)(x => if (x eq this.options) this else copy()(x))
+  }
+  object WithTokenizerOptions {
+    def apply(input: Input)(implicit options: Option[TokenizerOptions]): Input = options
```
bjaglin, line 130: I found it a bit cumbersome that you need to wrap options in `Some` here (as well as in the `withTokenizerOptions ` above). I assume it's because the default implicit expose in `TokenizerOptions` has that signature? Could we drop the `Option` wrapper here and there (putting the actual `default` instance as default implicit)?

bjaglin (commented):
Some questions / tweaks suggestions I had after trying out https://github.com/scalameta/scalameta/pull/3907

scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/InternalTrees.scala
```diff
@@ -64,20 +64,26 @@ trait InternalTree extends Product {
   }
 
   def tokenizeFor(dialect: Dialect): Tokens =
-    if (origin.dialectOpt.contains(dialect)) tokensOpt.get else retokenize(dialect)
+    if (origin.dialectOpt.contains(dialect)) tokensOpt.get else retokenize(dialect, None)
```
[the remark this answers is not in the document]
kitbellew: good point!

scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
kitbellew: will add an overload

scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/InternalTrees.scala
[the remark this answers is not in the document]
kitbellew: #3911

scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
kitbellew: #3912

PR #3907, opened 2024-08-23
WhitespaceTokenizer: add `Granular` definition

kitbellew:
Fixes #3786, as a final follow-on.


scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/WhitespaceTokenizer.scala
```diff
@@ -19,7 +19,14 @@ object WhitespaceTokenizer {
   def apply(input: Input, dialect: Dialect)(implicit
       options: Option[TokenizerOptions],
       tokens: java.util.Collection[Token]
-  ): WhitespaceTokenizer = new Grouping(input, dialect)
+  ): WhitespaceTokenizer =
+    if (options.forall(_.groupWhitespace)) new Grouping(input, dialect) else new Granular
```
bjaglin, line 23: I don't mind to force custom options in Scalafix, but shouldn't `Granular` be the default to preserve the <4.9.8 behavior, at least in the 4.9.x line?

```suggestion
    if (exists(_.groupWhitespace)) new Grouping(input, dialect) else new Granular
```

scalameta/tokenizers/shared/src/main/scala/scala/meta/tokenizers/TokenizerOptions.scala
```diff
@@ -4,6 +4,7 @@ import scala.util.DynamicVariable
 
 final class TokenizerOptions(
     // options which control how tokens are emitted
+    val groupWhitespace: Boolean = true
```
bjaglin, line 7: same question as above, could it be an opt-in rather than an opt-out?
```suggestion
    val groupWhitespace: Boolean = false
```

bjaglin (commented):
I have tested this in Scalafix and I managed to get all tests relying on the historical tokeniser behavior green 👍

bjaglin (approved):
Tested on scalafix, tests are green out of the box, without any explicit options

xuwei-k:
- https://github.com/scalameta/scalameta/issues/3979#issuecomment-2387539371

PR #3911, opened 2024-08-24
Origin: internally expose possible input

kitbellew:
In InternalTree, use it to tokenize, before resorting to syntax parsing. Also, if that Input contained any TokenizerOptions, they would naturally be used. Follow-on to #3786.

### DIFF
--- a/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/InternalTrees.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/InternalTrees.scala
@@ -81,7 +81,7 @@ trait InternalTree extends Product {
     case Lit.String(value) =>
       val input = Input.VirtualFile("<InternalTrees.tokens>", value)
       Tokens(Array(Constant.String(input, dialect, 0, value.length, value)))
-    case _ => tokenize(textAsInput, dialect).get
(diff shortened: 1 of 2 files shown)


scalameta/trees/shared/src/main/scala/scala/meta/internal/trees/InternalTrees.scala
```diff
@@ -64,19 +64,24 @@ trait InternalTree extends Product {
...
+      dialect: Dialect
+  )(implicit tokenize: Tokenize, tokenizerOptions: Option[TokenizerOptions]): Tokens = this match {
     case Lit.String(value) =>
       val input = Input.VirtualFile("<InternalTrees.tokens>", value)
       Tokens(Array(Constant.String(input, dialect, 0, value.length, value)))
-    case _ => implicitly[Tokenize].apply(textAsInput, dialect).get
+    case _ => tokenize(origin.inputOpt.getOrElse(textAsInput), dialect).get
   }
 
   private[meta] def textAsInput(implicit
```
bjaglin, line 87: not visible in the diff, but another implicit `Option[TokenizerOptions]` is captured - it's probably more correct to get it from the input just like above?

bjaglin (commented):
LGTM to favor the options from the potential Input (the same should probably be done in textAsInput), but I question the implicit captures as a fallback (I guess that the intent?) to get options if we can't find the Input.

PR #3912, opened 2024-08-24
Input: add overload with actual TokenizerOptions

kitbellew:
Follow-on to #3786.

### DIFF
--- a/scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
+++ b/scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
@@ -15,8 +15,10 @@ sealed trait Input extends Product with Serializable with inputs.InternalInput {
 
   private[meta] def tokenizerOptions: Option[TokenizerOptions] = None
   def withoutTokenizerOptions: Input = this
-  def withTokenizerOptions(implicit options: Option[TokenizerOptions]): Input = Input
-    .WithTokenizerOptions(this)
+  final def withTokenizerOptions(implicit options: Option[TokenizerOptions]): Input =
+    withTokenizerOptions(options.orNull)
+  def withTokenizerOptions(options: TokenizerOptions): Input =
+    if (options eq null) this else Input.WithTokenizerOptions(this, options)
 }
 
 object Input {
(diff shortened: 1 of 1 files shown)


scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
```diff
@@ -15,8 +15,10 @@ sealed trait Input extends Product with Serializable with inputs.InternalInput {
 
   private[meta] def tokenizerOptions: Option[TokenizerOptions] = None
   def withoutTokenizerOptions: Input = this
-  def withTokenizerOptions(implicit options: Option[TokenizerOptions]): Input = Input
-    .WithTokenizerOptions(this)
+  final def withTokenizerOptions(implicit options: Option[TokenizerOptions]): Input =
+    withTokenizerOptions(options.orNull)
```
bjaglin, line 19: Does this even have to be publicly exposed? Is the implicit `Option[TokenizerOptions]`just sugar for tests?

Anyway, I didn't quite get how the options get propagated (and the need for the DynamicVariable to back the default implicit), so I am probably missing a piece.

bjaglin (approved):

scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Input.scala
kitbellew: in your case, since i restored the original behaviour, you don't need to use it. but i will use it in scalafmt, hence i would have to use `input.withTokenizerOptions(...)` before using the parser.

and yes, I will probably set the global value since scalafmt can occasionally use the parser multiple times (for rewrite rules, mostly), and the global value can be implicitly visible.

does that make sense?

bjaglin-swile: So if I get this right,you'll add `withTokenizerOptions()` to all input instance throughout the code in scalafmt, and to avoid passing a custom TokenizerOptions to opt-in grouping whitespaces each time, you'll control the global implicit?

I am still contemplating that Option wrapper.... If the globlal implicit was a `TokenizerOptions` instead of an `Option[Tokenizer]`, wouldn't the overload you added here (after adding an implicit on the argument) be enough for both cases (providing an explicit option or capturing the default ones that might be custom)?

bjaglin: So if I get this right, you'll add `withTokenizerOptions()` to all Input instance throughout the code in scalafmt, and to avoid passing a custom `TokenizerOptions` to opt-in grouping whitespaces each time, you'll control the global implicit?

I am still contemplating that `Option` wrapper.... If the globlal implicit was a `TokenizerOptions` instead of an `Option[Tokenizer]`, wouldn't the overload you added here (after adding an implicit on the argument) be enough for both cases (providing an explicit option or capturing the default ones that might be custom)?

kitbellew: i originally tried without option but then got bigger down somewhere, don't remember where 🙂 i can try to check again, before we release

bjaglin: OK - anyway, not a blocker for me obviously, just curiosity and willingness to reduce public API if possible

kitbellew: #3919

PR #3930, opened 2024-08-25
Scaladoc: define and parse markdown code span

### LINKED ISSUE #3908: Code expressions in Scaladoc parsing
The Scaladoc parser expects code expressions to be formatted as `{{{x + y}}}`. There are two problems with that:

- The current Scaladoc generates a _code block_ from it, even when `{{{` is not followed by a newline.
- The Scaladoc documentation suggests _not_ to use wikidoc in new projects but to switch to markdown instead.

So, `{{{x + y}}}` should really be `` `x + y` `` in new code, but this is _not_ parsed as a `CodeExpr` token by Scalameta. This causes problems in tools that rely on Scalameta, such as Scalafmt (see <https://github.com/scalameta/scalafmt/issues/4147>).

I think `` `x + y` `` should be parsed as a new `MdCodeExpr` that extends `TextPart` like `CodeExpr` (there is already `MdCodeBlock` as an alternate to `CodeBlock`).

kitbellew:
Fixes #3908.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScaladocParser.scala
@@ -93,6 +93,15 @@ object ScaladocParser {
     }
   }
 
+  private def mdCodeSpanParser[$: P]: P[MdCodeSpan] = P {
+    val tick = "`"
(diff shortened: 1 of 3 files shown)


kitbellew:
@charpov fyi

tgodzik (approved):
LGTM!

PR #3936, opened 2024-08-27
bugfix: Don't throw when broken string interpolation

tgodzik:
Related to https://github.com/scalameta/metals/issues/6699

It seems the code is perfectly reachable after all, so maybe it's better to default to invalid.

The other option I was thinking about (or an additional) is to finish interpolation splice right away with and End, but that might be too complex for a case that might not pop up that often.

### DIFF
--- a/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
+++ b/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
@@ -277,10 +277,7 @@ class ScalametaTokenizer(input: Input, dialect: Dialect)(implicit options: Token
 
       case EOF => new Token.EOF(input, dialect)
       case SHEBANG => new Token.Shebang(input, dialect, curr.offset, curr.endOffset, curr.strVal)
-      case INVALID => getInvalid(curr, curr.strVal)
-
-      // the rest of tokens shouldn't be obtained via this method
-      case _ => unreachable(debug(curr))
+      case _ => getInvalid(curr, curr.strVal)
     }
   }
 
--- a/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
+++ b/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
@@ -60,13 +60,15 @@ class ScalametaTokenizer(input: Input, dialect: Dialect)(implicit options: Token
           pushPart(beg)
           pushToken(Token.Interpolation.SpliceStart(input, dialect, dollarOffset, dollarOffset + 1))
           val splice = nextToken()
-          if (splice.token != STRINGPART) {
-            pushToken(getToken(splice))
-            if (splice.token == LBRACE) loop(braceBalance = 1)
-            val end = nextToken()
-            pushToken(Token.Interpolation.SpliceEnd(input, dialect, end.offset, end.offset))
-            emitContents(end)
-          } else emitContents(splice)
+          val end =
+            if (splice.token == STRINGPART) splice
+            else {
+              pushToken(getToken(splice))
+              if (splice.token == LBRACE) loop(braceBalance = 1)
+              nextToken()
+            }
+          pushToken(Token.Interpolation.SpliceEnd(input, dialect, end.offset, end.offset))
+          emitContents(end)
         } else beg
 
       // NOTE: before emitStart, curr is the first token that follows INTERPOLATIONID
@@ -279,7 +281,8 @@ class ScalametaTokenizer(input: Input, dialect: Dialect)(implicit options: Token
 
       case EOF => new Token.EOF(input, dialect)
       case SHEBANG => new Token.Shebang(input, dialect, curr.offset, curr.endOffset, curr.strVal)
-      case _ => getInvalid(curr, curr.strVal)
+      case INVALID => getInvalid(curr, curr.strVal)
+      case _ => getInvalid(curr, s"Unexpected token id ${curr.token}, contents:\n" + curr.strVal)
     }
   }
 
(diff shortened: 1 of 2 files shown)


scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
```diff
@@ -277,10 +277,7 @@ class ScalametaTokenizer(input: Input, dialect: Dialect)(implicit options: Token
 
       case EOF => new Token.EOF(input, dialect)
       case SHEBANG => new Token.Shebang(input, dialect, curr.offset, curr.endOffset, curr.strVal)
-      case INVALID => getInvalid(curr, curr.strVal)
-
-      // the rest of tokens shouldn't be obtained via this method
-      case _ => unreachable(debug(curr))
+      case _ => getInvalid(curr, curr.strVal)
```
kitbellew, line 280: so what type of token was attempted here? just want to be sure we're not masking a problem elsewhere.so what type of token was attempted here? just want to be sure we're not masking a problem elsewhere.

tgodzik: STRINGPART since we have invalid token on single $ we then have a normal string contents

tests/shared/src/test/scala/scala/meta/tests/tokenizers/TokenizerSuite.scala
```diff
@@ -2102,5 +2102,40 @@ class TokenizerSuite extends BaseTokenizerSuite {
...
+                  |""".stripMargin.replace("'''", "\"\"\"")
+
+    val struct = """|BOF [0..0)
+                    |LF [0..1)
+                    |Interpolation.Id(s) [1..2)
+                    |Interpolation.Start(''') [2..5)
+                    |Interpolation.Part(|Available languages (default = ) [5..37)
+                    |Interpolation.SpliceStart [37..38)
+                    |Invalid(Not one of: `$'_, `$$', `$'ident, `$'this, `$'BlockExpr) [38..38)
+                    |Invalid():
```
tgodzik, line 2124:
```suggestion
                    |StringPart():
```
alternatively we could try having string part and just have the single invalid token.

tests/shared/src/test/scala/scala/meta/tests/tokenizers/TokenizerSuite.scala
```diff
@@ -2102,5 +2102,37 @@ class TokenizerSuite extends BaseTokenizerSuite {
...
+                  |'''.stripMargin
+                  |""".stripMargin.replace("'''", "\"\"\"")
+
+    val struct = """|BOF [0..0)
+                    |LF [0..1)
+                    |Interpolation.Id(s) [1..2)
+                    |Interpolation.Start(''') [2..5)
+                    |Interpolation.Part(|Available languages (default = ) [5..37)
+                    |Interpolation.SpliceStart [37..38)
+                    |Invalid(Not one of: `$'_, `$$', `$'ident, `$'this, `$'BlockExpr) [38..38)
```
tgodzik, line 2123: I changed it a bit and I think it's currently the best option. Splice starts, but is broken, so we continue with normal string part and an invalid token.

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
```diff
@@ -60,11 +60,13 @@ class ScalametaTokenizer(input: Input, dialect: Dialect)(implicit options: Token
...
-          val end = nextToken()
-          pushToken(Token.Interpolation.SpliceEnd(input, dialect, end.offset, end.offset))
-          emitContents(end)
+          if (splice.token != STRINGPART) {
+            pushToken(getToken(splice))
+            if (splice.token == LBRACE) loop(braceBalance = 1)
+            val end = nextToken()
+            pushToken(Token.Interpolation.SpliceEnd(input, dialect, end.offset, end.offset))
+            emitContents(end)
+          } else emitContents(splice)
```
kitbellew, line 69: should we do something like this, perhaps:
```
val end = 
  if (splice.token == STRINGPART) splice
  else {
    pushToken(getToken(splice))
    if (splice.token == LBRACE) loop(braceBalance = 1)
    nextToken()
  }
  pushToken(Token.Interpolation.SpliceEnd(input, dialect, end.offset, end.offset))
  emitContents(end)
```
that way, we always close SpliceStart with a SpliceEnd.

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
kitbellew: keeping `getInvalid` is a good idea, just in case. do you think we should adjust `getInvalid` to include the `LegacyToken` that was parsed? for debugging purposes?

kitbellew (changes requested):

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
tgodzik: Done!

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/ScalametaTokenizer.scala
tgodzik: I split the cases again and added a proper message which includes that info whenever unexpected token was found

kitbellew (approved):

kitbellew:
@tgodzik i merged but perhaps you'd like to remove your branch if it's not automatically removed.

PR #3959, opened 2024-09-12
Add basic support for `tracked` parameters

KacperFKorban:
This PR adds basic support for the experimental `tracked` modifier.

Relevant links:
- https://dotty.epfl.ch/docs/reference/experimental/modularity.html
- https://github.com/scala/scala3/blob/e9d48313a2e11685a148e105f038da368485aa66/docs/_docs/internals/syntax.md?plain=1#L382

Any feedback will be appreciated.

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
@@ -161,7 +161,9 @@ final class Dialect private[meta] (
     // https://docs.scala-lang.org/scala3/reference/other-new-features/control-syntax.html
     val allowQuietSyntax: Boolean,
     // Are binary literals allowed? SIP-42.
-    val allowBinaryLiterals: Boolean
+    val allowBinaryLiterals: Boolean,
+    // Are tracked parameters allowed? docs: https://dotty.epfl.ch/docs/reference/experimental/modularity.html
+    val allowTrackedParameters: Boolean
 ) extends Product with Serializable {
 
   // NOTE(olafur) checklist for adding a new dialect field in a binary compatible way:
@@ -256,7 +258,8 @@ final class Dialect private[meta] (
     allowParamClauseInterleaving = false,
     allowFewerBraces = false,
     allowQuietSyntax = false,
-    allowBinaryLiterals = false
+    allowBinaryLiterals = false,
+    allowTrackedParameters = false
     // NOTE(olafur): declare the default value for new fields above this comment.
   )
 
@@ -393,6 +396,9 @@ final class Dialect private[meta] (
   def withAllowBinaryLiterals(newValue: Boolean): Dialect =
     privateCopy(allowBinaryLiterals = newValue)
 
+  def withAllowTrackedParameters(newValue: Boolean): Dialect =
+    privateCopy(allowTrackedParameters = newValue)
+
   // NOTE(olafur): add the next `withX()` method above this comment. Please try
   // to use consistent formatting, use `newValue` as the parameter name and wrap
   // the body inside curly braces.
@@ -458,7 +464,8 @@ final class Dialect private[meta] (
       allowParamClauseInterleaving: Boolean = this.allowParamClauseInterleaving,
       allowFewerBraces: Boolean = this.allowFewerBraces,
       allowQuietSyntax: Boolean = this.allowQuietSyntax,
-      allowBinaryLiterals: Boolean = this.allowBinaryLiterals
+      allowBinaryLiterals: Boolean = this.allowBinaryLiterals,
+      allowTrackedParameters: Boolean = this.allowTrackedParameters
       // NOTE(olafur): add the next parameter above this comment.
   ): Dialect = {
     val notForUnquote = unquoteType eq UnquoteType.None
@@ -521,7 +528,8 @@ final class Dialect private[meta] (
       allowParamClauseInterleaving = allowParamClauseInterleaving,
       allowFewerBraces = allowFewerBraces,
       allowQuietSyntax = allowQuietSyntax,
-      allowBinaryLiterals = allowBinaryLiterals
+      allowBinaryLiterals = allowBinaryLiterals,
+      allowTrackedParameters = allowTrackedParameters
     )
     if (equivalent) return this // RETURN!
     new Dialect(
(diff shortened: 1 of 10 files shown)


tgodzik (changes requested):
You need to update two suites:
[error] 	scala.meta.tests.trees.ReflectionSuite
[error] 	scala.meta.tests.api.SurfaceSuite

And maybe we should add this as a dialect option?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
```diff
@@ -18,6 +18,7 @@ class SoftKeywords(dialect: Dialect) {
   object KwInfix extends IsWithName(dialect.allowInfixMods, "infix")
   object KwExtension extends IsWithName(dialect.allowExtensionMethods, "extension")
   object KwErased extends IsWithName(dialect.allowErasedDefs, "erased")
+  object KwTracked extends IsWithName(true, "tracked")
```
kitbellew, line 21: @tgodzik already mentioned that you need to identify a dialect flag for this, and enable it for the appropriate dialect instances.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3086,6 +3087,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
       case _ => if (hasExplicitMods) syntaxErrorExpected[KwVal]; None
     }
     varOrVarParamMod.foreach(mods += _)
+    if (!mods.has[Mod.ValParam] && !mods.has[Mod.VarParam])
+      rejectMod[Mod.Tracked](mods, "`tracked' modifier can only be used val/var parameters in classes")
```
kitbellew, line 3091: i am generally leaning towards removing validation unless it resolves ambiguity. if the code is correct, parser doesn't need to check; and if it isn't, the compiler will error out.

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/LanguageExtensionsSuite.scala
```diff
@@ -0,0 +1,41 @@
+package scala.meta.tests.parsers.dotty
+
+import scala.meta._
+
+class LanguageExtensionsSuite extends BaseDottySuite {
```
kitbellew, line 5: i don't know if we need a separate suite for just one modifier and 4 tests. isn't there already a suitable one?

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -1330,6 +1330,8 @@ object Mod {
   class Transparent() extends Mod
   @ast
   class Erased() extends Mod
+  @ast
```
kitbellew, line 1333: either here, or in Dialect when you add an appropriate flag, could you please add a link to the documentation which describes this feature?

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/LanguageExtensionsSuite.scala
KacperFKorban: I created a new suite, so that we can put all the other new language extension features there. Is that ok, or should I move it somewhere?
There are quite a few extensions missing here, so the test suite can fill up quickly.

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
KacperFKorban: Added it to the dialect flag comment

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/SoftKeywords.scala
KacperFKorban: Done, one thing to remember will be to add it to `3.5.0` once it is added. (Right now, I only added it to 3.future)

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/LanguageExtensionsSuite.scala
kitbellew: would you like to add these tests, too? 😁

KacperFKorban: I meant the ones that aren't implemented here yet.

e.g.
- `into`
- `namedTuples`
- `captureChecking`
- `namedTypeArguments`
- etc.

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -161,7 +161,9 @@ final class Dialect private[meta] (
     // https://docs.scala-lang.org/scala3/reference/other-new-features/control-syntax.html
     val allowQuietSyntax: Boolean,
     // Are binary literals allowed? SIP-42.
-    val allowBinaryLiterals: Boolean
+    val allowBinaryLiterals: Boolean,
+    // Are tracked parameters allowed? docs: https://dotty.epfl.ch/docs/reference/experimental/modularity.html
+    val allowTrackedParameters: Boolean
```
kitbellew, line 166: I'm sorry, is this feature part of an approved SIP, or released scala version? adding experimental stuff that may or may not be adopted is premature otherwise.

we have not yet implemented 3.4 parsing so I would start with that.

tgodzik: I think it would be safe to this to ScalaFuture as it will not be too much problem to remove it in the future. I would agree in terms of bigger features, but here it's rather safe.

I can follow up with 3.4.0 and 3.5.0 dialects later on.

tgodzik (approved):
LGTM from me

@kitbellew are you ok with merging? I can follow up with making sure that the dialects are up to date.

kitbellew:
> LGTM from me
> 
> @kitbellew are you ok with merging? I can follow up with making sure that the dialects are up to date.

overall, looks ok, but the code is not properly formatted yet 🙂

KacperFKorban:
@kitbellew I just noticed it too. I pushed a new update right now :sweat_smile:

PR #3964, opened 2024-09-13
ScalametaParser: remove modifier validation

kitbellew:


tgodzik (approved):
LGTM

I don't think we as a parser should verify it and it might be a problem keeping it up to date always

PR #3978, opened 2024-09-24
bugfix: Fix infinite loop when invalid character was used

tgodzik:
Fixes https://github.com/scalameta/metals-vscode/issues/1540


kitbellew (approved):
oh! should we immediately release 10.1?

tgodzik:
> oh! should we immediately release 10.1?

It was present earlier, but might be good to release. I will go ahead and do it later today

Issue #3993, opened 2024-10-15
Support named tuples related syntax


KacperFKorban:
Scalameta should be able to support new syntax related to named tuples. This includes:
- named tuple types
- named tuple terms
- named tuples in pattern matching

Syntax changes from the SIP:
```
SimpleType        ::=  ...
                    |  ‘(’ NameAndType {‘,’ NameAndType} ‘)’
NameAndType       ::=  id ':' Type

SimpleExpr        ::=  ...
                    |  '(' NamedExprInParens {‘,’ NamedExprInParens} ')'
NamedExprInParens ::=  id '=' ExprInParens

Patterns          ::=  Pattern {‘,’ Pattern}
                    |  NamedPattern {‘,’ NamedPattern}
NamedPattern      ::=  id '=' Pattern
```

Useful links:
- implementation PR: https://github.com/scala/scala3/pull/19174/files#diff-89e03cdd16b8e5fd6435d2df4b44a3e58c408a22adb7683c149015ab0d64f4bf
- SIP PR: https://github.com/scala/improvement-proposals/pull/72/files

KacperFKorban:
@tgodzik Thanks! :heart:

PR #4001, opened 2024-10-17
feature: Add support for named tuples

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/3993

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2470,6 +2470,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
   trait SeqContextSensitive extends PatternContextSensitive {
     // is a sequence pattern _* allowed?
     def isSequenceOK: Boolean
+    def isNamedTupleOk: Boolean = false
 
     def patterns(): List[Pat] = commaSeparated(pattern())
 
@@ -2636,6 +2637,8 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
               syntaxError(s"pattern must be a value or have parens: $sid$x", at = currToken)
             }
             sid match {
+              case name: Term.Name if isNamedTupleOk && acceptOpt[Equals] =>
+                Pat.Assign(name, noSeqWithNamed.pattern())
               case name: Term.Name.Quasi => name.become[Pat]
               case name: Term.Name =>
                 if (soft.StarSplice(currToken) && tryAheadNot[Ident]) Pat.Repeated(name)
@@ -2661,7 +2664,7 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
         case _: Xml.Start => xmlPat()
         case _: LeftParen =>
           val lpPos = currIndex
-          val patterns = inParensOnOpenOr(noSeq.patterns())(Nil)
+          val patterns = inParensOnOpenOr(noSeqWithNamed.patterns())(Nil)
           makeTuple(lpPos, patterns, Lit.Unit(), Pat.Tuple.apply) {
             case t if !isRhs => Right(t)
             case t @ Pat.Tuple(_ :: Nil) => Right(t)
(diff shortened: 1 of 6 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2631,6 +2632,14 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
               targs.fold(ref)(x => autoEndPos(sid)(Term.ApplyType(ref, x))),
               autoPos(argumentPatterns().reduceWith(Pat.ArgClause.apply))
             )
+          } else if (at[Equals] && isNamedTupleOk) {
```
tgodzik, line 2635: Should we check dialect here? All other cases are already covered and parse.

kitbellew: if there's no confusion (`=` can't be interpreted differently), i am in favor of skipping as otherwise you'd have to error out if the dialect doesn't allow.

unrelated: see below for an optional implementation alternative.

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -766,6 +766,8 @@ object Pat {
     })
   }
   @ast
+  class Assign(name: Term.Name, rhs: Pat) extends Pat
```
kitbellew, line 769: could you also extend `Tree.WithBody`, just like `Term.Assign`? (I wonder why Term.Assign allows not just Name).

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/NamedTuplesSuite.scala
```diff
@@ -0,0 +1,96 @@
...
+        Term.Tuple(List(
+          Term.Assign(Term.Name("a"), Lit.Int(123)),
+          Term.Assign(Term.Name("d"), Lit.Double(2.01d))
+        ))
+      )
+    )))
+  }
+
+  test("simple-named-type") {
+    runTestAssert[Type]("(a: Int, b: String)")(Type.Tuple(List(
```
kitbellew, line 28: is this from the same sip, or in the same vein?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2631,6 +2632,13 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
...
+            next()
+            sidToken match {
+              case ident: Ident => Pat.Assign(Term.Name(ident.value), noSeqWithNamed.pattern())
+              case _ => syntaxError(s"pattern must be an identifier: $sid", at = currToken)
+
+            }
```
kitbellew, line 2641: instead of new code above, we can put it below:

```
sid match {
  case name: Term.Name if isNamedTupleOk && acceptOpt[Equals] =>
    Pat.Assign(name, noSeqWithNamed.pattern())
  // rest is as before
  case name: Term.Name.Quasi => ...
}
```

that way, we handle possible targs, sid being Quasi or Select, etc.

kitbellew: sorry, forgot to "Save" this comment earlier.

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/NamedTuplesSuite.scala
tgodzik: It's the same sip, it's the type of the named tuples, but it already works, so decided to not add a new AST node.

kitbellew: of course, don't need new trees

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
tgodzik: Done!

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Good idea, changed

kitbellew (approved):
beautiful!

kitbellew:
merged, feel free to delete the branch.

Issue #3994, opened 2024-10-15
Support new context bound syntax


KacperFKorban:
Scalameta should support new context bound syntax introduced in SIP-64. These changes include:
- naming context bounds, as in `A: Monoid as a`
 - a new syntax for multiple context bounds, as in `A: {Monoid, Ord}`
 - context bounds for type members
 - a cleaner syntax for given definitions

Summary of syntax changes from the SIP:
```
TmplDef           ::=  'given' GivenDef
GivenDef          ::=  [id ':'] GivenSig
GivenSig          ::=  GivenImpl
                    |  '(' ')' '=>' GivenImpl
                    |  GivenConditional '=>' GivenSig
GivenImpl         ::=  GivenType ([‘=’ Expr] | TemplateBody)
                    |  ConstrApps TemplateBody
GivenConditional  ::=  DefTypeParamClause
                    |  DefTermParamClause
                    |  '(' FunArgTypes ')'
                    |  GivenType
GivenType         ::=  AnnotType1 {id [nl] AnnotType1}
TypeDef           ::=  id [TypeParamClause] TypeAndCtxBounds
TypeParamBounds   ::=  TypeAndCtxBounds
TypeAndCtxBounds  ::=  TypeBounds [‘:’ ContextBounds]
ContextBounds     ::=  ContextBound | '{' ContextBound {',' ContextBound} '}'
ContextBound      ::=  Type ['as' id]
FunType           ::=  FunTypeArgs (‘=>’ | ‘?=>’) Type
                    |  DefTypeParamClause '=>' Type
FunExpr           ::=  FunParams (‘=>’ | ‘?=>’) Expr
                    |  DefTypeParamClause ‘=>’ Expr
```

Links:
- implementation PR: https://github.com/scala/scala3/pull/20061/files
- SIP PR: https://github.com/scala/improvement-proposals/pull/81/files

PR #4002, opened 2024-10-17
feature: Add improved bounds syntax

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/3994

Fortunately, the changed given syntax is working without an issue.

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
@@ -52,6 +52,7 @@ final class Dialect private[meta] (
     // Are view bounds supported by this dialect?
     // def f[A <% Int](a: A)
     // Removed in Dotty.
+    @deprecated("allowViewBounds unneeded, it was only used for an error", ">4.10.2")
     val allowViewBounds: Boolean,
     // Are XML literals supported by this dialect?
     // We plan to deprecate XML literal syntax, and some dialects
@@ -165,7 +166,7 @@ final class Dialect private[meta] (
     // Are tracked parameters allowed? docs: https://dotty.epfl.ch/docs/reference/experimental/modularity.html
     val allowTrackedParameters: Boolean,
     // https://github.com/scala/improvement-proposals/pull/81/files
-    val allowImprovedBoundsAndGivens: Boolean
+    val allowImprovedTypeClassesSyntax: Boolean
 ) extends Product with Serializable {
 
   // NOTE(olafur) checklist for adding a new dialect field in a binary compatible way:
@@ -193,7 +194,7 @@ final class Dialect private[meta] (
       allowTrailingCommas: Boolean,
       allowTraitParameters: Boolean,
       allowTypeLambdas: Boolean,
-      allowViewBounds: Boolean,
+      allowViewBounds: Boolean, // unused
       allowXmlLiterals: Boolean,
       toplevelSeparator: String // unused
   ) = this(
@@ -214,7 +215,7 @@ final class Dialect private[meta] (
     allowTrailingCommas = allowTrailingCommas,
     allowTraitParameters = allowTraitParameters,
     allowTypeLambdas = allowTypeLambdas,
-    allowViewBounds = allowViewBounds,
+    allowViewBounds = true,
     allowXmlLiterals = allowXmlLiterals,
     toplevelSeparator = "", // unused
     allowNumericLiteralUnderscoreSeparators = false,
@@ -262,7 +263,7 @@ final class Dialect private[meta] (
     allowQuietSyntax = false,
     allowBinaryLiterals = false,
     allowTrackedParameters = false,
-    allowImprovedBoundsAndGivens = false
+    allowImprovedTypeClassesSyntax = false
     // NOTE(olafur): declare the default value for new fields above this comment.
   )
 
@@ -299,7 +300,8 @@ final class Dialect private[meta] (
   def withAllowTraitParameters(newValue: Boolean): Dialect =
     privateCopy(allowTraitParameters = newValue)
   def withAllowTypeLambdas(newValue: Boolean): Dialect = privateCopy(allowTypeLambdas = newValue)
-  def withAllowViewBounds(newValue: Boolean): Dialect = privateCopy(allowViewBounds = newValue)
+  @deprecated("allowViewBounds unneeded, it was only used for an error", ">4.10.2")
+  def withAllowViewBounds(newValue: Boolean): Dialect = this
   def withAllowXmlLiterals(newValue: Boolean): Dialect = privateCopy(allowXmlLiterals = newValue)
   @deprecated("toplevelSeparator has never been used", ">4.4.35")
   def withToplevelSeparator(newValue: String): Dialect = this
@@ -402,8 +404,8 @@ final class Dialect private[meta] (
   def withAllowTrackedParameters(newValue: Boolean): Dialect =
     privateCopy(allowTrackedParameters = newValue)
 
-  def withAllowImprovedBoundsAndGivens(newValue: Boolean): Dialect =
-    privateCopy(allowImprovedBoundsAndGivens = newValue)
+  def withAllowImprovedTypeClassesSyntax(newValue: Boolean): Dialect =
+    privateCopy(allowImprovedTypeClassesSyntax = newValue)
 
   // NOTE(olafur): add the next `withX()` method above this comment. Please try
   // to use consistent formatting, use `newValue` as the parameter name and wrap
@@ -424,7 +426,6 @@ final class Dialect private[meta] (
       allowTrailingCommas: Boolean = this.allowTrailingCommas,
       allowTraitParameters: Boolean = this.allowTraitParameters,
       allowTypeLambdas: Boolean = this.allowTypeLambdas,
-      allowViewBounds: Boolean = this.allowViewBounds,
       allowXmlLiterals: Boolean = this.allowXmlLiterals,
       allowNumericLiteralUnderscoreSeparators: Boolean =
         this.allowNumericLiteralUnderscoreSeparators,
@@ -472,7 +473,7 @@ final class Dialect private[meta] (
       allowQuietSyntax: Boolean = this.allowQuietSyntax,
       allowBinaryLiterals: Boolean = this.allowBinaryLiterals,
       allowTrackedParameters: Boolean = this.allowTrackedParameters,
-      allowImprovedBoundsAndGivens: Boolean = this.allowImprovedBoundsAndGivens
+      allowImprovedTypeClassesSyntax: Boolean = this.allowImprovedTypeClassesSyntax
       // NOTE(olafur): add the next parameter above this comment.
   ): Dialect = {
     val notForUnquote = unquoteType eq UnquoteType.None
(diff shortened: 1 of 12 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3114,35 +3118,56 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
...
                   val msg = "Use an implicit parameter instead.\n" +
                     "Example: Instead of `def f[A <% Int](a: A)` " +
                     "use `def f[A](a: A)(implicit ev: A => Int)`."
                   syntaxError(s"View bounds are not supported. $msg", at = currToken)
                 }
-                doWhile(vbounds += getBound())(at[Viewbound])
+                doWhile(vbounds += getBound())(acceptOpt[Viewbound])
+              }
+              def addContextBound() = cbounds += getBound(allowBoundsAlias = true)
+              currToken match {
```
kitbellew, line 3144: if there are no other match cases, here's an alternative:
```
if (acceptOpt[Colon]) {
  def addContextBounds[T : ClassTag]: Unit =
    doWhile(addContextBound(allowBoundsAlias = false))(acceptOpt[T])
  if (acceptOpt[LeftBrace]) inBracesAfterOpen(addContextBounds[Comma])
  else addContextBounds[Colon]
}
```

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3114,35 +3118,56 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
...
-                currToken match {
-                  case t: Ellipsis => ellipsis[Type](t, 1)
-                  case _ => typ()
-                }
+              def getBound(allowBoundsAlias: Boolean = false): Type = currToken match {
+                case t: Ellipsis => ellipsis[Type](t, 1)
+                case _ => contextBoundOrAlias(allowBoundsAlias)
               }
-              if (at[Viewbound]) {
+              if (acceptOpt[Viewbound]) {
```
kitbellew, line 3134: looking at the code now, and in the context of parsing resiliency discussions we've had, I would:
- get rid of the `ctxBoundsAllowed` parameter to this method
- not add the new dialect flag

either there's a ViewBound and/or Colon token, in which case we'll add the bounds, or there isn't. these tokens can't be parsed differently based on either of those two flags above.

i'd even deprecate and ignore the `allowViewBounds` dialect flag and remove the error just below.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3114,35 +3118,56 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
...
             val cbounds = new ListBuffer[Type]
             if (ctxBoundsAllowed) {
               @inline
-              def getBound(): Type = {
-                next()
-                currToken match {
-                  case t: Ellipsis => ellipsis[Type](t, 1)
-                  case _ => typ()
-                }
+              def getBound(allowBoundsAlias: Boolean = false): Type = currToken match {
```
kitbellew, line 3130: the method is used in two places only, so there's no prevailing default, thus i'd remove the default value and specify `getBound(allowAlias = xxx)` explicitly. Also, I'd rename the parameter to `allowAlias` as `Bound` is redundant in the context of this method.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3114,35 +3118,56 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
...
             val tbounds = typeBounds()
             val vbounds = new ListBuffer[Type]
             val cbounds = new ListBuffer[Type]
             if (ctxBoundsAllowed) {
               @inline
-              def getBound(): Type = {
-                next()
-                currToken match {
-                  case t: Ellipsis => ellipsis[Type](t, 1)
-                  case _ => typ()
```
kitbellew, line 3127: side comment, mostly rhetorical: i was surprised to find that `typ()` doesn't handle `Ellipsis`... why wouldn't it?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3114,35 +3118,56 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
...
     }
   }
 
+  def contextBoundOrAlias(allowBoundsAlias: Boolean) = {
+    val tpe = typ()
+    tpe match {
+      case Type.ApplyInfix(nm: Type.Name, Type.Name("as"), alias: Type.Name)
+          if allowBoundsAlias && dialect.allowImprovedBoundsAndGivens => Type.BoundsAlias(alias, nm)
+      case _ => tpe
+    }
```
kitbellew, line 3168: minor alternative:
```
typ() match {
  ...
  case tpe => tpe
}
```

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -501,7 +501,6 @@ object Type {
   class ArgClause(values: List[Type]) extends Member.ArgClause
   @ast
   class CasesBlock(cases: List[TypeCase] @nonEmpty) extends Tree.CasesBlock
-
```
kitbellew, line 504: I'd keep the blank before Name

kitbellew (changes requested):
all good but i'd propose not adding new dialect flag and removing existing `typeParam()` method flag.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: actually, could even inline the `addContextBound` method here as well, it's not used anywhere else. so
```
if (acceptOpt[Colon]) {
  def addContextBounds[T : ClassTag]: Unit =
    doWhile(cbounds += getBound(allowBoundsAlias = true))(acceptOpt[T])
  if (acceptOpt[LeftBrace]) inBracesAfterOpen(addContextBounds[Comma])
  else addContextBounds[Colon]
}
```
and `addContextBound` didn't take a parameter :)

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: I would keep the dialect flag since `A as otherName` would be an infix type before, so this would change parsing for Scala 2 even

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
```diff
@@ -163,7 +163,9 @@ final class Dialect private[meta] (
     // Are binary literals allowed? SIP-42.
     val allowBinaryLiterals: Boolean,
     // Are tracked parameters allowed? docs: https://dotty.epfl.ch/docs/reference/experimental/modularity.html
-    val allowTrackedParameters: Boolean
+    val allowTrackedParameters: Boolean,
+    // https://github.com/scala/improvement-proposals/pull/81/files
```
kitbellew, line 167: should we use something more readable, like https://docs3.scala-lang.org/sips/sips/typeclasses-syntax.html?

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: ah, logical :)

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: We can do a follow up removing it and check, but I wouldn't do it here

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Thanks! Good idea, changed

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: I removed allowViewBounds as it's not really useful

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
tgodzik: Renamed and removed default

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
tgodzik: Added back

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
tgodzik: Changed to allowImprovedTypeclassSyntax

tgodzik:
@kitbellew anything I missed here or is it ok to merge? This is coming in 3.6.x, so wanted to support it in metals also

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
kitbellew: to answer your question: it still looks to me that the old parameter is not needed, only the dialect flag to distinguish between alias and type infix.

tgodzik: Right, makes sense! I removed ctxBoundsAllowed param

scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
kitbellew: sorry, I meant here not to change the name of the flag, but to include a different URL for reference :) profuse apologies.

the scala-lang doc is more readable than the textual pull request. the original flag name was absolutely fine, can be restored if you like.

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -3114,38 +3105,39 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
...
-                doWhile(vbounds += getBound())(at[Viewbound])
-              }
-              while (at[Colon]) cbounds += getBound()
+            @inline
+            def getBound(allowAlias: Boolean): Type = currToken match {
+              case t: Ellipsis => ellipsis[Type](t, 1)
+              case _ => contextBoundOrAlias(allowAlias)
+            }
+            if (acceptOpt[Viewbound])
+              doWhile(vbounds += getBound(allowAlias = false))(acceptOpt[Viewbound])
```
kitbellew, line 3118: hmm, perhaps this might be simpler, sorry i didn't realize it earlier:
```
while (acceptOpt[Viewbound])
  vbounds += getBound(allowAlias = false)
```

tgodzik: Fixed!

kitbellew (approved):

tgodzik:
I will go ahead with a release since we need the new features for Scala 3.6 support

PR #4013, opened 2024-10-30
ScalametaParser: allow named types within `types()`

### LINKED ISSUE #4007: Named tuples parser error on latest metals snapshot
Metals server version: `1.4.0+6-73dc6db8-SNAPSHOT` which already includes `Scalameta v4.11.0` that added support for named tuples.

Code:
```scala
val y = summon[Tuple2[Int, String] <:< (x: Int, y: String)]
```

Error:
```
`)` expected but `:` found
```

kitbellew:
Fixes #4007. Alternative to #4009.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -668,27 +668,28 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
       // We need to accept `(T, U) => W`, `(x: T): x.U` and also support unquoting.
       var hasImplicits = false
 
-      @tailrec
-      def paramOrType(modsBuf: mutable.Builder[Mod, List[Mod]]): Type = currToken match {
+      def modOrType(modsBuf: mutable.Builder[Mod, List[Mod]]): Type = currToken match {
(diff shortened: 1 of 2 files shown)


scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -780,6 +781,26 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
...
+        val mods = fmods
+        val modPos = if (mods.isEmpty) startPos else mods.head.begIndex
+        val name = atPos(startPos)(Type.Name(t.value))
+        val tpe = maybeParamType(allowFunctionType)
+        Some(autoEndPos(modPos)(Type.TypedParam(name, tpe, mods)))
+      case _ => None
+    }
+
+    private def maybeNamedType(): Type = {
+      val ok = dialect.allowImprovedTypeClassesSyntax
```
tgodzik, line 800: That's not really related to the change. That's named tuples, but we didn't add a separate dialect for it.

kitbellew: didn't realize that. so named tuples are ok to be parsed for any dialect, correct? ok, and other than, what do you think about this alternative?

PR #3939, opened 2024-08-28
improvement: Add cross publish for Scala 3

tgodzik:
This is very much a work in progress, but at least we are down to 11 failing tests.

PR #4020, opened 2024-11-07
refactor: Implement custom xml scanner instead of using fastparse

tgodzik:
While trying to cross compile to Scala 3 it become evident that I can't move XmlParser from the trees module. On the other hand, I am not able to reimplement all the macros there, so instead I decided to reimplement the parser.

Why not keep using fastparse?

If using fastparse in trees we would get fastparse_2.13 to all the Scala 3 modules and that would conflict with fastparse_3 that we would need there.

So did I implement the parser?

No, since it's hard, but I wonder if a scanner would not be enough for us:
- the additional validation can be checked by the compiler
- scanner should be faster (I can improve the implementation for sure here)
- we can add the parser back for validation later while actually parsing, but since xml is ~deprecated~ discouraged in Scala 3 I wonder if this is worth the effort. Maybe it's fine to just leave out the additional validation.

The code is not yet perfect, but if the approach is ok, I will work on it some more.


kitbellew:
and it's not possible to shade fastparse?

tgodzik:
Hmm... I guess that's another option that I didn't think about. Though then we would have both unshaded and shaded fastparse , which seems wrong.

But maybe that's still a safer option, I will check that out tomorrow

tgodzik:
Looks like we are already shading them, but that is only done on the release, so if we want to run the tests then it will still not work. We will get:
```
[error] Modules were resolved with conflicting cross-version suffixes in ProjectRef(uri("file:/home/tgodzik/Documents/scalameta/"), "parsersJVM"):
[error]    com.lihaoyi:geny _3, _2.13
[error]    com.lihaoyi:sourcecode _3, _2.13
[error]    com.lihaoyi:fastparse _3, _2.13
```
even if we shade _2.13 and _3 under separate package names.

So I have to exclude fasparse_2.13 so that we can compile the modules that use fasparse_3. But then we miss one class from 2.13 that is needed and is not present in fastparse_3.

We would have to go back to publishing fastparse separately and shading it differently for Scala 2 to use in the trees module.

We could possibly make a minimal sbt project that only does it and then we can depend on it. What do you think?

tgodzik:
> why are there both _2.13 and _3 versions? i assumed that if we build with scala2, we'd have _2.13, and with scala3 only _3

We want to build `trees` with 2.13.x and use that version for Scala 3. This means trees brings in `fastparse_2.13`, but then `parsers` module also uses fasparse and they are cross compiled for Scala 2.13 and 3. So for `parsers_3` we need `fastparse_3` as it uses macros. So we end up having both fastparse versions and sbt is unhappy with it:
- fastparse_2.13 is coming from trees_2.13
- fastparse_3 is coming from parsers_3 which depends on trees_2.13

So the options are to not use fastparse at all in trees or have a totally shaded module. Anything else I tried fails since 

> do you have a branch in your fork that shows what you have been trying to do?

That's basically https://github.com/scalameta/scalameta/pull/3939 <- I am working on making that ready and start porting changes. The Scala 3 is still not working, I need to separate the semanticdb tests and figure out how to deal with the XmlParser.

> should we ask if there's something that the author of fastparse could suggest for this case?

That's not really related to fastparse itself but to sbt build definition itself

> while i don't mind implementing our own scanner, i'd strong prefer to avoid it, more debt

I know, but it also felt this is actually the most maintainable in the long run. It's much simpler than a full parser and should be faster. Though we might get some bugs along the way. I see spark repo had some xml that we test in the community build and  we have a bunch of tests, so this should be safe.

It also feels weird to me that we have parser in the scanner which we don't actually need at all as we don't have xml trees.

kitbellew:
maybe this is somehow evidents... but why wouldn't parsers_3 depend on trees_3 (and not trees_2.13)?

tgodzik:
> maybe this is somehow evident... but why wouldn't parsers_3 depend on trees_3 (and not trees_2.13)?

That's the thing, publishing trees_3 would require rewriting all the ast macros to something different since Scala 3 macros do not allow for annotations yet. And as Scala 3 can just use 2.13 libraries and the macros are contained within trees itself, we should be fine (at least from what I checked) to use 2.13. So trees would not be cross compiled to Scala 3

kitbellew:
ah, interesting. in that case, why do we even need to compile parsers for 3?

kitbellew:
... or are you slowly porting to 3, and trees is something you'd like to leave for dessert? :)

tgodzik:
> or are you slowly porting to 3, and trees is something you'd like to leave for dessert? :)

Yes, though I wouldn't call it exactly a dessert :sweat_smile: 

My hope is that at some point I will figure out which way will be easiest:
- wait for the macro annotations
- change to generating code 

I think the second part would probably be safer in the long run.

tgodzik:
> you said that parsers_3 requires fastparse_3 but you also said that scala3 can use 2.13 libraries; can parsers_3 use fastparse_2.13 instead (if that gets rid of the problem)?

fastparse uses macros which are not compatible between 2 and 3. There 2 different implementations.

> if for some reason not, can we move scaladoc parsing stuff (both own and the older contrib version) into the trees module (as that's the only code that uses fastparse within parser)?

That could work,  together with shading it would mean downstream users would not see any issue

tgodzik:
Ok, it seems to work. I had move a couple of classes in https://github.com/scalameta/scalameta/pull/3939/files#diff-f689a3ed75de1834609a68720ce3ef5eb1369e1d928b4e7c1c61d1ecde356c67 like:

image

Do you think that's ok? I think I can start moving some changes out of that PR.

kitbellew:
> Ok, it seems to work. I had move a couple of classes in https://github.com/scalameta/scalameta/pull/3939/files#diff-f689a3ed75de1834609a68720ce3ef5eb1369e1d928b4e7c1c61d1ecde356c67 like:
> Do you think that's ok? I think I can start moving some changes out of that PR.

yes, of course, as we discussed, moving these files is perfectly appropriate. looks like you're also moving some tests to their dedicated project, also fine. i didn't look through the rest of the change, though :)

tgodzik:
I plan to extract most of it to smaller pieces since the pr was mostly my experiment

Issue #3996, opened 2024-10-15
Support capture checking related syntax


KacperFKorban:
Scalameta should support syntax related to capture checking.
These changes include:
- syntax for indicating captures `tpe^` and `tpe^{...}`
- syntax for capturing functions `tpe ->^{...} tpe`

Links:
- implementation PR: https://github.com/scala/scala3/pull/15877/files#diff-7b0af84ac42fe737a4f3700b2ce9838c02f7bc1f7d4d266f44b07c6b33199abb

kitbellew:
@KacperFKorban this is a complex feature and requires explicit documentation, including changes to EBNF.

so far, the only document I have found is this: https://www.scala-lang.org/api/3.3.3/docs/docs/reference/experimental/cc.html

this document makes no mention of EBNF changes -- and does not show in its examples your `^{....}`. so, at this point its under-documented to be implemented, at least not without deprecations or binary-breaking changes soon after.

kitbellew:
@KacperFKorban @tgodzik are pure functions possible without capture checking? does it make sense to add separate `Dialect` flags, one for pure functions and another for capture checking?

one scenario that i thought was described somewhere was that perhaps `->` will be supported in scala2 but captures will be done via an annotation...

KacperFKorban:
@kitbellew Yes, pure functions are possible without capture checking. They are enabled by a separate language feature/import. https://github.com/scala/scala3/blob/d78516aa97b986024681cb59ebf7817c205742bb/library/src/scala/runtime/stdLibPatches/language.scala#L78

PR #4030, opened 2024-11-13
ScalametaParser: parse pure function types

kitbellew:
Helps with #3996.

### DIFF
--- a/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
+++ b/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
@@ -112,7 +112,7 @@ package object dialects {
   implicit val Scala3: Dialect = Scala36
 
   implicit val Scala3Future: Dialect = Scala3.withAllowUnderscoreAsTypePlaceholder(true)
-    .withAllowTrackedParameters(true)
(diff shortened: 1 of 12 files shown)


kitbellew:
@KacperFKorban fyi

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -104,6 +104,7 @@ package object dialects {
     .withAllowParamClauseInterleaving(true).withAllowParameterTypeConversions(true)
 
   implicit val Scala34: Dialect = Scala33.withAllowQuotedTypeVariables(true)
```
tgodzik, line 106: Since this is available via import, should we not add it to Scala3Future?

tgodzik (approved):

scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala
kitbellew: is the same true for `into` as well?

tgodzik: I think so

PR #4040, opened 2024-11-19
Trees: add Term.SelectPostfix

kitbellew:
Just like ApplyInfix, which largely reflects form rather than content, and could have been expressed as a Select followed by an Apply instead, let's introduce SelectPostfix.

This also makes it clear that any Select trees are properly constructed having a dot in their syntax, as recent scala releases started requiring an explicit import to allow `postfixOps`.

### DIFF
--- a/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
@@ -285,8 +285,15 @@ object Term {
     def value = ""
     checkParent(ParentChecks.AnonymousImport)
   }
+  @branch
+  trait SelectLike extends Term.Ref with Pat {
+    def qual: Term
+    def name: Term.Name
+  }
   @ast
-  class Select(qual: Term, name: Term.Name) extends Term.Ref with Pat
+  class Select(qual: Term, name: Term.Name) extends SelectLike
+  @ast
+  class SelectPostfix(qual: Term, name: Term.Name) extends SelectLike
   @ast
   class Interpolate(prefix: Name, parts: List[Lit] @nonEmpty, args: List[Term]) extends Term {
     checkFields(parts.length == args.length + 1)
--- a/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/InfixSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/InfixSuite.scala
@@ -258,10 +258,10 @@ class InfixSuite extends BaseDottySuite {
          |""".stripMargin,
       """|{
          |  freezing
-         |  |.boiling
+         |  |boiling
          |}
          |""".stripMargin
-    )(blk("freezing", tselect("|", "boiling")))
+    )(blk("freezing", tpostfix("|", "boiling")))
   }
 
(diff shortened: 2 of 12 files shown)


scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -285,8 +285,15 @@ object Term {
...
+  @branch
+  trait SelectLike extends Term.Ref with Pat {
+    def qual: Term
+    def name: Term.Name
+  }
+  @ast
+  class Select(qual: Term, name: Term.Name) extends SelectLike
   @ast
-  class Select(qual: Term, name: Term.Name) extends Term.Ref with Pat
+  class SelectPostfix(qual: Term, name: Term.Name) extends SelectLike
```
tgodzik, line 296: We should put that into breaking section when releasing as people should update from just Select to SelectLike

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/InfixSuite.scala
```diff
@@ -256,14 +256,92 @@ class InfixSuite extends BaseDottySuite {
...
+         |""".stripMargin,
+      """|{
+         |  freezing
+         |  `|`|
+         |}
+         |""".stripMargin
+    )(blk("freezing", tpostfix("|", "|")))
+  }
+
+  test("scala3 infix syntax 3.2.5") {
```
tgodzik, line 331: Do we need to test 3.2.x ? People should be using at least LTS 3.3.x version, so it's highly unexpected for them to be using 3.2.x

tgodzik (approved):
The change makes sense, just two minor comments otherwise LGTM

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
kitbellew: can you explain what you mean here? are you talking about release notes, or about the code itself?

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/InfixSuite.scala
kitbellew: this is not a scala version but a numbered list of test cases.

there were several examples shown on the documentation page, and i created tests for each (1, 2, 3). then i added variations and numbered them 3.1 and 3.2, for instance.

here, i am adding variations of variations, hence 3.2.1 etc.

scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
tgodzik: I just mean we should have a special section in the release notes about potential breaking changes people could encounter since their code will now parse differently

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/InfixSuite.scala
tgodzik: Aaa that makes sense, I could not figure out why this changed in 3.2.x line :sweat_smile:

PR #4053, opened 2024-12-03
improvement: Don't use format since it seems to break at some chars

tgodzik:
Related to https://github.com/scalameta/metals/issues/7003

Looks like there is at least 6 stack traces formatting error for InvalidToken and this looks like a problem.

We get things like:
```
	at java.util.regex.Pattern$BmpCharProperty.match(java.base@11.0.24/Pattern.java:3966)
	at java.util.regex.Pattern$GroupHead.match(java.base@11.0.24/Pattern.java:4807)
	at java.util.regex.Pattern$Branch.match(java.base@11.0.24/Pattern.java:4752)
	at java.util.regex.Pattern$Branch.match(java.base@11.0.24/Pattern.java:4750)
	at java.util.regex.Pattern$BranchConn.match(java.base@11.0.24/Pattern.java:4716)
	at java.util.regex.Pattern$GroupTail.match(java.base@11.0.24/Pattern.java:4866)
	at java.util.regex.Pattern$BmpCharPropertyGreedy.match(java.base@11.0.24/Pattern.java:4347)
	at java.util.regex.Pattern$GroupHead.match(java.base@11.0.24/Pattern.java:4807)
	at java.util.regex.Pattern$Branch.match(java.base@11.0.24/Pattern.java:4752)
	at java.util.regex.Pattern$BranchConn.match(java.base@11.0.24/Pattern.java:4716)
	at java.util.regex.Pattern$GroupTail.match(java.base@11.0.24/Pattern.java:4866)
	at java.util.regex.Pattern$BmpCharPropertyGreedy.match(java.base@11.0.24/Pattern.java:4347)
	at java.util.regex.Pattern$GroupHead.match(java.base@11.0.24/Pattern.java:4807)
	at java.util.regex.Pattern$Branch.match(java.base@11.0.24/Pattern.java:4752)
	at java.util.regex.Pattern$Branch.match(java.base@11.0.24/Pattern.java:4750)
	at java.util.regex.Pattern$BmpCharProperty.match(java.base@11.0.24/Pattern.java:3967)
	at java.util.regex.Pattern$Start.match(java.base@11.0.24/Pattern.java:3622)
	at java.util.regex.Matcher.search(java.base@11.0.24/Matcher.java:1729)
	at java.util.regex.Matcher.find(java.base@11.0.24/Matcher.java:773)
	at java.util.Formatter.parse(java.base@11.0.24/Formatter.java:2702)
	at java.util.Formatter.format(java.base@11.0.24/Formatter.java:2655)
```

and we should be as efficient as possible when just printing an error. Especially if we have a lot of errors then this might become inefficient.

### DIFF
--- a/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala
+++ b/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala
@@ -234,7 +234,7 @@ class LegacyScanner(input: Input, dialect: Dialect) {
 
     @inline
     def reportIllegalCharacter(): Unit = curr
-      .setInvalidToken(s"illegal character '\\u${"%04x".format(ch)}'")
+      .setInvalidToken("illegal unicode codepoint: 0x" + ch.toHexString)
 
     (ch: @switch) match {
       case ' ' =>
--- a/tests/shared/src/test/scala/scala/meta/tests/tokenizers/TokenizerSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/tokenizers/TokenizerSuite.scala
@@ -2136,6 +2136,7 @@ class TokenizerSuite extends BaseTokenizerSuite {
          |""".stripMargin.replace("'''", "\"\"\"").nl2lf
     assertTokenizedAsStructureLines(code, struct)
   }
+
   test("unexpected-character") {
     val code = """|
                   |val 。 = 123
@@ -2145,7 +2146,7 @@ class TokenizerSuite extends BaseTokenizerSuite {
     val struct = s"""|BOF [0..0)
                      |LF [0..1)
                      |KwVal [1..4)
-                     |Invalid(illegal character '\\u3002') [5..6)
+                     |Invalid(illegal unicode codepoint: 0x3002) [5..6)
                      |MultiHS(3) [4..7)
                      |Equals [7..8)
                      |Space [8..9)


scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala
```diff
@@ -233,8 +233,10 @@ class LegacyScanner(input: Input, dialect: Dialect) {
     }
 
     @inline
-    def reportIllegalCharacter(): Unit = curr
-      .setInvalidToken(s"illegal character '\\u${"%04x".format(ch)}'")
+    def reportIllegalCharacter(): Unit = {
+      val output = "\\u" + Integer.toHexString(ch | 0x10000).substring(1);
```
tgodzik, line 237: This is a bit less clear but doesn't involve regex underneath

kitbellew: can you, perhaps, add a comment in the code explaining what this does?

kitbellew (commented):
while I agree we should be very efficient, still it doesn't really explain the core underlying problem. 

could it be that metals is a bit too aggressive constantly parsing too many files at once? only one is presumably active at any given time.

also, if metals is capable of pointing out the location of the illegal character in a given file, can the user share what exactly was being parsed there which triggered the error? should we change something in the scanner to help identify that location?

is there perhaps a bug in the scanner that leads to valid code being interpreted as invalid in that user's case?

tgodzik:
> could it be that metals is a bit too aggressive constantly parsing too many files at once? only one is presumably active at any given time.

This is a possibility, but what surprised me is that every stack trace I saw was at `reportIllegalCharacter` which might mean this is what takes longest. If there is a lot of such characters then it will tokenize a longer time.

> also, if metals is capable of pointing out the location of the illegal character in a given file, can the user share what exactly was being parsed there which triggered the error? should we change something in the scanner to help identify that location?

So the problem is that we don't get any errors and I am not sure why. We don't also have access to the codebase, so figure I will try first here and see if that helps.

> is there perhaps a bug in the scanner that leads to valid code being interpreted as invalid in that user's case?

Might be, but we don't know which file that is since we use the scanner not only for open files bu dependencies as well.

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala
tgodzik: Maybe? The stack trace goes later like:
```
at java.lang.String.format(java.base@11.0.24/String.java:2897)
	at scala.collection.StringOps$.format$extension(StringOps.scala:991)
	at scala.meta.internal.tokenizers.LegacyScanner.reportIllegalCharacter$1(LegacyScanner.scala:237)
```

It's here https://gist.github.com/felher/9939f409ab3f7d0df5e5abe8f06ec1e2

tgodzik: Ok, I changed it to not use `s"` and added proper explanations.

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala
```diff
@@ -232,9 +232,22 @@ class LegacyScanner(input: Input, dialect: Dialect) {
...
+     * The algorithm for calculating unicode is as follows:
+     *   - `ch | 0x10000` gets us a new binary number: 1 at 17 position, then zeroes and then the
+     *     actual char number.
+     *   - `Integer.toHexString(ch | 0x10000)` gets us the hexagonal unicode plus the one from 17
+     *     position
+     *   - `Integer.toHexString(ch | 0x10000).substring(1)` drops that added one
+     *
+     * This way we don't have to prefix with 0s since they will be included in the hex number
+     * string.
+     */
```
kitbellew, line 245: ah, now i understand it! well, in that case, this assumes that `ch` fits into 16 bits, and it doesn't: https://stackoverflow.com/questions/27415935/does-unicode-have-a-defined-maximum-number-of-code-points

especially since illegal characters likely are larger.

here are two alternatives i can propose (need to write a short test, of course):
```
/**
 * Prints exactly the number of bytes in `ch` and has no extra memory allocations.
 * Borrows logic from [[java.lang.Integer.toHexString]] and [[java.util.HexFormat.toHexDigits]]
 */
def one() = {
  val sb = new java.lang.StringBuilder(32)
  sb.append("illegal character '\\u")
  val hexFormat = java.util.HexFormat.of()
  // computes bytes (min 2) from number of non-zero bits
  var numBytes = 2.max((Integer.SIZE - Integer.numberOfLeadingZeros(ch) + 7) / 8)
  while (numBytes > 0) {
    numBytes -= 1
    val curByte = (ch >>> (8 * numBytes)) & 0xf
    sb.append(hexFormat.toHighHexDigit(curByte))
    sb.append(hexFormat.toLowHexDigit(curByte))
  }
  sb.append('\'')
  curr.setInvalidToken(sb.toString)
}

/**
  * Prints 2 bytes if `ch` fits, or all 4 otherwise.
  * Has an extra memory allocation from `toHexDigits`.
  */
def two() = {
  val hexFormat = java.util.HexFormat.of()
  val output =
    if (Integer.numberOfLeadingZeros(ch) < 16)
      hexFormat.toHexDigits(ch)
    else
      hexFormat.toHexDigits(ch.toShort)
  curr.setInvalidToken("illegal character '\\u" + output + "'")
}
```

kitbellew (changes requested):
Looks like we might need to add a test or two, including those which take more than 16 bits.

ChatGPT says the first 16 bits are occupied by Basic Multilingual Plane, and shows examples from others:

Supplementary Multilingual Plane (SMP, U+010000–U+01FFFF)

- U+1F600: 😀 (Grinning Face Emoji)
- U+1F4A9: 💩 (Pile of Poo Emoji)
- U+1D11E: 𝄞 (Musical Symbol G Clef)
- U+1F680: 🚀 (Rocket Emoji, SMP)
- U+1F9D1: 🧑 (Person Emoji, SMP)

tgodzik:
> Looks like we might need to add a test or two, including those which take more than 16 bits.
> 
> ChatGPT says the first 16 bits are occupied by Basic Multilingual Plane, and shows examples from others:
> 
> Supplementary Multilingual Plane (SMP, U+010000–U+01FFFF)
> 
> * U+1F600: 😀 (Grinning Face Emoji)
> * U+1F4A9: 💩 (Pile of Poo Emoji)
> * U+1D11E: 𝄞 (Musical Symbol G Clef)
> * U+1F680: 🚀 (Rocket Emoji, SMP)
> * U+1F9D1: 🧑 (Person Emoji, SMP)

Curiously, all those are legal as identifiers. I am not sure now why `。` is not :sweat_smile:

tgodzik:
> Curiously, all those are legal as identifiers. I am not sure now why `。` is not 😅

And it's not possible to write \u with more than 4 characters.

What if we drop all that and just toString that character?

kitbellew:
> > Curiously, all those are legal as identifiers. I am not sure now why `。` is not 😅
> 
> And it's not possible to write \u with more than 4 characters.
> 
> What if we drop all that and just toString that character?

`toString` is likely not so good since it show that little circle (for all invalid-surrogate codepoints).

I think the simplest solution is:
```
  curr.setInvalidToken("illegal unicode codepoint: 0x" + ch.toHexString)
```

if you don't like `\u` with more than 4 characters :)

tgodzik:
I think the compiler itself only deals with 4 

https://github.com/scala/scala/blob/06d9be6e31c62c3ddeb03e8b802daef250d0dbc1/src/compiler/scala/tools/nsc/ast/parser/Scanners.scala#L846

tgodzik:
> > I think the compiler itself only deals with 4
> > https://github.com/scala/scala/blob/06d9be6e31c62c3ddeb03e8b802daef250d0dbc1/src/compiler/scala/tools/nsc/ast/parser/Scanners.scala#L846
> 
> are you referring to `%04x` again? if so, what i remember, it means "print hex, but at least 4, pad with zeros if there are fewer than 4".

ach, that's quite magical. I changed to ` curr.setInvalidToken("illegal unicode codepoint: 0x" + ch.toHexString)` solution. What do you think?

kitbellew:
> > > I think the compiler itself only deals with 4
> > > https://github.com/scala/scala/blob/06d9be6e31c62c3ddeb03e8b802daef250d0dbc1/src/compiler/scala/tools/nsc/ast/parser/Scanners.scala#L846
> > 
> > 
> > are you referring to `%04x` again? if so, what i remember, it means "print hex, but at least 4, pad with zeros if there are fewer than 4".
> 
> ach, that's quite magical. I changed to ` curr.setInvalidToken("illegal unicode codepoint: 0x" + ch.toHexString)` solution. What do you think?

i think it's good... but you didn't upload a new patch.

tgodzik:
> i think it's good... but you didn't upload a new patch.

My bad, I was sure I did it!

kitbellew:
looks good

kitbellew:
since this has been merged, could delete the branch.

@olafurpg is it possible to give one of us permissions to modify this repo's settings (such as auto deleting branches on merge)?

olafurpg:
I would have done it already if I wasn’t on the phone

tgodzik:
Will do!

PR #4077, opened 2024-12-09
SBT: upgrade munit-1.0.3, scalajs-1.17.0 plugins

kitbellew:


build.sbt
```diff
@@ -457,11 +457,14 @@ lazy val isScala3 = isScalaBinaryVersion("3")
...
 
 lazy val munitLibrary = Def.setting {
+  // TODO: this is getting ridiculous; any way to fix dependency on munit's
+  // scala version upgrades, specifically for 2.13 (and not 2.12 or earlier)?
   val munitV =
     if (isScala211.value) "0.7.29"
     else if (scalaVersion.value == "2.13.11") "1.0.0-M10"
     else if (scalaVersion.value == "2.13.12") "1.0.0-M11"
     else if (scalaVersion.value == "2.13.13") "1.0.0"
+    else if (scalaVersion.value == "2.13.14") "1.0.2"
```
kitbellew, line 467: any thoughts on why this is only necessary for 2.13 and not other minor versions?

tgodzik: Scala 2.13.x will be able to break binary compat for the library by adding new methods etc. This will not happen for 2.12.x, since the work there will mostly be bugfixes and no new features

build.sbt
```diff
@@ -457,11 +457,14 @@ lazy val isScala3 = isScalaBinaryVersion("3")
 def isScala213or3 = Def.setting(isScala213.value || isScala3.value)
 
 lazy val munitLibrary = Def.setting {
+  // TODO: this is getting ridiculous; any way to fix dependency on munit's
```
tgodzik, line 460: We could use coursier and fetch like in https://github.com/scalameta/metals/blob/main/project/SemanticDbSupport.scala#L18 to generate a map of supported versions :thinking: 

Or I could stop updating Scala 2.13 in munit for the time being if not needed

build.sbt
kitbellew: so for each 2.13 patch we build, we will continue maintaining the appropriate version of munit?

and why only munit, and not any other dependency?

PR #4085, opened 2024-12-12
Account for ClassInfoType within toSemanticTpe

### LINKED ISSUE #4084: SemanticDB: Calling `CompilationUnit.toTextDocument` on `AnyRef` throws an exception
## Description of Issue
Calling `scala.meta.internal.semanticdb.scalac.TextDocumentOps$XtensionCompilationUnitDocument.toTextDocument`
on
```scala
/*
 * Scala (https://www.scala-lang.org)
 *
 * Copyright EPFL and Lightbend, Inc.
 *
 * Licensed under Apache License 2.0
 * (http://www.apache.org/licenses/LICENSE-2.0).
 *
 * See the NOTICE file distributed with this work for
 * additional information regarding copyright ownership.
 */

package scala

/** Class `AnyRef` is the root class of all ''reference types''.
 *  All types except the value types descend from this class.
 *  @template
 */
trait AnyRef extends Any {

  /** The equality method for reference types.  Default implementation delegates to `eq`.
   *
   *  See also `equals` in [[scala.Any]].
   *
   *  @param  that    the object to compare against this object for equality.
   *  @return         `true` if the receiver object is equivalent to the argument; `false` otherwise.
   */
  def equals(that: Any): Boolean = this eq that

  /** The hashCode method for reference types.  See hashCode in [[scala.Any]].
   *
   *  @return   the hash code value for this object.
   */
  def hashCode: Int = sys.error("hashCode")

  /** Creates a String representation of this object.  The default
   *  representation is platform dependent.  On the java platform it
   *  is the concatenation of the class name, "@", and the object's
   *  hashcode in hexadecimal.
   *
   *  @return     a String representation of the object.
   */
  def toString: String = sys.error("toString")

  /** Executes the code in `body` with an exclusive lock on `this`.
   *
   *  @param    body    the code to execute
   *  @return           the result of `body`
   */
  def synchronized[T](body: => T): T = sys.error("synchronized")

  /** Tests whether the argument (`that`) is a reference to the receiver object (`this`).
   *
   *  The `eq` method implements an [[https://en.wikipedia.org/wiki/Equivalence_relation equivalence relation]] on
   *  non-null instances of `AnyRef`, and has three additional properties:
   *
   *   - It is consistent: for any non-null instances `x` and `y` of type `AnyRef`, multiple invocations of
   *     `x.eq(y)` consistently returns `true` or consistently returns `false`.
   *   - For any non-null instance `x` of type `AnyRef`, `x.eq(null)` and `null.eq(x)` returns `false`.
   *   - `null.eq(null)` returns `true`.
   *
   *  When overriding the `equals` or `hashCode` methods, it is important to ensure that their behavior is
   *  consistent with reference equality.  Therefore, if two objects are references to each other (`o1 eq o2`), they
   *  should be equal to each other (`o1 == o2`) and they should hash to the same value (`o1.hashCode == o2.hashCode`).
   *
   *  @param  that    the object to compare against this object for reference equality.
   *  @return         `true` if the argument is a reference to the receiver object; `false` otherwise.
   */
  final def eq(that: AnyRef): Boolean = sys.error("eq")

  /** Equivalent to `!(this eq that)`.
   *
   *  @param  that    the object to compare against this object for reference equality.
   *  @return         `true` if the argument is not a reference to the receiver object; `false` otherwise.
   */
  final def ne(that: AnyRef): Boolean = !(this eq that)

  /** The expression `x == that` is equivalent to `if (x eq null) that eq null else x.equals(that)`.
   *
   *  @param    that  the object to compare against this object for equality.
   *  @return         `true` if the receiver object is equivalent to the argument; `false` otherwise.
   */
  final def ==(that: Any): Boolean =
    if (this eq null) that.asInstanceOf[AnyRef] eq null
    else this equals that

  /** Create a copy of the receiver object.
   *
   *  The default implementation of the `clone` method is platform dependent.
   *
   *  @note   not specified by SLS as a member of AnyRef
   *  @return a copy of the receiver object.
   */
  protected def clone(): AnyRef

  /** Called by the garbage collector on the receiver object when there
   *  are no more references to the object.
   *
   *  The details of when and if the `finalize` method is invoked, as
   *  well as the interaction between `finalize` and non-local returns
   *  and exceptions, are all platform dependent.
   *
   *  @note   not specified by SLS as a member of AnyRef
   */
  protected def finalize(): Unit

  /** Wakes up a single thread that is waiting on the receiver object's monitor.
   *
   *  @note   not specified by SLS as a member of AnyRef
   */
  final def notify(): Unit

  /** Wakes up all threads that are waiting on the receiver object's monitor.
   *
   *  @note   not specified by SLS as a member of AnyRef
   */
  final def notifyAll(): Unit

  /** See [[https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#wait--]].
   *
   *  @note   not specified by SLS as a member of AnyRef
   */
  final def wait (): Unit

  /** See [[https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#wait-long-int-]]
   *
   * @param timeout the maximum time to wait in milliseconds.
   * @param nanos   additional time, in nanoseconds range 0-999999.
   * @note not specified by SLS as a member of AnyRef
   */
  final def wait (timeout: Long, nanos: Int): Unit

  /** See [[https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#wait-long-]].
   *
   * @param timeout the maximum time to wait in milliseconds.
   * @note not specified by SLS as a member of AnyRef
   */
  final def wait (timeout: Long): Unit
}
```

raises an exception in `scala.meta.internal.semanticdb.scalac.TypeOps$XtensionGTypeSType.toSemanticTpe`

## Stack trace:
```
scala.sys.package$.error(package.scala:27)
	scala.meta.internal.semanticdb.scalac.TypeOps$XtensionGTypeSType.loop$1(TypeOps.scala:74)
	scala.meta.internal.semanticdb.scalac.TypeOps$XtensionGTypeSType.toSemanticTpe(TypeOps.scala:76)
	scala.meta.internal.semanticdb.scalac.TypeOps$XtensionGTypeSType.loop$2(TypeOps.scala:104)
	scala.meta.internal.semanticdb.scalac.TypeOps$XtensionGTypeSType.toSemanticSig(TypeOps.scala:127)
	scala.meta.internal.semanticdb.scalac.SymbolInformationOps$XtensionGSymbolMSymbolInformation.sig(SymbolInformationOps.scala:116)
	scala.meta.internal.semanticdb.scalac.SymbolInformationOps$XtensionGSymbolMSymbolInformation.toSymbolInformation(SymbolInformationOps.scala:196)
	scala.meta.internal.semanticdb.scalac.TextDocumentOps$XtensionCompilationUnitDocument$traverser$3$.saveSymbol$1(TextDocumentOps.scala:176)
	scala.meta.internal.semanticdb.scalac.TextDocumentOps$XtensionCompilationUnitDocument$traverser$3$.trySymbolDefinition(TextDocumentOps.scala:178)
	scala.meta.internal.semanticdb.scalac.TextDocumentOps$XtensionCompilationUnitDocument$traverser$3$.tryFindMtree(TextDocumentOps.scala:273)
	scala.meta.internal.semanticdb.scalac.TextDocumentOps$XtensionCompilationUnitDocument$traverser$3$.traverse(TextDocumentOps.scala:554)
	scala.meta.internal.semanticdb.scalac.TextDocumentOps$XtensionCompilationUnitDocument$traverser$3$.traverse(TextDocumentOps.scala:165)
	scala.reflect.api.Trees$Traverser.$anonfun$traverseStats$2(Trees.scala:2523)
	scala.reflect.api.Trees$Traverser.atOwner(Trees.scala:2532)
	scala.reflect.api.Trees$Traverser.$anonfun$traverseStats$1(Trees.scala:2523)
	scala.reflect.api.Trees$Traverser.traverseStats(Trees.scala:2522)
	scala.reflect.internal.Trees$PackageDef.traverse(Trees.scala:349)
	scala.reflect.internal.Trees.itraverse(Trees.scala:1613)
	scala.reflect.internal.Trees.itraverse$(Trees.scala:1612)
	scala.reflect.internal.SymbolTable.itraverse(SymbolTable.scala:28)
	scala.reflect.internal.SymbolTable.itraverse(SymbolTable.scala:28)
	scala.reflect.api.Trees$Traverser.traverse(Trees.scala:2500)
	scala.meta.internal.semanticdb.scalac.TextDocumentOps$XtensionCompilationUnitDocument$traverser$3$.traverse(TextDocumentOps.scala:625)
	scala.meta.internal.semanticdb.scalac.TextDocumentOps$XtensionCompilationUnitDocument.toTextDocument(TextDocumentOps.scala:628)
	scala.meta.internal.pc.SemanticdbTextDocumentProvider.textDocument(SemanticdbTextDocumentProvider.scala:54)
	scala.meta.internal.pc.ScalaPresentationCompiler.$anonfun$semanticdbTextDocument$1(ScalaPresentationCompiler.scala:469)
```

## Exception:
```
java.lang.RuntimeException: unsupported type Any {
def i n i t (): Unit
def equals(that: Any): Boolean
def hashCode: Int
def toString: String
def synchronized[T](body: => T): T
final def eq(that: AnyRef): Boolean
final def ne(that: AnyRef): Boolean
final def ==(that: Any): Boolean
protected def clone(): AnyRef
protected def finalize(): Unit
final def notify(): Unit
final def notifyAll(): Unit
final def wait(): Unit
final def wait(timeout: Long, nanos: Int): Unit
final def wait(timeout: Long): Unit
}: ClassInfoType(List(TypeRef(ThisType(scala), scala.Any, List())), Scope(TermName("$init$"), TermName("equals"), TermName("hashCode"), TermName("toString"), TermName("synchronized"), TermName("eq"), TermName("ne"), TermName("$eq$eq"), TermName("clone"), TermName("finalize"), TermName("notify"), TermName("notifyAll"), TermName("wait"), TermName("wait"), TermName("wait")), TypeName("AnyRef"))
```

## Versions
`semanticDb`: `4.11.0`
`scala`: `2.13.14`

## Additional Context:
This issue came up using metals in a work project--see the [metals issue](https://github.com/scalameta/metals/issues/6816) for more context and discussion. I've opened up a draft MR that simply accounts for `ClassInfoType` within `toSemanticTpe`, returning `NoType` and this, at least,
prevents Metals from crashing and doesn't seem to have an apparent negative effect; I haven't found a good place to add a test to "spec" this though I'd be happy to hear any suggestions. I have a feeling 
that that is probably a "hacky fix." If anyone would have any suggestions for troubleshooting this further to come up with
a more satisfactory solution, I'd be very happy to help :).

masonedmison:
Seems to fix #4084 though I haven't figured out a way to unit test this yet--suggestions welcome!

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TypeOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TypeOps.scala
@@ -71,6 +71,7 @@ trait TypeOps {
          */
         case g.WildcardType => s.NoType
         case _: g.BoundedWildcardType => s.NoType
+        case _: g.ClassInfoType => s.NoType
         case gother => sys.error(s"unsupported type $gother: ${g.showRaw(gother)}")
       }
       loop(gtpe)


tgodzik (dismissed):
Thanks for looking into this!

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TypeOps.scala
```diff
@@ -71,6 +71,7 @@ trait TypeOps {
          */
         case g.WildcardType => s.NoType
         case _: g.BoundedWildcardType => s.NoType
+        case _: g.ClassInfoType => s.NoType
```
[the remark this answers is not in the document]
tgodzik: If you don't have the time to add the test, do let me know!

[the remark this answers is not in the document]
masonedmison: @tgodzik Sounds good! I should have some time to try to spec this later this afternoon.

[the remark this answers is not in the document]
masonedmison: @tgodzik Maybe a stupid question, but I can't seem to actually run any of the test cases that use `targeted` from `SemanticdbSuite`, e.g.,
```
sbt:scalametaRoot> testsSemanticdb/testOnly *Issue2024Suite
[info] Passed: Total 0, Failed 0, Errors 0, Passed 0
```

It seems that none of the tests defined using `targeted` are excited :thinking:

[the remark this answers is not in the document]
masonedmison: I added a first attempt at a test as part of [this commit](https://github.com/scalameta/scalameta/commit/03249846bc08a16b650bf74eee90e2665e79c74a) but simply pasting in `AnyRef.scala` fails to compile with
```
<input>:4:7: error illegal cyclic inheritance involving type AnyRef
```.

I could probably use some help here with this if you would have the time.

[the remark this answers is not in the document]
Comment (tgodzik) [reply semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TypeOps.scala:74]: You might need to change the name to not conflict with the standard AnyRef

As for the tests this might work:
`testsSemanticdb/testOnly -- *Issue2024Suite`

Or a full name of the suite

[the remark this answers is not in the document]
Comment (kitbellew) [reply semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TypeOps.scala:74]: > You might need to change the name to not conflict with the standard AnyRef

I think the whole point is that code exists which uses the standard `AnyRef`, and @masonedmison is trying to make sure semanticdb doesn't fail on it.

As to the error on cyclic inheritance: do a web search or ask an LLM what to do about it, they will likely help.

Is it possible that excluding the entire `ClassInfoType` is not the right thing to do, as it doesn't just pertain to this case but to all inheritance cases? Perhaps something more narrow?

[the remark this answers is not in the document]
Comment (masonedmison) [reply semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TypeOps.scala]: @kitbellew I just pushed up a commit that matches the `typeSymbol` of the `ClassInfoType` value. Please correct me if there is a better way to do this.

File: semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TypeOps.scala [hunk 2]
```diff
@@ -71,6 +71,7 @@ trait TypeOps {
          */
         case g.WildcardType => s.NoType
         case _: g.BoundedWildcardType => s.NoType
+        case g.ClassInfoType(_, _, sym) if sym.name == g.TypeName("AnyRef") => s.NoType
```
tgodzik, line 74: I think it's better to ignore than fail in this case, but let me double check what is going on here

masonedmison: @tgodzik Also, FWIW, [this](https://github.com/scalameta/scalameta/blob/main/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolInformationOps.scala#L108) is where a `TypeBounds` type is being set (since `symbolOf[AnyRef].isAliasType` returns true). When computing the [signature](https://github.com/scalameta/scalameta/blob/main/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TypeOps.scala#L102) for `TypeBounds` we eventually call `toSemanticTpe` on `slo` which is where the failure occurs.

We could potentially instead try to address this issue upstream of here, e.g., in the `preprocess` method.

masonedmison: > chatgpt says the `illegal cyclic reference` occurs when encountering symbol `AnyRef` aliased to `Object` while in a compiler which defines `AnyRef` differently (i.e. deriving from `Any`). basically, clash of different versions of scala. not sure what can be done about this.

@kitbellew Sorry, was this in response to [this](https://github.com/scalameta/scalameta/pull/4085#discussion_r1887098236)?

kitbellew: caveat: i don't know anything about this code.

here's what i see: whether we use `TypeBounds(tpe, tpe)` or just `tpe`, `toSemanticTpe` will be called anyway (where you linked, or on line 125 a bit below).

perhaps in `SymbolInformationOps`, in the `preprocess` method, we should call `dealias` (or `dealiasWiden` which looks to be recursive)? i mean, if it's an alias... another option is, in `scala.reflect.internal.Symbols`, there's some code which checks for `isAliasType` and then calls `resultType`, perhaps that's the way.

something like:
```
def preprocess(info: g.Type): g.Type = info.dealiasWiden match {
  case g.PolyType(tparams, tpe) => g.PolyType(tparams, preprocess(tpe))
  case tpe => g.TypeBounds(tpe, tpe)
}
```

masonedmison: @kitbellew Ah apologies, I lazily glanced at the git blame and saw your name, but it was just a "formatting commit." It looks like this `preprocess`ing of  `Symbols` which are alias types was added as part of commit `6da0dbd88564` which was introduced in early 2018. Being that this has been in place for so long, and given the committer, I think I'm leaning again towards catching the failure within `toSemanticTpe`, as originally proposed.

masonedmison: @tgodzik 

> I think it's better to ignore than fail in this case, but let me double check what is going on here

Have you had a chance to think more about the above?

tgodzik: I am looking into that right now, but since the test is not working properly I wend ahead to try and add it to expect tests, but that is somehow broken (most likely by me)

tgodzik: Good news is that I actually managed to fix save-expect, bad news this just doesn't compile at all :sweat_smile:

tgodzik: Let's just leave the case and give up on the test. It's already fixing things and should not break anything.

It's becoming a bit of waste of time. We can stop using semanticdb on library file altogether in Metals. We wouldn't have that case.

masonedmison: @tgodzik Do you want the to generally catch the `ClassInfoType`, or specifically target the case where `ClassInfoType(_, _, <anyref>)`?

masonedmison: Just let me know, and I'll clean up the commits and force push so we can get this in :)

tgodzik: Generally catch ClassInfoType. We would fail in this case, but I haven't seen that failure being useful ever.

masonedmison: I just pushed up the original commit, where we generally catch `ClassInfoType`. Let me know if you disagree

masonedmison: Done

tgodzik: No worries! It's great that you dig in so deep here, I hope it was at least a bit educational :sweat_smile: We've hit the usual case of trying to compile something we should not have to compile ever since AnyRef is something the compiler creates. This is why it got so complicated

tgodzik (approved):
Thanks again!

PR #4088, opened 2024-12-14
ScannerTokens: relax trailing comma NL constraint

kitbellew:
Once again, the role of this parser is not to complain about unambiguous contexts, relaxing this constraint will allow interpreting code which is "almost" correct and can be corrected easily by a tool like scalafmt or scalafix.


tgodzik (approved):
Makes sense, we will be able to automatically format instead of complaining.

PR #4104, opened 2025-01-02
TextDocumentOps: traverse function properly

### LINKED ISSUE #3975: Fix scala.meta.tests.semanticdb.OccurrenceSuite -> example/Synthetic.scala
It seems that yield now returns an additional local symbol for Scala 2.13.15, which is not correct.

```
  for {
    b <- scala.concurrent.Future.successful(2)
    if a < b
  } yield a
```
last a is found to be a separate symbol (which might be true internally)

kitbellew:
In a recent version of compiler, it uses incorrect position values for anonymous functions which leads to incorrect symbol matching for that function instead of its body.

Hence, make sure we don't allow the function to get to the fallback case in the match and explicitly traverse parameters and body separately.

Fixes #3975.


tgodzik (approved):
LGTM! Thanks for looking into this!

PR #4105, opened 2025-01-03
ScalametaParser: support named tuples in extractor patterns

Florian3k:
This PR adds support for named patterns in extractor patterns (e.g. when matching case classes).

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2820,6 +2820,11 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
   }
 
   /* In addition to above named tuple patterns are allowed*/
+  object seqOKWithNamed extends SeqContextSensitive {
+    val isSequenceOK = true
+    override val isNamedTupleOk: Boolean = true
+  }
+
   object noSeqWithNamed extends SeqContextSensitive {
     val isSequenceOK = false
     override def isNamedTupleOk: Boolean = true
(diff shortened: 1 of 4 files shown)


Florian3k:
Related: https://github.com/scalameta/scalafmt/issues/4713

scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
```diff
@@ -2820,6 +2820,11 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect) {
   }
 
   /* In addition to above named tuple patterns are allowed*/
+  object seqOKWithNamed extends SeqContextSensitive {
+    val isSequenceOK = true
+    override def isNamedTupleOk: Boolean = true
```
kitbellew, line 2825: why not add this to `seqOK` instead? if not possible, please add a test which demonstrates it.

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/NamedTuplesSuite.scala
```diff
@@ -116,4 +116,36 @@ class NamedTuplesSuite extends BaseDottySuite {
       )
     ))
   }
+
```
kitbellew, line 119: could you add more tests (if, of course, they are applicable), such as:
```
  test("1") {
    val code = "val Foo(name = name, id = id) = Foo(name = "123", id = 456)"
    val tree = ...
    runTestAssert[Stat](code)(tree)
  }

  test("2") {
    val code = """case (a = 123, b = (c = "123", d = 123)) =>"""
    val tree = ...
    runTestAssert[Case](code)(tree)
  }

  // in SuccessSuite
  test("3") {
    q"val Foo(name = $name, id = $id)" = q"""Foo(name = "123", id = 456)"""
    // assert on name and id
  }


```

kitbellew (changes requested):
thank you (and excellent use of CommonTrees shortcuts, too).

in addition to comments below, I would ask you to amend this PR by splitting into two commits:
- first adds tests and likely uses `runTestError` to demonstrate what doesn't work today
- second adds functionality and fixes the tests.

kitbellew (approved):

kitbellew:
@Florian3k thanks for the submission, i added the tests i mentioned and merged. feel free to remove your branch.

PR #4112, opened 2025-01-04
TextDocumentOps: capture lambdas for SAM

### LINKED ISSUE #3394: Support for SAM in symbol for semanticdb
Currently, when using code such as:
```
// applies to abstract classes too
trait MyTestTrait {
  def compute(input: Int): Int
}

// MyTestObject could be a class, trait, etc.
object MyTestObject {

  val test1: MyTestTrait = (input: Int) => input + 1
  def test2: MyTestTrait = (input: Int) => input + 2

  val test3: MyTestTrait = new MyTestTrait {
    def compute(input: Int): Int = input + 3
  }
  def test4: MyTestTrait = new MyTestTrait {
    def compute(input: Int): Int = input + 4
  }

}
```

There is no new symbol entry for the anonymous class coming from SAM.

We should do it similar as we for anonymous classes.

kitbellew:
Fixes #3394.


tests-semanticdb/src/test/resources/example/SingleAbstractMethod.scala
```diff
@@ -0,0 +1,16 @@
...
+
+class SingleAbstractMethod/*<=example.SingleAbstractMethod#*/ {
+
+  trait SAM/*<=example.SingleAbstractMethod#SAM#*/ {
+    def foo/*<=example.SingleAbstractMethod#SAM#foo().*/(a/*<=example.SingleAbstractMethod#SAM#foo().(a)*/: Int/*=>scala.Int#*/): Int/*=>scala.Int#*/
+  }
+  def withSAM/*<=example.SingleAbstractMethod#withSAM().*/(a/*<=example.SingleAbstractMethod#withSAM().(a)*/: SAM/*=>example.SingleAbstractMethod#SAM#*/) = a/*=>example.SingleAbstractMethod#withSAM().(a)*/.foo/*=>example.SingleAbstractMethod#SAM#foo().*/(0)
+
+  withSAM/*=>example.SingleAbstractMethod#withSAM().*/(_ + 1/*<=example.SingleAbstractMethod#SAM#*/)
+  withSAM/*=>example.SingleAbstractMethod#withSAM().*/((x: Int) => x - 1/*<=example.SingleAbstractMethod#SAM#*/)
```
tgodzik, line 11:
```suggestion
  withSAM/*=>example.SingleAbstractMethod#withSAM().*/((x: Int) => x - 1/*<=local0*/)
```
Otherwise we will find SAM type here, while it's not occurring here and it's just the type of the expression.

For annonymous classes we do locals also.

We would need to make sure isSemanticdbLocal method treats this as a local (since we can never really access this instance from the outside or anywhere for that metter )

tgodzik (changes requested):
Thanks for working on this! I think we need to change it a bit, do let me know if you need any help with it!

tests-semanticdb/src/test/resources/example/SingleAbstractMethod.scala
tgodzik: > first, since withSAM expects a SAM, isn't this a definition of a SAM type? if somebody was looking for places where SAM was instantiated, isn't this one of them? and if it's not, then i might not have fully understood the point of the original problem

This is essentially both a definition on a new SAM type and of an instance of that type. We want occurrence of that instance and type definition in symbols section.

> second, about anonymous classes. is that a separate point or related to the first?

I just mentioned it since it;s a very similar case and we do this already for them. 

> and regarding isSemanticdbLocal: what is that used for?

That's in SymbolOps class, the method is used to know which instances are local to a scope, so no need to name them properly (they are not accessible from the outside scope.) Though this might be more complex than just handling the case in that new method.

kitbellew: all addressed in the new patch, please check.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -77,6 +79,15 @@ trait TextDocumentOps {
       def addOccurrence(mpos: m.Position, gsym: g.Symbol, role: Role): Unit =
         addOccurrenceFromSemantic(mpos, gsym.toSemantic, role)
 
+      def addSamOccurrence(mpos: m.Position, gtree: g.Function): Boolean = gtree.tpe match {
+        case tpe: g.TypeRef if tpe.args.isEmpty =>
+          tpe.sym.toSemantic match {
+            case Symbols.None => false
+            case ssym => samoccurrences.update(mpos, (ssym, Role.DEFINITION)); true
```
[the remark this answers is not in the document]
tgodzik: Yes, it should be a definition, thanks!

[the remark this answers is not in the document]
kitbellew: should i interpret this as an approval, or is the full review still to be done? :)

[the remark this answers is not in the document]
kitbellew: > should i interpret this as an approval, or is the full review still to be done? :)

never mind :)

tests-semanticdb/src/test/resources/example/SingleAbstractMethod.scala
```diff
@@ -0,0 +1,16 @@
+package example
+
+class SingleAbstractMethod/*<=example.SingleAbstractMethod#*/ {
+
+  trait SAM/*<=example.SingleAbstractMethod#SAM#*/ {
+    def foo/*<=example.SingleAbstractMethod#SAM#foo().*/(a/*<=example.SingleAbstractMethod#SAM#foo().(a)*/: Int/*=>scala.Int#*/): Int/*=>scala.Int#*/
+  }
+  def withSAM/*<=example.SingleAbstractMethod#withSAM().*/(a/*<=example.SingleAbstractMethod#withSAM().(a)*/: SAM/*=>example.SingleAbstractMethod#SAM#*/) = a/*=>example.SingleAbstractMethod#withSAM().(a)*/.foo/*=>example.SingleAbstractMethod#SAM#foo().*/(0)
+
+  withSAM/*=>example.SingleAbstractMethod#withSAM().*/(_ + 1/*<=local0*//*<=example.SingleAbstractMethod#SAM#*/)
```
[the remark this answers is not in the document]
kitbellew: i can easily remove it, but if someone is looking for all derivations of `SAM`, is this nor required? or is there some other place where this happens?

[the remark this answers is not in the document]
kitbellew: and we don't need to `saveSymbol` or anything else like that for this `Function`? just add the `local` occurrence, and that's it?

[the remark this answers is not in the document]
tgodzik: I think that's it, at least that's what we are doing in the similar case of anonymous classes, so it would be consistent. And this is basically an anonymous class, which is being made from a function.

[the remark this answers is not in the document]
kitbellew: an anonymous class is a `DefTree` for which symbols are saved via `trySymbolDefinition` (and for `Function`, it's not called), and your logic was to register occurrence after the symbol was saved in `trySymbolDefinition`: https://github.com/scalameta/scalameta/pull/1902/files#diff-841ef40ee21ef3b23412eef14c7a959d54b607713a2e285a8793e2d3efbae6fb

not saying we should do anything here, as my understanding of this is extremely superficial, but observing the difference.

kitbellew:
@tgodzik updated the patch, it's now very simple. also tried adding symbols, as a test, got some `$anonfun` so probably not very interesting.

tests-semanticdb/src/test/resources/example/SingleAbstractMethod.scala
```diff
@@ -0,0 +1,16 @@
...
+  trait SAM/*<=example.SingleAbstractMethod#SAM#*/ {
+    def foo/*<=example.SingleAbstractMethod#SAM#foo().*/(a/*<=example.SingleAbstractMethod#SAM#foo().(a)*/: Int/*=>scala.Int#*/): Int/*=>scala.Int#*/
+  }
+  def withSAM/*<=example.SingleAbstractMethod#withSAM().*/(a/*<=example.SingleAbstractMethod#withSAM().(a)*/: SAM/*=>example.SingleAbstractMethod#SAM#*/) = a/*=>example.SingleAbstractMethod#withSAM().(a)*/.foo/*=>example.SingleAbstractMethod#SAM#foo().*/(0)
+
+  withSAM/*=>example.SingleAbstractMethod#withSAM().*/(_ + 1/*<=local0*/)
+  withSAM/*=>example.SingleAbstractMethod#withSAM().*/((x: Int) => x - 1/*<=local2*/)
+
+  def funcSAM/*<=example.SingleAbstractMethod#funcSAM().*/(y/*<=example.SingleAbstractMethod#funcSAM().(y)*/: Int/*=>scala.Int#*/) = y/*=>example.SingleAbstractMethod#funcSAM().(y)*/ */*=>scala.Int#`*`(+3).*/ 2
+  withSAM/*=>example.SingleAbstractMethod#withSAM().*/(funcSAM/*=>example.SingleAbstractMethod#funcSAM().*/)
```
tgodzik, line 14:
```suggestion
  withSAM/*=>example.SingleAbstractMethod#withSAM().*/(funcSAM/*=>example.SingleAbstractMethod#funcSAM().*//*<=local3*/)
```
any chance to also get the occurrence here?

tests-semanticdb/src/test/resources/metac.expect_2.13
```diff
@@ -82,6 +82,75 @@ Diagnostics:
...
+[5:12..5:13): a <= example/SingleAbstractMethod#SAM#foo().(a)
+[5:15..5:18): Int => scala/Int#
+[5:21..5:24): Int => scala/Int#
+[7:6..7:13): withSAM <= example/SingleAbstractMethod#withSAM().
+[7:14..7:15): a <= example/SingleAbstractMethod#withSAM().(a)
+[7:17..7:20): SAM => example/SingleAbstractMethod#SAM#
+[7:24..7:25): a => example/SingleAbstractMethod#withSAM().(a)
+[7:26..7:29): foo => example/SingleAbstractMethod#SAM#foo().
+[9:2..9:9): withSAM => example/SingleAbstractMethod#withSAM().
+[9:10..9:15): _ + 1 <= local0
```
tgodzik, line 138: We are missing local0 in symbols, so there is no way to know what is it's parent

tests-semanticdb/src/test/resources/metac.expect_2.13
```diff
@@ -82,6 +82,75 @@ Diagnostics:
...
+[9:12..9:13): + => scala/Int#`+`(+4).
+[10:2..10:9): withSAM => example/SingleAbstractMethod#withSAM().
+[10:10..10:27): (x: Int) => x - 1 <= local2
```
tgodzik, line 141: We are also missing local2 in symbols section

tgodzik (changes requested):

tests-semanticdb/src/test/resources/metac.expect_2.13
kitbellew: added, and in other places.

tests-semanticdb/src/test/resources/metac.expect_2.13
```diff
@@ -82,6 +82,79 @@ Diagnostics:
...
+  Int => scala/Int#
+example/SingleAbstractMethod#funcSAM().(y) => param y: Int
+  Int => scala/Int#
+example/SingleAbstractMethod#withSAM(). => method withSAM(a: SAM): Int
+  a => example/SingleAbstractMethod#withSAM().(a)
+  SAM => example/SingleAbstractMethod#SAM#
+  Int => scala/Int#
+example/SingleAbstractMethod#withSAM().(a) => param a: SAM
+  SAM => example/SingleAbstractMethod#SAM#
+local0 => val local $anonfun
```
kitbellew, line 118: is that a useful symbol?

tests-semanticdb/src/test/resources/example/SingleAbstractMethod.scala
```diff
@@ -0,0 +1,16 @@
...
+  trait SAM/*<=example.SingleAbstractMethod#SAM#*/ {
+    def foo/*<=example.SingleAbstractMethod#SAM#foo().*/(a/*<=example.SingleAbstractMethod#SAM#foo().(a)*/: Int/*=>scala.Int#*/): Int/*=>scala.Int#*/
+  }
+  def withSAM/*<=example.SingleAbstractMethod#withSAM().*/(a/*<=example.SingleAbstractMethod#withSAM().(a)*/: SAM/*=>example.SingleAbstractMethod#SAM#*/) = a/*=>example.SingleAbstractMethod#withSAM().(a)*/.foo/*=>example.SingleAbstractMethod#SAM#foo().*/(0)
+
+  withSAM/*=>example.SingleAbstractMethod#withSAM().*/(_ + 1/*<=local0*/)
+  withSAM/*=>example.SingleAbstractMethod#withSAM().*/((x: Int) => x - 1/*<=local1*/)
+
+  def funcSAM/*<=example.SingleAbstractMethod#funcSAM().*/(y/*<=example.SingleAbstractMethod#funcSAM().(y)*/: Int/*=>scala.Int#*/) = y/*=>example.SingleAbstractMethod#funcSAM().(y)*/ */*=>scala.Int#`*`(+3).*/ 2
+  withSAM/*=>example.SingleAbstractMethod#withSAM().*/(funcSAM/*=>example.SingleAbstractMethod#funcSAM().*//*<=local3*/)
```
kitbellew, line 14: added a local here as well

tests-semanticdb/src/test/resources/metac.expect_2.13
tgodzik: ideally it would point to `SingleAbstractMethod` as a parent such as:

```suggestion
local0 => val local: AnyRef with SingleAbstractMethod {}
  AnyRef => scala/AnyRef#
  SingleAbstractMethod => _empty_/SingleAbstractMethod#
```

kitbellew: done... i think.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -77,6 +79,28 @@ trait TextDocumentOps {
...
+          val gpos = gtree.pos
+          val gsym = gtree.symbol
+          def atPos(mpos: m.Position): Boolean = gsym.toSemantic match {
+            case Symbols.None => false
+            case ssym =>
+              samoccurrences.update(mpos, (ssym, Role.DEFINITION))
+              if (!shouldNotSaveSymbol(gsym) && !shouldNotSaveSemanticSymbol(ssym))
+                saveSymbolFromSemantic(gsym.setInfo(tpe), ssym)
+              true
+          }
```
tgodzik, line 93:
```suggestion
      def addSamOccurrence(gtree: g.Function): Boolean = gtree.tpe match {
        case tpe: g.TypeRef if tpe.args.isEmpty =>
          val gpos = gtree.pos
          val gsym = gtree.symbol
          val gsymInfo = gtree.attachments.get[global.SAMFunction] match {
            case None => tpe
            case Some(global.SAMFunction(tpe, sym, sym2)) =>
              sym2.info
          }
          def atPos(mpos: m.Position): Boolean = gsym.toSemantic match {
            case Symbols.None => false
            case ssym =>
              samoccurrences.update(mpos, (ssym, Role.DEFINITION))
              if (!shouldNotSaveSymbol(gsym) && !shouldNotSaveSemanticSymbol(ssym))
                saveSymbolFromSemantic(gsym.setInfo(gsymInfo), ssym)
              true
          }
```
this will get the type to be the actual anonymous class that is generated

tgodzik: We could also potentially use that  attachment instead of the other conditions to be sure we have a sam type

kitbellew: done. thank you for finding this.

tgodzik:
Looks like 2.11.12 doesn't have this, we can create a trait that gets the attachment type and for 2.11 does only return the type of the tree

tests-semanticdb/src/test/resources/metac.expect_2.13
```diff
@@ -82,6 +82,85 @@ Diagnostics:
...
+  Int => scala/Int#
+example/SingleAbstractMethod#withSAM(). => method withSAM(a: SAM): Int
+  a => example/SingleAbstractMethod#withSAM().(a)
+  SAM => example/SingleAbstractMethod#SAM#
+  Int => scala/Int#
+example/SingleAbstractMethod#withSAM().(a) => param a: SAM
+  SAM => example/SingleAbstractMethod#SAM#
+local0 => val local $anonfun: $anon
+  $anon => local1
+local1 => class $anon extends Object with SAM { +1 decls }
```
tgodzik, line 120: I think all three local that are shown in the test above should have the same definition.  And local1 doesn't exist in occurrences

tgodzik (commented):
It seemed to work when testing today. I can take a further look tomorrow

tests-semanticdb/src/test/resources/metac.expect_2.13
kitbellew: 1. should add `local1` to occurrences?
2. why should they have the same definition? i mean, they derive from the same synthetic type (local1) but they are different functions.

kitbellew: > 1. should add `local1` to occurrences?

specifically: `local1` doesn't have a fixed position, so where exactly should it be defined? I can't "put" it with the definition of `SAM` trait since that trait could possibly come from another file.

another option is this: instead of `gsym.setInfo(sam.tpe)` and saving the `sam` symbol, i can use `gsym.setInfo(sam.info)` and not save any additional symbols, it will look like this:
`local0 => val local $anonfun:  extends Object with SAM { +1 decls }` (two spaces after `:` and no `$anon` name for the synthetic class). that way, no position and no local for the synthetic class, and you get your `extends Object with SAM` (not `AnyRef` but `Object` though it's the same).

kitbellew:
> Looks like 2.11.12 doesn't have this, we can create a trait that gets the attachment type and for 2.11 does only return the type of the tree

do we need it for 2.11? if it doesn't have it, then is it possible that code using SAM would even exist?

tgodzik:
> > Looks like 2.11.12 doesn't have this, we can create a trait that gets the attachment type and for 2.11 does only return the type of the tree
> 
> do we need it for 2.11? if it doesn't have it, then is it possible that code using SAM would even exist?

I think we don't need it for 2.11.12 yeah

tests-semanticdb/src/test/resources/metac.expect_2.13
tgodzik: Currently, local0 etc. point to functions which actually get replaced later on with SAM type, so I don't think we need to save them at all into semanticdb. Just save the anonymous class with occurrence at the function position

kitbellew: i am sorry, i am not sure i understand it. could you take over the commit and adjust it yourself?

here's how i interpret what we have:
```
// this is the position range which defines `local0` which is...
[9:10..9:15): _ + 1 <= local0
// ... an instance of type `$anon` also known ...
local0 => val local $anonfun: $anon
  $anon => local1
// ... as anonymous class deriving from SAM
local1 => class $anon extends Object with SAM { +1 decls }
  Object => java/lang/Object#
  SAM => example/SingleAbstractMethod#SAM#
```

so `local0` doesn't point to a function but to a instance of a SAM type.

but i am not an expert in semanticdb, as you know, just very superficial, so if you know what you'd like to see there, to be able to use within metals, then please help me :)

tgodzik: I can finish it up. I mostly think local0 should have the type of local1, because local0 will not exist in the generated code.

kitbellew: so going back to yesterday's state: if they won't exist in the generated code, and if you don't prefer the more verbose current solution, which is -- importantly -- not too hacky, why can't local0 simply have the type `SAM`?

also, local1 is not a type, it's a label we assigned to something, so I'd volunteer a guess that, at best, it's a matter of how this is printed out.

tgodzik: > so going back to yesterday's state: if they won't exist in the generated code, and if you don't prefer the more verbose current solution, which is -- importantly -- not too hacky, why can't local0 simply have the type SAM?

The type of it is never really just sam. First it's a function then it's transformed by the compiler into the anonymous class. It would also not work currently with metals, where it's primarily used as we don't show references to type if that type doesn't exist in source code. We can only make it work with go to implementation, which this effectively is.

> also, local1 is not a type, it's a label we assigned to something, so I'd volunteer a guess that, at best, it's a matter of how this is printed out.

sure that might not matter actually that much, but we need it in occurrences to actually show it somewhere. We can't use local0 since there is no relation between the two.

kitbellew: > > also, local1 is not a type, it's a label we assigned to something, so I'd volunteer a guess that, at best, it's a matter of how this is printed out.
> 
> sure that might not matter actually that much, but we need it in occurrences to actually show it somewhere. We can't use local0 since there is no relation between the two.

ah, if the problem is showing `local1` in occurrences, that should be fairly easy, i think.

tgodzik: I have a fix ready, will open a separate PR in a bit.

tgodzik: Ok, so this is what will be most of use to us https://github.com/scalameta/scalameta/pull/4127/commits/df4b7b0a0a05525abbf7ebd5477f8157e264c8ec

Though, the 2.13 expect hasn't updated, so maybe something is still off

tgodzik (approved):
Thanks!

PR #4121, opened 2025-01-08
Refactor: simplify Metadata.registry annotation

jchyb:
We replace the List with varargs, so that we do not have to "unapply" the List apply signature in the quasiquote when reading the registry in a macro.


kitbellew (approved):
Lovely! Cryptic stuff is gone :)

kitbellew:
@jchyb feel free to delete your branch

PR #4127, opened 2025-01-10
TextDocumentOps: capture lambdas for SAM vol 2

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/3394.

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
@@ -86,12 +86,11 @@ trait TextDocumentOps {
           case Symbols.None =>
           case ssym =>
             samoccurrences.update(mpos, (ssym, Role.DEFINITION))
-            if (!shouldNotSaveSymbol(gsym) && !shouldNotSaveSemanticSymbol(ssym)) {
-              saveSymbolDo(sam)
-              saveSymbolFromSemantic(gsym.setInfo(sam.tpe), ssym)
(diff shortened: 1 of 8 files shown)


semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -86,12 +86,11 @@ trait TextDocumentOps {
...
             samoccurrences.update(mpos, (ssym, Role.DEFINITION))
-            if (!shouldNotSaveSymbol(gsym) && !shouldNotSaveSemanticSymbol(ssym)) {
-              saveSymbolDo(sam)
-              saveSymbolFromSemantic(gsym.setInfo(sam.tpe), ssym)
+            if (!shouldNotSaveSymbol(sam) && !shouldNotSaveSemanticSymbol(ssym)) {
+              saveSymbolFromSemantic(sam, ssym)
             }
         }
-        if (gpos.isDefined && (gsym ne null)) {
+        if (gpos.isDefined && (gsym ne null) && (sam ne null)) {
```
kitbellew, line 93: is it possible for `sam` to be null?

kitbellew (commented):
looks very simple :) i guess i was overthinking it.

one request: could you fold the third commit into the second (so we have one commit adding the examples, and the other implementing functionality and showing how the examples have changed)?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
tgodzik: Sure! Will do it once I get back to a computer today.

kitbellew:
i updated my own pr with the same change, just in case

Issue #4128, opened 2025-01-10
Please publish for Scala 2.13.16


SethTisue:
which is now on Maven Central, as per https://contributors.scala-lang.org/t/scala-2-13-16-release-planning/6773/21

bjaglin:
Hey @tgodzik ! The latest stable version of scalafix depends on 4.10.1, I hope you can backpublish that far. I'll tag a new Scalafix release with the the current or upcoming Scalameta release shortly for full 3.6.2 syntax support anyway, so if it's too much hassle, don't worry.

tgodzik:
> The latest stable version of scalafix depends on 4.10.1, I hope you can backpublish that far.

That's no problem! I just need to figure out some issues beforhand

tgodzik:
Looks like we are getting Scala JS issues:
```
[error] Found unknown label apply at source-/home/tgodzik/Documents/scalameta/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala,line-563,offset=18279: default8()
[error] Error while emitting LegacyScanner.scala
[error] Found unknown label apply at source-/home/tgodzik/Documents/scalameta/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyScanner.scala,line-563,offset=18279: default8()
[error] two errors found
[error] (treesJS / Compile / compileIncremental) Compilation failed
```

not sure how to fix that, but I can work around it

tgodzik:
I backpublished semanticdb for 4.12.x and for 4.10.1. Let me know if we need anything more

bjaglin:
All good for 2.13.16!

An unrelated question, that will help me sync the upcoming scalafix release: are you planning to cut 4.12.4 any time soon? It looks like all outstanding SIP-64 syntax issues have been addressed since 4.12.3.

kitbellew:
> All good for 2.13.16!
> 
> An unrelated question, that will help me sync the upcoming scalafix release: are you planning to cut 4.12.4 any time soon? It looks like all outstanding SIP-64 syntax issues have been addressed since 4.12.3.

in a couple of days, unless something else pops up

bjaglin:
> in a couple of days, unless something else pops up

Time flies :smile: Thanks!

PR #4129, opened 2025-01-10
chore: Add support for Scala 2.13.16

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/4128

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,16 @@ jobs:
         java:
           - '8'
           - '17'
+        exclude:
+          - type: 'JS'
+            scala: '2.13.16'
+        include:
+          - type: 'JS'
+            scala: '2.13.15'
+            java: 17
+          - type: 'JS'
+            scala: '2.13.15'
+            java: 8
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         command:
-          - "++2.13.15; download-scala-library; testsSemanticdb/test"
+          - "++2.13.16; download-scala-library; testsSemanticdb/test"
           - "communitytest/test"
           - "mima"
     steps:
@@ -38,7 +38,7 @@ jobs:
           - 'Native'
         scala:
           - '2.12.20'
-          - '2.13.15'
+          - '2.13.16'
           - '3.3.4'
         java:
           - '8'
@@ -69,9 +69,9 @@ jobs:
           - '2.12.16'
           - '2.12.17'
           - '2.12.18'
-          - '2.13.12'
           - '2.13.13'
           - '2.13.14'
+          - '2.13.15'
     steps:
       - uses: actions/checkout@v4
         with:
--- a/build.sbt
+++ b/build.sbt
@@ -77,8 +77,8 @@ Global / resolvers +=
   "scala-integration".at("https://scala-ci.typesafe.com/artifactory/scala-integration/")
 
 val commonJsSettings = Seq(
-  crossScalaVersions := List(LatestScala213, LatestScala212),
-  scalaVersion := LatestScala213,
+  crossScalaVersions := List("2.13.15", LatestScala212),
+  scalaVersion := "2.13.15",
   bspEnabled := false,
   scalaJSLinkerConfig := StandardConfig().withBatchMode(true),
   scalacOptions ++= {
--- a/build.sbt
+++ b/build.sbt
@@ -77,8 +77,8 @@ Global / resolvers +=
   "scala-integration".at("https://scala-ci.typesafe.com/artifactory/scala-integration/")
 
 val commonJsSettings = Seq(
-  crossScalaVersions := List("2.13.15", LatestScala212),
-  scalaVersion := "2.13.15",
+  crossScalaVersions := List(LatestScala213ForJS, LatestScala212),
+  scalaVersion := LatestScala213ForJS,
   bspEnabled := false,
   scalaJSLinkerConfig := StandardConfig().withBatchMode(true),
   scalacOptions ++= {
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
@@ -141,7 +141,14 @@ trait TextDocumentOps {
             names.result()
           }
           private def indexArgNames(mapp: m.Tree): Unit = mapp match {
-            case t: m.Init => margnames(t.pos.start) = getAssignLhsAndIndexRhs(t.argClauses)
+            case t: m.Init =>
+              val names = getAssignLhsAndIndexRhs(t.argClauses)
+              margnames(t.pos.start) = names
+              // 2.13.16 changed position
+              t.parent.foreach {
+                case n: m.Term.New => margnames(n.pos.start) = names
+                case _ =>
+              }
 
             case t: m.Term.Apply =>
               margnames(t.fun.pos.end) = getAssignLhsAndIndexRhs(t.argClause :: Nil)
--- a/tests-semanticdb/src/test/resources/metac.expect_2.13
+++ b/tests-semanticdb/src/test/resources/metac.expect_2.13
@@ -455,7 +455,7 @@ Occurrences:
 [42:21..42:25): head => scala/collection/IterableOps#head().
 
 Diagnostics:
-[3:0..3:33) [warning] Unused import of deprecated lazy value higherKinds: higherKinds no longer needs to be imported explicitly
+[3:22..3:33) [warning] Unused import of deprecated lazy value higherKinds: higherKinds no longer needs to be imported explicitly
 
 Synthetics:
 [39:2..39:9): locally => *[Unit]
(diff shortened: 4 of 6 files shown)


kitbellew:
1. the compiler moving around trees and positions is lovely.
2. native and js not published? @SethTisue

tgodzik:
> 1. the compiler moving around trees and positions is lovely.

That can happen, the current algorithm is a bit too fragile, but I think I have a fix.

> 2. native and js not published? @SethTisue

Yeah, we need to wait probably till monday, so I can work on the issue over the weekend

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -141,7 +141,14 @@ trait TextDocumentOps {
...
           }
           private def indexArgNames(mapp: m.Tree): Unit = mapp match {
-            case t: m.Init => margnames(t.pos.start) = getAssignLhsAndIndexRhs(t.argClauses)
+            case t: m.Init => 
+              // 2.13.16 changed position
+              t.parent.foreach{
+                case n: m.Term.New =>
+                  margnames(n.pos.start) = getAssignLhsAndIndexRhs(t.argClauses)
+              }
+              margnames(t.pos.start) = getAssignLhsAndIndexRhs(t.argClauses)
```
kitbellew, line 150: i am not sure if `foreach` will not throw if parent is not `new`... also, we call `getAssign...` twice.

perhaps this is safer:
```
val tree = t.parent match {
  case Some(p: m.Term.New) => p
  case _ => t
}
margnames(tree.pos.start) = getAssignLhsAndIndexRhs(t.argClauses)
```
unless you want to set to both, in which case:
```
val names = getAssignLhsAndIndexRhs(t.argClauses)
margnames(t.pos.start) = names
t.parent match {
  case Some(p: m.Term.New) => margnames(p.pos.start) = names
  case _ =>
}
```

kitbellew (changes requested):

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
kitbellew: should we also do it for `Ctor.Block` and `Mod.Annot`? the code that uses this is:
```
            if (gsym != null && (gsym.isMethod || gsym.isConstructor)) for {
              margnames <- margnames.get(gstart) ++ margnames.get(gpoint)
              margname <- margnames
              gparams <- gsym.paramss
              gparam <- gparams.find(_.name.decoded == margname.value)
            } success(margname, gparam)
```
so it's any method or ctor with argument names.

p.s. perhaps there's a simpler way to do this...

SethTisue:
Scala.js has published (1.18.1, anyway). But no idea if Wojciech will get to Native before Monday. (But that's why we won't announce before Wednesday.O

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
tgodzik: Looks like the issue only happens with New, but I can check other possible problems

tgodzik: I added some tests to TargetedSuite and it seems to work. I can't figure out how to get Ctor.Body actually

tests-semanticdb/src/test/resources/metac.expect_2.13
```diff
@@ -1816,12 +1819,55 @@ Synthetics:
       b,
       c,
       d
-    ) == (1, 2, 3, 4))}).map[Tuple2[Tuple2[Int, Int], Tuple4[Int, Int, Int, Int]]]({(local8) => orig(e = (
+    ) == (1, 2, 3, 4))}).map[Tuple2[Tuple2[Int, Int], Tuple4[Int, Int, Int, Int]]]({(local8) => orig((
```
kitbellew, line 1822: looks like this logic had multiple redundant checks for `if (a, b, c, d) == (1, 2, 3, 4)` that got optimized away... is that how you understand it?

tests-semanticdb/src/test/resources/metac.expect_2.13
```diff
@@ -1816,12 +1819,55 @@ Synthetics:
...
+      b,
+      c,
+      d
+    ) == (1, 2, 3, 4)
+    e = (
       a,
       b,
       c,
       d
-    ))}).withFilter({(local9) => orig(e == (1, 2, 3, 4))}).flatMap[Tuple6[Int, Int, Int, Int, Tuple4[Int, Int, Int, Int], Tuple4[Int, Int, Int, Int]]]({(local10) => orig(f <- List(e)
```
kitbellew, line 1824: well, one extra check still remains here

.github/workflows/ci.yml
```diff
@@ -43,6 +43,16 @@ jobs:
...
+        exclude:
+          - type: 'JS'
+            scala: '2.13.16'
+        include:
+          - type: 'JS'
+            scala: '2.13.15'
+            java: 17
+          - type: 'JS'
+            scala: '2.13.15'
+            java: 8
```
kitbellew, line 55: we could also do (if i read the [doc](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#expanding-or-adding-matrix-configurations) correctly):
```
matrix:
  type: // no change
    ...
  scala: // remove 2.13
    - '2.12.20'
    - '3.3.4'
  java: // no change
    ...
  include:
    - scala '2.13.16'
      type: [ 'JVM', 'Native' ] 
    - scala '2.13.15'
      type: 'JS'
```

build.sbt
```diff
@@ -77,8 +77,8 @@ Global / resolvers +=
   "scala-integration".at("https://scala-ci.typesafe.com/artifactory/scala-integration/")
 
 val commonJsSettings = Seq(
-  crossScalaVersions := List(LatestScala213, LatestScala212),
```
kitbellew, line 80: should we add a `val LatestScala213ForJS`, perhaps? :)

tests-semanticdb/src/test/scala/scala/meta/tests/semanticdb/TargetedSuite.scala
```diff
@@ -238,6 +238,61 @@ class TargetedSuite extends SemanticdbSuite {
...
+         |_empty_/CLS#`<init>`(+1).(name2) => param name2: Int
+         |  Int => scala/Int#
+         |_empty_/CLS#name. => private[this] val method name: String
+         |  String => scala/Predef.String#
+         |local0 => val local x$1: "123"
+         |local1 => val local x$2: String
+         |  String => scala/Predef.String#
+         |
+         |Occurrences:
+         |[0:1..0:11): deprecated => _empty_/
```
kitbellew, line 278: why is this empty? is that expected?

tests-semanticdb/src/test/resources/metac.expect_2.13
tgodzik: I would guess so, but we don't really use those anyway.

tests-semanticdb/src/test/scala/scala/meta/tests/semanticdb/TargetedSuite.scala
tgodzik: Hmm... not really, but looks like an older bug, so maybe I will raise a separate issue? Seems to only happen when no package is defined

tests-semanticdb/src/test/resources/metac.expect_2.13
tgodzik: I think we could remove for comprehensions from synthetics, they are overly complex and unused.

.github/workflows/ci.yml
tgodzik: Doesn't work unfortunately, you need to specify everything and can't use an array, so this will be longer that now

build.sbt
tgodzik: Done

tests-semanticdb/src/test/scala/scala/meta/tests/semanticdb/TargetedSuite.scala
kitbellew: btw, looks like in 2.11 and 2.12 it's not empty but i don't know about 2.13.15... that's why i like to add tests first, to establish baseline, before making changes.

tests-semanticdb/src/test/resources/metac.expect_2.13
tgodzik: https://github.com/scalameta/scalameta/issues/4135

tests-semanticdb/src/test/scala/scala/meta/tests/semanticdb/TargetedSuite.scala
tgodzik: Uch, looks like that is broken also :/ This is a very unlucky release

tgodzik: `2.13.15` is the same. Will fix it separately

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
tgodzik: Looks like it's already covered in the test actually. I was using an older version of scalameta

tests-semanticdb/src/test/scala/scala/meta/tests/semanticdb/TargetedSuite.scala
tgodzik: https://github.com/scalameta/scalameta/issues/4136

kitbellew (approved):

kitbellew:
@tgodzik thank you, merged. can remove branch.

PR #4131, opened 2025-01-10
SymbolOps: synthetic local params are useless

kitbellew:


tests-semanticdb/src/test/resources/metac.expect_2.12
```diff
@@ -1662,7 +1654,7 @@ Synthetics:
   withFilter => scala/collection/TraversableLike#withFilter().
   b => local5
   flatMap => scala/collection/generic/FilterMonadic#flatMap().
-  check$ifrefutable$1 => local6
+  local6 => local6
```
kitbellew, line 1657: while here it was "used", the rest of them ("local7", "local8", 9, 10. etc) are all `localX => localX`, so this is consistent.

tests-semanticdb/src/test/resources/metac.expect_2.13
```diff
@@ -126,8 +126,6 @@ local3 => param x: Int
 local4 => class $anon extends Object with SAM { +1 decls }
   Object => java/lang/Object#
   SAM => example/SingleAbstractMethod#SAM#
-local5 => param y: Int
```
kitbellew, line 129: this is the one that started bothering me. it was here but not in occurrences (and you made a point before that a symbol without occurrences is useless).

tgodzik (approved):
Makes sense! LGTM

tests-semanticdb/src/test/resources/metac.expect_2.13
tgodzik: > and you made a point before that a symbol without occurrences is useless

Especially local ones, since we will never really find them or use them. Might be a bit different in thing like objects for case classes, but that should be at least in synthetics section

PR #4138, opened 2025-01-11
bugfix: Don't publish shared module for specific version

tgodzik:
It's currently not fully cross published and caused https://github.com/scalameta/scalameta/actions/runs/12726385364/job/35474709067


kitbellew (approved):
is it not used anywhere?

tgodzik:
It is, but it's also already published, so sonatype shows an error.

kitbellew:
is this specifically for backpublished tags? not for regular releases?

tgodzik:
> is this specifically for backpublished tags? not for regular releases?

Yes, that's right

PR #4173, opened 2025-01-28
TextDocumentOps: get pattern occurrences in ValDef

### LINKED ISSUE #4169: Wrong symbol in local deconstruction
```
class PatternMatching {
  def localDeconstruction = {
    val some = Option(123)
    val Some(number2/*local2*/) =
      some
    number2/*local1*/
    number2/*local1*/
  }
}
```

Semanticdb produces different symbols for number2 in `Some(...)` and the occurrences later on.

Looks like the first local variable in line 4 is from a synthetic tree that gets contructed:
```
val number2: Int = (some: Option[Int] @unchecked) match {
  case (value: Int): Some[Int]((number2 @ _)) => number2
}
```
My guess is that it actually points to number2 inside the synthetic match.

This got introduced in https://github.com/scalameta/scalameta/pull/4118/files

I added a test case in https://github.com/tgodzik/scalameta/pull/new/add-test-case

It's all a bit of a mess and I was not able to fix it aside from reverting the commit, but that breaks the new tests. In Metals however we use presentation compiler for any local symbols so in reality this should not break anything, but is a bug from perspective of any tool that would use it. I just have no idea how to solve it sensibly right now.

kitbellew:
Also, always traverse ValDef rhs first, as it helps with capturing pattern occurrences from LHS since for pattern occurrence conflicts the last one wins. Fixes #4169.

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SyntheticOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SyntheticOps.scala
@@ -43,6 +43,11 @@ trait SyntheticOps {
     def toRange: s.Range = toMeta.toRange
     def toSemanticOriginal: s.OriginalTree = toRange.toSemanticOriginal
 
+    def hasSyntheticSymbol: Boolean = gTree.symbol match {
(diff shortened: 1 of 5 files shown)


tgodzik (approved):
Thanks! LGTM!

PR #4176, opened 2025-01-30
ScannerTokens: handle `given` pattern after `case`

kitbellew:
Helps with scalameta/scalafmt#4764.


tgodzik (approved):
I thought we already supported that :O

PR #4179, opened 2025-02-03
scalajs: upgrade to v1.18.2

kitbellew:


tgodzik (approved):

tgodzik:
I think we should also bump minor so that people see that something more change.

kitbellew:
> I think we should also bump minor so that people see that something more change.

@bjaglin would you be ok with that?

tgodzik:
Actually, we should have done it on the 0.17 -> 0.18 bump but I guess that ship has sailed

bjaglin:
sure, I won't propagate the minor bump to the scalafix version since there is no impact (unless we use the opportunity of the minor bump to actually introduce other changes obviously, but I'll be looking at the changelog)

kitbellew:
> Actually, we should have done it on the 0.17 -> 0.18 bump but I guess that ship has sailed

who has version 0.18?

tgodzik:
I mean `1.18.0`, sorry about that. We did upgrade it a while back. Since people with 1.17.x will no longer be able to use scalameta libs it's a bit of an incompatibility.

PR #4225, opened 2025-03-31
improvement: Explain how to back publish a version

tgodzik:
And remove additional code

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,3 +126,12 @@ yarn start
 
 Now visit http://localhost:3000 and you should see a local version of the
 website. New changes should trigger a reload in the browser.
+
+## Back publishing
+
+To release a new version of the SemanticDB compiler plugin, you need to back publish it for a specific Scala version. To do this push a tag to the scalameta repository in a form of:
+
+  `v4.13.43@2.13.17@releaseSemanticdb`
+
+This will trigger a release workflow that will publish the plugin artifact to
+Sonatype. Replace `v4.13.4` and `2.13.17` with the version of the release you are making.
--- a/build.sbt
+++ b/build.sbt
@@ -39,6 +39,16 @@ enablePlugins(ScalaUnidocPlugin)
 addCommandAlias("benchAll", benchAll.command)
 addCommandAlias("benchLSP", benchLSP.command)
 addCommandAlias("benchQuick", benchQuick.command)
+commands += Command.command("releaseSemanticdb")(s =>
+  List(
+    "semanticdbSharedJVM",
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -126,3 +126,12 @@ yarn start
 
 Now visit http://localhost:3000 and you should see a local version of the
 website. New changes should trigger a reload in the browser.
+
+## Back publishing
+
+To release a new version of the SemanticDB compiler plugin, you need to back publish it for a specific Scala version. To do this push a tag to the scalameta repository in a form of:
+
+  `v4.13.43@2.13.17@semanticdbScalacCore/publishSigned@semanticdbScalacPlugin/publishSigned`
```
kitbellew, line 134: what causes this to be processed as described? also, it looks like this includes some sbt targets, but lists core in addition to scalacPlugin (and not metac etc. that were in the code above), why is that?

tgodzik: It's here https://github.com/sbt/sbt-ci-release?tab=readme-ov-file#back-publishing-support 

I can add all the sbt targets, I haven't because they are not needed for Metals in reality.

kitbellew: > It's here https://github.com/sbt/sbt-ci-release?tab=readme-ov-file#back-publishing-support

- in this document, it seems that only one command is allowed, am i misreading this?
- does `semanticdbScalacCore/publishSigned@semanticdbScalacPlugin/publishSigned` really work, in practice?
- and why isn't `semanticdbScalacPlugin/publishSigned` sufficient for publishing just the plugin?
- if we need to run multiple -- or lengthy -- commands, would it make sense to add an alias to `build.sbt`, to avoid "fat fingers"?

tgodzik: I think it works by switching @ to ;, but it might be a good idea to just create a command.

tgodzik: Changed to using a single command, should be much nicer to work with now.

build.sbt
```diff
@@ -39,6 +39,11 @@ enablePlugins(ScalaUnidocPlugin)
 addCommandAlias("benchAll", benchAll.command)
 addCommandAlias("benchLSP", benchLSP.command)
 addCommandAlias("benchQuick", benchQuick.command)
+commands += Command.command("releaseSemanticdb")(s =>
+  "semanticdbSharedJVM/publishSigned" :: "semanticdbScalacPlugin/publishSigned" ::
```
kitbellew, line 43: this begs for `(a :: b :: ...).map(_ + "/publishSigned") ++ s`, or something like that

tgodzik: done!

PR #4231, opened 2025-04-20
Trees: Term.CapSetName for capture polymorphism

kitbellew:


scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -295,6 +295,8 @@ object Term {
   class Super(thisp: sm.Name, superp: sm.Name) extends Term.Ref
   @ast
   class Name(value: Predef.String @nonEmpty) extends sm.Name with Term.Ref with Pat
+  @ast // https://scala-lang.org/api/3.x/docs/docs/reference/experimental/cc.html#capability-polymorphism-1
```
kitbellew, line 298: another alternative is not to introduce a new tree but instead use `Type.Name("X^")` for type parameters and `Term.Name("X^")` for capabilities within a `Type.Capturing`... basically, parse `Token.Ident("X")` and `Token.Ident("^")` and merge them into a name tree.

perhaps it's simpler.

tgodzik (approved):
LGTM

PR #4241, opened 2025-05-12
Fix issue with not recognizing dialects in post-3.4.0 versions of Scala

jchyb:
It looks like implicit vals in scala 3.3.x are unpickled into a methods, and in later versions they are unpickled into vals. To confirm that the imported dialect can be found in the proper package, we now just check whether it's a declaration there, regardless of it being a method or a val.

I did not add any regression tests here, as the problem appeared only in the scala versions not used in the build, so they would be ineffective here.


kitbellew:
> I did not add any regression tests here, as the problem appeared only in the scala versions not used in the build, so they would be ineffective here.

in that case, I'd start by adding a newer scala version into the build, then a test that _fails_, then a fix showing that the test becomes successful.

tgodzik:
Let me add that separately.

tgodzik:
Ok, so if you switch to version 3.7.0 things don't even compile without this change. So instead I added Scala 3.7.0 to tests and run it as an additional case. Though there is some weirdness around  sbt here, so might need to dig in a bit more

tgodzik:
Ok, tests seems to work. Without @jchyb fix it wouldn't actually even compile the tests

build.sbt
```diff
@@ -517,7 +518,8 @@ lazy val sharedSettings = Def.settings(
     else List(compilerPlugin("org.scalamacros" % "paradise" % "2.1.1" cross CrossVersion.full))
   },
   scalacOptions ++= { if (isScala213.value) List("-Ymacro-annotations") else Nil },
-  scalacOptions ++= { if (isScala213or3.value) List("-Xfatal-warnings") else Nil },
+  scalacOptions ++=
+    { if (isScala213or3.value && !isScala3Next.value) List("-Xfatal-warnings") else Nil },
```
tgodzik, line 522: There is a bunch of migration warnings, that we don't want to change yet

kitbellew:
is this not something that could be tested, explicitly, first with `typecheckError` (like in `quasiquotes/ErrorSuite`), and then, after the fix, moved to `DottySuccessSuite`? if not, then lgtm.

tgodzik (approved):

PR #4242, opened 2025-05-15
Fix quasiquote parsing in inlined code

jchyb:
After inlining, the position shown by `Position.ofMacroExpansion` would change, breaking the scalameta quasiquote parser and having it result with errors. Now we use the StringContext position, as that does not seem to be affected by inlining.

### DIFF
--- a/scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ReificationMacros.scala
+++ b/scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ReificationMacros.scala
@@ -169,29 +169,18 @@ class ReificationMacros(using val topLevelQuotes: Quotes) {
     reifySkeleton(skeleton, mode, qType, dialectExpr, input)
   }
 
-  private def metaInput() = {
-    val pos = Position.ofMacroExpansion
+  private def metaInput()(using strCtxExpr: Expr[StringContext]) = {
+    val pos = strCtxExpr.asTerm.pos
     val reflectInput = pos.sourceFile
     val content = reflectInput.content.getOrElse(throwSourceFileError())
-    val start = {
-      var i = pos.start
-      while (content(i) != '"') i += 1 // skip method name
-      while (content(i) == '"') i += 1 // skip quotations
-      i
-    }
-    val end = {
-      var i = pos.end - 1
-      while (content(i) == '"') i -= 1 // skip quotations
-      i + 1
-    }
     val metaInput = Input.VirtualFile(reflectInput.path, content)
-    Input.Slice(metaInput, start, end)
+    Input.Slice(metaInput, pos.start, pos.end)
   }
 
-  private def isMultiline() = {
-    val pos = Position.ofMacroExpansion
+  private def isMultiline()(using strCtxExpr: Expr[StringContext]) = {
+    val pos = strCtxExpr.asTerm.pos
     val content = pos.sourceFile.content.getOrElse(throwSourceFileError())
-    content(pos.start + 1) == '"' && content(pos.start + 2) == '"'
+    content(pos.start - 1) == '"' && content(pos.start - 2) == '"'
   }
 
   private def extractModeTerm(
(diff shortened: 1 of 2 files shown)


tests/shared/src/test/scala-3/scala/meta/tests/quasiquotes/Scala3SpecificSuccessSuite.scala
kitbellew, line 1: why do we need a separate for for this, and not DottySuccessSuite?

tests/shared/src/test/scala-3/scala/meta/tests/quasiquotes/Scala3SpecificSuccessSuite.scala
```diff
@@ -0,0 +1,24 @@
+package scala.meta.tests
+
+package quasiquotes
+
+import scala.meta._
+import scala.meta.dialects.Scala3
+
+class Scala3SpecificSuccessSuite extends TreeSuiteBase {
+
+  test("multiline - parse when inlined") {
```
kitbellew, line 10: where's the failed assert here? also, when the fix is added in the second commit, why doesn't this test change, to show improvement?

kitbellew (changes requested):

tests/shared/src/test/scala-3/scala/meta/tests/quasiquotes/Scala3SpecificSuccessSuite.scala
jchyb: Because scala2 does not support the  inline keyword, so we would end up with compilation failure. DottySuccessSuite only seems to relate to the used dialect, not the scala version running.

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ReificationMacros.scala
```diff
@@ -167,29 +167,23 @@ class ReificationMacros(using val topLevelQuotes: Quotes) {
   private def throwSourceFileError() = // should never be reached
     throw new Exception("Source file contents could not be read while expanding quasiquote")
 
-  private def metaInput() = {
-    val pos = Position.ofMacroExpansion
+  private def metaInput(strCtx: Expr[StringContext]) = {
+    val pos = strCtx.asTerm.pos
     val reflectInput = pos.sourceFile
     val content = new String(reflectInput.content.getOrElse(throwSourceFileError()))
-    val start = {
```
kitbellew, line 174: why is start replaced with pos.start, but end is not replaced with pos.end? judging by code in isMultiline, pos.start points to quotes, yet we used to skip them in `start` (and still do in `end`).

tests/shared/src/test/scala-3/scala/meta/tests/quasiquotes/Scala3SpecificSuccessSuite.scala
```diff
@@ -13,22 +13,12 @@ class Scala3SpecificSuccessSuite extends TreeSuiteBase {
         class $name:
           val a: Int = 1
       """
-    assertEquals(
-      compileErrors("""inlinedMultiline(Type.Name("AAA"))"""),
-      s"""error: double quotes are not allowed in single-line quasiquotes
-         |package quasiquotes
-         |^""".stripMargin
-    )
+    inlinedMultiline(Type.Name("AAA"))
```
kitbellew, line 16: if there's no `assrtNoThrow`, i'd at least add a comment that _that's_ what we are testing.

kitbellew (changes requested):

scalameta/quasiquotes/shared/src/main/scala-3/scala/meta/internal/quasiquotes/ReificationMacros.scala
kitbellew: @jchyb not sure if you saw this question

kitbellew (approved):

kitbellew:
@jchyb  feel free to remove branch

PR #4246, opened 2025-05-26
Deterministic .semanticdb files

reimai:
We are using bazel as a build tool, which relies on compilation being idempotent. Unordered sequences in .semanticdb files result in unnecessary jar files changes and build cache misses. 

Here I've added some order to TextDocument serialization to speed up bazel builds with semanticdb.

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
@@ -43,19 +43,6 @@ trait TextDocumentOps {
   implicit class XtensionCompilationUnitDocument(unit: g.CompilationUnit) {
     def toTextDocument: s.TextDocument = toTextDocument(None)
 
-    private def orderOrElse[T](one: Ordering[T], another: Ordering[T]): Ordering[T] =
-      (x: T, y: T) => {
-        val res1 = one.compare(x, y)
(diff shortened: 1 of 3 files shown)


kitbellew:
@reimai how do these unordered sequences change between runs? isn't there some natural, _deterministic_ order of symbol traversal?

tgodzik:
Looks like all the cases we have were already sorted, since no changes were needed. Do you have an example where the order is not correct? I would assume like @kitbellew that the order depends on tree traversal which is stable.

reimai:
> @reimai how do these unordered sequences change between runs? isn't there some natural, _deterministic_ order of symbol traversal?

They do (we're using version 4.9.3, but I don't see changes since then), not every run, but the order changes, especially between builds on different machines.
 
```metap``` presents semanticdb content sorted, but if I look into these files using protoc, the order of occurrences is different:
[0a58b830-8df8-409e-88c0-cf2fe9d54769.txt](https://github.com/user-attachments/files/20480066/0a58b830-8df8-409e-88c0-cf2fe9d54769.txt)
[41d2fec1-5ba9-495f-862b-73f157488ffa.txt](https://github.com/user-attachments/files/20480067/41d2fec1-5ba9-495f-862b-73f157488ffa.txt)

reimai:
Yes, I've run sbt testsSemanticdb/test successfully after fixing format issues.

kitbellew (approved):

tgodzik:
We'll probably do a release soon, since there was a bunch of changes needed upstream.

PR #4253, opened 2025-06-03
ScalametaParser: case body can be a partial

### LINKED ISSUE #4252: Exception with valid Scala 3.6 code (nested case)
## Required: Configuration ##

```
version = "3.9.7"
runner.dialect = scala3
```

## Required: Command-line parameters ##

When I run scalafmt via CLI like this: `scalafmt [path to file]` (`.scalafmt.conf` being in `.`)

## Steps

Given code like this:
```scala
object FmtKo:
  def f: (Int, Int) => PartialFunction[String, Unit] =
    case (x, y) =>
      case "abc" => ()
```

## Problem

Scalafmt fails with the following error:
```
[path to file]:4: error: [dialect scala3] `outdent` expected but `case` found
      case "abc" => ()
      ^
```

## Expectation

I would like scalafmt to succeed. 

To be overly fastidious: and to keep the code in a compiling state, which I believe means not removing the line-break and indent before the second `case`.

## Workaround

Wrapping the inner case in `{}` allows scalafmt to succeed.

## Notes

I verified that the above code compiles (Scala 3.6.4).

kitbellew:
Fixes #4252.


tgodzik (approved):
LGTM!

PR #4265, opened 2025-06-05
Fix parsing of doc param with body on next line

### LINKED ISSUE #4264: Scaladoc param name is incorrect when description is started on next line
The following doc comment is currently not parsed correctly.
```
/**
  *   @param x
  *       This is a really long description
  */
case class A(x: String)
```

The name is parsed to `x\nThis` and the body `is a really long description` correct should be that the name is `x` and the body should be `This is a really long description`.

eejbyfeldt:
Fixes #4264


kitbellew (commented):
separate comment: is there any reason why you are using this older, unsupported scaladoc parser in contrib/? there's a newer, "official" version.

kitbellew:
>... Maybe marking deprecated could help people discover which parser to use.

#4266

scalameta/trees/shared/src/main/scala/scala/meta/contrib/ScaladocParser.scala
```diff
@@ -84,10 +84,10 @@ object ScaladocParser {
         require(kind.numberParameters == 2)
         // Multiple parameter doc tokens
         def parser[$: P] = {
-          def nameParser: P[String] = ((AnyChar ~ !" ").rep ~ AnyChar).!.map(_.trim)
+          def nameParser: P[String] = ((AnyChar ~ !(" " | "\n")).rep ~ AnyChar).!.map(_.trim)
```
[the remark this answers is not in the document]
eejbyfeldt: So by "showing how it fails" you mean that the spec should reflect the current (wrong) behavior instead of showing that it fails by failing?

tests/shared/src/test/scala/scala/meta/tests/contrib/ScaladocParserSuite.scala
```diff
@@ -236,4 +236,19 @@ class ScaladocParserSuite extends FunSuite {
...
   }
 
+  test("multiline param") {
+    val description = "This description starts on the next line."
+    val result = parseString(
+      s"""|/** @param a
+            *    $description
+            */
+         |case class foo(a: String)
+         |""".stripMargin
```
kitbellew, line 246: please run `./bin/scalafmt` on each commit and resubmit

eejbyfeldt: Done.

PR #4271, opened 2025-06-14
Trees: define Decl.GivenAnonymous

### LINKED ISSUE #4268: Abstract givens in trait cannot be anonymous error with scala 3.7.1 and scalafmt 3.9.7
Hello

and thank you very much for your nice work.

The following trait is ok for scala 3.7.1 but causes scalafmt to complain: 
`error: [dialect scala3] abstract givens cannot be anonymous`

```
  trait AbstractGivenTest {
    given ExecutionContext
  }
```

Best regards,
  Sven

kitbellew:
The `name` field of Decl.Given should have been a Name, not Term.Name, but can't change it now. Fixes #4268.


tgodzik (approved):
LGTM!

PR #2464, opened 2021-09-05
Add ApplyUsingTree that represents context parameter application

tanishiking:
Close https://github.com/scalameta/scalameta/issues/2460

Added Scala's `Synthetic` part:
Screen Shot 2021-09-05 at 16 40 24

PR #4272, opened 2025-06-16
Add ApplyContextTree for context parameter application

### LINKED ISSUE #2460: Improve the Synthetic's tree so it can represent the context param application (with `using`)
See the original discussion: https://github.com/lampepfl/dotty/pull/13288#discussion_r697623068

## Background
Scala3's SemanticDB extractor extracts context params application as `Synthetic`. https://github.com/lampepfl/dotty/pull/13288 

For example
```scala
def foo(x: Int)(using Int) = ???
import Foo.given Int
def m() = foo(0)

// synthetics
def m() = foo(0)(x)

Synthetic(
  <foo(0)>,
  ApplyTree(
    OriginalTree(<foo(0)>),
    List(
      IdTree(<x>),
    )
  )
)
```

## Potential problem
- In the above example, there's no way to tell the synthesized symbol `x` is an implicit parameter or context params.
  - Because the given argument is imported by `import Foo.given Int` and there's no SymbolInformation for `x` is available from the `TextDocument`.
  - If `x` is defined in the same `TextDocument` (e.g. `def m(using x: Int) = foo(0)`), we can lookup the SymbolInformation and see it's properties.
- Lack of the symbol's property could be a problem for
  - `metals`'s `show implicit arguments` feature will show `foo(0)(x)` which doesn't compile, instead of `foo(0)(using x)`.
  - Maybe in the future, some tools will suffer from the lack of information?

So the discussion is `Should Synthetic have complete information to re-construct the original program from the tree?`

## Counter argument
It might not a big problem for tooling. 

- Though `metals`'s `show implicit arguments` feature ends up showing `foo(0)(x)` which is uncompilable decoration, the decorations are never compiled.
- And there are no tools that use the Synthetic for transforming the original source, as far as I know.

## Potential solution
- Add a flag to `ApplyTree` so we can tell it is a context-param or implicit param.
- Insert `SymbolInformation` for the context-param argument.
  - which contradicts with the original design of `SymbolInformation`, but I think it works well.
    - > "Symbols" is a section of a TextDocument that stores information about Symbols that are defined in the underlying document. 
- Add a flag to `IdTree` ?

natsukagami:
As a continuation of #2464. Closes #2460. One step towards scala/scala3#22936.

## Changes

- [x] semanticdb: Add context apply tree
- [x] metap: Add support for printing ApplyContextTree
- [x] Specification

### DIFF
--- a/docs/semanticdb/specification.md
+++ b/docs/semanticdb/specification.md
@@ -1037,6 +1037,7 @@ message Tree {
     OriginalTree original_tree = 6;
     SelectTree select_tree = 7;
     TypeApplyTree type_apply_tree = 8;
+    ApplyContextTree apply_context_tree = 9;
   }
 }
 ```
@@ -1121,6 +1122,15 @@ message TypeApplyTree {
 
 A `TypeApplyTree` represents the type application of a method, providing that
 method with type arguments.
+
+An `ApplyContextTree` represents a contextual method application, using Scala 3 `using` syntax.
+
+```protobuf
+message ApplyContextTree {
+  Tree function = 1;
+  repeated Tree arguments = 2;
+}
+```
 
 ## Data Schemas
 
--- a/semanticdb/semanticdb/shared/src/main/proto/semanticdb.proto
+++ b/semanticdb/semanticdb/shared/src/main/proto/semanticdb.proto
@@ -396,10 +396,16 @@ message Tree {
     OriginalTree original_tree = 6;
     SelectTree select_tree = 7;
     TypeApplyTree type_apply_tree = 8;
+    ApplyContextTree apply_context_tree = 9;
   }
 }
 
 message ApplyTree {
+  Tree function = 1;
+  repeated Tree arguments = 2;
+}
+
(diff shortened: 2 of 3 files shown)


tgodzik (approved):
LGTM!

tgodzik:
I guess this would be easier for tools, since they wouldn't have to support a totally new tree node.

natsukagami:
Based on the earlier PR, we agreed on a new node instead of a property to be consistent with existing nodes.

I don't have strong preferences about this.

kitbellew:
> Based on the earlier PR, we agreed on a new node instead of a property to be consistent with existing nodes.
> 
> I don't have strong preferences about this.

1. which earlier PR?
2. if we were talking about a completely new entity, one that is today not represented by any tree, we could consider a new tree; but if this is already parsed as an apply, and the only thing missing is _what type_ of apply, then I'd argue that adding a field would be less disruptive to those who don't care and sufficient for those who do.

natsukagami:
#2464 was the earlier attempt

PR #4275, opened 2025-06-20
Semanticdb: add properties bitmask to ApplyTree

kitbellew:
Also remove recently added ApplyContextTree as no longer necessary. Follow-on to #4272.


kitbellew:
@natsukagami fyi

PR #4290, opened 2025-08-05
chore: Test with 3.7.3

tgodzik:


kitbellew:
build fails with a warning which is now treated as an error (since #4296). fix the problem in #4299.

project/Versions.scala
```diff
@@ -7,7 +7,7 @@ object Versions {
   val Scala213Versions = getVersions2(13, 13 to 16)
   val Scala3Versions = getVersions3(3 -> 6, 7 -> 2)
   val Scala2ReleaseCandidates = getReleaseCandidates(2)
-  val Scala3ReleaseCandidates = getReleaseCandidates(3)
+  val Scala3ReleaseCandidates = getReleaseCandidates(3, (7, 3, "RC2"))
```
kitbellew, line 10: you need to undo this line, and change to `7 -> 3` on line 8 above.

PR #4294, opened 2025-08-29
chore: Test with 3.3.7-RC1

tgodzik:


tgodzik:
Not sure why 3.7.2 failed, but it somehow started using 2.13, might be an issue with RC. I will take a look once the final release is out if this is still a problem

kitbellew:
> Not sure why 3.7.2 failed, but it somehow started using 2.13, might be an issue with RC. I will take a look once the final release is out if this is still a problem

a bit puzzling. you are not changing 3.7.x here but 3.3.x; also, your [3.7.3-rcX](#4290) works fine.

kitbellew:
according to chatgpt, we might not be able to do this until 3.7.3 (or add both 3.3.6 and 3.3.7-rc1 to EarliestScalaVersions in project/Versions).

here's the full answer, see if it's true:

1. Scala 3 “binary versioning”

In Scala 2 you probably know the rule: 2.13.0, 2.13.1, … all publish as _2.13 because they are binary compatible across the whole series.
Scala 3 does something similar, but with an extra layer:
	•	Each published artifact is suffixed _3 and includes a “binary compatible base version” internally.
	•	That base version isn’t just 3; it’s whatever the Scala compiler says it can read classfiles from.

So it’s not “every 3.x works with every other 3.x” — instead, the compiler team explicitly wires which ranges of Scala 3 versions are mutually compatible.

⸻

2. Where does this rule live?

It’s in the Scala 3 compiler itself, in the ScalaRelease tables inside the compiler code. Concretely:
	•	The compiler has a notion of the “Scala binary version” (scalaBinaryVersion in sbt) and also a “Scala epoch / patch compatibility list”.
	•	For each released compiler, the team declares which older minor versions it can consume artifacts from.

For example, in 3.7.2 the compiler was built with a table that says:
	•	“I can read artifacts built with the 3.3.x series up to 3.3.6.”
	•	So _3.3.6 is considered a valid cross-versioned artifact when compiling/running on 3.7.2.

But 3.7.2 has no knowledge of 3.3.7-RC1 — because that version didn’t exist yet when 3.7.2 was released.
So the compiler won’t claim compatibility, and sbt won’t consider _3.3.7-RC1 artifacts usable.

⸻

3. Why does this matter?

That explains the asymmetry you noticed:
	•	✅ 3.7.2 ↔ 3.3.6 → works, because the compiler compatibility table includes 3.3.6.
	•	❌ 3.7.2 ↔ 3.3.7-RC1 → fails, because no released Scala 3.7.x knows that 3.3.7-RC1 is binary-compatible.

Once 3.3.7 final comes out, newer Scala 3 compilers (e.g. 3.7.3 or 3.8.0) will update their tables and declare it compatible.

tgodzik:
This doesn't seem true. We don't have any such compatibility table. Newer versions do not know anything about 3.3.x 

My guess is that this will work with non RC version and just that there is some logic that doesn't work with non stable versions.

kitbellew:
i made a couple of changes to scala3 warnings (#4296) and which versions are built (#4295) and rebased this. now 3.3.7-RC1 build fails; the difference with #4290 build is this:

```
# 3.7.3-RC2
[info] Setting Scala version to 3.3.6 on 3 projects.
[info] Setting Scala version to 3.7.3-RC2 on 1 projects.
[info] Excluded 34 projects, run ++ 3.7.3-RC2 -v for more details.

# 3.3.7-RC1
[info] Setting Scala version to 3.3.7-RC1 on 1 projects.
[info] Excluded 37 projects, run ++ 3.3.7-RC1 -v for more details.
```

not sure why 3.3.6 is not used for 3.3.7-RC1 anymore.

PR #4315, opened 2025-09-18
Add Protobuf symbol kinds to SemanticDB

olafurpg:
Previously, SemanticDB only covered Java and Scala symbols. I am working
on adding Protobuf support in Metals so that you can navigate from Scala
to Protobuf, and I've hacked it by reusing Scala kinds, but it's cleaner
if we define Protobuf as part of the spec.

### DIFF
--- a/semanticdb/semanticdb/shared/src/main/proto/semanticdb.proto
+++ b/semanticdb/semanticdb/shared/src/main/proto/semanticdb.proto
@@ -269,9 +269,7 @@ message SymbolInformation {
     TRAIT = 14;
     INTERFACE = 18;
 
-    // =======================
     // Protobuf-specific kinds
-    // =======================
     MESSAGE = 22;
     // IMPORTANT! Java enums are classes with *enum* property, see
     // https://scalameta.org/docs/semanticdb/specification.html#symbolinformation-2


semanticdb/semanticdb/shared/src/main/proto/semanticdb.proto
```diff
@@ -267,6 +268,19 @@ message SymbolInformation {
     CLASS = 13;
     TRAIT = 14;
     INTERFACE = 18;
+
+    // =======================
+    // Protobuf-specific kinds
+    // =======================
+    MESSAGE = 19;
```
tgodzik, line 275: Looks like they are not properly sorted, should be
```suggestion
    MESSAGE = 22;
```

tgodzik: Looks similar to https://github.com/sbt/sbt/issues/6392 Eugene I think complained at some point that sbt installed by coursier causes a lot of issues, so that could be it as well

olafurpg: I see the CI errors, will get it fixed

PR #4319, opened 2025-10-01
improvement: Add an ability to release specific semanticdb version

tgodzik:
This enables us:
- release for a specific Scala version from a branch with needed fixes. Sometimes those are needed, in the most recent case to fix semanticdb plugin and update sbt for releasing
- not have to remember what exactly needs to be release for semanticdb
- set isSnapshot to false, which was breaking the release previously (might be sbt-ci-release bug)

This is only needed for releasing semanticdb, so I removed the project input parameter.

I am already testing it in https://github.com/scalameta/scalameta/actions/runs/18155195434/job/51673342558

### DIFF
--- a/.github/workflows/release-semanticdb.yml
+++ b/.github/workflows/release-semanticdb.yml
@@ -1,0 +1,48 @@
+name: Release semanticdb
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Scalameta version (e.g., 1.2.3)"
+        required: true
+      branch:
+        description: "Branch or tag to release from (e.g., v4.10.13)"
+        required: true
+      scala:
+        description: "Scala version (e.g., 2.13.16)"
+        required: false
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out tag v${{ github.event.inputs.version }}
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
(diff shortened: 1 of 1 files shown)


.github/workflows/release-custom.yml
```diff
@@ -1,16 +1,13 @@
-name: Release specific project/tag
+name: Release semanticdb
```
kitbellew, line 1: why not duplicate this workflow and then customize for semanticdb? i used it to release trees specifically in the past, so it's not just for semanticdb (and, judging by the change, not at all for semanticdb).

kitbellew (changes requested):

.github/workflows/release-custom.yml
tgodzik: ok sure, I didn't realise it was useful for trees. Added a separate workflow then.

.github/workflows/release-semanticdb.yml
```diff
@@ -0,0 +1,48 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Scalameta version (e.g., 1.2.3)"
+        required: true
+      branch:
+        description: "Branch or tag to release from (e.g., v4.10.13)"
+        required: true
```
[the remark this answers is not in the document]
tgodzik: It's almost always a new branch that I need to release it from, so I think I will leave it as required as a sanity check (meaning "are we sure we don't need to fix the tag?")

PR #4352, opened 2025-11-13
ast: use DialectOnly origin in `copy`, for `apply`

### LINKED ISSUE #4351: exception with nested inline definition when copy the outer definition
After importing the `Scala3` dialect, I define a method "m" with a nested _inline_ method "n", and do a noop copy - the same body of "m". Then I get an exception:

```scala
scala> import scala.meta._
                                                                                                                                      
scala> import dialects.Scala3

scala> Defn.Def(Nil, Term.Name("m"), Nil, None, Term.Block(Defn.Def(Mod.Inline() :: Nil, Term.Name("n"), Nil, None, Lit.Unit()) :: Lit.Unit() :: Nil))
val res0: scala.meta.Defn.Def = def m = {
  inline def n = ()
  ()

}
scala> res0.copy(body = res0.body)
java.lang.UnsupportedOperationException: Scala213 doesn't support inline modifiers

```

A workaround is to avoid `inline`-ing nested  methods.

kitbellew:
Fixes #4351.

### DIFF
--- a/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
+++ b/scalameta/common/shared/src/main/scala/scala/meta/internal/trees/ast.scala
@@ -140,6 +140,10 @@ class AstNamerMacros(val c: Context) extends Reflection with CommonNamerMacros {
           istats1 += declareGetter(p.name, p.tpt, mods)
           p
         }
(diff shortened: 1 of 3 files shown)


tgodzik (approved):
LGTM!

PR #4386, opened 2025-12-09
Trees: Type.Capturing now takes a Type.Captures

### LINKED ISSUE #4385: Empty capture sets
When working with capture sets, which scalameta appears to handle quite well, I've ran into an issue with empty capture sets - that is, things of the shape `Foo^{}`.

It's a bit of weird construct, but it's found in the [standard library](https://github.com/scala/scala3/blob/07883c1c2ded087826897c465940ff07379a3bac/library/src/scala/util/boundary.scala#L39), for example.

This confuses scalameta, which I've found out while using scalafmt and getting the following error:

```
[error] (Compile / scalafmt) org.scalafmt.sbt.ScalafmtSbtReporter$ScalafmtSbtError: scalafmt: [3.10.2] /path/to/Fail.scala:6: error: [dialect scala3future] `identifier` expected but `}` found
[error] private case class Fail(label: Label ^{}, message: String) extends ControlThrowable with NoStackTrace
```

kitbellew:
Capture checking is still an experimental, fluid feature, and in our earlier implementation we failed to anticipate empty capabilities sets so let's fix that now. Fixes #4385.

### DIFF
--- a/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
+++ b/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
@@ -812,10 +812,19 @@ object Type {
     final def stats: List[Tree] = typeDefs
   }
 
+  @branch
+  trait Captures extends Tree
+
+  @ast
+  class CapturesAny() extends Captures
+
+  @ast
+  class CapturesSet(values: List[Term.Ref]) extends Captures
+
   @ast
   class Capturing(
       tpe: Type,
-      caps: List[Term.Ref] // [cap]tures or [cap]abilities
+      caps: Captures // [cap]tures or [cap]abilities
   ) extends Type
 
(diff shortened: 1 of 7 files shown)


scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala
```diff
@@ -812,10 +812,19 @@ object Type {
...
+  class CapturesAny() extends Captures
+
+  @ast
+  class CapturesSet(values: List[Term.Ref]) extends Captures
+
   @ast
   class Capturing(
       tpe: Type,
-      caps: List[Term.Ref] // [cap]tures or [cap]abilities
+      caps: Captures // [cap]tures or [cap]abilities
```
tgodzik: Shouldn't this fail MiMa actually?

PR #4391, opened 2025-12-13
Tokenizer: add {Int,Float}XL tokens and trees

### LINKED ISSUE #4204: Metals may report false errors when the experimental Numeric Literals feature is enabled
### Describe the bug

When trying to use the Scala 3 experimental feature called "Numeric Literals" (see [https://docs.scala-lang.org/scala3/reference/experimental/numeric-literals.html](https://docs.scala-lang.org/scala3/reference/experimental/numeric-literals.html)), metals seems to have its own checks that defeats (some of) the purpose with that feature.

When the experimental feature is enabled, numeric literals should not be restricted in length. If it is not restricted in length, it looks like the compiler already yields an error (in addition to the error Metals reports).

Taking the example from the page describing Numeric Literals, we have something like:
```scala
package example

import scala.language.experimental.genericNumberLiterals
import scala.util.FromDigits

case class BigFloat(mantissa: BigInt, exponent: Int):
    override def toString = s"${mantissa}e${exponent}"

object BigFloat:
    import scala.util.FromDigits

    def apply(digits: String): BigFloat =
        val (mantissaDigits, givenExponent) =
            digits.toUpperCase.split('E') match
                case Array(mantissaDigits, edigits) =>
                    val expo =
                        try FromDigits.intFromDigits(edigits)
                        catch
                            case ex: FromDigits.NumberTooLarge =>
                                throw FromDigits.NumberTooLarge(s"exponent too large: $edigits")
                    (mantissaDigits, expo)
                case Array(mantissaDigits) =>
                    (mantissaDigits, 0)
        val (intPart, exponent) =
            mantissaDigits.split('.') match
                case Array(intPart, decimalPart) =>
                    (intPart ++ decimalPart, givenExponent - decimalPart.length)
                case Array(intPart) =>
                    (intPart, givenExponent)
        BigFloat(BigInt(intPart), exponent)

    given FromDigits.Floating[BigFloat]:
        def fromDigits(digits: String) = apply(digits)
end BigFloat

package example

import scala.language.experimental.genericNumberLiterals
import scala.util.FromDigits

case class BigFloat(mantissa: BigInt, exponent: Int):
    override def toString = s"${mantissa}e${exponent}"

object BigFloat:
    import scala.util.FromDigits

    def apply(digits: String): BigFloat =
        val (mantissaDigits, givenExponent) =
            digits.toUpperCase.split('E') match
                case Array(mantissaDigits, edigits) =>
                    val expo =
                        try FromDigits.intFromDigits(edigits)
                        catch
                            case ex: FromDigits.NumberTooLarge =>
                                throw FromDigits.NumberTooLarge(s"exponent too large: $edigits")
                    (mantissaDigits, expo)
                case Array(mantissaDigits) =>
                    (mantissaDigits, 0)
        val (intPart, exponent) =
            mantissaDigits.split('.') match
                case Array(intPart, decimalPart) =>
                    (intPart ++ decimalPart, givenExponent - decimalPart.length)
                case Array(intPart) =>
                    (intPart, givenExponent)
        BigFloat(BigInt(intPart), exponent)

    given FromDigits.Floating[BigFloat]:
        def fromDigits(digits: String) = apply(digits)
end BigFloat

val myValue: BigFloat = 1e10_0000_000_000
```

This compiles fine with SBT from a terminal, since the experimental feature has been enabled using: `import scala.language.experimental.genericNumberLiterals`

In VS Code using Metals, an error occurs: malformed floating-point number: Too many nonzero exponent digits.

Notice that if we remove the experimental feature import, we get one error compiling with SBT from a terminal:
```
36 |val myValue: BigFloat = 1e10_0000_000_000
   |                        ^^^^^^^^^^^^^^^^^
   |                        number too large
```
And inside VS Code with Metals, we get two errors:
 * number too large
 * malformed floating-point number: Too many nonzero exponent digits.

### Expected behavior

When the experimental feature is in scope, Metals should _not_ yield the "malformed floating-point number: Too many nonzero exponent digits."
While the error message from Metals is arguably slightly better then the one issued by the Scala compiler, maybe it should be considered to rely on the compiler yielding the proper error messages. Then you don't have to take things like imported experimental features into account.

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v1.5.1

### Extra context or search terms

I'm using Scala version 3.6.3 and SBT version 1.10.4.
The issue is the same no matter if I use Bloop or SBT as build server in Metals.

kitbellew:
In tokenizer, extract fractional-value tokens as Constant.FloatXL unless they have an explicit suffix identifying it as Float or Double, or if they are within Double range; at the same time, integer tokens that do not have a suffix will continue to be extracted as Constant.Int if they are within 32-bit, and as Constant.IntXL otherwise.

Add a new AnyDecimal "numeric" type which refers to arbitrary-precision decimal values.

Fixes #4204.

### DIFF
--- a/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyTokenData.scala
+++ b/scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyTokenData.scala
@@ -25,6 +25,8 @@ class LegacyTokenData {
   /** the base of a number */
   var base: Int = 0
 
+  var decimalExponent = ""
+
   override def toString =
     s"{token = $token, position = [$offset,$endOffset), strVal = $strVal, base = $base}"
 
@@ -33,7 +35,18 @@ class LegacyTokenData {
    */
   def charVal: Char = if (strVal.isEmpty) 0 else strVal.charAt(0)
 
-  private def toBigInt(prevToken: Token, maxBitLength: Int, what: String): Either[String, BigInt] =
+  /**
+   * @return
+   *   - left: error string
+   *   - right: parsed BigInt
+   *     - left: if the value is out of range
+   *     - right: if the value is correct and in range
+   */
+  private def toBigInt(
+      prevToken: Token,
+      maxBitLength: Int,
+      what: String
+  ): Either[String, Either[BigInt, BigInt]] =
     try {
       val value = BigInt(strVal, base)
       val isNegativeSign = prevToken match {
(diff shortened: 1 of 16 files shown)


scalameta/tokens/shared/src/main/scala/scala/meta/tokens/Token.scala
```diff
@@ -227,10 +227,14 @@ object Token {
...
     @freeform("long constant")
     class Long(value: BigInt) extends NumericConstant[BigInt]
+    @freeform("arbitrary integer constant")
+    class IntXL(value: BigInt) extends NumericConstant[BigInt]
     @freeform("float constant")
     class Float(value: BigDecimal) extends NumericConstant[BigDecimal]
     @freeform("double constant")
     class Double(value: BigDecimal) extends NumericConstant[BigDecimal]
+    @freeform("arbitrary decimal constant")
+    class FltXL(value: AnyDecimal) extends NumericConstant[AnyDecimal]
```
tgodzik:
```suggestion
    class FloatXL(value: AnyDecimal) extends NumericConstant[AnyDecimal]
```
Nit, but doesn't seem worth losing the two letters to make it shorter

tests/shared/src/test/scala-2/scala/meta/tests/prettyprinters/QuasiSyntacticSuite.scala
```diff
@@ -158,7 +158,6 @@ class QuasiSyntacticSuite extends scala.meta.tests.parsers.ParseSuite {
          |""".stripMargin
     assertStruct(q"val x = 1d")(expected)
     assertStruct(q"val x = 1D")(expected)
-    assertStruct(q"val x = 1.0")(expected)
```
tgodzik: why remove this one? Seems a valid case

scalameta/tokenizers/shared/src/main/scala/scala/meta/internal/tokenizers/LegacyTokenData.scala
```diff
@@ -42,28 +44,32 @@ class LegacyTokenData {
       prevToken: Token,
       maxBitLength: Int,
       what: String
-  ): Either[String, BigInt] = {
+  ): Either[String, Either[BigInt, BigInt]] = {
```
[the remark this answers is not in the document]
kitbellew: yes, we do.

PR #4429, opened 2026-01-15
ScannerTokens: further limit start of Given region

### LINKED ISSUE #4428: Parsing fails for named given arguments when used in import statement
The following code successfully compiles, but parsing fails when trying to format using scalafmt 3.10.3.

```scala
class Foo {
  def x(): Unit = ()
}

object Foo {
  def withFoo(f: Foo => Unit): Unit = ()

  // Here!!
  withFoo { case foo @ given Foo => 
    import foo.* // Works fine if I remove this line
    x()
  }
}
```

Output:

```
error: [dialect scala3] `}` expected but `identifier` found
    x()
```

Also fails with indentation-based syntax:

```scala
  withFoo:
    case foo @ given Foo =>
      import foo.*
      x()
```

Workaround is to either remove the `import` statement, or wrap `given Foo` in parenthesis as below:

```scala
  withFoo { case foo @ (given Foo) => 
    import foo.*
    x()
  } 
```

kitbellew:
Specify when it could happen, instead of excluding when it couldn't. Fixes #4428.


tgodzik (approved):
LGTM!

PR #4482, opened 2026-02-03
Comment out ellipsis tokens for quasiquotes

### LINKED ISSUE #4481: Quasiquotes using `..$` produce incorrect `text` and `syntax`
Quasiquotes with substitutions using `..$` are missing strings in their syntax representation. This is likely related to #4434 , which was resolved by https://github.com/scalameta/scalameta/pull/4435

```scala
import scala.meta._

val terms = List(q"Foo")
val code = q"""
..${terms}; "any message"
"""

println(code.syntax)
```

Results in
```
{
  Foo

}
```

You can reproduce this with `scala-cli`

```
scala-cli https://gist.github.com/zainab-ali/4a266814dcb02fb99b67d8210a913d0b
```

Or by running a test in  the scalameta codebase: https://github.com/scalameta/scalameta/compare/main...zainab-ali:quasiquote-bug

zainab-ali:
We are already backquoting the unquote token, thus replacing it with an Ident; we could also backquote the ellipsis as well but if the output were to be used not just for tokenization but also for parsing, it might result in two consecutive idents where in reality only one entity would be expected.

This is an attempt to reproduce and resolve #4481 by using the same logic as #4435 .

### DIFF
--- a/scalameta/tokens/shared/src/main/scala/scala/meta/internal/prettyprinters/TokensToString.scala
+++ b/scalameta/tokens/shared/src/main/scala/scala/meta/internal/prettyprinters/TokensToString.scala
@@ -40,6 +40,7 @@ object TokensToString {
             }
           }) idx += 1
         sb.append('`')
+      case t: Token.Ellipsis => sb.append("/* ").append(chars, t.start, t.len).append(" */")
       case t => sb.append(chars, t.start, t.len)
     }
     sb.toString
--- a/tests/shared/src/test/scala/scala/meta/tests/quasiquotes/DottySuccessSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/quasiquotes/DottySuccessSuite.scala
@@ -231,20 +231,20 @@ class DottySuccessSuite extends TreeSuiteBase {
     assertTokensAsStructureLines(
       quoted.tokens,
       """|BOF [0..0)
-         |Dot [0..1)
-         |Dot [1..2)
-         |Ident($terms) [2..10)
-         |Semicolon [10..11)
-         |Space [11..12)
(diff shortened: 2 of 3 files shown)


tests/shared/src/test/scala/scala/meta/tests/quasiquotes/DottySuccessSuite.scala
```diff
@@ -223,11 +223,33 @@ class DottySuccessSuite extends TreeSuiteBase {
     assertTree(quoted)(blk(tname("Foo"), lit("any message")))
   }
 
-  test("quasiquote in list substitution") {
+  test("#4481: quasiquote with ellipsis") {
     import dialects.Scala3
     val terms = List(q"Foo")
     val quoted: Tree = q"""..${terms}; "any message""""
 
+    assertTokensAsStructureLines(
```
kitbellew: the second commit should not contain any changes to tests other than to demonstrate what was fixed compared to the first commit. each commit must fully pass all tests on its own. 

here, you are changing the title, adding asserts on tokens but, most importantly, not changing any asserts which means either there was no bug to fix, or you didn't show it in the first commit -- it would fail tests.

scalameta/tokens/shared/src/main/scala/scala/meta/internal/prettyprinters/TokensToString.scala
```diff
@@ -40,6 +40,7 @@ object TokensToString {
             }
           }) idx += 1
         sb.append('`')
+      case t: Token.Ellipsis => sb.append("/* ").append(chars, t.start, t.len).append(" */")
```
kitbellew: i changed this part, see comments.

kitbellew (approved):

Issue #1334, opened 2018-02-17
Add support for Annotation.arguments


xeno-by:
This is a tough one. Storing full ASTs will blow up the schema and make it virtually impossible to specify the format comprehensively (specifying desugarings is known to be at least very hard and probably outright impractical). 

We probably should think about workarounds. Maybe introduce something similar to `ClassfileAnnotArg` in scalac? Maybe use the `Synthetic` loophole?

xeno-by:
Since we now have `Constant` data structures in our schema, it should be easy to support literal annotation arguments. That won't provide full support even for `ClassfileAnnotation` annotations, but it would be a quick win anyway.

kitbellew:
looks like this is related to #2281 and #4487 (which add a field but don't populate, though).

PR #4487, opened 2026-02-04
Semanticdb: add arguments to Annotation message

kitbellew:
Also, define AssignTree which is likely to be the parameter for it. Based on #2281, but without the MiMa-incompatible class rename. Helps with #1334.

### DIFF
--- a/semanticdb/semanticdb/shared/src/main/proto/semanticdb.proto
+++ b/semanticdb/semanticdb/shared/src/main/proto/semanticdb.proto
@@ -334,8 +334,10 @@ message Documentation {
   Format format = 2;
 }
 
+// TODO; rename to AnnotationTree
 message Annotation {
   Type tpe = 1;
+  repeated Tree arguments = 2;
 }
 
 message AssignTree {
@@ -418,6 +420,7 @@ message Tree {
     SelectTree select_tree = 7;
     TypeApplyTree type_apply_tree = 8;
     AssignTree assign_tree = 9;
+    Annotation annotation_tree = 10;
   }
 }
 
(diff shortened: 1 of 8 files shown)


semanticdb/semanticdb/shared/src/main/proto/semanticdb.proto
```diff
@@ -334,6 +334,12 @@ message Documentation {
 
 message Annotation {
   Type tpe = 1;
+  repeated Tree parameters = 2;
```
kitbellew: - should this be arguments, instead of parameters?
- @Strum355 , how will you populate this field, though, if it's added?

tgodzik: Probably arguments fits better. 

I think it's already implemented in https://github.com/sourcegraph/scip-java/pull/148/changes#diff-942a0d71385e5c2ab8ea74e2c3212a73c9e311c288d58ad283bbe54c0d03570e

kitbellew:
added code setting these arguments. ptal.

tests-semanticdb/src/test/resources/metac-metacp.diff_2.12
```diff
@@ -1414,6 +1426,7 @@ types/Test.C#annType1.
 -                type_ref {
 -                  prefix {
 -                  symbol: "scala/Int#"
+-          arguments {
```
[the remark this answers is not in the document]
kitbellew: scalapb prints these messages like that. i can't control how it does that.

https://github.com/scalapb/ScalaPB/blob/8efa08e57366279512c168b2b95a027cf6cccc76/scalapb-runtime/src/main/scala/scalapb/textformat/Printer.scala#L32

PR #4545, opened 2026-04-03
chore: Test with Scala 3.8.4

tgodzik:


tgodzik:
We now have a final 3.8.4 release.

PR #4563, opened 2026-04-10
bugfix: Fix into keyword on enums

### LINKED ISSUE #4561: 'into' is not accepted before 'enum'
The following reports an error at the Scalameta level, but the Scala compiler accepts it (3.8.x with `-preview`):
```scala
into enum Foo:
  case Bar
```
with the message "`;` expected but `enum` found" located on the `enum` keyword.

Note that `enum class` is correctly parsed.

tgodzik:
Fixes https://github.com/scalameta/scalameta/issues/4561


kitbellew (approved):

sjrd:
Is it expected that this fix is not yet available in Metals 1.68.0, which was released 3 weeks ago?

`into` is about to become stable in Scala 3.9.0.

tgodzik:
I need to make a release, which I intended to do before my vacation, but didn't manage to. Doing it this week!

Issue #4146, opened 2025-01-14
[Scala 3] Implement transversers


tgodzik:
Currently aside from trees modules I think we are only missing transversers for Scala 3 which can be used to transform or traverse trees.

https://github.com/scalameta/scalameta/tree/main/scalameta/transversers/shared/src/main/scala/scala/meta/transversers

It's actually quite simple to traverse trees manually, but these are useful utilities

tgodzik:
This is quite a hard to implement, so I would not start with this issue.

ltrojanowski:
@tgodzik I have a draft PR which relates to this issue but need guidance. https://github.com/scalameta/scalameta/pull/4620. If you'd rather use the PR as a starting point, and do it yourself I'm also fine with that

kitbellew:
#4620 fixed this issue.

PR #4599, opened 2026-04-29
transversers: split off SimpleApi, for scala3

kitbellew:


bjaglin:
can we consider that this closes https://github.com/scalameta/scalameta/issues/4146?

kitbellew:
> can we consider that this closes #4146?

not yet. because this only covers simple traversers, but the macro-based one -- and, more importantly, the transformer -- is not covered.

PR #4603, opened 2026-05-05
TextDocumentOps: add SAM methods to symbols and occurrences

### LINKED ISSUE #4596: "Go to implementations" ignores single abstract methods
### Describe the bug

Just like in the title - "go to implementations" doesn't seem to notice SAM implementations.

Reproduction steps:

1. write this

```scala
//> using scala "2.13.10"

trait Operation {
  def perform(s: String): Int
}

object Operation {

  val normal: Operation =
    new Operation {
      def perform(s: String): Int = s.length
    }

  val sam: Operation = s => s.length
  val anonSam: Operation = _.length
}
```

2. focus on "perform" in the Operation trait
3. Go to implementations
4. Only the first implementation shows up

### Expected behavior

All three implementations are highlighted.

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

0.11.9+90-31235494-SNAPSHOT

### Extra context or search terms

_No response_

iigaz:
This PR builds upon #4112 in an attempt to fix https://github.com/scalameta/metals/issues/4596.

Adding methods to symbols and occurrences allows metals to pick them up in "Go to implementations" of a method (you could previously see SAMs among implementations of a trait, but in my opinion it's more intuitive to search for implementations of a method inside that trait, and the issue above also concerns searching by methods).

Besides autotests, I also tested it with the latest v1 metals (with patched scalameta v4.15.2), and it works as expected.

<details>
<summary>Screenshots of metals with suggested changes</summary>

#### Code lens

lens

#### Go to implementations

gotoimpl
</details>

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
@@ -52,6 +52,8 @@ trait TextDocumentOps {
       clearSymbolPointsCache()
       val occurrences = emptyOccurrenceMap()
       val samoccurrences = emptyOccurrenceMap()
+      // SAM methods share positions with their classes.
+      val sammethodoccurrences = emptyOccurrenceMap()
       val symbols = mutable.Map[String, s.SymbolInformation]()
       val synthetics = mutable.ListBuffer[s.Synthetic]()
       val syntheticTreeCache = new mutable.HashMap[g.Tree, Option[s.Synthetic]]
@@ -95,7 +97,7 @@ trait TextDocumentOps {
             }
             ok
           }
-          set(sam, samoccurrences)
+          set(sam, samoccurrences) && set(sam.info.decls.elems.sym, sammethodoccurrences)
         }
         val gsym = gt.symbol
         val gpos = gt.pos
@@ -654,7 +656,7 @@ trait TextDocumentOps {
       val finalOccurrences = {
         val buf = List.newBuilder[s.SymbolOccurrence]
         for {
-          map <- Iterator(occurrences, mpatoccurrences, samoccurrences)
+          map <- Iterator(occurrences, mpatoccurrences, samoccurrences, sammethodoccurrences)
           (pos, (sym, role)) <- map
           flatSym <- sym.asMulti
         } buf += s.SymbolOccurrence(Some(pos.toRange), flatSym, role)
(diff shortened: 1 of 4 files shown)


semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -85,12 +87,23 @@ trait TextDocumentOps {
 
       def addSamOccurrence(gt: g.Function) = getSyntheticSAMClass(gt).foreach { sam =>
         val gsym = gt.symbol
+        // Get the only declared method of the SAM class.
+        // Remove its position to create a new local on every call.
+        val gsymMethod = sam.info.decls.elems.sym.setPos(g.NoPosition)
```
iigaz: `elems` is a `var private[scala]`, if this code is not allowed to use it an alternative can be `.toList.head` or `headOption`

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -52,6 +52,8 @@ trait TextDocumentOps {
       clearSymbolPointsCache()
       val occurrences = emptyOccurrenceMap()
       val samoccurrences = emptyOccurrenceMap()
+      // SAM methods share positions with their classes.
+      val sammethodoccurrences = emptyOccurrenceMap()
```
iigaz, line 56: I don't think I understand the naming scheme here yet, if some names are not appropriate I will change them

tgodzik (dismissed):
LGTM from me, @kitbellew what do you think?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
kitbellew: @iigaz i have a bit of a concern about this.

one, we are modifying a symbol, and i don't know what side effects this change brings.

two, there's one thing i don't understand. i checked out the code and moved this line to the next patch (so it's defined right before being used), and that broke the tests. which means, this setting of NoPosition has some magical effect only if it's invoked before `gsym.toSemantic`, and I don't know why -- and find it a bit unreliable.

however, i have an alternative proposal which might in fact be even better. i will send in a patch soon.

iigaz: > we are modifying a symbol

Sorry, completely missed that. I guess I thought it would be fine since we're only changing a synthetic class. Another option would be to clone this symbol before updating its position, at the cost of extra `local`s.

> i checked out the code and moved this line to the next patch (so it's defined right before being used), and that broke the tests.

Yes, I can confirm this is caused by side effects of changing position. As to why, I don't think  `gsym.toSemantic` matters in this case, but rather that `toSymbolInfo` in `saveSymbolFromSemantic` might call `.toSemantic` on the method before its position was cleared. When placing the definition before it, tests pass just fine.

Thanks for taking a closer look on this.

iigaz: Well, I realized I only really need to create "empty" `local`s, and I can get rid of changing the position entirely by exposing a way to generate new locals without adding them to `pointsCache`. Suggested changes will be described in the second commit.

Edit: ~~messed up and didn't test properly, tests won't pass. Will fix asap.~~ fixed.

kitbellew: very nice observation, it took me a while to figure this out earlier 😀

I'm refactoring a bit...

kitbellew:
@iigaz thank you for this submission, it was merged. feel free to delete the branch.

PR #4618, opened 2026-05-13
Cyclic jar fix for comments

### LINKED ISSUE #2086: StackOverFlow when indexing Jars with self-referencing `Class-Path:`
Discovered when using ScalaMetals (v0.9.1):
https://gist.github.com/faizk/df7cde49b2c581df9971ea5f58c8eec4

On closer examination, I found at least one third-party JAR
(my `build.sbt` is basically a bunch of `unmanagedJars in Compile : = .. `, etc) that is self-referential:

Given a jar file `MyJar.jar`:
```
Manifest-Version: 1.0
Class-Path: MyJar.jar
Created-By: 1.7.0_06 (Oracle Corporation)
```

I'm not sure that Jars such as these are even valid, but they do break tools, so perhaps handling such situations
is a good idea (either silently skip but warn about it, or throw an exception describing the problematic input).

I can't imagine there would be a specification that says that the `Class-Path:` attr can never resolve to the name of the original jar file - that doesn't make sense, so I think protecting against just a situation is called for,

Likely - 
https://github.com/scalameta/scalameta/blob/master/semanticdb/metacp/src/main/scala/scala/meta/internal/classpath/ClasspathIndex.scala#L134-L165

I just tried a quick fix on the lines of 
```scala
     if (abspath.isFile || abspath.isDirectory) {
       if (abspath != jarpath) 
          expandEntry(abspath)
     }
```
Which fixes the issue for me, so I can send a PR (to stop recursing and maybe log a warning) - unless I find after some more research on this that there's a better way.

skestle:
I decided to play around with #2086, and here's the outline of a maximal fix:

* Passed reporter in to be able to log a warning when cyclic dependency found - this replaces the exception with a message. If not done now, it will never get implemented as it will be an addition: in this fix, it's a reduction [of output].
* Tracked what's been indexed. Note that **the implementation is wrong**. It needs to be a stack to correctly identify the cycles (using recursion). The set will still be handy in a correct solution for a fast exit so that repeatedly referenced jars aren't processed multiple times.

Fixes #2086.


kitbellew:
@skestle I split your commit into three, as we prefer to have each commit achieve a focused goal (and a test commit, by itself, to demonstrate the error so that it's clear how it later improves).

can you please amend the second commit to fix your claim that "the implementation is wrong"?

as to your questions: i don't see a problem with using the reporter (but made it optional); adding a small jar to resources looks fine, too.

skestle:
> adding a small jar to resources looks fine, too.

I figured that, but adding a->c and b->c jars will be another 3 jars (to verify it's not considered a cycle). It could be considered clearer to have the physical files (and allow an easy regression test with all jars), but creating temp jars would be very declarative in-line code. I'm torn, so look to existing contributors to weigh in.

And then there's a->b->c->b which should also exist as a test (ensuring b->c->b is reported as the cycle)

skestle:
> we prefer to have each commit achieve a focused goal

BTW - music to my ears. The amount of times I drilled into co-workers about commit "SRP" - ideally being able to point to 1 file or even 1 line of code as the reason for the commit. Each MR should aim to have 1 reason.
Sorry, it's just been a long time since I needed to have clean commits.

kitbellew:
> > adding a small jar to resources looks fine, too.
> 
> I figured that, but adding a->c and b->c jars will be another 3 jars (to verify it's not considered a cycle). It could be considered clearer to have the physical files (and allow an easy regression test with all jars), but creating temp jars would be very declarative in-line code. I'm torn, so look to existing contributors to weigh in.

a couple of files less than 1k each seem trivial to be torn about. @tgodzik what do you think about including a couple of small jars as resources for testing?

kitbellew:
@skestle are you planning to close the loop on this? i hope i have answered your earlier question, on adding resources.

skestle:
Sorry yes - I've come back to it a few times, but not for any meaningful time. I do expect to complete it - but if it's messy do close it, and I'll reopen when updated.

PR #4620, opened 2026-05-19
Add transversers to scala3

ltrojanowski:
A while ago I started work to add the collection like support back to the scala3 version of scalameta. I did however get stuck and need some guidance to get this over the line.

When developing I noticed that nothing seems to trigger the following branches of the original `transformer.scala` macro:
```scala
      case tpe @ OptionListTreeTpe(_) =>
        optionTransformer(fname, tpe, listTransformer(_, _, treeTransformer))
      case tpe @ ListListTreeTpe(_) =>
        listTransformer(fname, tpe, listTransformer(_, _, treeTransformer))
```
I'm now unsure what to do, is this a bug? is this something which should be removed from the scala3 version and only exists for some historic reason in the scala2x codebase? I don't know what part of the syntax tree this was supposed to match

As far as the approach taken to port this macro's functionality over it's pretty much the same as for scala3-tree-lifts


tgodzik:
@jchyb do you remember the code around this? Are you able to help out with some hints?

jchyb:
The specific code is in [transversers](https://github.com/scalameta/scalameta/blob/e9226042dde7b38e3d3845d58789bf6c01ff8354/scalameta/common2/shared/src/main/scala/scala/meta/internal/transversers/transformer.scala#L69-L72), which I didn't really touch, so I can't comment for sure.

But looking at the overall approach, it seems right to me. scala3-tree-lifts theoretically could have been translated as a macro instead (and ideally probably would have better to do it that way), but there were some issues with how the scala-2 compiled Trees ADT and the annotations were read in a scala-3 macro which I wasn't able to work around. I suspect the situation would be the same here.

I suppose you could either add printlns in those two cases and see if they trigger anywhere during macro expansion, or add `-Vprint:typer` to whatever module calls the transverser macro, which would prints the compiled code (including the one generated by a macro) and see if you can find the related stuff.

<s>Generally, for scala3-tree-lifts, I worked with -Vprint and directly translated the result of that to Scala 3, so I would recommend doing that here as well for the easiest approach.</s> (EDIT: nvm, I misremembered I think, sorry about that)

ltrojanowski:
> The specific code is in [transversers](https://github.com/scalameta/scalameta/blob/e9226042dde7b38e3d3845d58789bf6c01ff8354/scalameta/common2/shared/src/main/scala/scala/meta/internal/transversers/transformer.scala#L69-L72), which I didn't really touch, so I can't comment for sure.
> 
> But looking at the overall approach, it seems right to me. scala3-tree-lifts theoretically could have been translated as a macro instead (and ideally probably would have better to do it that way), but there were some issues with how the scala-2 compiled Trees ADT and the annotations were read in a scala-3 macro which I wasn't able to work around. I suspect the situation would be the same here.
> 
> I suppose you could either add printlns in those two cases and see if they trigger anywhere during macro expansion, or add `-Vprint:typer` to whatever module calls the transverser macro, which would prints the compiled code (including the one generated by a macro) and see if you can find the related stuff.
> 
> Generally, for scala3-tree-lifts, I worked with -Vprint and directly translated the result of that to Scala 3, so I would recommend doing that here as well for the easiest approach. (EDIT: nvm, I misremembered I think, sorry about that)

I did expand the macro and inspect the output manually, I became suspicious why I don't see any generated code matching the template for those two branches. I added println and confirmed that they are not getting hit. My first suspicion was, that this gets hit only for some scala versions, due to syntax tree differences, but compiling for the latest versions of 2.11, 2.12, 2.13 gives the same result. The code path doesn't get exercised.
I guess what I was really hoping for, was that you @tgodzik would know and could reach out to @xeno-by to confirm what this was for

albertikm:
> I did expand the macro and inspect the output manually, I became suspicious why I don't see any generated code matching the template for those two branches...

this likely has to do with the fact that we used to have method arguments as `List[List[Term]]` but now the inner list has become `Term.ArgClause`, to include positions of parentheses and existence of modifiers such as `implicit`. other cases similarly.

so if it's not used, then so be it, not a bug. whether to keep it going forward... only preference is Scala 2 and 3 should do the same.

kitbellew:
my alter ego wrote the previous comment 🙂 need to remember to switch.

PR #4622, opened 2026-05-27
Specify class loader when loading `Tokenize` service.

### LINKED ISSUE #4621: `Tokenize` instance not found in SBT builds
The following error occurs when using scalameta versions `4.16.1` and above in SBT builds:

```
java.lang.UnsupportedOperationException: No implementation of Tokenize service is provided
```

In `4.16.1`, the tokenizer is [loaded using the JVM ServiceLoader](https://github.com/scalameta/scalameta/pull/4585/changes). The [ServiceLoader.load](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/ServiceLoader.html#load(java.lang.Class)) method uses the current thread context class loader, which does not load the `Tokenize` instance.
 
This is resolved by passing in the correct class loader with `getClass.getClassLoader`.

## Reproduction

Create an SBT project with the following files:

```scala
// build.sbt
scalaVersion := "3.3.6"

import scala.meta._

Compile / sourceGenerators += Def.task {
  val target = ((Compile / sourceManaged).value / "demo")

  val code = source"""
package demo

object Demo {}""".toString

  IO.write(target, code)
  Seq(target)
}.taskValue
```

```scala
// project/build.properties
sbt.version = 1.12.11
```

```scala
// project/dependencies.sbt
libraryDependencies += "org.scalameta" %% "scalameta" % "4.17.0"
```

Then run `sbt compile` to generate sources. Source generation fails with:

```
[error] (Compile / managedSources) java.lang.UnsupportedOperationException: No implementation of Tokenize service is provided
```

zainab-ali:
This resolves #4621 by specify the class loader.

As a general rule, `getClass.getClassLoader` should be used, and `Thread.currentThread.getContextClassLoader` should be avoided.

[This StackOverflow answer](https://stackoverflow.com/questions/1771679/difference-between-threads-context-class-loader-and-normal-classloader) has a good summary of class loading best practices.

[The ServiceLoader docs](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/ServiceLoader.html#load(java.lang.Class)) contain more details on the specifics. I haven't looked through these in detail, so there might be a better approach.

I couldn't think of a way to unit test this, so didn't add tests to reproduce / resolve the issue.

### DIFF
--- a/scalameta/trees2/jvm/src/main/scala/scala/meta/tokenizers/PlatformCompat.scala
+++ b/scalameta/trees2/jvm/src/main/scala/scala/meta/tokenizers/PlatformCompat.scala
@@ -5,7 +5,7 @@ import java.util.ServiceLoader
 object PlatformCompat {
 
   lazy val loadTokenize: Option[Tokenize] = {
-    val factories = ServiceLoader.load(classOf[Tokenize]).iterator()
+    val factories = ServiceLoader.load(classOf[Tokenize], getClass().getClassLoader()).iterator()
     if (factories.hasNext) Some(factories.next()) else None
   }
 


scalameta/trees2/jvm/src/main/scala/scala/meta/tokenizers/PlatformCompat.scala
```diff
@@ -5,7 +5,7 @@ import java.util.ServiceLoader
 object PlatformCompat {
 
   lazy val loadTokenize: Option[Tokenize] = {
-    val factories = ServiceLoader.load(classOf[Tokenize]).iterator()
+    val factories = ServiceLoader.load(classOf[Tokenize], getClass().getClassLoader()).iterator()
```
kubukoz, line 8: can we make this settable in a function parameter? I take responsibility for the idea to use `getClassLoader`, but in hindsight it's probably just as limiting as the previous behavior.

kitbellew: @kubukoz the only way with a parameter is in #4623, take a look.

kubukoz: sounds good 👍

zainab-ali:
Closing in favour of #4623

PR #4631, opened 2026-06-04
TextDocumentOps: wrap full infix term in type-application sugar

### LINKED ISSUE #1045: Weird position for infix operator type annotation sugar
```scala
(2 :: x)
```
has the sugar `*[Int]` with position starting and ending at the end offset for `x`

Example

```
[104..106): :: => _root_.scala.collection.immutable.List#`::`(Ljava/lang/Object;)Lscala/collection/immutable/List;.
[107..108): x => _root_.fix.ExpandSugars.x.
...
[108..108): *[Int]
  [0..1): * => _star_.
```

I expected the sugar to be

```
[106..106): *[Int]
  [0..1): * => _star_.
```

jozanek:
Closes #1045, which reported that the inferred type-argument sugar `*[Int]` for an infix expression like `(2 :: x)` had a degenerate **zero-width** position.

That was a symptom of the pre-2018 SemanticDB text format. The Jan-2025 synthetics rewrite already fixed it: `*` (the `OriginalTree`) now spans the function `Select` the inferred type args attach to — consistent with the prefix case `List(1).map => *[Int]` — rather than a degenerate point.

This PR reduces to a regression test capturing that behavior and closes the issue. No functional change.


semanticdb/integration/src/main/scala/example/Synthetic.scala
```diff
@@ -40,8 +46,7 @@ class Synthetic {
   } println(a)
   for {
     a <- scala.concurrent.Future.successful(1)
-    b <- scala.concurrent.Future.successful(2)
-    if a < b
+    b <- scala.concurrent.Future.successful(2) if a < b
   } yield a
```
kitbellew, line 45: why change these lines?

semanticdb/integration/src/main/scala/example/Synthetic.scala
```diff
@@ -14,6 +14,10 @@ class Synthetic {
   val lst = 1 #:: 2 #:: Stream.empty
   lst + "foo"
 
+  // See https://github.com/scalameta/scalameta/issues/1045
+  val ys: List[Int] = Nil
+  2 :: ys
+
```
kitbellew: i'd prefer it if you could:
- put this expression at the end so it doesn't cause the expect files to change so drastically
- add just this (or any other test change) in the first commit, along with expect changes, and then, in a separate commit, modify the functionality and show how the test results have changed.

that is, if we end up proceeding.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -565,7 +565,17 @@ trait TextDocumentOps {
             case t: g.TypeApply => getTypeApply(t, isWithinFor = false)
             case t: g.Apply =>
               val res = cachedIfSelect(t.fun)
-              if (res.isEmpty) forApply(t) else None
+              if (res.nonEmpty) None
+              else t.fun match {
+                // infix operator with inferred type args (e.g. `2 :: xs`, `x -> x`): the other
```
kitbellew: in `a op b`, `op` is the operator and `b` is the operand. which other operand are you referring to?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
```diff
@@ -610,6 +624,24 @@ trait TextDocumentOps {
           }
         }
 
+        // An infix application desugars to `Apply(TypeApply(Select(qual, op), targs), args)`; the
+        // method argument is the operand NOT covered by the `Select` span (`qual` + `op` in source
```
kitbellew: "operand" here is `args` or `targs`? and why should it be covered?

kitbellew (changes requested):
perhaps i don't fully understand what you're trying to do here... but why would a non-existent, inferred type arg clause occupy not a zero-width position somewhere between the operator and the operand, but the entire infix expression (including, it looks like, all parent infix expressions)? how's that consistent with an actual, non-inferred type arg clause which doesn't do the same?

you seem to suggest that #1045 has already been fixed by some other change, which is good if it's so; we can add a test to verify it and close it.

jozanek:
> perhaps i don't fully understand what you're trying to do here... but why would a non-existent, inferred type arg clause occupy not a zero-width position somewhere between the operator and the operand, but the entire infix expression (including, it looks like, all parent infix expressions)? how's that consistent with an actual, non-inferred type arg clause which doesn't do the same?
> 
> you seem to suggest that #1045 has already been fixed by some other change, which is good if it's so; we can add a test to verify it and close it.

You're right on all counts - thanks for the review. I updated PR to include new test to verify current behaviour.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
jozanek: removed

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
jozanek: outdated

semanticdb/integration/src/main/scala/example/Synthetic.scala
jozanek: moved

kitbellew (approved):

PR #4632, opened 2026-06-05
skip inaccessible paths in recursive file walks

### LINKED ISSUE #1168: Loading semanticsdb attempts to visit paths that are not accessible
Here is the corresponding exception when running scala-meta 2.1.2:

```
java.nio.file.AccessDeniedException: /home/guest/test/./private
java.io.UncheckedIOException: java.nio.file.AccessDeniedException:  /home/guest/test/./private
    at java.nio.file.FileTreeIterator.fetchNextIfNeeded(FileTreeIterator.java:88)
    at java.nio.file.FileTreeIterator.hasNext(FileTreeIterator.java:104)
    at java.util.Iterator.forEachRemaining(Iterator.java:115)
    at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
    at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
    at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
    at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
    at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
    at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
    at org.langmeta.internal.io.PlatformFileIO$.listAllFilesRecursively(PlatformFileIO.scala:40)
    at org.langmeta.internal.io.FileIO$.listAllFilesRecursively(FileIO.scala:28)
    at org.langmeta.io.Multipath.$anonfun$deep$1(Multipath.scala:43)
    at org.langmeta.io.Multipath.$anonfun$deep$1$adapted(Multipath.scala:16)
    at scala.collection.immutable.List.foreach(List.scala:389)
    at org.langmeta.io.Multipath.deep(Multipath.scala:16)
    at org.langmeta.io.Multipath.deep$(Multipath.scala:14)
    at org.langmeta.io.Classpath.deep(Multipath.scala:68)
    at org.langmeta.internal.semanticdb.vfs.Database$.load(Database.scala:13)
    at scala.meta.internal.SemanticdbPipeline$SemanticdbJvmComponent$PersistSemanticdbPhase.run(SemanticdbPipeline.scala:104)
```

The problem is that the `PlatformFileIO.listAllFilesRecursively` is using `files.walk`. This method _visits all levels of the file tree_, regardless of the permissions.

jozanek:
Closes #1168.

`Files.walk` returns a lazy `Stream<Path>` that throws `UncheckedIOException` (wrapping `AccessDeniedException`) the moment it reaches a directory the process can't read - aborting the whole traversal. The stream is also never closed.

This replaces `Files.walk` with `Files.walkFileTree` + a `SimpleFileVisitor` whose:

- `visitFileFailed` skips **`AccessDeniedException`** (the reported bug) and **`FileSystemLoopException`**, and lets any other `IOException` propagate - so  a genuine I/O failure aborts loudly instead of returning a partial listing as if it were complete.
- `FOLLOW_LINKS` is set so a symlink's *target* attributes are used, i.e. `attrs.isRegularFile()` lists symlinked files (matching the old `Files.walk` + `Files.isRegularFile`). N.B. this also traverses symlinked directories, whose cycle case is the `FileSystemLoopException` skipped above. 

Since #4635, `Locator` (the SemanticDB load path) and the benchmark `FileFixtures` both go through `FileIO.listAllFilesRecursively`, so this fix makes them resilient too. JS/Native keep their own recursive implementations and are unaffected.

Structured as two commits per review:
1. a JVM test that reproduces the `AccessDeniedException`;
2. the fix, which flips that test to assert the unreadable directory is skipped.

### DIFF
--- a/scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformFileIO.scala
+++ b/scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformFileIO.scala
@@ -2,12 +2,12 @@ package scala.meta.internal.io
 
 import scala.meta.io._
 
+import java.io.IOException
 import java.net.URI
 import java.nio.charset.Charset
-import java.nio.file.{FileSystem, FileSystemAlreadyExistsException, FileSystems, Files, OpenOption,
-  Path}
+import java.nio.file._
+import java.nio.file.attribute.BasicFileAttributes
 import java.util
-import java.util.stream.Collectors
 
 object PlatformFileIO {
 
@@ -46,11 +46,27 @@ object PlatformFileIO {
 
   def listAllFilesRecursively(root: AbsolutePath): ListFiles = {
     val relativeFiles = List.newBuilder[RelativePath]
-    val iter = Files.walk(root.toNIO).iterator()
-    while (iter.hasNext) {
-      val path = iter.next()
-      if (Files.isRegularFile(path)) relativeFiles += RelativePath(root.toNIO.relativize(path))
-    }
+    // FOLLOW_LINKS so a symlink's target attributes are used (symlinked files are listed); note this
+    // also traverses symlinked directories.
+    Files.walkFileTree(
+      root.toNIO,
+      util.EnumSet.of(FileVisitOption.FOLLOW_LINKS),
+      Integer.MAX_VALUE,
+      new SimpleFileVisitor[Path] {
+        override def visitFile(file: Path, attrs: BasicFileAttributes): FileVisitResult = {
+          if (attrs.isRegularFile) relativeFiles += RelativePath(root.toNIO.relativize(file))
+          FileVisitResult.CONTINUE
+        }
+        // #1168: skip unreadable paths (AccessDeniedException); FOLLOW_LINKS can also raise
+        // FileSystemLoopException on a symlink cycle, so skip that too. Any other I/O error
+        // propagates, so a partial scan is never reported as complete. postVisitDirectory is left
+        // to SimpleFileVisitor, which likewise rethrows a non-null exception.
+        override def visitFileFailed(file: Path, exc: IOException): FileVisitResult = exc match {
+          case _: AccessDeniedException | _: FileSystemLoopException => FileVisitResult.CONTINUE
+          case _ => throw exc
+        }
+      },
+    )
     ListFiles(root, relativeFiles.result())
   }
 
--- a/tests/jvm/src/test/scala/scala/meta/tests/io/PlatformFileIOSuite.scala
+++ b/tests/jvm/src/test/scala/scala/meta/tests/io/PlatformFileIOSuite.scala
@@ -10,22 +10,27 @@ import java.nio.file.{FileSystems, Files}
 import munit.FunSuite
 
 class PlatformFileIOSuite extends FunSuite {
-  // #1168: listAllFilesRecursively aborts the walk when it meets an unreadable directory
-  test("listAllFilesRecursively aborts on unreadable directories") {
+  // #1168: an unreadable subdirectory must be skipped, not abort the whole traversal
+  test("listAllFilesRecursively skips unreadable directories") {
     assume(
       FileSystems.getDefault.supportedFileAttributeViews.contains("posix"),
       "POSIX file permissions not supported on this platform",
     )
     val root = AbsolutePath(Files.createTempDirectory("scalameta-1168"))
+    val ok = root.resolve("ok.txt")
+    Files.write(ok.toNIO, "ok".getBytes)
     val secret = root.resolve("secret")
     Files.createDirectory(secret.toNIO)
     Files.setPosixFilePermissions(secret.toNIO, PosixFilePermissions.fromString("---------"))
     // running as root bypasses permission checks, so the directory stays readable: skip then
     assume(!Files.isReadable(secret.toNIO), "directory still readable (running as root?)")
-    try intercept[java.io.UncheckedIOException](FileIO.listAllFilesRecursively(root))
-    finally {
+    try {
+      val obtained = FileIO.listAllFilesRecursively(root) // must not throw
+      assert(obtained.contains(ok), s"$ok missing from:\n${obtained.mkString("\n")}")
+    } finally {
       // restore permissions and delete the temporary tree
       Files.setPosixFilePermissions(secret.toNIO, PosixFilePermissions.fromString("rwx------"))
+      Files.delete(ok.toNIO)
       Files.delete(secret.toNIO)
       Files.delete(root.toNIO)
     }
--- a/tests/jvm/src/test/scala/scala/meta/tests/io/PlatformFileIOSuite.scala
+++ b/tests/jvm/src/test/scala/scala/meta/tests/io/PlatformFileIOSuite.scala
@@ -1,0 +1,33 @@
+package scala.meta.tests
+package io
+
+import scala.meta._
+import scala.meta.internal.io._
+
+import java.nio.file.attribute.PosixFilePermissions
+import java.nio.file.{FileSystems, Files}
+
+import munit.FunSuite
+
+class PlatformFileIOSuite extends FunSuite {
+  // #1168: listAllFilesRecursively aborts the walk when it meets an unreadable directory
+  test("listAllFilesRecursively aborts on unreadable directories") {
+    assume(
+      FileSystems.getDefault.supportedFileAttributeViews.contains("posix"),
+      "POSIX file permissions not supported on this platform",
+    )
+    val root = AbsolutePath(Files.createTempDirectory("scalameta-1168"))
+    val secret = root.resolve("secret")
+    Files.createDirectory(secret.toNIO)
+    Files.setPosixFilePermissions(secret.toNIO, PosixFilePermissions.fromString("---------"))
+    // running as root bypasses permission checks, so the directory stays readable: skip then
+    assume(!Files.isReadable(secret.toNIO), "directory still readable (running as root?)")
+    try intercept[java.io.UncheckedIOException](FileIO.listAllFilesRecursively(root))
+    finally {
+      // restore permissions and delete the temporary tree
+      Files.setPosixFilePermissions(secret.toNIO, PosixFilePermissions.fromString("rwx------"))
+      Files.delete(secret.toNIO)
+      Files.delete(root.toNIO)
+    }
+  }
+}


bench/semanticdb/src/main/scala/scala/meta/internal/bench/FileFixtures.scala
```diff
@@ -1,14 +1,11 @@
...
 import java.nio.file._
 
 trait FileFixtures {
   lazy val scalapDir: Path = Paths.get(s"${BuildInfo.sourceroot}/bench/corpus/scalap")
 
-  lazy val scalapFiles: List[Path] = {
-    val all = Files.walk(scalapDir).iterator.toScala.toList
-    all.filter(p => !Files.isDirectory(p))
-  }
+  lazy val scalapFiles: List[Path] = collectFiles(scalapDir)(file => !Files.isDirectory(file))
```
kitbellew: use `FileIO.listAllFilesRecursively`.

scalameta/io/shared/src/main/scala/scala/meta/internal/io/package.scala
```diff
@@ -2,11 +2,37 @@ package scala.meta.internal
...
 
 package object io {
 
+  /**
+   * Recursively collects the files (never directories) under `root` for which `accept` holds,
+   * silently skipping any path that can't be accessed. This is the resilient replacement for
+   * `Files.walk`, which aborts the whole traversal on the first unreadable directory.
+   * Symlinks are not followed during traversal.
+   */
+  def collectFiles(root: Path)(accept: Path => Boolean): List[Path] = {
```
kitbellew: move this algorithm into JVM version of PlatformFileIO.listAllFilesRecursively.

if this algorithm works on JS and Native (the previous one didn't), then move it to FileIO instead and remove the PlatformFileIO versions from all three platforms.

kitbellew (changes requested):
can you please structure this PR as two commits:
- first commit adds a test which shows the exception you are fixing
- second commit fixes the issue and shows how that test is changed?

jozanek:
> can you please structure this PR as two commits:
> 
> * first commit adds a test which shows the exception you are fixing
> * second commit fixes the issue and shows how that test is changed?

Restructured into the two commits.

bench/semanticdb/src/main/scala/scala/meta/internal/bench/FileFixtures.scala
jozanek: updated

scalameta/io/shared/src/main/scala/scala/meta/internal/io/package.scala
jozanek: It can't go into shared FileIO. Scala.js has no runtime java.nio.file.Files (the repo only stubs Paths/Path), so a shared walkFileTree impl fails Scala.js linking. Native's javalib likewise lacks the tree-walk APIs, which is why the three platform-specific listAllFilesRecursively exist. So I inlined the resilient walkFileTree into the JVM PlatformFileIO and removed collectFiles. FileFixtures now uses FileIO.listAllFilesRecursively, and Locator (the actual semanticdb-load path) reuses it too.

tests/jvm/src/test/scala/scala/meta/tests/io/PlatformFileIOSuite.scala
```diff
@@ -0,0 +1,34 @@
...
+
+class FileIOAccessDeniedSuite extends FunSuite {
+  // #1168: listAllFilesRecursively aborts the whole walk when it meets an unreadable
+  // directory. This test documents that exception; the fix flips it to skip instead.
+  test("listAllFilesRecursively aborts on unreadable directories") {
+    assume(
+      FileSystems.getDefault.supportedFileAttributeViews.contains("posix"),
+      "POSIX file permissions not supported on this platform",
+    )
+    val root = AbsolutePath(Files.createTempDirectory("scalameta-1168"))
```
kitbellew, line 19: don't we need to remove this? or it disappears automatically?

scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformFileIO.scala
```diff
@@ -46,11 +47,21 @@ object PlatformFileIO {
...
-    while (iter.hasNext) {
-      val path = iter.next()
-      if (Files.isRegularFile(path)) relativeFiles += RelativePath(root.toNIO.relativize(path))
-    }
+    Files.walkFileTree(
+      root.toNIO,
+      new SimpleFileVisitor[Path] {
+        override def visitFile(file: Path, attrs: BasicFileAttributes): FileVisitResult = {
+          // Files.isRegularFile follows symlinks, matching the historical Files.walk behavior.
+          if (Files.isRegularFile(file)) relativeFiles += RelativePath(root.toNIO.relativize(file))
```
kitbellew: why not `attrs.isRegularFile()`?

scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformFileIO.scala
```diff
@@ -46,11 +47,21 @@ object PlatformFileIO {
...
-    val iter = Files.walk(root.toNIO).iterator()
-    while (iter.hasNext) {
-      val path = iter.next()
-      if (Files.isRegularFile(path)) relativeFiles += RelativePath(root.toNIO.relativize(path))
-    }
+    Files.walkFileTree(
+      root.toNIO,
+      new SimpleFileVisitor[Path] {
+        override def visitFile(file: Path, attrs: BasicFileAttributes): FileVisitResult = {
+          // Files.isRegularFile follows symlinks, matching the historical Files.walk behavior.
```
kitbellew: why do you sometimes terminate your comment with a dot, and sometimes not?

elsewhere in the code, it seems, multi-sentence comments have dots, but short comments like yours do not.

scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformFileIO.scala
```diff
@@ -2,12 +2,13 @@ package scala.meta.internal.io
 
 import scala.meta.io._
 
+import java.io.IOException
 import java.net.URI
 import java.nio.charset.Charset
-import java.nio.file.{FileSystem, FileSystemAlreadyExistsException, FileSystems, Files, OpenOption,
-  Path}
+import java.nio.file.attribute.BasicFileAttributes
+import java.nio.file.{FileSystem, FileSystemAlreadyExistsException, FileSystems, FileVisitResult,
```
kitbellew: optional: i think it's time to do a wildcard import. probably should have done even earlier.

bench/semanticdb/src/main/scala/scala/meta/internal/bench/FileFixtures.scala
```diff
@@ -1,14 +1,13 @@
...
 import java.nio.file._
 
 trait FileFixtures {
   lazy val scalapDir: Path = Paths.get(s"${BuildInfo.sourceroot}/bench/corpus/scalap")
 
-  lazy val scalapFiles: List[Path] = {
-    val all = Files.walk(scalapDir).iterator.toScala.toList
-    all.filter(p => !Files.isDirectory(p))
-  }
+  lazy val scalapFiles: List[Path] = FileIO.listAllFilesRecursively(AbsolutePath(scalapDir))
```
kitbellew: i separated the switch to FileIO in a separate PR, #4635.

semanticdb/semanticdb/shared/src/main/scala/scala/meta/internal/semanticdb/Locator.scala
```diff
@@ -10,15 +11,17 @@ object Locator {
     .foreach(path => apply(path)(fn))
 
   def apply(path: Path)(fn: (Path, TextDocuments) => Unit): Unit = if (Files.exists(path))
-    if (Files.isDirectory(path)) Files.walk(path).iterator().toScala
```
kitbellew: i separated the switch to FileIO in a separate PR, #4635.

tests-semanticdb/src/test/scala/scala/meta/tests/semanticdb/Issue1168Suite.scala
```diff
@@ -0,0 +1,38 @@
+package scala.meta.tests.semanticdb
+
+import scala.meta.internal.semanticdb.{Locator, TextDocuments}
+
+import java.nio.file.attribute.PosixFilePermissions
+import java.nio.file.{FileSystems, Files, Path}
+
+import munit.FunSuite
+
+class Issue1168Suite extends FunSuite {
```
kitbellew: is this test somehow different from FileIOAccessDeniedSuite? if not really, then i wouldn't keep both.

tests/jvm/src/test/scala/scala/meta/tests/io/PlatformFileIOSuite.scala
kitbellew, line 1: Since this will not later be an access-denied test, should we call it PlatformFileIOSuite instead? because it tests FileIO for the JVM platform?

kitbellew (changes requested):

scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformFileIO.scala
jozanek: see comment above, it doesn't follow symlinks and I wanted to match old `Files.walk` behaviour

scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformFileIO.scala
jozanek: no particular reason :) I guess because I don't have any preference.

scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformFileIO.scala
kitbellew: there is a an overload of walkFileTree which takes two more parameters, one of them is about symlinks. would that help?

tests/jvm/src/test/scala/scala/meta/tests/io/PlatformFileIOSuite.scala
jozanek: Updated: the demonstrating commit no longer creates `ok.txt`. The fix commit adds it where it's asserted.

tests-semanticdb/src/test/scala/scala/meta/tests/semanticdb/Issue1168Suite.scala
jozanek: Removed. After #4635 your Locator walks dirs via `FileIO.listAllFilesRecursively`, so PlatformFileIOSuite already covers that path.

tests/jvm/src/test/scala/scala/meta/tests/io/PlatformFileIOSuite.scala
jozanek: Renamed to `PlatformFileIOSuite`

tests/jvm/src/test/scala/scala/meta/tests/io/PlatformFileIOSuite.scala
kitbellew: i am not sure i understood. how does that explain whether `root`, the temporary directory, is deleted?

scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformFileIO.scala
kitbellew: repeat the comment about doing a wildcard import here.

scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformFileIO.scala
```diff
@@ -46,11 +47,21 @@ object PlatformFileIO {
...
-    while (iter.hasNext) {
-      val path = iter.next()
-      if (Files.isRegularFile(path)) relativeFiles += RelativePath(root.toNIO.relativize(path))
-    }
+    Files.walkFileTree(
+      root.toNIO,
+      new SimpleFileVisitor[Path] {
+        override def visitFile(file: Path, attrs: BasicFileAttributes): FileVisitResult = {
+          // Files.isRegularFile follows symlinks, matching the previous Files.walk behavior
+          if (Files.isRegularFile(file)) relativeFiles += RelativePath(root.toNIO.relativize(file))
```
kitbellew: repeat the comment about using `walkFileTree` overload which takes additional parameters, one of which instructs following symlinks: would that make it possible to use attrs.isRegularFile?

kitbellew (changes requested):
ideally, each comment is addressed in some way.

tests/jvm/src/test/scala/scala/meta/tests/io/PlatformFileIOSuite.scala
jozanek: haha, I missed and thought it's related to ok.txt which was having same issue, now both are deleted

scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformFileIO.scala
jozanek: oh my, I only understand now what you meant, see comment about symlink traversal

scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformFileIO.scala
jozanek: added

jozanek: done

scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformFileIO.scala
jozanek: done

scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformFileIO.scala
```diff
@@ -2,12 +2,12 @@ package scala.meta.internal.io
 
 import scala.meta.io._
 
+import java.io.IOException
 import java.net.URI
 import java.nio.charset.Charset
-import java.nio.file.{FileSystem, FileSystemAlreadyExistsException, FileSystems, Files, OpenOption,
-  Path}
+import java.nio.file._
+import java.nio.file.attribute.BasicFileAttributes
```
kitbellew, line 9: if you refer to it as `attribute.BasicFileAttributes`, you won't need this import. if using `attribute.` makes that line 56 split over multiple lines, then don't do it.

jozanek:
@kitbellew sorry for making this review messy (keeping 2 commits makes it difficult to have meaningful conversation), waiting for another round of review...

kitbellew (approved):
minor nitpick.

kitbellew:
> @kitbellew sorry for making this review messy (keeping 2 commits makes it difficult to have meaningful conversation), waiting for another round of review...

what can i say... not a big fan of github, prefer [gerrit](https://www.gerritcodereview.com/). :)

scalameta/io/jvm/src/main/scala/scala/meta/internal/io/PlatformFileIO.scala
jozanek: yeah, unfortunately it breaks it into 4 lines

PR #4636, opened 2026-06-07
Add erased subtyping query to the SemanticDB symbol table

### LINKED ISSUE #1213: Add support to test erased subtyping
Currently, it's not possible to query whether a symbol inherits another symbol. For example, it would be nice to have the ability check if `scala.reflect.io.FileOperationException` is a subtype of `Throwable` or if some `FooSuite` is a subtype of `org.scalatest.FunSuite`.

It's possible to workaround this limitation by using a small hack with JMV reflection

```scala
@ def toFqn(symbol: Symbol) = symbol.syntax.replaceAll("\\.$", "\\$").stripSuffix("#").stripPrefix("_root_.")
defined function toFqn

@ val cl = this.getClass.getClassLoader
cl: ClassLoader = ammonite.runtime.SpecialClassLoader@214b199c

// companion object
@ classOf[Throwable].isAssignableFrom(cl.loadClass(toFqn(Symbol("_root_.scala.reflect.io.FileOperationException."))))
res33: Boolean = false

// class
@ classOf[Throwable].isAssignableFrom(cl.loadClass(toFqn(Symbol("_root_.scala.reflect.io.FileOperationException#"))))
res34: Boolean = true
```
However, this solution does not work in for example Scala.js.

This functionality would be useful to implement scalafix rules like

* https://github.com/scalacenter/scalafix/issues/531

jozanek:
## Summary

Adds `SymbolTable.isSubtypeOf(child, parent): Boolean` — an *erased* subtyping query (symbols
compared, type arguments ignored) over the SemanticDB symbol table. Closes #1213.

## What changed

A default method on the `SymbolTable` trait (inherited by `GlobalSymbolTable`, `LocalSymbolTable`,
`AggregateSymbolTable`), walking `ClassSignature.parents` transitively via `info`:

- **Reflexive** for resolvable symbols; **cycle-safe**.
- **Follows type aliases** (`TypeSignature.upperBound`) and unwraps
  structural/annotated/existential/universal wrappers — needed because a class's direct parent is
  often an alias (`scala/MatchError#` → `scala/package.RuntimeException#` → `java/lang/RuntimeException#`).
- **Fails closed** — unresolvable symbols (incl. a missing `child`, even when `child == parent`)
  return `false`, never throw.

## Compatibility

Internal JVM-only API (`scala.meta.internal.symtab`, in `semanticdb-scalac-core`): no public-API /
MiMa impact, no `save-expect` change. JVM-only is intentional — it reads platform-independent
SemanticDB and its consumers run on the JVM.

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
@@ -1,10 +1,56 @@
 package scala.meta.internal.symtab
 
-import scala.meta.internal.semanticdb.SymbolInformation
+import scala.meta.internal.semanticdb.{AnnotatedType, ClassSignature, ExistentialType,
+  IntersectionType, StructuralType, SymbolInformation, Type, TypeRef, TypeSignature, UniversalType,
+  WithType}
+
+import scala.collection.mutable
 
 trait SymbolTable {
 
   /** Returns the SymbolInformation for the given symbol, or None if the symbol is missing. */
   def info(symbol: String): Option[SymbolInformation]
+
+  /**
+   * Tests whether `child` is an erased subtype of `parent`.
+   *
+   * "Erased" means only symbols are compared and type arguments are ignored — the semantics of
+   * `java.lang.Class.isAssignableFrom`, which is what the motivating Scalafix rules need (see
+   * scalameta/scalameta#1213). Unlike that JVM-reflection approach it reads the compiler-emitted
+   * SemanticDB rather than loading the analyzed program's classes, so it also covers targets whose
+   * classes are not JVM-loadable (e.g. Scala.js/Native projects); the query itself runs on the JVM.
+   *
+   * It is reflexive for resolvable symbols and walks `ClassSignature.parents` transitively,
+   * following type aliases through `TypeSignature.upperBound` (e.g. the scala-library parent
+   * `scala/package.RuntimeException#` resolves to `java/lang/RuntimeException#`) and unwrapping
+   * structural/annotated/existential/universal type wrappers. Symbols not resolvable via [[info]]
+   * (including a missing `child`, even when `child == parent`) yield `false` rather than throwing,
+   * and the traversal is cycle-safe.
+   */
+  def isSubtypeOf(child: String, parent: String): Boolean = {
+    val seen = mutable.Set.empty[String]
+    def loop(sym: String): Boolean = seen.add(sym) && info(sym)
+      .exists(i => sym == parent || parentSymbols(i).exists(loop))
+    loop(child)
+  }
+
+  private def parentSymbols(sinfo: SymbolInformation): List[String] = sinfo.signature match {
+    case cs: ClassSignature => cs.parents.iterator.flatMap(typeSymbols).toList
+    case ts: TypeSignature => typeSymbols(ts.upperBound) // follow type-alias upper bound
+    case _ => Nil
+  }
+
+  private def typeSymbols(tpe: Type): List[String] = tpe match {
+    case TypeRef(_, symbol, _) => symbol :: Nil
+    case WithType(types) => types.iterator.flatMap(typeSymbols).toList
+    case IntersectionType(types) => types.iterator.flatMap(typeSymbols).toList
+    // unwrap type wrappers to the underlying type(s); e.g. a refined parent is encoded as
+    // StructuralType(WithType(...)) (see scalacp/TypeOps.scala).
+    case StructuralType(t, _) => typeSymbols(t)
+    case AnnotatedType(_, t) => typeSymbols(t)
+    case ExistentialType(t, _) => typeSymbols(t)
+    case UniversalType(_, t) => typeSymbols(t)
+    case _ => Nil
+  }
 
 }
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
@@ -1,9 +1,8 @@
 package scala.meta.internal.symtab
 
-import scala.meta.internal.semanticdb.{AnnotatedType, ClassSignature, ExistentialType,
-  IntersectionType, StructuralType, SymbolInformation, Type, TypeRef, TypeSignature, UniversalType,
-  WithType}
+import scala.meta.internal.semanticdb._
 
+import scala.annotation.tailrec
 import scala.collection.mutable
 
 trait SymbolTable {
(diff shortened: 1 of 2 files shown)


semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -1,10 +1,56 @@
 package scala.meta.internal.symtab
 
-import scala.meta.internal.semanticdb.SymbolInformation
+import scala.meta.internal.semanticdb.{AnnotatedType, ClassSignature, ExistentialType,
```
kitbellew: would a wildcard import cause any conflicts? if not, could we switch to it?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -1,10 +1,56 @@
...
+   * following type aliases through `TypeSignature.upperBound` (e.g. the scala-library parent
+   * `scala/package.RuntimeException#` resolves to `java/lang/RuntimeException#`) and unwrapping
+   * structural/annotated/existential/universal type wrappers. Symbols not resolvable via [[info]]
+   * (including a missing `child`, even when `child == parent`) yield `false` rather than throwing,
+   * and the traversal is cycle-safe.
+   */
+  def isSubtypeOf(child: String, parent: String): Boolean = {
+    val seen = mutable.Set.empty[String]
+    def loop(sym: String): Boolean = seen.add(sym) && info(sym)
+      .exists(i => sym == parent || parentSymbols(i).exists(loop))
```
kitbellew:
```suggestion
    def loop(sym: String): Boolean = sym == parent ||
      seen.add(sym) && info(sym).exists(parentSymbols(_).exists(loop))
```

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -1,10 +1,56 @@
...
+    loop(child)
+  }
+
+  private def parentSymbols(sinfo: SymbolInformation): List[String] = sinfo.signature match {
```
kitbellew: return Iterator

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -1,10 +1,56 @@
...
+    case WithType(types) => types.iterator.flatMap(typeSymbols).toList
+    case IntersectionType(types) => types.iterator.flatMap(typeSymbols).toList
+    // unwrap type wrappers to the underlying type(s); e.g. a refined parent is encoded as
+    // StructuralType(WithType(...)) (see scalacp/TypeOps.scala).
+    case StructuralType(t, _) => typeSymbols(t)
+    case AnnotatedType(_, t) => typeSymbols(t)
+    case ExistentialType(t, _) => typeSymbols(t)
+    case UniversalType(_, t) => typeSymbols(t)
+    case _ => Nil
+  }
```
kitbellew:
```suggestion
  @tailrec
  private def typeSymbols(tpe: Type): Iterator[String] = tpe match {
    case TypeRef(_, symbol, _) => Iterator(symbol)
    case WithType(types) => typeSymbolsFlatten(types)
    case IntersectionType(types) => typeSymbolsFlatten(types)
    // unwrap type wrappers to the underlying type(s); e.g. a refined parent is encoded as
    // StructuralType(WithType(...)) (see scalacp/TypeOps.scala).
    case StructuralType(t, _) => typeSymbols(t)
    case AnnotatedType(_, t) => typeSymbols(t)
    case ExistentialType(t, _) => typeSymbols(t)
    case UniversalType(_, t) => typeSymbols(t)
    case _ => Iterator.empty
  }

  private def typeSymbolsFlatten(types: Iterable[Type]): Iterator[String] =
    types.iterator.flatMap(typeSymbols)
```

tests-semanticdb/src/test/scala/scala/meta/tests/symtab/SubtypingSuite.scala
```diff
@@ -0,0 +1,87 @@
...
+ * path, and cycle safety. Uses `assert` per project convention.
+ */
+class SubtypingSuite extends FunSuite {
+
+  private val classpath = Library.scalaLibrary.classpath() ++
+    Classpath(BuildInfo.databaseClasspath +: BuildInfo.classDirectories: _*)
+  private val symtab = GlobalSymbolTable(classpath, includeJdk = true)
+
+  private def isSubtype(child: String, parent: String)(implicit loc: munit.Location): Unit =
+    test(s"$child <: $parent")(assert(symtab.isSubtypeOf(child, parent)))
```
kitbellew: would it be a stupid idea to check `isSubtypeOf` twice? the first time i saw your `seen` cycle-protector, i thought, it returns `false` if something is added to it a second time, and we need to test running the method a second time; obviously, it's local to that method but the thought persisted. :)

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -1,10 +1,56 @@
...
+   * classes are not JVM-loadable (e.g. Scala.js/Native projects); the query itself runs on the JVM.
+   *
+   * It is reflexive for resolvable symbols and walks `ClassSignature.parents` transitively,
+   * following type aliases through `TypeSignature.upperBound` (e.g. the scala-library parent
+   * `scala/package.RuntimeException#` resolves to `java/lang/RuntimeException#`) and unwrapping
+   * structural/annotated/existential/universal type wrappers. Symbols not resolvable via [[info]]
+   * (including a missing `child`, even when `child == parent`) yield `false` rather than throwing,
+   * and the traversal is cycle-safe.
+   */
+  def isSubtypeOf(child: String, parent: String): Boolean = {
```
kitbellew, line 34: why do we take Strings and not something more structured? we are added this functionality but no real-life usage examples.

kitbellew (changes requested):
bigger question:

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: done

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: updated

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: yep!

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: applied

tests-semanticdb/src/test/scala/scala/meta/tests/symtab/SubtypingSuite.scala
jozanek: Not a stupid idea at all.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: Strings are the native form of SemanticDB symbols. Every field this walk traverses (`SymbolInformation.symbol`, `TypeRef.symbol`) is a string in the protobuf schema, so a wrapper would be unwrapped again at the next hop. It also mirrors the trait's only other method, `info(symbol: String)`.

Real-life usage is the scalafix rule from #1213 (scalacenter/scalafix#531): `symtab.isSubtypeOf(returnTypeSym, "java/lang/Throwable#")`, where `returnTypeSym` comes straight from a `TypeRef`. I've added that example to the scaladoc.

kitbellew (approved):

PR #4637, opened 2026-06-08
Add regression test for parenthesized infix synthetic position (#1223)

### LINKED ISSUE #1223: Parenthesis affecting infix operator parsing

jozanek:
Wrapping an infix application in parentheses must not shift the implicit synthetic off the application. In scalameta 2.1.2 (reported in #1223) the `CanBuildFrom` synthetic for `(a zip b)` was attached at the second operand instead of spanning the whole application.

The current range-based synthetics already handle this correctly. This adds a focused `Issue1223Suite` (extending `PrintSuiteBase`) asserting the exact synthetic ranges:

- `(2 :: Nil)` → `[2:3..2:9): :: Nil => *[Int]` — cross-version; the type-application tracks `:: Nil`, shifted only by the leading `(`.
- `(nums zip strs)` on **2.12** → `[2:46..2:59): nums zip strs => *(Seq.canBuildFrom[Tuple2[Int, String]])` — the implicit-arg synthetic spans the whole parenthesized application (the exact 2.1.2 failure).
- Same snippet on **2.13** → `[2:46..2:54): nums zip => *[String]` (`zip` no longer takes `CanBuildFrom`); the positioning invariant still holds.

Closes #1223

### DIFF
--- a/tests-semanticdb/src/test/scala/scala/meta/tests/semanticdb/Issue1223Suite.scala
+++ b/tests-semanticdb/src/test/scala/scala/meta/tests/semanticdb/Issue1223Suite.scala
@@ -1,0 +1,40 @@
+package scala.meta.tests.semanticdb
+
+// https://github.com/scalameta/scalameta/issues/1223
+// Wrapping an infix application in parentheses must not shift the synthetic off
+// the application. In scalameta 2.1.2 the implicit-argument synthetic was attached
+// at the second operand instead of spanning the whole `a zip b` application.
+//
+// Snippets are wrapped in `object Wrapped { ... }` by PrintSuiteBase, so source
+// lines start at line 2. Synthetic ranges are printed by Print.synthetic (Compact).
+class Issue1223Suite extends PrintSuiteBase {
+
+  // Cross-version: the `*[Int]` type-application tracks `:: Nil` (cols 3..9),
+  // shifted by exactly the leading `(` relative to an un-parenthesized `2 :: Nil`.
+  checkSynthetics(
+    "(2 :: Nil)",
+    """|[2:3..2:9): :: Nil => *[Int]
+       |""".stripMargin,
+  )
+
+  private val zip = "class C(nums: Seq[Int], strs: Seq[String]) { (nums zip strs) }"
+
+  if (ScalaVersion.is212)
+    // 2.12: zip takes an implicit CanBuildFrom. The synthetic spans the full
+    // parenthesized application `nums zip strs`, not just the second operand.
+    checkSynthetics(
+      zip,
+      """|[2:46..2:59): nums zip strs => *(Seq.canBuildFrom[Tuple2[Int, String]])
+         |[2:46..2:54): nums zip => *[Int, String, Seq[Tuple2[Int, String]]]
+         |""".stripMargin,
+    )
+  else
+    // 2.13: zip is a plain type-application (no CanBuildFrom). Same invariant:
+    // the synthetic tracks the infix application, not the `(`.
+    checkSynthetics(
+      zip,
+      """|[2:46..2:54): nums zip => *[String]
+         |""".stripMargin,
+    )
+
+}


tests-semanticdb/src/test/scala/scala/meta/tests/semanticdb/Issue1223Suite.scala
```diff
@@ -0,0 +1,40 @@
...
+  )
+
+  private val zip = "class C(nums: Seq[Int], strs: Seq[String]) { (nums zip strs) }"
+
+  if (ScalaVersion.is212)
+    // 2.12: zip takes an implicit CanBuildFrom. The synthetic spans the full
+    // parenthesized application `nums zip strs`, not just the second operand.
+    checkSynthetics(
+      zip,
+      """|[2:46..2:59): nums zip strs => *(Seq.canBuildFrom[Tuple2[Int, String]])
```
kitbellew, line 27: shouldn't it be inserted at [59:59), instead of [46,59) (or wherever `strs` stops)? i am not sure i remember what a synthetic is supposed to convey and over which range.

also, how does that reflect the example in the issue it closes, with one position inside parens and the other outside?

jozanek: Good question — the range is not an insertion point, it's the source span of the original expression the synthetic applies to, with `*` being an `OriginalTree` placeholder for that original.

So `[46:59): nums zip strs => *(Seq.canBuildFrom[…])` reads as: *the original application `nums zip strs` (cols 46–59) receives a synthesized implicit argument list `(Seq.canBuildFrom[…])`.* The `*` is the `orig(nums zip strs)` subtree; `checkTrees` prints it as `orig(nums zip strs)(Seq.canBuildFrom[…])`.

This is the established convention for implicit-argument synthetics — e.g. `PrintSuite` (2.12) asserts:

```
[2:0..2:18): List(1).map(_ + 2) => *(List.canBuildFrom[Int])
orig(List(1).map(_ + 2))(List.canBuildFrom[Int])
```

i.e. the range spans the whole `List(1).map(_ + 2)`, not a zero-width point after it. A `[59:59)` would be inconsistent with every other implicit-args synthetic, and would lose the association back to the user's expression that tooling (Metals, etc.) relies on to render the inferred implicit.

> how does that reflect the example in the issue, with one position inside parens and the other outside?

That framing was specific to 2.1.2, whose synthetics were *point/text insertions* — the reported bug was that the point landed at offset 74 (inside the parens, at the second operand `b`) instead of reflecting the application. The current SemanticDB format replaced point-insertion with a structured tree carrying the **range of the original application**, so the inside-vs-outside-parens choice no longer exists: the synthetic is anchored to the range of `nums zip strs` itself. It sits inside the parens because the parens are just grouping and aren't part of the typed `Apply` node — consistent with `(2 :: Nil)` giving `[2:3..2:9): :: Nil` (the application, not the `(`). The test asserts exactly that invariant: parentheses don't displace the synthetic off the application.

Happy to add a short comment in the suite spelling out what the range conveys, if that helps readers.

kitbellew: @tgodzik mind chiming in, please?

tgodzik: I think it's fine as is, but honestly we no longer use synthetics for anything, there might be some scalafix rule that uses it, but none that I know of. Let's document the current status and if needed we'll change it.

Issue #1494, opened 2018-04-12
Scalacp does not emit Type for final val in scala-library


olafurpg:
Observe that there are no `tpe { .. }` values below
```
metap -proto $(metacp $(coursier fetch org.scala-lang:scala-library:2.12.4)) | grep 'duration.package.DAYS' -A 10
        declarations: "scala.concurrent.duration.package.DAYS(?)."
        declarations: "scala.concurrent.duration.package.DAYS."
        declarations: "scala.concurrent.duration.package.HOURS(?)."
        declarations: "scala.concurrent.duration.package.HOURS."
        declarations: "scala.concurrent.duration.package.MICROSECONDS(?)."
        declarations: "scala.concurrent.duration.package.MICROSECONDS."
        declarations: "scala.concurrent.duration.package.MILLISECONDS(?)."
        declarations: "scala.concurrent.duration.package.MILLISECONDS."
        declarations: "scala.concurrent.duration.package.MINUTES(?)."
        declarations: "scala.concurrent.duration.package.MINUTES."
        declarations: "scala.concurrent.duration.package.NANOSECONDS(?)."
        declarations: "scala.concurrent.duration.package.NANOSECONDS."
--
    symbol: "scala.concurrent.duration.package.DAYS(?)."
    kind: METHOD
    properties: 1032
    name: "DAYS"
    accessibility {
      tag: PUBLIC
      symbol: ""
    }
    owner: "scala.concurrent.duration.package."
    language: SCALA
  }
--
    symbol: "scala.concurrent.duration.package.DAYS."
    kind: FIELD
    properties: 1032
    name: "DAYS"
    accessibility {
      tag: PRIVATE_THIS
      symbol: ""
    }
    owner: "scala.concurrent.duration.package."
    language: SCALA
```

In particular, I expected the `DAYS(?).` symbol to have a `tpe`.

xeno-by:
Maybe this happens because final vals are supposed to be inlined, so their signatures are not fully computed? Another hypothesis is that Scalap doesn't correctly understand ConstantType types.

gabro:
Digging deeper, the low-level reason why it crashes is that `.info` returns 50 for that symbol and that's not handle by the `typeEntry` parser here:

https://github.com/scala/scala/blob/6ff3facb1791f54ad10fbf44059d892e5fd00510/src/scalap/scala/tools/scalap/scalax/rules/scalasig/ScalaSig.scala#L213-L231

[...]

PR #4638, opened 2026-06-08
Remove stale SuperType FIXME in metacp TypeOps (closes #1291)

### LINKED ISSUE #1291: Support SuperType in metacp
Not implemented now because scalap doesn't support this.

jozanek:
Closes #1291.

`metacp` reads only pickled *signatures* (the declared `info` of symbols) out of `.class` files via the upstream `scalap` library. `SuperType` (`This.super[Mixin]`) is the type of `super` references - an **expression-level** type. scalac never pickles it (pickles store declaration signatures, not method bodies), and scalap's `ScalaSig` type parser has no `SUPERtpe` case in its `Type` ADT, so `metacp` can never receive a `SuperType`. The scalac/`TextDocumentOps` path handles `SuperType` because it walks fully-typed source trees; the pickle path provably doesn't need to.

So #1291 is a can't-happen case, not a missing feature. This replaces the stale commented-out FIXME in `scalacp/TypeOps.scala` (which used an invalid `g.` prefix copied from the scalac-side TypeOps and would not compile here) with an accurate explanatory comment. The block's second half — the `#1494` JAVA_ENUM concern - is already handled by the live `ConstantType(underlying: ExternalSymbol)` case just below it, so it's dropped.

Comment-only change: no behavioral change, no MiMa impact, no expect-file regeneration. A test is not expressible, since no input can make scalap emit a `SuperType`; the existing `case other => sys.error(...)` catch-all already covers any surprise.

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/scalacp/TypeOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/scalacp/TypeOps.scala
@@ -37,14 +37,11 @@ trait TypeOps {
         case ThisType(sym) =>
           val ssym = sym.ssym
           s.ThisType(ssym)
-        // // FIXME: https://github.com/scalameta/scalameta/issues/1291
-        // case g.SuperType(gpre, gmix) =>
-        //   ???
-        // // FIXME: Could it be that this missing pattern matching case may have been
-        // // the reason for mysterious Metacp crashes?
-        // // https://github.com/scalameta/scalameta/issues/1494
-        // case g.ConstantType(g.Constant(sym: g.TermSymbol)) if sym.hasFlag(gf.JAVA_ENUM) =>
-        //   ???
+        // NOTE: There is deliberately no SuperType case. SuperType (`This.super[Mixin]`) is the
+        // type of `super` references, an expression-level type that scalac never pickles (pickles
+        // store declaration signatures, not method bodies), and scalap has no SUPERtpe parser -- so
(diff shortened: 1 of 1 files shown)


semanticdb/scalac/library/src/main/scala/scala/meta/internal/scalacp/TypeOps.scala
```diff
@@ -37,14 +37,11 @@ trait TypeOps {
...
-        // case g.SuperType(gpre, gmix) =>
-        //   ???
-        // // FIXME: Could it be that this missing pattern matching case may have been
-        // // the reason for mysterious Metacp crashes?
-        // // https://github.com/scalameta/scalameta/issues/1494
-        // case g.ConstantType(g.Constant(sym: g.TermSymbol)) if sym.hasFlag(gf.JAVA_ENUM) =>
-        //   ???
+        // NOTE: There is deliberately no SuperType case. SuperType (`This.super[Mixin]`) is the
+        // type of `super` references, an expression-level type that scalac never pickles (pickles
+        // store declaration signatures, not method bodies), and scalap has no SUPERtpe parser -- so
```
kitbellew: is `SUPERtpe` intentional?

jozanek: Yes. It's the pickle-format tag constant (scala.reflect.internal.pickling.PickleFormat, like NOtpe/THIStpe). scalap's reader has no case for it, so it can't decode one. Pushed a follow-up rewording the comment.

kitbellew (approved):

Issue #1298, opened 2018-02-10
SemanticDB find polymorphic candidates


MasseGuillaume:
```scala
object A {
  def foo(x: String): Int = 1     // M1
  def foo(y: Int): Int = 2          // M2
  val a = foo(1) // M2
//              ^
}
```

I would like to query the apply `foo(1)` and know the alternatives: M1, M2.

context: https://github.com/scalacenter/scalafix/pull/594#issuecomment-364429865

MasseGuillaume:
I'm currently investigating the scalac api to know if it's possible to do this.

MasseGuillaume:
This is one possible solution: https://github.com/scalacenter/scalafix/pull/594#issuecomment-364657589

PR #4639, opened 2026-06-09
Add SymbolTable.alternatives for same-owner overload candidates

jozanek:
Partial fix for #1298.

At a resolved call site scalac picks a single overload, so the SemanticDB occurrence points to one disambiguated symbol (e.g. `A.foo(+1).`). This adds an internal `SymbolTable.alternatives(symbol): Option[List[String]]` that recovers the sibling overload candidates declared in the same owner (`A.foo().`, `A.foo(+1).`).

The contract (`Option` so callers tell "could not inspect" from "no other overload") and the same-owner scope limitation (why inherited/split overloads are excluded) are documented in the method's scaladoc. It's a concrete default method on the trait, so `Global`/`Local`/`Aggregate` tables inherit it, and lives in `scala.meta.internal.*`, so it's MiMa-exempt.

Heads up: the scaladoc is on the longer side — it grew as I iterated and kept hitting one more edge case (partiality, never fabricating results, hardlinks, the inheritance/override limits). I'd rather over-document the contract than leave it implicit, but happy to shorten it if it reads as too verbose.


semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -1,10 +1,54 @@
...
+   *     from a symbol alone); or a method whose declaring owner is unavailable here or does not
+   *     declare it.
+   *
+   * Scope: only overloads declared in the *same owner* are returned. Overloads inherited from, or
+   * split across, supertypes are intentionally excluded — a classpath symbol table cannot resolve
+   * them safely: `overriddenSymbols` is `Nil` outside the compiler plugin, so override and overload
+   * are indistinguishable when walking parents, and the full receiver-type candidate set further
+   * needs the call site's static receiver type, which a symbol alone does not carry.
+   */
+  def alternatives(symbol: String): Option[List[String]] =
```
kitbellew: 1. why not call it overloads? it doesn't return alternatives (meaning, all **other** overloads, excluding this one)
2. why not drop `Option`? a `Nil` response means symbol was not found, non-`Nil` shows all overloads including this one.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -1,10 +1,54 @@
...
+    if (symbol.isNone) None
+    else if (symbol.isMulti) Some(symbol.asMulti)
+    else symbol.desc match {
+      case d.Method(name, _) => info(symbol.owner).map(_.signature).flatMap {
+          case c: ClassSignature =>
+            val sameName = c.declarations.symbols.filter(_.desc match {
+              case d.Method(v, _) => v == name
+              case _ => false
+            })
+            if (sameName.contains(symbol)) Some(sameName) else None
```
kitbellew: what does this check do? does it make sure that `symbol` does not just _look_ like a method with this name but actually corresponds to one of them?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -1,10 +1,54 @@
...
+   *     declare it.
+   *
+   * Scope: only overloads declared in the *same owner* are returned. Overloads inherited from, or
+   * split across, supertypes are intentionally excluded — a classpath symbol table cannot resolve
+   * them safely: `overriddenSymbols` is `Nil` outside the compiler plugin, so override and overload
+   * are indistinguishable when walking parents, and the full receiver-type candidate set further
+   * needs the call site's static receiver type, which a symbol alone does not carry.
+   */
+  def alternatives(symbol: String): Option[List[String]] =
+    if (symbol.isNone) None
```
kitbellew: optional alternative:
```
    if (symbol.isMulti) Some(symbol.asMulti)
    else if (symbol.isGlobal) {
      val (desc, owner) = DescriptorParser(symbol)
      desc match {
        case Descriptor.Method(name, _) => info(owner).map(_.signature).flatMap {
            case c: ClassSignature =>
              val sameName = c.declarations.symbols.filter(_.desc match {
                case Descriptor.Method(`name`, _) => true
                case _ => false
              })
              if (sameName.contains(symbol)) Some(sameName) else None
            case _ => None
          }
        case _ => None
      }
    } else None
```

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: 1. I took naming from issue's author, but makes more sense to accept your proposal.
2. Also makes sense, Nil is basically None.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: Yes, exactly. It guards against fabrication. if the owner resolves, but doesn't actually declare this symbol (e.g. `scala/Predef.thisIsNotAMethod().`), `sameName` won't contain it, so we return empty rather than inventing a candidate.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: updated

tests-semanticdb/src/test/scala/scala/meta/tests/symtab/SymbolTableSuite.scala
```diff
@@ -85,47 +85,45 @@ class SymbolTableSuite extends FunSuite {
 
   // https://github.com/scalameta/scalameta/issues/1298
   // `scala.Predef.assert` is overloaded: assert(assertion) and assert(assertion, message).
-  private val assertAlternatives = Set("scala/Predef.assert().", "scala/Predef.assert(+1).")
+  private val assertOverloads = Set("scala/Predef.assert().", "scala/Predef.assert(+1).")
```
kitbellew: i apologise, i meant to ask: why do you convert to a set? is the `overloads` order non-deterministic, with +2 following +1 etc.?

jozanek: It's deterministic. Declaration order matches the disambiguator order (base, then +1, +2). I used `Set` because I wasn't sure metacp preserved that order. It does (confirmed on 2.12/2.13), so switched to the ordered `List` and documented it.

kitbellew (approved):

PR #4640, opened 2026-06-09
Add TextDocument.build_target

### LINKED ISSUE #1307: Add TextDocument.buildTarget
Looks like it would be beneficial to add something like `TextDocument.buildTarget` to keep track of build targets that generated the text document.

jozanek:
Closes #1307.

Adds an optional `build_target` field (tag 13) to the SemanticDB `TextDocument` so a payload can record which build target produced it. The scalac plugin populates it from a new `-P:semanticdb:buildtarget:<id>` option; `metap` prints it; the spec documents it.

Design:
- First-class string field, resolving the thread's "first-class field vs. uri query-param" question in favor of a field. Optional/opt-in, so existing producers are unaffected.
- Backward/forward compatible: a new optional proto3 string at an unused tag needs no new `Schema` version.
- Spec recommends a stable, URI-like identifier (e.g. BSP `BuildTargetIdentifier.uri`).

Scope: covers the scalac source-document path (snapshot dedup across platforms/Scala versions). Dependency/library provenance via metacp is intentionally out of scope and left as a follow-up — it needs per-classpath-entry attribution (a single metacp run spans many jars) and revisiting the schema-vs-index question.


semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/ConfigOps.scala
```diff
@@ -68,6 +71,7 @@ object SemanticdbConfig {
   private val SetExclude = "exclude:(.*)".r
   private val SetSourceroot = "sourceroot:(.*)".r
   private val SetTargetroot = "targetroot:(.*)".r
+  private val SetBuildtarget = "buildtarget:(.*)".r
```
kitbellew: should it be `.+`, since it cannot be empty?

semanticdb/semanticdb/shared/src/main/proto/semanticdb.proto
```diff
@@ -25,6 +25,7 @@ message TextDocument {
   string text = 3;
   string md5 = 11;
   Language language = 10;
+  string build_target = 13;
```
kitbellew: nit: why are these fields not sorted by their field id?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/ConfigOps.scala
jozanek: agree

semanticdb/semanticdb/shared/src/main/proto/semanticdb.proto
jozanek: Existing fields were never id-sorted. They're grouped by kind (scalars then repeated sections). I placed `build_target=13` in the scalar group, now updated to last position.

kitbellew (approved):

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/ConfigOps.scala
```diff
@@ -13,6 +13,7 @@ case class SemanticdbConfig(
     fileFilter: FileFilter,
     sourceroot: AbsolutePath,
     targetroot: AbsolutePath,
+    buildTarget: String,
```
tgodzik, line 16: Looks like this might have caused issues like:
```
java.lang.NoSuchMethodError: 'java.lang.String scala.meta.internal.semanticdb.scalac.SemanticdbConfig.copy$default$6()' 
```
for scalafix. We might need to revert this change or use unroll 🤔 

The issue popped up in https://github.com/scalameta/metals/actions/runs/28714688800/job/85153908853?pr=8644

But will probably show up any time scalafix's semanticdb differs from the one set in sbt. Weirdly enough Scalafix doesn't use SemanticdbConfig class directly 🤔

kitbellew: #4678

PR #4642, opened 2026-06-10
chore: Test Scala 3.9.0

tgodzik:


kitbellew:
@tgodzik i pushed a change into your pr branch, after the build refactoring with versions. in case you were planning to re-use it for release PR.

project/Versions.scala
```diff
@@ -14,9 +14,8 @@ object Versions {
     // the first two are tested in pre-merge for JVM and JS, and in post-merge for Native
     // NB: use one of these two lines for RC testing, as long as it's not merged
     "3.3.8" -> "3_lts",
-    "3.8.4" -> "3_next",
+    "3.9.0" -> "3_next",
```
kitbellew, line 17: i already added 3.9.0 as its own line, below. i thought 3.10 will be next next. 😀

tgodzik: ach, I totally missed it. Github doesn't show this :/ 

I can close then, will open 3.10 after the first RC is out

PR #4643, opened 2026-06-10
chore: Add Scala 3.9 dialect

tgodzik:


tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/TypeSuite.scala
```diff
@@ -567,7 +567,7 @@ class TypeSuite extends BaseDottySuite {
   }
 
   test("into-type") {
-    implicit val dialect: Dialect = dialects.Scala3Future
+    implicit val dialect: Dialect = dialects.Scala39
```
kitbellew: since this new dialect doesn't change any flags, should this be an earlier dialect instead?

scalameta/dialects2/shared/src/main/scala/scala/meta/dialects/package.scala
```diff
@@ -100,7 +100,9 @@ package object dialects {
 
   implicit val Scala38: Dialect = Scala37.withAllowParameterTypeConversions(true)
 
-  implicit val Scala3: Dialect = Scala38
+  implicit val Scala39: Dialect = Scala38
```
kitbellew, line 103: side comment: will the next one be called 310 or 3_10?

tgodzik: I would say 310.

tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/TypeSuite.scala
tgodzik: Right, changed to Scala38

kitbellew (approved):

PR #4650, opened 2026-06-11
Emit constructor parameter occurrences at class parameter definition sites

### LINKED ISSUE #1327: Consider using multisymbols for class parameter definitions in SymbolOccurrences
```scala
class C(val x: Int)
```

In https://github.com/scalameta/scalameta/pull/1326, the `SymbolOccurrence` for `x` is associated with `_empty_.C#x()` - the getter for `x`. This is done to make sure that most references to `x` have the same symbol as the definition of `x`, so that it's easier to migrate existing SemanticDB-based tools.

However, strictly speaking the definition occurrence should be a multisymbol `_empty_.C#x().;_empty_.C#x.`, because the spec says that `x` defines multiple symbols at once. Let's consider what we should do here.

jozanek:
Fixes #1327.

For `class C(val x: Int)`, the definition of `x` introduces two symbols — the accessor (`C#x.`) and the constructor parameter (``C#`<init>`().(x)``) - but only the accessor received a `SymbolOccurrence`. As agreed in #1327, the definition site now emits multiple `SymbolOccurrence`s covering the same range, one per defined symbol. Internally the pair is staged as a `Symbols.Multi` and expanded by the existing `asMulti` step in `finalOccurrences`, so multi-symbol strings never reach the payload.

The accessor and the parameter are paired by name: scalac keeps no symbol-level link between them, and name matching is the same relation `scalacp.SymbolInformationOps` uses in the opposite direction.

Consumer impact: references are unchanged - they still resolve to the accessor only. At constructor-parameter definition sites there are now two DEFINITION occurrences sharing one range, so "first occurrence at position" lookups see only one of the two symbols. Tools that want both must collect all occurrences at the range. SymbolInformation payloads are unchanged.

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/TextDocumentOps.scala
@@ -244,6 +244,22 @@ trait TextDocumentOps {
               setterInfos.foreach(info => symbols(info.symbol) = info)
             }
           }
+          // A `val`/`var` parameter of a primary constructor (and any plain parameter that scalac
+          // turns into a field) gives rise to more than one symbol: the accessor (the getter, or the
+          // underlying field) and the constructor PARAMETER symbol. References resolve to the
+          // accessor (see the #1538 FIXME below), but per
+          // https://github.com/scalameta/scalameta/issues/1327 the definition site covers both, so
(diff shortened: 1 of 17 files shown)


kitbellew (approved):
with all these changes, scala3 semanticdb will fall hopelessly behind. any thoughts on doing stuff there in parallel? 😂

kitbellew (approved):

jozanek:
> with all these changes, scala3 semanticdb will fall hopelessly behind. any thoughts on doing stuff there in parallel? 😂

@kitbellew done in https://github.com/scala/scala3/pull/26410

PR #4655, opened 2026-06-14
Add SYNTHETIC property to SemanticDB symbols

### LINKED ISSUE #1492: Missing SYNTHETIC or CASE property for generated case class methods

jozanek:
Closes #1492.

## What & why

SemanticDB had no way to mark a symbol as compiler-synthesized, so generated members (case class / `AnyVal` methods, pattern-match temporaries, SAM `$anon` classes) were indistinguishable from user-written code and downstream tools (Metals, Scalafix) couldn't filter them.

This adds a dedicated `SYNTHETIC = 0x800000` property. Per the #1492 discussion it's a new property rather than an overloaded `CASE`, and it takes the next free slot after `OVERRIDE` (reserved `0x1`/`0x2` untouched), so it's a backward- and forward-compatible schema addition.

## Worth a closer look

- The property is set on any `gsym.isSynthetic` symbol in **both** producers (metac plugin and metacp/scalacp scalasig reader) to keep them consistent.
- **Self parameters are excluded**: scalac flags them synthetic even when written in source (`self =>`), and the spec grants self params no properties.
- Spec adds `SYNTHETIC` to the per-kind property lists (value/variable/method/parameter/class/object) and the previously-missing `OVERRIDE` table row.

## Out of scope

Java enum `values()`/`valueOf` are not marked — they aren't `ACC_SYNTHETIC` in bytecode, so metacp's Java path can't detect them and marking in metac alone would diverge from metacp. Left for a follow-up if desired.


semanticdb/scalac/library/src/main/scala/scala/meta/internal/scalacp/SymbolInformationOps.scala
```diff
@@ -55,6 +55,7 @@ trait SymbolInformationOps {
       def isAbstractClass = sym.isClass && sym.isAbstract && !sym.isTrait
       def isAbstractMethod = sym.isMethod && sym.isDeferred
       def isAbstractType = sym.isType && !sym.isParam && sym.isDeferred
+      if (!sym.isPackage && sym.isSynthetic) flip(p.SYNTHETIC)
```
kitbellew: looks like this method should simply start with
```
  if (sym.isPackage) return 0 // XXX: returning
```

also, can synthetic be set on a java symbol?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolInformationOps.scala
```diff
@@ -61,6 +61,9 @@ trait SymbolInformationOps {
       def isAbstractMethod = gsym.isMethod && gsym.isDeferred
       def isAbstractType = gsym.isType && !gsym.isParameter && gsym.isDeferred
       def isObject = gsym.isModule && !gsym.hasFlag(gf.PACKAGE)
+      // NOTE: self parameters carry the synthetic flag even when written in source
```
kitbellew: just like above, this method could also start with
```
  if (gsym.hasFlag(gf.PACKAGE)) return 0 // XXX: returning
```

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolInformationOps.scala
```diff
@@ -61,6 +61,9 @@ trait SymbolInformationOps {
...
+      // (e.g. `self =>`), and the spec grants them no properties, so exclude them.
+      if (!gsym.hasFlag(gf.PACKAGE) && !gsym.isSelfParameter && gsym.isSynthetic) flip(p.SYNTHETIC)
```
kitbellew: this self-parameter check is not present in scalacp version of this logic, why?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolInformationOps.scala
jozanek: done

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolInformationOps.scala
jozanek: scalacp never emits self-parameter symbols. Its `kind` has the `SELF_PARAMETER` case commented out (scalap doesn't model self-params), so there's nothing to guard. Added a code comment noting this.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/scalacp/SymbolInformationOps.scala
jozanek: Done. It could be set on Java symbols, so I moved the check into the `else` (Scala) branch - `SYNTHETIC` is now Scala-only. I will have a look and add it in folow up PR.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolInformationOps.scala
kitbellew: two lines above, there's another package check, should we remove it?

kitbellew (changes requested):

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolInformationOps.scala
jozanek: Good catch, removed it (and also inlined `isObject`).

kitbellew (approved):

PR #4656, opened 2026-06-14
Add hierarchy-aware SymbolTable.overloads(tpe, name)

jozanek:
Partial fix for #1298, follow-up to #4639.

#4639's `overloads(symbol)` only finds overloads declared in a method's own owner. This adds the hierarchy-aware `SymbolTable.overloads(tpe, name)`: given a receiver type symbol and a method name, it walks `tpe`'s linearization, collects every same-named method (own + inherited), and override-deduplicates by erased parameter signature — substituting parent type arguments first, so a generic override (`Base[A]` / `Sub extends Base[String]`) collapses to one while genuine overloads survive. Constructors aren't inherited, so `<init>` stays on the receiver type.

Reuses the type-hierarchy walk shape from #4639's `isSubtypeOf` (left untouched); the contract and caveats live in the method's scaladoc. Internal (`scala.meta.internal.symtab`) → MiMa-exempt.

The remaining piece — extracting the static receiver type at the call site — is Scalafix-side.

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
@@ -36,17 +36,11 @@ trait SymbolTable {
     else if (symbol.isGlobal) {
       val (desc, owner) = DescriptorParser(symbol)
       desc match {
-        case d.Method(name, _) => info(owner).map(_.signature) match {
-            case Some(c: ClassSignature) =>
-              // keep only members that really are same-named methods of this owner, so a symbol
-              // that merely parses like one (but isn't declared here) isn't fabricated into the set
-              val sameName = c.declarations.symbols.filter(_.desc match {
-                case d.Method(`name`, _) => true
-                case _ => false
-              })
-              if (sameName.contains(symbol)) sameName else Nil
-            case _ => Nil
-          }
+        case d.Method(name, _) =>
+          // keep `owner`'s same-name methods only if `symbol` is actually one of them, so a symbol
+          // that merely parses like a method of `owner` (but isn't declared there) isn't fabricated
+          val sameName = membersNamed(owner, name)
+          if (sameName.contains(symbol)) sameName else Nil
         case _ => Nil
       }
     } else Nil
@@ -81,12 +75,10 @@ trait SymbolTable {
     // constructors are not inherited, so do not walk parents for the `<init>` name
     val bases = if (name == "<init>") List(tpe -> Map.empty[String, String]) else linearization(tpe)
     val seenSignatures = mutable.Set.empty[List[String]]
-    bases.flatMap { case (owner, env) => membersNamed(owner, name).map(_ -> env) }.filter {
-      case (method, env) => erasedSignature(method, env) match {
-          case Some(signature) => seenSignatures.add(signature)
-          case None => true // unresolved parameter types: keep rather than risk a wrong merge
-        }
-    }.map(_._1)
+    bases.flatMap { case (owner, env) =>
+      // an unresolved erased signature (None) keeps the method rather than risk a wrong merge
+      membersNamed(owner, name).filter(erasedSignature(_, env).forall(seenSignatures.add))
+    }
   }
 
   /**
@@ -117,28 +109,34 @@ trait SymbolTable {
     loop(child)
   }
 
-  private def parentSymbols(sinfo: SymbolInformation): Iterator[String] = sinfo.signature match {
-    case cs: ClassSignature => typeSymbolsFlatten(cs.parents)
-    case ts: TypeSignature => typeSymbols(ts.upperBound) // follow type-alias upper bound
-    case _ => Iterator.empty
-  }
+  // Parent (symbol, type-arguments) pairs from a class signature, or the alias target for a type.
+  private def parentRefs(sinfo: SymbolInformation): Iterator[(String, Seq[Type])] =
+    sinfo.signature match {
+      case cs: ClassSignature => typeRefsFlatten(cs.parents)
+      case ts: TypeSignature => typeRefs(ts.upperBound) // follow type-alias upper bound
+      case _ => Iterator.empty
+    }
 
   @tailrec
-  private def typeSymbols(tpe: Type): Iterator[String] = tpe match {
-    case TypeRef(_, symbol, _) => Iterator(symbol)
-    case WithType(types) => typeSymbolsFlatten(types)
-    case IntersectionType(types) => typeSymbolsFlatten(types)
+  private def typeRefs(tpe: Type): Iterator[(String, Seq[Type])] = tpe match {
+    case TypeRef(_, symbol, args) => Iterator(symbol -> args)
+    case WithType(types) => typeRefsFlatten(types)
+    case IntersectionType(types) => typeRefsFlatten(types)
     // unwrap type wrappers to the underlying type(s); e.g. a refined parent is encoded as
     // StructuralType(WithType(...)) (see scalacp/TypeOps.scala).
-    case StructuralType(t, _) => typeSymbols(t)
-    case AnnotatedType(_, t) => typeSymbols(t)
-    case ExistentialType(t, _) => typeSymbols(t)
-    case UniversalType(_, t) => typeSymbols(t)
+    case StructuralType(t, _) => typeRefs(t)
+    case AnnotatedType(_, t) => typeRefs(t)
+    case ExistentialType(t, _) => typeRefs(t)
+    case UniversalType(_, t) => typeRefs(t)
     case _ => Iterator.empty
   }
 
-  private def typeSymbolsFlatten(types: Iterable[Type]): Iterator[String] = types.iterator
-    .flatMap(typeSymbols)
+  private def typeRefsFlatten(types: Iterable[Type]): Iterator[(String, Seq[Type])] = types.iterator
+    .flatMap(typeRefs)
+
+  // symbol-only views, used where type arguments aren't needed (e.g. `isSubtypeOf`)
+  private def parentSymbols(sinfo: SymbolInformation): Iterator[String] = parentRefs(sinfo).map(_._1)
+  private def typeSymbols(tpe: Type): Iterator[String] = typeRefs(tpe).map(_._1)
 
   // `tpe` plus its transitive parents, most-derived first and cycle-safe, each paired with the
   // substitution of its type parameters to erased symbols (so generic overrides compare correctly).
@@ -153,25 +151,6 @@ trait SymbolTable {
     }
     loop(tpe, Map.empty)
     seen.toList
-  }
-
-  // Parent (symbol, type-arguments) pairs from a class signature, or the alias target for a type.
-  private def parentRefs(sinfo: SymbolInformation): Iterator[(String, Seq[Type])] =
-    sinfo.signature match {
-      case cs: ClassSignature => cs.parents.iterator.flatMap(typeRefs)
-      case ts: TypeSignature => typeRefs(ts.upperBound)
-      case _ => Iterator.empty
-    }
-
-  private def typeRefs(tpe: Type): Iterator[(String, Seq[Type])] = tpe match {
-    case TypeRef(_, symbol, args) => Iterator(symbol -> args)
-    case WithType(types) => types.iterator.flatMap(typeRefs)
-    case IntersectionType(types) => types.iterator.flatMap(typeRefs)
-    case StructuralType(t, _) => typeRefs(t)
-    case AnnotatedType(_, t) => typeRefs(t)
-    case ExistentialType(t, _) => typeRefs(t)
-    case UniversalType(_, t) => typeRefs(t)
-    case _ => Iterator.empty
   }
 
   // Map `owner`'s type parameters to the erased symbols of `args` (resolved through the caller's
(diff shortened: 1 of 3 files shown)


semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -51,6 +51,44 @@ trait SymbolTable {
...
+  def overloads(tpe: String, name: String): List[String] = {
+    // constructors are not inherited, so do not walk parents for the `<init>` name
+    val bases = if (name == "<init>") List(tpe -> Map.empty[String, String]) else linearization(tpe)
+    val seenSignatures = mutable.Set.empty[List[String]]
+    bases.flatMap { case (owner, env) => membersNamed(owner, name).map(_ -> env) }.filter {
+      case (method, env) => erasedSignature(method, env) match {
+          case Some(signature) => seenSignatures.add(signature)
+          case None => true // unresolved parameter types: keep rather than risk a wrong merge
+        }
+    }.map(_._1)
```
kitbellew:
```suggestion
    bases.flatMap { case (owner, env) =>
      // unresolved parameter types: use `forall` to keep rather than risk a wrong merge
      membersNamed(owner, name).filter(erasedSignature(_, env).forall(seenSignatures.add))
    }
```

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -102,4 +140,102 @@ trait SymbolTable {
...
+      env: Map[String, String],
+  ): Map[String, String] =
+    if (args.isEmpty) Map.empty
+    else info(owner).map(_.signature) match {
+      case Some(c: ClassSignature) => c.typeParameters.symbols.iterator.zip(args.iterator)
+          .flatMap { case (param, arg) => headSymbol(arg, env).map(param -> _) }.toMap
+      case _ => Map.empty
+    }
+
+  private def membersNamed(owner: String, name: String): List[String] =
```
kitbellew: this is the same exact logic you added in the previous commit, lines 39-49, so let's use it there.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -102,4 +140,102 @@ trait SymbolTable {
...
+  private def typeRefs(tpe: Type): Iterator[(String, Seq[Type])] = tpe match {
+    case TypeRef(_, symbol, args) => Iterator(symbol -> args)
+    case WithType(types) => types.iterator.flatMap(typeRefs)
+    case IntersectionType(types) => types.iterator.flatMap(typeRefs)
+    case StructuralType(t, _) => typeRefs(t)
+    case AnnotatedType(_, t) => typeRefs(t)
+    case ExistentialType(t, _) => typeRefs(t)
+    case UniversalType(_, t) => typeRefs(t)
+    case _ => Iterator.empty
+  }
```
kitbellew: i would recommend instead updating (or renaming) parentSymbols, typeSymbols, and typeSymbolFlatten above, to return args as well. it's the same exact logic, DRY.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -102,4 +140,102 @@ trait SymbolTable {
   private def typeSymbolsFlatten(types: Iterable[Type]): Iterator[String] = types.iterator
     .flatMap(typeSymbols)
 
+  // `tpe` plus its transitive parents, most-derived first and cycle-safe, each paired with the
+  // substitution of its type parameters to erased symbols (so generic overrides compare correctly).
+  // A type always precedes its own ancestors, so dedup-by-first-seen keeps the most-derived method.
+  private def linearization(tpe: String): List[(String, Map[String, String])] = {
+    val seen = mutable.LinkedHashMap.empty[String, Map[String, String]]
+    def loop(sym: String, env: Map[String, String]): Unit = if (!seen.contains(sym)) {
+      seen(sym) = env
```
kitbellew: instead of
```
if (!seen.contains(sym)) {
  seen(sym) = env
```

i'd use
```
if (env eq seen.getOrElseUpdate(sym, env)) {
```
or
```
val size = seen.size()
seen.getOrElseUpdate(sym, env)
if (seen.size() != size) {
```

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -102,4 +140,102 @@ trait SymbolTable {
...
+  private def erasedSignature(method: String, env: Map[String, String]): Option[List[String]] =
+    info(method).map(_.signature) match {
+      case Some(m: MethodSignature) =>
+        val types = m.parameterLists.iterator.flatMap(paramInfos).map(_.signature match {
+          case v: ValueSignature => headSymbol(v.tpe, env)
+          case _ => None
+        }).toList
+        if (types.forall(_.isDefined)) Some(types.flatten) else None
+      case _ => None
+    }
```
kitbellew, line 191:
```suggestion
  private def erasedSignature(method: String, env: Map[String, String]): Option[List[String]] =
    info(method).map(_.signature) match {
      case Some(m: MethodSignature) =>
        val symInfos = m.parameterLists.iterator.flatMap(paramInfos)
        val res = List.newBuilder[String]
        @tailrec
        def iter(): Option[List[String]] =
          if (symInfos.hasNext) symInfos.next().signature match {
            case v: ValueSignature => headSymbol(v.tpe, env) match {
                case Some(sym) =>
                  res += sym
                  iter()
                case None => None
              }
            case _ => None
          }
          else Some(res.result())
        iter()
      case _ => None
    }
```

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -102,4 +140,102 @@ trait SymbolTable {
...
+
+  private def paramInfos(scope: Scope): Iterator[SymbolInformation] =
+    if (scope.symlinks.nonEmpty) scope.symlinks.iterator.flatMap(info) else scope.hardlinks.iterator
+
+  // Reduce a type to its erased head symbol: unwrap by-name/repeated wrappers, then resolve a type
+  // parameter through `env` (substitution) or, failing that, follow its alias/upper bound.
+  private def headSymbol(tpe: Type, env: Map[String, String]): Option[String] = {
+    @tailrec
+    def resolve(sym: String, fuel: Int): String = env.get(sym) match {
```
kitbellew: what is `fuel`? and is 16 a magic number?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -102,4 +140,102 @@ trait SymbolTable {
...
+          case _ => sym
+        }
+      case None => sym
+    }
+    def head(t: Type): Option[String] = t match {
+      case ByNameType(u) => head(u)
+      case RepeatedType(u) => head(u)
+      case _ => typeSymbols(t).toList.headOption
+    }
+    head(tpe).map(resolve(_, 16))
```
kitbellew:
```suggestion
    @tailrec
    def getInnerType(t: Type): Option[String] = t match {
      case ByNameType(u) => getInnerType(u)
      case RepeatedType(u) => getInnerType(u)
      case _ => t
    }
    val innerTypeSymbols = typeSymbols(getInnerType(tpe))
    if (innerTypeSymbols.hasNext) Some(resolve(innerTypeSymbols.next(), 16)) else None
```

tests-semanticdb/src/test/scala/scala/meta/tests/symtab/SymbolTableSuite.scala
```diff
@@ -162,4 +162,160 @@ class SymbolTableSuite extends FunSuite {
...
+    assertEquals(
+      genericHierarchy("scala/Predef.String#").overloads("_empty_/Sub#", "foo"),
+      List("_empty_/Sub#foo()."),
+    ),
+  )
+
+  test("overloads(tpe, name): a generic hierarchy keeps genuine overloads")(
+    // Sub#foo(Int) is a real overload of Base#foo(String-after-substitution), so both survive
+    assertEquals(
+      genericHierarchy("scala/Int#").overloads("_empty_/Sub#", "foo").toSet,
```
kitbellew: why are we doing .toSet again? can you please take into account previous comments, or, if you are using a friend, share the context with that friend.

kitbellew (changes requested):

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: Yeah, your version is cleaner,udpated.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: done

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: good call

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: magic number eliminated, replaced with seen cycle

tests-semanticdb/src/test/scala/scala/meta/tests/symtab/SymbolTableSuite.scala
jozanek: shame, sorry for this one, I did this work in parallel with previous PR.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: IDK here. The `forall` version is shorter, and the short-circuit only saves work when an early parameter is unresolvable (rare, and parameter lists are tiny) so the payoff doesn't justify the extra code. Your call.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: Also unsure here. By-name/repeated nest at most one level. I will adopt your version only if you prefer it.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
kitbellew: then keep it as is. i wrote that code _to understand_ what yours was doing 😁

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
kitbellew: if you can add tailrec to yours and not do toList just to get the first element, then I'm ok with that.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: Used your `size`/`getOrElseUpdate` variant.

JFYI - the first form (`env eq …`): `Map.empty` is a singleton, so it's the env for the root and every no-type-args ancestor. On a revisit `getOrElseUpdate` returns that same `Map.empty` and `env eq existing` is `true`, so it'd recurse again, unbounded on `A extends B; B extends A` (the cycle test would StackOverflow).

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: tailrec added, toList removed

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -79,27 +109,118 @@ trait SymbolTable {
...
+    def resolve(sym: String, seen: Set[String]): String = env.get(sym) match {
+      case Some(bound) => bound
+      case None => info(sym).map(_.signature) match {
+          case Some(ts: TypeSignature) => firstSymbol(typeSymbols(ts.upperBound)) match {
+              case Some(next) if !seen(next) => resolve(next, seen + next)
+              case _ => sym
+            }
+          case _ => sym
+        }
+    }
```
kitbellew:
```suggestion
    val seen = mutable.Set.empty[String]
    @tailrec
    def resolve(sym: String): String = env.get(sym) match {
      case Some(bound) => bound
      case None => info(sym).map(_.signature) match {
          case Some(ts: TypeSignature) => firstSymbol(typeSymbols(ts.upperBound)) match {
              case Some(next) if seen.add(next) => resolve(next)
              case _ => sym
            }
          case _ => sym
        }
    }
```

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -79,27 +109,118 @@ trait SymbolTable {
...
+    @tailrec
+    def head(t: Type): Option[String] = t match {
+      case ByNameType(u) => head(u)
+      case RepeatedType(u) => head(u)
+      case _ => firstSymbol(typeSymbols(t))
+    }
+    head(tpe).map(resolve(_, Set.empty))
```
kitbellew:
```suggestion
    @tailrec
    def head(t: Type): Option[String] = t match {
      case ByNameType(u) => head(u)
      case RepeatedType(u) => head(u)
      case _ =>
        val symbols = typeSymbols(t)
        if (symbols.hasNext) Some(resolve(symbols.next()) else None
    }
    head(tpe)
```

kitbellew: this one is optional, but while i understand why you needed to define firstSymbol, it's not generic enough, and without it, this version makes more sense to me.

of course, if you really want to add some sort of extension method somewhere under scala-2.12 defining .nextOption...

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
kitbellew: you usually defined a mutable set (hashset?), so this one is a bit out of form

kitbellew:
@jozanek one request, if you don't mind: rather than working on many tracks at once, would you be able to iterate of fewer ones but quickly? in the past couple of days, this and another pr have been dormant, post my latest review, but there was new work on some other issues.

jozanek:
@kitbellew ooops, I have just opened new one :) I agree to narrow the focus and open only up to 2 PRs at the same time.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: used your mutable-set form

semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
jozanek: dropped the helper

albertikm:
you can open as many as you want, but as long as older ones get iterated on quickly. i keep cycling through prs to see if there have been updates so we can nitpick a bit more or merge 😁

kitbellew (approved):

PR #4657, opened 2026-06-15
Remove dead ScalaSigParserError catch for resolved #1494

### LINKED ISSUE #1494: Scalacp does not emit Type for final val in scala-library

jozanek:
Investigating #1494 ("Scalacp does not emit Type for final val in scala-library"), I found it's **already resolved** — `final val` constants now emit correct types via `metacp`: literal constants become a `ConstantType`, and the reported Java-enum alias (`DAYS = TimeUnit.DAYS`) becomes a `SingleType` to the enum symbol via `ConstantType(underlying: ExternalSymbol)` in `TypeOps`.

Both are regression-tested by `types.Test.Literal` in the `example` corpus (incl. `javaEnum = LinkOption.NOFOLLOW_LINKS`), with the scalacp/classpath path matching the compiler-plugin path (`metacp.expect` == `metac.expect`, enforced by `MetacpExpect`).

The only leftover is the stale catch in `SymbolInformationOps.sig` that silently swallows scalap parse failures into `NoSignature`:
```scala
case ScalaSigParserError("Unexpected failure") =>
  // FIXME: https://github.com/scalameta/scalameta/issues/1494
  s.NoSignature
```
I confirmed (temporary instrumentation) it never fires across the supported scala-library versions, scala-reflect, scala-compiler, and cats-core, so this removes dead code and the misleading silent fallback.

### DIFF
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/scalacp/SymbolInformationOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/scalacp/SymbolInformationOps.scala
@@ -8,7 +8,6 @@ import scala.meta.internal.semanticdb.{Language => l}
 import scala.meta.internal.{semanticdb => s}
 
 import scala.collection.mutable
-import scala.tools.scalap.scalax.rules.ScalaSigParserError
 import scala.tools.scalap.scalax.rules.scalasig._
 
 trait SymbolInformationOps {
@@ -158,9 +157,6 @@ trait SymbolInformationOps {
           if (settings.logBrokenSignatures) reporter.err
             .println(s"broken signature for ${sym.ssym}: ${ex.getMessage}")
           if (settings.stubBrokenSignatures) s.NoSignature else throw ex
-        case ScalaSigParserError("Unexpected failure") =>
-          // FIXME: https://github.com/scalameta/scalameta/issues/1494
-          s.NoSignature
       }
 
     private def annotations: List[s.AnnotationTree] = {
--- a/tests-semanticdb/src/test/scala-2.12/scala/meta/tests/metacp/FinalValConstantSuite.scala
+++ b/tests-semanticdb/src/test/scala-2.12/scala/meta/tests/metacp/FinalValConstantSuite.scala
@@ -1,0 +1,103 @@
+package scala.meta.tests.metacp
+
+import scala.meta.cli._
+import scala.meta.internal.io._
+import scala.meta.internal.semanticdb.TextDocuments
+import scala.meta.internal.{semanticdb => s}
+import scala.meta.io._
+import scala.meta.metacp.{Settings => MetacpSettings}
+
+import java.nio.file.{Files, Path}
+
+import scala.reflect.internal.util.BatchSourceFile
+import scala.reflect.io.{Directory, PlainDirectory}
+import scala.tools.nsc.reporters.StoreReporter
+import scala.tools.nsc.{Global, Settings => CompilerSettings}
+
+import munit.FunSuite
+
+// Regression test for https://github.com/scalameta/scalameta/issues/1494:
+// metacp (the classpath -> SemanticDB path, i.e. scalacp) must emit a type for
+// `final val` constants. A literal constant becomes a ConstantType, and a `final val`
+// aliasing a Java enum constant (the originally reported `DAYS = TimeUnit.DAYS` shape)
+// becomes a SingleType to that enum symbol.
+//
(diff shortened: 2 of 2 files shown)


albertikm:
two questions: 
- @tgodzik could you chime in? looks non controversial
- @jozanek thank you for going over the list methodically but perhaps it would be more useful to start with more recent stuff first? if something was reported years ago and not re-reported again, chances are, it's not that important anymore

jozanek:
@albertikm thanks for having a look. Honestly, I have no idea what is more useful as I am new to this repo and I haven't yet build "the roadmap" vibe. I would appreciate any insight. Regarding the importance/time correlation, I just fixed 7 years old bug in bloop (junit5 integration). Also I am strong believer that old comments shall be reviewed if they are still applicable (this PR basically). But I get your point. What do you say maintainers go over current bugs and close ones that they consider obsolete?

kitbellew:
> @albertikm thanks for having a look. Honestly, I have no idea what is more useful as I am new to this repo and I haven't yet build "the roadmap" vibe. I would appreciate any insight. Regarding the importance/time correlation, I just fixed 7 years old bug in bloop (junit5 integration). Also I am strong believer that old comments shall be reviewed if they are still applicable (this PR basically). But I get your point. What do you say maintainers go over current bugs and close ones that they consider obsolete?

recent maintainers are @tgodzik and me, and my understanding of the semantic side of things is cursory at best. perhaps original brains (@xeno-by , @tanishiking , @olafurpg ) could put their thumbs on this scale.

from the standpoint of a tool like scalafix, consistency between scala2 and scala3 semantic db might be important. it would be fantastic to close most issues, one way or another, but the ranking is unknown to me.

tgodzik:
I agree that some of the older issues might have lost relevance, but I think it's still worth to close any that are simple enough to fix. I would hesitate to work on anything larger, best to ask before contributing any time there.

tgodzik (commented):
> I found it's already resolved — final val constants now emit correct types via metacp: literal constants become a ConstantType, and the reported Java-enum alias (DAYS = TimeUnit.DAYS) becomes a SingleType to the enum symbol via ConstantType(underlying: ExternalSymbol) in TypeOps.

Do we have tests for that? If we don't could you add them here?

jozanek:
> > I found it's already resolved — final val constants now emit correct types via metacp: literal constants become a ConstantType, and the reported Java-enum alias (DAYS = TimeUnit.DAYS) becomes a SingleType to the enum symbol via ConstantType(underlying: ExternalSymbol) in TypeOps.
> 
> Do we have tests for that? If we don't could you add them here?

@tgodzik Already covered by goldens (`types.Test.Literal` via `MetacpExpect`). Added `FinalValConstantSuite` on top, `assert`-based: literal → `ConstantType`, Java-enum alias → `SingleType`, non-final → widened `TypeRef`.

tgodzik (approved):
Makes sense then! We should be ok to merge

PR #4658, opened 2026-06-17
Mark Java enum values/valueOf as SYNTHETIC

jozanek:
Follow-up to #1492: mark Java enum `values()`/`valueOf(String)` as `SYNTHETIC`, matched by name + exact descriptor so overloads like `valueOf(int)` aren't misclassified.


semanticdb/scalac/library/src/main/scala/scala/meta/internal/javacp/Javacp.scala
```diff
@@ -212,8 +214,22 @@ object Javacp {
...
 
-        val methodInfo =
-          addInfo(methodSymbol, methodKind, methodDisplayName, methodSig, method.node.access)
+        // The JLS-mandated enum methods `values`/`valueOf` are compiler-synthesized (not in
+        // source) yet are not flagged ACC_SYNTHETIC in bytecode. Match them by name AND exact
+        // generated descriptor so an unrelated overload (e.g. `valueOf(int)`) isn't misclassified.
+        val isSyntheticEnumMethod = isEnum &&
+          (method.node.name == "values" && method.node.desc == s"()[L${node.name};" ||
+            method.node.name ==
+            "valueOf" && method.node.desc == s"(Ljava/lang/String;)L${node.name};")
```
kitbellew:
```suggestion
        val isSyntheticEnumMethod = isEnum && (methodDisplayName match {
          case "values" => method.node.desc == s"()[L${node.name};"
          case "valueOf" => method.node.desc == s"(Ljava/lang/String;)L${node.name};"
          case _ => false
        })
```

kitbellew: however, isn't it possible to look at method parameters in a structured way, without matching on `.desc`?

kitbellew (changes requested):
does it only work for java enums, not scala3 enums?

semanticdb/scalac/library/src/main/scala/scala/meta/internal/javacp/Javacp.scala
jozanek: rewritten to structured signature detection

jozanek:
@kitbellew yes, my original intention was for java enums only. But I tried running metacp on a compiled Scala 3 enum and it emits nothing.

kitbellew (approved):

semanticdb/scalac/library/src/main/scala/scala/meta/internal/javacp/Javacp.scala
```diff
@@ -212,8 +214,28 @@ object Javacp {
...
+        // The JLS-mandated enum methods `values`/`valueOf` are compiler-synthesized (not in
+        // source) yet are not flagged ACC_SYNTHETIC in bytecode. Match them by name AND exact
+        // generated signature so an unrelated overload (e.g. `valueOf(int)`) isn't misclassified.
+        def enumType = ClassTypeSignature.simple(node.name)
+        val isSyntheticEnumMethod = isEnum &&
+          (methodDisplayName match {
+            case "values" => method.signature.params.isEmpty &&
+              method.signature.result == ArrayTypeSignature(enumType)
+            case "valueOf" =>
+              method.signature.params == List(ClassTypeSignature.simple("java/lang/String")) &&
```
kitbellew, line 226: i am assuming you have a test which covers this specific comparison.

kitbellew (approved):

semanticdb/scalac/library/src/main/scala/scala/meta/internal/javacp/Javacp.scala
jozanek: Nah, added it in #4661

PR #4665, opened 2026-06-27
ScalametaParser: `if` cond in parens can be tuple

### LINKED ISSUE #4664: Issues parsing tuple in new syntax
```
object MainNeverCalled {
   if (a, b) == pair then ???
}
```

It seems we expect the closing parenthesis not a tuple.

kitbellew:
Fixes #4664.


tgodzik (approved):
LGTM! Thanks for looking into this!

PR #4675, opened 2026-07-05
parsers: support scala 3 match sub-cases

### LINKED ISSUE #4673: Support new scala-3 sub-case matching
```scala
    context.lift(idx) match {
      case Some(m) if m.content match case c: ChatMessageContent.AssistantToolCall => ???
      case _ =>
        ???
    }
```

valid code, compiles, but the formatter breaks

https://docs.scala-lang.org/scala3/reference/experimental/sub-cases.html

on another note, I tried to wrap the whole thing in a `// format: off ... // format: on` block, hoping that would allow me to at least use the feature, but alas, still a parse error. Not sure if theres anything that could be done there, to allow "malformed" code in a formatted off block.

Because with the current state of things, I basically cant use this cool new scala feature, because my CICD pipeline enforces formatting, and theres seemingly no way to make the formatter allow this

kitbellew:
Parse Scala 3 sub-cases: a `match` in a case guard whose sub-cases fall through to the enclosing match, e.g.
`case Some(m) if m.content match case c: T => ...`.

Mirrors dotty's representation: `Case(pat, cond, body)` is unchanged; the sub-match is a new `Term.SubMatch` node (a MatchLike sibling of Match/SelectMatch) held in `body`. It is distinct from a plain `Term.Match` body because it has different syntax (`if e match`, no `=>`) and semantics (falls through rather than throwing when no sub-case matches).

No dialect flag: the guard is parsed normally (a trailing `match` becomes a match-as-operator expression), then disambiguated by what follows -- a second `if`, or the absence of `=>`, reinterprets the guard `match` as a sub-match; a guard followed by `=>` stays an ordinary boolean guard (which may itself be a match). Both new forms were previously syntax errors, so this is unambiguous and additive.

Scanner changes:
- getCaseIntro: when a same-line `case` follows a `match` sitting in a case guard, give the synthetic indent region a deeper indent so the single sub-case collapses at the enclosing level instead of absorbing the following outer case (the catch/match-body single-liner keeps its same-level leniency).
- RegionCaseExpr now extends RegionNonDelimNonIndented (it is a non-delimiter, non-indented region), so getOutdentInfo unwinds through a lingering case-guard region left on a sub-match indent (a sub-case has no `=>` to convert the guard region to a body). Without this, a braceless nested sub-sub-match's single multi-level dedent stopped short and absorbed the outer case.

Fixes #4673.

### DIFF
--- a/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
+++ b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala
@@ -2715,9 +2715,24 @@ class ScalametaParser(input: Input)(implicit dialect: Dialect, options: ParserOp
         else if (forceSingleExpr) expr(location = BlockStat, allowRepeated = false)
         else blockOnOther()
       }
-      @inline
-      def guard(): Option[Term] = if (at[KwIf]) Some(guardOnIf()) else None
-      Case(pattern(), guard(), caseBody())
+      val pat = pattern()
+      if (!at[KwIf]) Case(pat, None, caseBody())
+      else {
+        // Scala 3 sub-cases: the guard is parsed normally (a trailing `match` becomes a
+        // match-as-operator expression), then disambiguated by what follows -- a second
+        // `if`, or the absence of `=>`, marks the guard `match` as a sub-match. A guard
+        // followed by `=>` stays an ordinary boolean guard (which may itself be a match).
+        def asSubMatch(t: Term.Match) = copyPos(t)(Term.SubMatch(t.expr, t.casesBlock, t.mods))
+        val firstGuard = guardOnIf()
+        if (at[KwIf]) guardOnIf() match {
+          case t: Term.Match => Case(pat, Some(firstGuard), asSubMatch(t))
+          case t => syntaxError("`match` expected", at = t)
+        }
+        else firstGuard match {
+          case t: Term.Match if !at[RightArrow] => Case(pat, None, asSubMatch(t))
+          case _ => Case(pat, Some(firstGuard), caseBody())
+        }
+      }
     }
   }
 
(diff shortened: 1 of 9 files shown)


zielinsky (commented):
Thanks for tackling this! The logic looks good to me.

But are we sure we don't want a dialect flag here? Sub-cases only work on Scala 3.8+ (and even then only with `language:experimental.subCases`), so parsing with e.g. the `Scala33` dialect should probably still give a syntax error - same as the compiler would.

Other features also got a flag even though they were additive and unambiguous, e.g. `allowBinaryLiterals` (`Scala35`).

kitbellew:
> But are we sure we don't want a dialect flag here? Sub-cases only work on Scala 3.8+ (and even then only with `language:experimental.subCases`), so parsing with e.g. the `Scala33` dialect should probably still give a syntax error - same as the compiler would.

scalameta parser is not designed as a **code validation** tool. it assumes the code is correct and tries to convert to ast.

the syntax sometimes evolves where the parser in reality didn't need to change, if it hadn't been using a gate.

> Other features also got a flag even though they were additive and unambiguous, e.g. `allowBinaryLiterals` (`Scala35`).

Perhaps it's time to make those flags a no-op if it's really unambiguous.

in this case, adding a flag requires explicitly rejecting some syntax which we know how to parse, because there's nothing else it could mean, and for what purpose? so that some time in the future someone would complain that this feature should not be part of Scala3Future dialect but added to 3.10?

tgodzik (approved):
Thanks @zielinsky for taking a look (context: he was more familiar with the new feature)

I guess you are right that we shouldn't add flags if the added feature doesn't break some other. Let's go with not adding a flag.

kitbellew:
> Thanks @zielinsky for taking a look (context: he was more familiar with the new feature)
> 
> I guess you are right that we shouldn't add flags if the added feature doesn't break some other. Let's go with not adding a flag.

thanks! is the new feature more or less settled in terms of shape?

tgodzik:
I think so, but we need to push it through SiP, which is the biggest hurdle.

PR #4678, opened 2026-07-06
Fix bincompat of SemanticdbConfig via @unroll

kitbellew:
Adding buildTarget as a case-class field in #4640 broke binary compatibility: it renumbered/retyped the synthetic copy/apply/ctor accessors (copy$default$6 went from BinaryMode to String). When two semanticdb-scalac versions share a classpath (e.g. scalafix vs. sbt), the evicted-to newer jar no longer has the old-arity methods, so older compiled code fails with NoSuchMethodError.

Move buildTarget to the last constructor param with a default and annotate it @unroll, which regenerates the old-arity apply/copy/ctor forwarders. Restores copy$default$6 to BinaryMode and adds a new copy$default$13:String.


kitbellew:
@tgodzik full disclosure: my friend Klaudiusz has authored this :) i think the commit icon should show it. also see #4679.

tgodzik (approved):
Thanks! That's exactly what I meant! I also sometimes get help from that same friend 😅

PR #4679, opened 2026-07-06
Shade semanticdb-scalac plugin internals per version

kitbellew:
Relocate the plugin's internal implementation classes into a per-version package via assemblyShadeRules, so two different semanticdb-scalac versions (e.g. scalafix's vs. sbt's) can coexist on one -Xplugin classpath without their identically-named internal classes colliding at link time (the NoSuchMethodError class of #4640).

Only scala.meta.internal.semanticdb.scalac.** moves; SemanticdbPlugin (named in scalac-plugin.xml) stays put, and the shade rules rewrite its references to the relocated classes. The semanticdb-shared wire format (scala.meta.internal.semanticdb.**, TextDocument etc.) is intentionally not shaded -- it's public interchange and stays mima's responsibility.

A follow-on to #4678 (see #4640).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -603,6 +603,23 @@ lazy val mergeSettings = Def.settings(
   assembly / assemblyJarName :=
     name.value + "_" + scalaVersion.value + "-" + version.value + "-assembly.jar",
   assembly / assemblyOption ~= { _.withIncludeScala(false) },
+  // Relocate the plugin's internal implementation classes into a per-version
+  // package, so two different semanticdb-scalac versions can coexist on one
+  // -Xplugin classpath (e.g. scalafix's vs. sbt's) without their identically
+  // named internal classes colliding at link time (see #4640). Only the entry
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
Looks good, though I wonder if we need it together with the unroll annotation?

kitbellew:
the way i understood the problem, there were multiple competing plugins, and truly internal code all of a sudden became subject to binary compatibility rules. 

so I see two alternatives: enable mima on this code or hide it completely. the unroll solution is about mima. what do you prefer?

tgodzik:
I think unroll and mima is probably a way to go. Shading always comes with some weirdness, but I don't have a very strong preference.

kitbellew:
let's start with this and then figure out mima. that track would require duplicating all truly public stuff out of the internal package (i think even the plugin is in internal) and enabling mima on it, then waiting for adoption before removing/deprecating the internal stuff. so shading should allow us to go through this process without disruptions.

PR #4681, opened 2026-07-07
Add hand-built-tree fixture tests for TreeSyntax

jozanek:
Round-trip tests can't localize printer-only regressions or pin per-dialect rendering. Adds `InfixParensSuite` and `DialectSyntaxSuite`, asserting reprints of directly-constructed trees.

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/DialectSyntaxSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/prettyprinters/DialectSyntaxSuite.scala
@@ -1,0 +1,126 @@
+package scala.meta.tests
+package prettyprinters
+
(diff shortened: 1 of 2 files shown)


albertikm:
@jozanek can you explain the why a bit more? we have dedicated syntactic suites, and they should have enough validation already, although per dialect, but importantly with parsing. TreeSyntax should produce code which is parsable by us, in the same dialect.

jozanek:
@albertikm sure, the printer's interface is `Tree => String`, and its real callers (Scalafix, quasiquote splices) feed it trees that never went through the parser (some not even expressible as source). These suites test that interface directly, including the same tree printed under several dialects, which round-trips can't do.

kitbellew (approved):

Issue #1544, opened 2018-05-22
Companion objects sometimes appear in wrong order in metac and metacp


xeno-by:
```
object X {
  case class A()
  class B
  object A
}
```

```
    tpe {
      tag: CLASS_INFO_TYPE
      classInfoType {
        parents {
          tag: TYPE_REF
          typeRef {
            symbol: "scala.AnyRef#"
          }
        }
        declarations: "_empty_.X.A#"
        declarations: "_empty_.X.A."
        declarations: "_empty_.X.A."
        declarations: "_empty_.X.B#"
      }
    }
```

Note the duplicate decl for `object A`. This is another bug in metacp: https://github.com/scalameta/scalameta/issues/1543.

PR #4688, opened 2026-07-11
metac: order companion object declarations by source position

jozanek:
## Problem

metac emitted `ClassSignature.declarations` with an explicit companion `object` in the wrong place. The namer puts it in an eager scope slot instead of its source position. So declared members came out of source order (#1544).

## Root cause

For a case class, or a class with default constructor parameters, `Namer.ensureCompanionObject` enters a companion right after the class. An explicit companion declared later reuses that early slot. The plugin read declaration order straight from `Scope.sorted`, so it kept the wrong order.

## Why only modules

The first attempt sorted all positioned declarations by source position. That regressed Java overload disambiguators. Java declaration order is non-normative, and reordering it desyncs metac from metacp. It also reordered late-entered annotated fields. The bug is specific to companion modules. So the fix floats only those, and leaves every other member in scope order.

## Scope: metac only

metacp reads Scala 2 pickles, which have no source positions. It cannot recover source order, so it still emits the old order. The divergence is recorded in `metac-metacp.diff`. The metacp half stays open. Part of #1544.

### DIFF
--- a/semanticdb/integration/src/main/scala/example/Issue1544.scala
+++ b/semanticdb/integration/src/main/scala/example/Issue1544.scala
@@ -1,0 +1,8 @@
+// See https://github.com/scalameta/scalameta/issues/1544
+package example
+
+object Issue1544 {
+  case class A()
+  class B
+  object A
+}
--- a/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolOps.scala
+++ b/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolOps.scala
@@ -100,7 +100,7 @@ trait SymbolOps {
             val gbuf = List.newBuilder[g.Symbol]
             gdecls.sorted.filter(_.isUsefulSymbolInformation).foreach(gbuf.+=)
             if (sym.isJavaDefined) sym.companionModule.info.decls.filter(_.isUseful).foreach(gbuf.+=)
-            SemanticdbDecls(gbuf.result())
+            SemanticdbDecls(inSourceOrder(gbuf.result()))
           case _ => SemanticdbDecls(Nil)
         }
         loop(sym.info)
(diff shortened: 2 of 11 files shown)


semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolOps.scala
```diff
@@ -214,6 +214,30 @@ trait SymbolOps {
...
   }
 
+  // Move explicit companion objects to their source position, see
+  // https://github.com/scalameta/scalameta/issues/1544. Namer.ensureCompanionObject eagerly
+  // enters a companion right after its class (e.g. for a case class or default-parameter
+  // class), so one declared later in the source reuses that early scope slot and lands out of
+  // order. We float only these modules; other members keep scalac's scope order (avoiding
+  // needless divergence from metacp). Synthetic companions with no explicit `object` stay put.
+  private def inSourceOrder(gsyms: List[g.Symbol]): List[g.Symbol] = {
+    def isFloatable(sym: g.Symbol): Boolean = sym.isModule && !sym.isSynthetic &&
```
kitbellew: thank you, claude, for using fancy terminology. @jozanek can we rename it to something else? in our context, float more frequently suggests something to do with floating-point numbers, and this seems to be about something else. but despite the overly lengthy comment above, i don't know what.

semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolOps.scala
```diff
@@ -214,6 +214,30 @@ trait SymbolOps {
...
+      val (floats, anchors) = gsyms.partition(isFloatable)
+      floats.foldLeft(anchors)((acc, module) =>
+        acc
+          .indexWhere(anchor => anchor.pos.isDefined && anchor.pos.start > module.pos.start) match {
+          case -1 => acc :+ module
+          case i =>
+            val (before, after) = acc.splitAt(i)
+            before ::: module :: after
+        },
+      )
```
kitbellew: n^2 algo and a bit cryptic. why not explicit merge:
```
val (floaters, anchors) = gsyms.partition(isFloatable)
val as = anchors.iterator.buffered
val result = List.newBuilder[...]
floaters.foreach { floater =>
  while (as.hasNext && (as.head.pos.isEmpty || as.head.pos.start <= floater.pos.start))
    result += as.next()
  result += floater
}
result ++= as
result.result()
```

semanticdb/integration/src/main/scala/example/Issue1544.scala
```diff
@@ -0,0 +1,8 @@
+// See https://github.com/scalameta/scalameta/issues/1544
```
kitbellew, line 1: as usual, could you add this file in the first commit, along with all output files to show how it worked before, and then add the sorting fix and the changes to the outputs?

kitbellew (changes requested):

tests-semanticdb/src/test/scala/scala/meta/tests/semanticdb/DeclarationOrderSuite.scala
```diff
@@ -0,0 +1,149 @@
+package scala.meta.tests
```
kitbellew, line 1: shouldn't this test be added in the first commit as well? the usual request, that the fix commit shows how the tests are changing.

tests-semanticdb/src/test/resources/metac-metacp.diff_2.13
```diff
@@ -585,6 +585,18 @@ example/InstrumentTyper#existential().
                      display_name: "T"
 
 
+==================
```
kitbellew, line 588: this has no red part. so earlier there was no difference between metac and metacp, and now there is, and that's good for some reason? 😄

kitbellew (changes requested):

PR #4689, opened 2026-07-11
Restrict legacy tokenizer internals to the scala.meta namespace

jozanek:
Marks the legacy lexer internals (`LegacyScanner`, `LegacyTokenData`, `LegacyToken`, `XmlParser`, `UnexpectedInputEndException`, the `CharArrayReader` companion) `private[meta]`, per the `CharArrayReader` precedent (#3295): Metals keeps compiling, everything outside `scala.meta` is fenced out.

### DIFF
--- a/tests/shared/src/test/scala/scala/meta/tests/tokenizers/BaseTokenizerSuite.scala
+++ b/tests/shared/src/test/scala/scala/meta/tests/tokenizers/BaseTokenizerSuite.scala
@@ -3,8 +3,6 @@ package scala.meta.tests.tokenizers
 import scala.meta._
 import scala.meta.tests.{TestHelpers, TreeSuiteBase}
 
-import munit.Location
-
 abstract class BaseTokenizerSuite extends TreeSuiteBase {
 
   protected val dialect: Dialect = dialects.Scala211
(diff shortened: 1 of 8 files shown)


tests/shared/src/test/scala/scala/meta/tests/tokenizers/BaseTokenizerSuite.scala
```diff
@@ -21,26 +19,6 @@ abstract class BaseTokenizerSuite extends TreeSuiteBase {
     expected.lift(obtained).getOrElse(fail("Got unexpected tokens: " + obtained))
   }
 
-  def assertLegacyScannedAsStringLines(
```
kitbellew, line 24: what does removing that have to do with this change?

kitbellew: ok, to be exact, what i meant is: please don't remove this, i use it for debugging. and it has absolutely nothing to do with the goal of this multi-step exercise.

jozanek: @kitbellew restored now, I found no call sites, so I mistook it for dead code. I am also curious about your take on the follow-up proposal in the description 🙏

kitbellew: i am not sure i understood which problem we are fixing, but perhaps i need to see what it is you are planning to propose.

jozanek: @kitbellew the one that you removed from this PR's description :)

kitbellew:
what i removed, i had read. and i removed because i incorrectly assumed that that would be the description of the squashed commit, and left only a relevant (to commit) piece. somehow the description ended up completely empty, so i guess i don't understand how GitHub generated them.

PR #4721, opened 2026-08-09
trees: remove all invariance, field, parent checks

kitbellew:
We are gradually reducing all unambiguous correctness verifications in this module, leaving all of this to the compiler.


tgodzik (approved):
Makes sense, I always have doubts if we are really in sync with what the compiler allows.

PR #4806, opened 2026-09-18
TreeSyntax: indent a commented extension body

### LINKED ISSUE #4805: Printer moves an extension body out of the clause when it has a leading comment
`reprint(withComments = true)` moves the body of an `extension` clause out of the clause when the body carries a leading comment, so valid Scala prints as Scala that does not compile.

### Steps to reproduce

```scala
import scala.meta._
import scala.meta.prettyprinters.XtensionReprint

implicit val d: Dialect = dialects.Scala33

val src = """|extension [A](self: List[A])
             |  /** doc */
             |  def sizeTwice: Int = self.length * 2
             |""".stripMargin

println(src.parse[Source].get.reprint(withComments = true))
```

Obtained (`·` marks a trailing space):

```scala
extension [A](self: List[A]) ·
/** doc */
def sizeTwice: Int = self.length * 2
```

Expected: something that still parses as one extension clause, e.g. the body indented under the clause, or the brace-delimited form the printer already emits for two or more methods.

Compiling the obtained text gives:

```
[error] ./Bug.scala:3:1
[error] Extension without extension methods
[error] def sizeTwice: Int = self.length * 2
[error] ^^^
[error] ./Bug.scala:3:22
[error] Not found: self
[error] def sizeTwice: Int = self.length * 2
[error]                      ^^^^
```

(Scala 3.3.6.)

### Notes

* The body must still carry `Origin.Parsed` with a leading comment. `tree.transform { ... }.syntax` does *not* reproduce it: `transform` rebuilds every node and drops the origins, so the comment is never replayed.
* Two or more methods in the clause are fine — the body is then a `Stat.Block` and the printer emits braces. One method, no comment, is fine too. The single-method-with-comment case is the one that breaks.
* A `//` line comment breaks the same way.
* The trailing space after the parameter clause is also visible in the output above.

### Versions

* scalameta 4.17.4, and `main` at 3be3f52 (`v4.17.4-6-g3be3f52`)
* dialect `Scala33`; the printer runs the same way on 2.12/2.13/3
* Scala 2.13.16 host program

### Is this blocking?

No — there are workarounds: `reprint(withComments = false)`, or brace the clause in the source so the body parses as a `Stat.Block`.

It surfaced as a hard failure in a tool that reprinted whole source files with comments and then compiled them; that particular tool has since changed strategy, but any caller of `reprint(withComments = true)` still gets non-compiling Scala out of valid Scala.

I have a fix with tests ready and will open a PR.

kryptt:
Fixes #4805.

A leading comment is printed on its own line, so an extension body printed on the `extension` line lands at the start of the next line — outside the clause. `reprint(withComments = true)` turned

```scala
extension [A](self: List[A])
  /** doc */
  def sizeTwice: Int = self.length * 2
```

into

```scala
extension [A](self: List[A]) 
/** doc */
def sizeTwice: Int = self.length * 2
```

which does not compile: `Extension without extension methods`, and `A`/`self` are out of scope.

The body is only printed on the same line when it has no leading comment; with one it is indented, which keeps it inside the clause and drops the trailing space after the parameter clause. The brace-delimited form was not an option here: for a single method the body is not a `Stat.Block`, so braces would change the tree and break `parse(code) == parse(print(parse(code)))`. Indenting round-trips.

Bodies without a leading comment print exactly as before, so the `Stat.Block` (two or more methods) and one-line forms are untouched.

Three tests in `ExtensionMethodsSuite`, all of which fail without the change: a single method with a leading scaladoc, the same with a `//` comment, and a braced body whose block carries the comment.

`tests2_12`, `tests2_13`, `tests3_lts`, `tests3_next` `testFull` and `mima2_12`/`mima2_13`/`mima3_lts` pass locally; no other test's layout changed. Reformatted with `./bin/scalafmt`.


kitbellew (changes requested):
1. create two commits: first adds tests and pins current, incorrect behavior, second adds the fix and flips tests.
2. the solution is specific to extension body; why is the same not applicable to any Scala 3 multi-line/optional-braces body?

kitbellew:
instructions for claude:
- do not narrate table-stakes, do not explain how, focus only on why and what; do not describe that the build is green since CI establishes that. 
- you claim, without proof, that several tests fail without the change; a separate commit, showing how they fail, is the minimum proof
- also, you explain that an optional-braces, significant-indentation case with one extension was not parsed as a block, and proceed to fix the printer when you should fix the parser.

kitbellew:
please also fetch and rebase, have claude read newly added CONTRIBUTING.md and AGENTS.md before proceeding.

